PR #227, opened 2019-01-08
discovery

lzl124631x:
This is not finished / intended to be merged. I'm stuck on this issue https://github.com/libp2p/go-libp2p-kad-dht/issues/221. I tried to connect to the chat executable in the chat-with-ren.. example. But I got error `dial attempt failed: protocol not supported`.


LeoHChen:
no longer valid.

PR #266, opened 2019-01-13
Update README.md

LeoHChen:
add git submodule step.


LeoHChen:
update README.md after we added vendor modules.

PR #267, opened 2019-01-16
add state syncing registration

chaosma:
Add node registration for state syncing. If a node is out of sync, it will register itself to a couple of peers and waiting for new blocks to broadcast.


test/bc_config.json
```diff
@@ -0,0 +1,50 @@
+{
```
LeoHChen: please remove bc_config.json

node/node.go
```diff
@@ -119,9 +120,10 @@ type Node struct {
 	clientServer *clientService.Server
 
 	// Syncing component.
-	downloaderServer *downloader.Server
-	stateSync        *syncing.StateSync
-	syncingState     uint32
+	downloaderServer       *downloader.Server
+	stateSync              *syncing.StateSync
+	syncingState           uint32
+	peerRegistrationRecord map[string]int64 // record registration time (unixtime) of peers begin in syncing
```
LeoHChen: You may have map concurrent access issue. please consider using sync.Map data structure.

test/bc_config.json
chaosma: I didn't create this file

LeoHChen: yes, it's a beacon chain by-product.  don't check it in.  we should actually add it to .gitingore.

node/node.go
chaosma: I will add mutex for map success

PR #277, opened 2019-01-18
Fix README.md to reference scripts correctly

johnwhitton:
Fixing a small typo so that the correct scripts directory is referenced when building.


LeoHChen:
Can one of the admins verify this patch?

PR #280, opened 2019-01-18
fix peerstore in hostv2

LeoHChen:
add support of proper PeerID in hostv2 for libp2p support. The PeerID is generated from the public key of the key pair. It will be used in secure communication among nodes.
we were using insecure mode in libp2p before.  this changeset adds the proper key pair and the right peerID support.

It contains many changes as we have to propagate the PeerID in the p2p.peer structure and every node needs to get the other party's PeerID in order to talk to them.


cmd/client/txgen/main.go
```diff
@@ -59,19 +63,35 @@ func main() {
 	// Add GOMAXPROCS to achieve max performance.
 	runtime.GOMAXPROCS(1024)
 
-	var clientPeer *p2p.Peer
+	var BCPeer *p2p.Peer
```
rlan35: Better be "bcPeer"?

p2p/host/hostv2/hostv2.go
```diff
@@ -24,8 +24,38 @@ const (
...
+	targetAddr, err := multiaddr.NewMultiaddr(addr)
+	if err != nil {
+		log.Error("AddPeer NewMultiaddr error", "error", err)
+		return err
+	}
+
+	p.Addrs = append(p.Addrs, targetAddr)
+
+	host.Peerstore().AddAddrs(p.PeerID, p.Addrs, peerstore.PermanentAddrTTL)
+	fmt.Printf("AddPeer add to peerstore: %v\n", *p)
```
rlan35: Use log.Info

mikedoan:
Sorry. I have not finished the review yet.

p2p/p2pimpl/p2pimpl.go
```diff
@@ -1,25 +1,57 @@
...
+	"net"
+
 	"github.com/harmony-one/harmony/p2p"
 	"github.com/harmony-one/harmony/p2p/host"
 	"github.com/harmony-one/harmony/p2p/host/hostv1"
 	"github.com/harmony-one/harmony/p2p/host/hostv2"
+
+	"github.com/harmony-one/harmony/internal/utils"
+	peer "github.com/libp2p/go-libp2p-peer"
+	multiaddr "github.com/multiformats/go-multiaddr"
```
harmony-ek: [nit/style] This package is `ma` by default, and people's eyes have been trained to expect `ma` to mean the multiaddr package.  Do we have a strong reason to use an alternative package tag?

p2p/p2pimpl/p2pimpl.go
```diff
@@ -1,25 +1,57 @@
...
+// on the same host if the node software restarted. The peerstore has to be persistent as well.
+func NewHost(self *p2p.Peer) (host.Host, error) {
 	if Version == 1 {
-		h := hostv1.New(peer)
-		return h
+		h := hostv1.New(self)
+		return h, nil
+	}
+
+	selfAddr, err := multiaddr.NewMultiaddr(fmt.Sprintf("/ip4/0.0.0.0/tcp/%s", self.Port))
```
harmony-ek: Just a note for future that the port may be an ephemeral port later when we have node discovery, which is a node ID → node locator (such as a multiaddr) lookup service.  In that case, a node would just let the OS pick an unused port (in the socket interface one does this by saying port 0; unsure what the libp2p way is) and bind to it, then later register its node-to-locator association with the node discovery/rendezvous service, using its own IP address + the ephemeral port number.

(This logic gets further complicated when we consider NAT traversal, but let's not go too deep :-p)

p2p/p2pimpl/p2pimpl.go
```diff
@@ -1,25 +1,57 @@
...
+	if err != nil {
+		return nil, err
+	}
+	self.Addrs = append(self.Addrs, selfAddr)
+
+	// TODO (leo), change to GenKeyP2PRand() to generate random key. Right now, the key is predicable as the
```
harmony-ek:
```suggestion
	// TODO (leo), change to GenKeyP2PRand() to generate random key. Right now, the key is predictable as the
```

consensus/consensus_leader_msg_test.go
```diff
@@ -14,7 +14,7 @@ import (
 func TestConstructAnnounceMessage(test *testing.T) {
 	leader := p2p.Peer{IP: "127.0.0.1", Port: "19999"}
 	validator := p2p.Peer{IP: "127.0.0.1", Port: "55555"}
-	host := p2pimpl.NewHost(leader)
+	host, _ := p2pimpl.NewHost(&leader)
```
harmony-ek: Why ignore error?
```suggestion
	host, err := p2pimpl.NewHost(&leader)
	if err != nil {
		test.FailNow(fmt.Sprintf("cannot create host: %s", err))
	}
```

cmd/beaconchain/main.go
```diff
@@ -48,6 +48,9 @@ func main() {
 		beaconchain.SetSaveFile(*resetFlag)
 		bc = beaconchain.New(*numShards, *ip, *port)
 	}
+
+	fmt.Printf("Beacon Chain Started: /ip4/%s/tcp/%v/ipfs/%s\n", *ip, *port, bc.GetID().Pretty())
```
mikedoan: Just some thinking. do you think we need to use log for all the printing and make it consistently over the whole code base?

consensus/consensus_leader_test.go
```diff
@@ -81,17 +84,20 @@ func TestProcessMessageLeaderResponse(test *testing.T) {
...
+	host2, _ := p2pimpl.NewHost(&validator2)
+	consensusValidator2 := New(host2, "0", []p2p.Peer{validator1, validator2, validator3}, leader)
 	consensusValidator2.blockHash = [32]byte{}
 	_, msg = consensusValidator2.constructCommitMessage(consensus_proto.MessageType_COMMIT)
 	consensusLeader.ProcessMessageLeader(msg[1:])
 
-	consensusValidator3 := New(p2pimpl.NewHost(validator3), "0", []p2p.Peer{validator1, validator2, validator3}, leader)
+	host3, _ := p2pimpl.NewHost(&validator3)
+	consensusValidator3 := New(host3, "0", []p2p.Peer{validator1, validator2, validator3}, leader)
 	consensusValidator3.blockHash = [32]byte{}
```
harmony-ek: [nit/refactor] Can you roll the above section into a loop to reduce copy-and-paste and make it a bit more readable?

cmd/harmony.go
```diff
@@ -142,7 +167,16 @@ func main() {
 		ldb, _ = InitLDBDatabase(*ip, *port)
 	}
 
-	host := p2pimpl.NewHost(selfPeer)
+	host, err := p2pimpl.NewHost(&selfPeer)
+	if err != nil {
+		panic("unable to new host in harmony")
+	}
+
+	fmt.Printf("[Harmony] selfPeer: %v\n", selfPeer)
```
mikedoan: is it debug message? log.Debug could be better?

internal/newnode/newnode.go
```diff
@@ -72,32 +80,34 @@ func (node *NewNode) requestBeaconChain(BCPeer p2p.Peer) (err error) {
...
 	}
-	p := p2p.Peer{IP: node.Self.IP, Port: node.Self.Port}
-	nodeInfo := &bcconn.NodeInfo{Self: p, PubK: pubk}
+	nodeInfo := &proto_node.Info{IP: node.Self.IP, Port: node.Self.Port, PubKey: pubk, PeerID: node.host.GetID()}
 	msg := bcconn.SerializeNodeInfo(nodeInfo)
 	msgToSend := proto_identity.ConstructIdentityMessage(proto_identity.Register, msg)
 	gotShardInfo := false
-	timeout := time.After(300 * time.Second)
-	tick := time.Tick(5 * time.Second)
+	timeout := time.After(120 * time.Second)
```
harmony-ek: Maybe add a TODO saying that this should be configurable?  Also, maybe `2 * time.Minute` instead?

internal/newnode/newnode.go
```diff
@@ -72,32 +80,34 @@ func (node *NewNode) requestBeaconChain(BCPeer p2p.Peer) (err error) {
...
 				host.SendMessage(node.host, BCPeer, msgToSend, nil)
 			}
 		}
 	}
 	if !gotShardInfo {
 		err = errors.New("could not create connection")
-		node.log.Crit("Could not get sharding info after 5 minutes")
-		os.Exit(1)
+		node.log.Crit("Could not get sharding info after 2 minutes")
+		os.Exit(10)
```
mikedoan: can you elaborate this? 10 or 1?

p2p/p2pimpl/p2pimpl.go
```diff
@@ -1,25 +1,57 @@
...
+	peerID, err := peer.IDFromPublicKey(pubKey)
+
+	if err != nil {
+		return nil, err
+	}
+
+	self.PeerID = peerID
+	h := hostv2.New(*self, priKey)
+
+	fmt.Printf("NewHost => self:%s, PeerID: %v\n", net.JoinHostPort(self.IP, self.Port), self.PeerID)
```
mikedoan: log.Info - we have a logging singleton which can be used across all code base.

p2p/p2pimpl/p2pimpl.go
LeoHChen: @harmony-ek , please open an issue for further note.

consensus/consensus_leader_test.go
LeoHChen: don't want to change the original test code too much.  I'll open an issue to simplify the test case as a beginner's task. #284

cmd/beaconchain/main.go
LeoHChen: don't want to change this one now.  current beacon chain cmd will be obsolete soon by boot node.

internal/newnode/newnode.go
LeoHChen: make it 10 to make it easier to identify the failure reason in return code of the go test.  Again, newnode will be obsolete soon.

p2p/p2pimpl/p2pimpl.go
LeoHChen: fixed.

cmd/harmony.go
LeoHChen: debug message, removed.

PR #282, opened 2019-01-18
be verbose in travis checker

LeoHChen:
use -count=1 to disable test cache. This is useful, especially in the local test.

### DIFF
--- a/scripts/travis_checker.sh
+++ b/scripts/travis_checker.sh
@@ -37,7 +37,7 @@ go_dirs="${tmpdir}/go_dirs.txt"
 dirnames < "${go_files}" | sort -u -t/ > "${go_dirs}"
 
 echo "Running go test..."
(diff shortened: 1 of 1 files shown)


scripts/travis_checker.sh
```diff
@@ -37,7 +37,7 @@ go_dirs="${tmpdir}/go_dirs.txt"
 dirnames < "${go_files}" | sort -u -t/ > "${go_dirs}"
 
 echo "Running go test..."
-if go test ./...
+if go test -v ./... -count=1
```
astralblue:
```suggestion
if go test -v -count=1 ./...
```
So as to conform to the typical UNIX command line convention (positional arguments come after optional arguments).

LeoHChen: sure.

PR #288, opened 2019-01-21
Switch to disk based db

mikedoan:


test/leader_run.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
LeoHChen: do you really need to check in this script?

core/genesis.go
```diff
@@ -221,7 +223,9 @@ func (g *Genesis) configOrDefault(ghash common.Hash) *params.ChainConfig {
 // to the given database (or discards it if nil).
 func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
 	if db == nil {
-		db = ethdb.NewMemDatabase()
+		utils.GetLogInstance().Error("db should be initialized")
+		os.Exit(1)
+		// db = ethdb.NewMemDatabase()
```
LeoHChen: Please just remove the commented out statement here.

test/deploy.sh
```diff
@@ -24,6 +24,7 @@ function cleanup() {
    done
    # Remove bc_config.json before starting experiment.
    rm -f bc_config.json
+   rm -rf /tmp/harmony*
```
LeoHChen: is the default path of the db in /tmp? 

We should make it in a dedicated directory like $HOME  or just current directory because /tmp is usually mounted as tmpfs, which is still memory based file system.

mikedoan: done.

cmd/harmony.go
```diff
@@ -82,7 +81,7 @@ func main() {
 	port := flag.String("port", "9000", "port of the node.")
 	logFolder := flag.String("log_folder", "latest", "the folder collecting the logs of this execution")
 	attackedMode := flag.Int("attacked_mode", 0, "0 means not attacked, 1 means attacked, 2 means being open to be selected as attacked")
-	dbSupported := flag.Bool("db_supported", false, "false means not db_supported, true means db_supported")
+	dbSupported := flag.Bool("db_supported", true, "false means not db_supported, true means db_supported")
```
LeoHChen: Two changes requested.
1) I'd ask for a parameter to let user specify where to store the db file. 
2) Also, is there a way to reload the db from disk to memory?   we should support reload the db from file to memory.

PR #290, opened 2019-01-22
Go generate

harmony-ek:
This PR removes all autogenerated files (except in Ethereum core) from the repo, and instead regenerates them upon invocation.


harmony-ek:
Travis-CI build may fail; will address soon.

mikedoan:
I don't think removing go generated files is a good idea - we should keep them instead

harmony-ek:
Had a discussion with @rlan35 and @mikedoan.

A few points survived the discussion:

* If not checked in, developers and IDEs may not be able to inspect the generated code by default until they regenerate the code by manually running `go generate`.
* If checked in, developers may forget to regenerate after changing the source file, or developers may use an unapproved/wrong version of the generator tool (such as an older version of protoc that the developer neglected to keep up to date), hurting reproducibility of builds and hindering bug triaging (unnecessary CNR).

After the discussion, I also came up with another reason against not having generated files in the tree: It breaks `go get` (which purposefully avoids running any external tool for security reasons; since the purpose of `go generate` is to run external tools to generate code, `go get` also skips `go generate`).

So, we are going with the following approach (accepted by @mikedoan and @rlan35; others, please comment):

* Keep the generated files in tree.
* Clearly designate the version of the codegen tool that everyone should be using (mention them in CONTRIBUTING.md).
* Use the right version of the codegen tool to regenerate the files from Travis-CI and Jenkins-CI.
    * Fail the build if the checked-in version is different from regenerated version.
        * Such failure means that the developer must have checked in a version not properly generated by the recommended version of the codegen tool.
        * Coupled with no-pass-no-merge rule, this will prevent the improperly generated versions from hitting master.

harmony-ek:
Deprioritizing this because this no longer blocks gossip-related refactor work.

PR #296, opened 2019-01-22
don't close the stream after send message

LeoHChen:
hope it will fix the sendmessage errors

### DIFF
--- a/p2p/host/hostv2/hostv2.go
+++ b/p2p/host/hostv2/hostv2.go
@@ -114,7 +114,6 @@ func (host *HostV2) SendMessage(p p2p.Peer, message []byte) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


p2p/host/hostv2/hostv2.go
```diff
@@ -114,7 +114,6 @@ func (host *HostV2) SendMessage(p p2p.Peer, message []byte) error {
 		return err
 	}
 
-	defer s.Close()
```
mikedoan: can you give some context of this in p2p? 

Maybe the whole logic of this function is wrong: every time we send a message we open a new stream
s, err := host.h.NewStream(context.Background(), p.PeerID, ProtocolID)

without closing this - we'll have many streams open by sending Messages.

PR #298, opened 2019-01-22
fix send message error

LeoHChen:
tested locally on my AWS desktop and it works as expected.


test/deploy.sh
```diff
@@ -138,8 +138,11 @@ while IFS='' read -r line || [[ -n "$line" ]]; do
   IFS=' ' read ip port mode shardID <<< $line
 	#echo $ip $port $mode
   if [[ "$mode" == "leader" || "$mode" == "validator" ]]; then
-      $DRYRUN $ROOT/bin/harmony -ip $ip -port $port -log_folder $log_folder $DB -min_peers $MIN $HMY_OPT 2>&1 | tee -a $LOG_FILE &
-      sleep 0.5
+     $DRYRUN $ROOT/bin/harmony -ip $ip -port $port -log_folder $log_folder $DB -min_peers $MIN $HMY_OPT 2>&1 | tee -a $LOG_FILE &
+     sleep 0.5
+  fi
+  if [[ "$mode" == "newnode" && "$SYNC" == "true" ]]; then
+     (sleep 35 ; $DRYRUN $ROOT/bin/harmony -ip $ip -port $port -log_folder $log_folder $DB -min_peers $MIN $HMY_OPT 2>&1 | tee -a $LOG_FILE ) &
```
chaosma: If there are multiple newnodes, will all the new nodes sleep 35 seconds and join the network at almost similar time? We want to each of them delay by 35s.

LeoHChen: I see.  I'll add some logic here.

test/configs/local_config1.txt
```diff
@@ -1,5 +1,5 @@
 127.0.0.1 9000 leader 0
-127.0.0.1 9001 validator 0
+127.0.0.1 9001 leader 0
```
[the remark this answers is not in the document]
LeoHChen: not new one, just make it more clear that we use 2 shards by default and 9001 is another shard leader.

PR #302, opened 2019-01-23
introduce universal message

mikedoan:
We use protobuf because it supports backward and forward compatibility. See
https://www.beautifulcode.co/backward-and-forward-compatibility-protobuf-versioning-serialization

Why protobuf? Think about when we decide to change the messsage proto from v1 to v2,
 either when the server upgrades the implementation from v1 to v2 or only client side upgrades from v1 to v2 both sides still work well thank to protobuf capability.

With this introduction, we can use this universal message to wrap any type of communication message as long we need to specify the type and introduce its request respectively.

TODO work is to refactor all current messaging protocol, for example: consensus_leader.go consensus_leader_msg.go etc..


mikedoan:
TODO: we'll need to clean up all data structure in the proto directory.

api/proto/message/gen.sh
```diff
@@ -0,0 +1 @@
+protoc -I ./ message.proto --go_out=./
```
LeoHChen: why don't put the script in /scripts directory?

api/proto/message/message.proto
```diff
@@ -0,0 +1,33 @@
...
+  NEWNODE_BEACON = 3;
+  BEACON_NEWNODE = 4;
+}
+
+// This is universal message for all communication protocols.
+// There are different Requests for different message types.
+// As we introduce a new type of message just add a new MessageType and new type of request in Message.
+//
+// The request field will be either one of the structure corresponding to the MessageType type.
+message Message {
```
rlan35: Not sure if it's good to have a single wraper for all messages. Maybe it's too early to decide that. But migrating to grpc is good.

LeoHChen: I don't know if we need to migrate to grpc for all messages.  @rlan35 , you mean using protobuf?

rlan35: yea, protobuf.

mikedoan: I will use this for beacon chain first.

mikedoan: grpc is actually just one feature of protobuf

mikedoan:
I'm submitting because i'm not going to change everything soon. just use first for beacon change.

PR #304, opened 2019-01-23
gometalinter p2p/...

harmony-ek:
Mostly fixes error handling, but there are a few functional changes:

* `ReadMessageContent()` returns an error upon `SetReadDeadline()` failures.
* hostv2 `SendMessage()` returns an error upon write failures.

BTW, gometalinter = ❤️.  I'd strongly recommend using it regularly.

### DIFF
--- a/p2p/helper.go
+++ b/p2p/helper.go
@@ -2,7 +2,6 @@ package p2p
 
 import (
 	"bufio"
-	"bytes"
 	"encoding/binary"
 	"io"
 	"time"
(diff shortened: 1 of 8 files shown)


p2p/helper.go
```diff
@@ -28,46 +27,39 @@ content (n bytes) - actual message content
...
 	}
 	// TODO: check on msgType and take actions accordingly
 	//// Read 4 bytes for message size
 	fourBytes := make([]byte, 4)
-	n, err := r.Read(fourBytes)
-	if err != nil {
-		log.Error("Error reading the p2p message size field")
-		return contentBuf.Bytes(), err
-	} else if n < len(fourBytes) {
-		log.Error("Failed reading the p2p message size field: only read", "Num of bytes", n)
```
mikedoan: the code is so ugly - hopefully it's not mine.

p2p/host/hostv2/hostv2.go
```diff
@@ -108,13 +109,21 @@ func (host *HostV2) BindHandlerAndServe(handler p2p.StreamHandler) {
 
 // SendMessage a p2p message sending function with signature compatible to p2pv1.
 func (host *HostV2) SendMessage(p p2p.Peer, message []byte) error {
+	logger := log.New("from", host.self, "to", p, "PeerID", p.PeerID)
```
mikedoan: let's use utils.GetLogInstance() for all the logging here - it does not make sense to create just one object for SendMessage.

mikedoan:
I support to use gometalinter - it will be a lot of effort to switch to that level. golint is still our base. Maybe create a bounty gitcoin for gometalinter

p2p/host/hostv2/hostv2.go
harmony-ek: This child logger was used so that it adds the common parameters (from/to/PeerID) to all log messages emitted from this function.  However, I should've created it as a child of utils.GetLogInstance() and not log.Root().  Will keep in mind next time :-)

PR #308, opened 2019-01-24
Fetch and use (“gimme”) stable Go, not master

harmony-ek:
The Go installation (gimme) step of Travis-CI is super slow (~3m),
because we use master version of Go, which gimme needs to
download/compile/install out of the Git repository.

Use stable instead of master, which lets gimme to download/install the
binary version (if available), of the latest stable release, and fall
back to source only if that fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - master
+  - stable
 install:
   - export GOPATH=$HOME/gopath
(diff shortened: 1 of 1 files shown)


harmony-ek:
The gimme step of the PR build took 5s, down from 3m (-97%).  <3

alajko:
Do you mind adding some logs?

harmony-ek:
@alajko, [189.98s](https://travis-ci.org/harmony-one/harmony/builds/483670398#L437) → [5.03s](https://travis-ci.org/harmony-one/harmony/builds/483680648#L445).  ❤️

PR #319, opened 2019-01-24
add PeerID in beacon chain RPC message

LeoHChen:
wallet needs to use PeerID to talk to leaders


LeoHChen:
no longer need it in master, moving to release1901 branch.

PR #324, opened 2019-01-25
Vendor Kyber at pre-modularization

harmony-ek:
Kyber is going through version 3 release process, in which they are
switching to being a full module and requiring the import path to be
go.dedis.ch/kyber. This breaks fresh installs and likely CI builds too.
Peg at the version right before the modularization so that our code (yet
to be adjusted for the new import path) continues to build.

This is a temporary bandaid.  We will be getting rid of Kyber soon in
favor of https://github.com/herumi/bls.


harmony-ek:
retest this please

harmony-ek:
retest this please

PR #325, opened 2019-01-25
service management for Node

mikedoan:


node/service.go
```diff
@@ -0,0 +1,136 @@
...
+type ActionType byte
+
+// Constants for Action Type.
+const (
+	Start ActionType = iota
+	Stop
+)
+
+// Type is service type.
+type Type byte
```
rlan35: Better to be explicit "ServiceType"

node/service.go
```diff
@@ -0,0 +1,136 @@
...
+
+// SendAction sends action to action channel which is observed by service manager.
+func (node *Node) SendAction(action *Action) {
+	node.actionChannel <- action
+}
+
+// TakeAction is how service manager handles the action.
+func (node *Node) TakeAction(action *Action) {
+	if node.serviceStore == nil {
+		utils.GetLogInstance().Error("Service store is not initialized.")
```
chaosma: why not start it if it's nil?

mikedoan: because with nil it can not get the service.

node/service.go
mikedoan: Done

PR #326, opened 2019-01-25
Remove init on bls test

rlan35:


rlan35:
> As the proposer of the soft-fork maintenance policy ([harmony-one/go-ethereum#1](https://github.com/harmony-one/go-ethereum/pull/1)), I still think third-party libraries' behavior should be controlled from our side, but if enough people (4+) are OK with it, I'll fold.
> 
> Case in point (quoted from the proposed policy):
> 
> > **Note:** Since this repo is a new fork, we should decide upon some ground rules here.  I propose that we treat this as a **soft fork:** This is primarily _upstream's_ (that is, Ethereum's) repository, and _their_ minimum bar on quality/style applies as much as practically acceptable.  We use this code as an external user.  In a sense, we are a “visitor,” and try to keep our “footprint”—local changes—as light as possible.  Lean local changes reduce our ongoing maintenance burdens.
> > …
> > 
> > 1. **Do** try to influence code behavior from the outside, e.g. through the API, rather than by changing the code.
> >    
> >    1. In particular, **don't** alter the default behavior, if the behavior can be controlled from the outside.  Such changes are unlikely to be accepted by upstream, due to backward compatibility concerns.
> 
> @chaosma @LeoHChen @mikedoan @alajko @lzl124631x, please weigh in.

Do this have to apply to all repos? Maybe other repo can have different rules?

harmony-ek:
> Do this have to apply to all repos? Maybe other repo can have different rules?

As we will have more and more soft forks, it quickly becomes a hassle to juggle different policies for different soft forks.  I would very much prefer one unified policy, with exceptions made sparingly with a strong consensus.  The supermajority (4+ out of 6) consensus I'm trying to see here is for that, and I implore everyone to carefully weigh in and voice their assessment of pros and cons.

chaosma:
I agree with @harmony-ek : 

Do try to influence code behavior from the outside, e.g. through the API, rather than by changing the code.
In particular, don't alter the default behavior, if the behavior can be controlled from the outside. Such changes are unlikely to be accepted by upstream, due to backward compatibility concerns.

unless we do a hard fork, not change the 3rd library behavior

harmony-ek:
Had offline discussion with @alajko and @rlan35: Consensus is not to have this (and revert the `init()` patch in the soft fork).

@alajko had a further point: If we need some compound behavior from the library which could be implemented either in the library or in the client, we should do it on the client side unless we believe the maintainer is likely to accept the patch if we upstreamed it.

(That belief/assessment whether the upstream maintainer would accept our upstream patch is an educated guess, with no clear answer.  I'd say try once or twice and quickly gauge his/her interest level.  Who knows, the upstream maintainer may not really be receptive or even responsive…, but we wouldn't know until we try.)

PR #333, opened 2019-01-25
clean up import order & change service type

mikedoan:


alajko:
LGTM. But I am wondering why 13 more files got dragged into this commit? any idea?

PR #338, opened 2019-01-26
Remove deprecated ready field

harmony-ek:
Originally #320, which can't go through 'cause it somehow doesn't trigger Travis-CI builds anymore.


harmony-ek:
There.  :-)

PR #339, opened 2019-01-26
Add unit tests for bls utils

rlan35:


internal/utils/utils_test.go
```diff
@@ -48,6 +48,14 @@ func TestGenKey(t *testing.T) {
 	GenKey("3.3.3.3", "3456")
 }
 
+// Test for GenKeyBLS
+func TestGenKeyBLS(t *testing.T) {
+	priKey, pubKey := GenKeyBLS("3.3.3.3", "3456")
+	if priKey == nil || pubKey == nil {
```
LeoHChen: can you find a way to compare the key generated is correct or not?

crypto/bls/bls_test.go
```diff
@@ -0,0 +1,37 @@
+package bls
+
+import (
+	"testing"
+
+	"github.com/harmony-one/bls/ffi/go/bls"
+	"github.com/harmony-one/harmony/internal/utils"
+)
+
+func TestNewMask(test *testing.T) {
```
LeoHChen: please add a comment for this function.

internal/utils/utils_test.go
rlan35: There is not much special logic in the function. It's basically putting what is generated from GetUniqueIDFromIPPort as bytes into private key and call the GetPublicKey util of BLS. I just add result checking for it based on the output. But we are trusting the BLS package is working fine for now.

rlan35:
> what's the test coverage for bls package?

ok  	github.com/harmony-one/harmony/crypto/bls	0.038s	coverage: 43.5% of statements

PR #348, opened 2019-01-29
Basic multisig checking for consensus

rlan35:


rlan35:
> looks like the consensus runs slower now after the verification is added.

Yes, locally, seems it's 10% slower.

PR #352, opened 2019-01-29
Add multi-sig check on validator commit phase

rlan35:


consensus/consensus_validator.go
```diff
@@ -235,9 +238,20 @@ func (consensus *Consensus) processCommittedMessage(message consensus_proto.Mess
...
-	_ = multiSig
-	_ = bitmap
-	// TODO: verify multi signature
+	deserializedMultiSig := bls.Sign{}
+	err = deserializedMultiSig.Deserialize(multiSig)
+	if err != nil {
+		utils.GetLogInstance().Warn("Failed to deserialize the multi signature for commit phase", "Error", err, "leader ID", leaderID)
+		return
+	}
+	mask, err := bls_cosi.NewMask(consensus.PublicKeys, nil)
```
LeoHChen: no error check here?

rlan35: The check is in the next if clause.

PR #361, opened 2019-01-30
Refactor message population into a func

rlan35:
## Test
echo 2 shards, 104 consensus, 277 total TPS, 11 nodes 127.0.0.1, 52, 139.672

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -581,3 +581,15 @@ func (consensus *Consensus) GetPeerFromID(peerID uint32) (p2p.Peer, bool) {
 func (consensus *Consensus) SendMessage(peer p2p.Peer, message []byte) {
 	host.SendMessage(consensus.host, peer, message, nil)
 }
+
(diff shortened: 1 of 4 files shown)


consensus/consensus.go
```diff
@@ -581,3 +581,15 @@ func (consensus *Consensus) GetPeerFromID(peerID uint32) (p2p.Peer, bool) {
 func (consensus *Consensus) SendMessage(peer p2p.Peer, message []byte) {
 	host.SendMessage(consensus.host, peer, message, nil)
 }
+
+// Populates the common basic fields for all consensus message.
+func (consensus *Consensus) populateBasicFields(message *consensus_proto.Message) {
```
LeoHChen: IMHO, better name it as populateMessageFields

alajko:
LGTM.

consensus/consensus.go
rlan35: Done.

PR #367, opened 2019-01-31
the deposit contract

alajko:
## Issue

Creating the Deposit Contract that the beaconchain service will run.

#### Test Coverage Data

* Before

Not relevant.

* After
Not relevant.

#### Test/Run Logs

Not relevant.

## TODO

- [ ] Add contract to beaconchain service on init
- [ ] Write interaction in `go` for interacting with state of the smart contract.

### DIFF
--- a/contracts/DepositContract.sol
+++ b/contracts/DepositContract.sol
@@ -1,0 +1,75 @@
+pragma solidity ^0.4.22;
+
+contract DepositContract {
+
+  mapping(address => uint) private Stakes;
+  mapping(address => uint) private Indices;
+  address[] private nodeAddressList;
+
+  event LogNewNode(address indexed nodeAddress, uint index, uint stake);
+  event LogUpdateStake(address indexed nodeAddress, uint index,  uint stake);
+  event LogDeleteNode(address indexed nodeAddress, uint index);
+  
+  function isNode(address nodeAddress) public constant returns(bool isIndeed) 
+  {
+    if(nodeAddressList.length == 0) return false;
+    return (nodeAddressList[Indices[nodeAddress]] == nodeAddress);
+  }
+
+function deposit() public payable returns(uint index)
+  {
+    if(isNode(msg.sender)) {
+    	Stakes[msg.sender] += msg.value;
+    	emit LogUpdateStake(msg.sender, Indices[msg.sender], Stakes[msg.sender]);
+   		return Indices[msg.sender];
+    } else {
+    	Stakes[msg.sender]   = msg.value;
+	    Indices[msg.sender] = nodeAddressList.push(msg.sender) - 1;
+    	emit LogNewNode(msg.sender, Indices[msg.sender], Stakes[msg.sender]);
+    	return nodeAddressList.length-1;
+    }
+  }
+
+  function withdraw(uint withdrawAmount) public returns (uint remainingStake) 
+  {
+        if (withdrawAmount  <= Stakes[msg.sender]) {
+        	 Stakes[msg.sender] -= withdrawAmount;
+             msg.sender.transfer(withdrawAmount);
+             if (Stakes[msg.sender] == 0) { //If all stake has been withdrawn, remove the node.
+             	deleteNode(msg.sender);
+             }
+            return Stakes[msg.sender];
+        } else {
+        	return Stakes[msg.sender];  //Overdraft protection, no money is withdrawn!
+        }
+        
+    }
+  
+  function deleteNode(address nodeAddress) public returns(uint index)
+  {
+  	//Move the last index to deleted index.
+    uint indexToDelete = Indices[nodeAddress];
+    address lastIndexAddress = nodeAddressList[nodeAddressList.length-1];
+    nodeAddressList[indexToDelete] = lastIndexAddress;
+    Indices[lastIndexAddress] = indexToDelete; 
+    nodeAddressList.length--;
+    emit LogDeleteNode(nodeAddress,indexToDelete);
+    emit LogUpdateStake(lastIndexAddress, indexToDelete, Stakes[lastIndexAddress]);
+    return indexToDelete;
+  }
+
+
+  function getNodeCount() public constant returns(uint count)
+  {
+    return nodeAddressList.length;
+  }
+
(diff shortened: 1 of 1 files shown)


contracts/DepositContract.sol
```diff
@@ -0,0 +1,75 @@
...
+
+  mapping(address => uint) private Stakes;
+  mapping(address => uint) private Indices;
+  address[] private nodeAddressList;
+
+  event LogNewNode(address indexed nodeAddress, uint index, uint stake);
+  event LogUpdateStake(address indexed nodeAddress, uint index,  uint stake);
+  event LogDeleteNode(address indexed nodeAddress, uint index);
+  
+  function isNode(address nodeAddress) public constant returns(bool isIndeed)
```
rlan35: Why do we need to track node is deposit contract? I think it's better to be simple, just tracking deposits.

contracts/DepositContract.sol
```diff
@@ -0,0 +1,75 @@
...
+  {
+    if(nodeAddressList.length == 0) return false;
+    return (nodeAddressList[Indices[nodeAddress]] == nodeAddress);
+  }
+
+function deposit() public payable returns(uint index)
+  {
+    if(isNode(msg.sender)) {
+    	Stakes[msg.sender] += msg.value;
+    	emit LogUpdateStake(msg.sender, Indices[msg.sender], Stakes[msg.sender]);
```
rlan35: why emit these events?

alajko: The compiler was showing warnings without `emit`.

contracts/DepositContract.sol
alajko: The node here is just the node address.

contracts/DepositContract.sol
```diff
@@ -0,0 +1,75 @@
+pragma solidity ^0.4.22;
+
+contract DepositContract {
+
+  mapping(address => uint) private Stakes;
+  mapping(address => uint) private Indices;
+  address[] private nodeAddressList;
```
rlan35: Do we need this list? A single map is not enough?

contracts/DepositContract.sol
alajko: Changed the name.

contracts/DepositContract.sol
rlan35: That's weird. My faucet contract never have this problem. You can try remix. I used that before and it worked fine.

contracts/DepositContract.sol
alajko: Any key is trivially "present" in a solidity map (i.e. `map[k] = default_value`). The list is useful to delete a `nodeAddress` and keep track of live stakes.

contracts/DepositContract.sol
alajko: Yes I used remix. I used 4.24 compiler version. 
This is what happened.
`browser/DepositContract.sol:56:5: Warning: Invoking events without "emit" prefix is deprecated.
    LogDeleteNode(nodeAddress,indexToDelete);`

contracts/DepositContract.sol
rlan35: We can not delete the key from the map? I assume we can. Then no need for a separate list, right?

contracts/DepositContract.sol
```diff
@@ -0,0 +1,75 @@
...
+  
+  function isNodeAddress(address nodeAddress) public constant returns(bool isIndeed) 
+  {
+    if(nodeAddressList.length == 0) return false;
+    return (nodeAddressList[Indices[nodeAddress]] == nodeAddress);
+  }
+
+function deposit() public payable returns(uint index)
+  {
+    if(isNode(msg.sender)) {
```
mikedoan: definitely this code is not working - you can not submit such code

contracts/DepositContract.sol
alajko: A key is always present in a map.

contracts/DepositContract.sol
alajko: Forgot to move all changes from remix. Will fix with next commit.

alajko: Added now.

contracts/DepositContract.sol
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    function deposit() public payable returns(uint index) {
+        if(isAccount(msg.sender)) {
+            Stakes[msg.sender] += msg.value;
+            emit LogUpdateStake(msg.sender, Indices[msg.sender], Stakes[msg.sender]);
+            return Indices[msg.sender];
+        } else {
+            Stakes[msg.sender] = msg.value;
+            Indices[msg.sender] = accountList.push(msg.sender) - 1;
```
LeoHChen: do you need to check overflow?

contracts/DepositContract.sol
```diff
@@ -0,0 +1,62 @@
...
+            if (Stakes[msg.sender] == 0) { //If all stake has been withdrawn, remove the Account.
+                deleteAccount(msg.sender);
+            }
+            return Stakes[msg.sender];
+        } else {
+            return Stakes[msg.sender];  //Overdraft protection, no money is withdrawn!
+        }        
+    }
+  
+    function deleteAccount(address account) public returns(uint index) {
```
LeoHChen: can this function be called by other contracts?

contracts/DepositContract.sol
```diff
@@ -0,0 +1,62 @@
...
+      	//Move the last index to deleted index.
+        uint indexToDelete = Indices[account];
+        address lastIndexAddress = accountList[accountList.length-1];
```
LeoHChen: do you need to check if (accountList.length -1) is >= 0?

contracts/DepositContract.sol
```diff
@@ -0,0 +1,62 @@
...
+        accountList[indexToDelete] = lastIndexAddress;
+        Indices[lastIndexAddress] = indexToDelete; 
+        accountList.length--;
```
LeoHChen: is overflow check needed?

rlan35: Better to use SafeMath lib.

PR #375, opened 2019-02-01
make state syncing compatible with new BLS consensus code; last mile

chaosma:
Fixing state syncing breaking with new BLS consensus code. The last mile problem still need to be fix.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -92,7 +92,7 @@ type Consensus struct {
 	OnConsensusDone func(*types.Block)
 
 	// current consensus block to check if out of sync
-	ConsensusBlock chan *types.Block
+	ConsensusBlock chan *BFTBlockInfo
 	// verified block to state sync broadcast
 	VerifiedNewBlock chan *types.Block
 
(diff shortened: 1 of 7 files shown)


consensus/consensus.go
```diff
@@ -211,6 +227,29 @@ func (consensus *Consensus) checkConsensusMessage(message consensus_proto.Messag
 	// check consensus Id
 	if consensusID != consensus.consensusID {
 		utils.GetLogInstance().Warn("Wrong consensus Id", "myConsensusId", consensus.consensusID, "theirConsensusId", consensusID, "consensus", consensus)
+		if consensus.IsLeader {
+			return false
+		}
+		// validator send consensus block to state syncing
+		if val, ok := consensus.blocksReceived[consensusID]; ok {
```
rlan35: I would suggest move this logic to the caller level of checkConsensusMessage, by defining and returning different error types and checking the type on the caller side.

chaosma: Sure

consensus/consensus_validator.go
```diff
@@ -12,6 +12,29 @@ import (
 	"github.com/harmony-one/harmony/internal/utils"
 )
 
+func (consensus *Consensus) sendBFTBlockToStateSyncing(consensusID uint32) {
```
LeoHChen: comment please.

consensus/consensus_validator.go
```diff
@@ -12,6 +12,29 @@ import (
...
+	// validator send consensus block to state syncing
+	if val, ok := consensus.blocksReceived[consensusID]; ok {
```
LeoHChen: golang will usual check !ok and return early.

chaosma: This will also return if !ok

PR #376, opened 2019-02-01
Adding Place Holder for Staking Service

alajko:
## Test

Will add tests once there is more code.

#### Test Coverage Data

* Before
* After

#### Test/Run Logs

Will add later.

### DIFF
--- a/api/service/staking/service.go
+++ b/api/service/staking/service.go
@@ -1,0 +1,25 @@
+package staking
+
+import (
+	"github.com/harmony-benchmark/log"
+	"github.com/harmony-one/harmony/p2p"
+)
+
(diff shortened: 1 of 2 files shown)


api/service/staking/service.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// Service is the struct for staking service.
+type Service struct {
+	Host p2p.Host
+}
+
+//StartService starts the staking service.
+func (s *Service) StartService() {
+	log.Info("Starting staking service.")
```
mikedoan: utils.GetLogInstance()

alajko: Done!

PR #386, opened 2019-02-05
Staking Transactions [WIP] [review-only]

alajko:
## Issue

There needs to be a way to generate staking transactions by the node.

#### Test Coverage Data

* Before

Will add later.

* After
Will add later.
## TODO

- [ ] Add testing and logging to note the receipt (and withdrawal) of funds.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -164,6 +164,10 @@ type Node struct {
 	// Service manager.
 	serviceManager *service_manager.Manager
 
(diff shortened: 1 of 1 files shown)


alajko:
Node does not have a ECDSA PriKey. This code also adds a way to generate it. This will be needed to sign the staking transactions. @rlan35 @harmony-ek @chaosma

PR #387, opened 2019-02-06
Refactor wallet code

mikedoan:
## Issue

I was reading wallet code for beacon work and feel that we need to refactor this code to make it looking better.

## Test

NA


alajko:
Some of these changes are clear. You made the `switch` case readable by having functions. However I don't understand why change some functions from uppercase to lower case.

mikedoan:
lowercase makes the functions themselves private.

Issue #373, opened 2019-01-31
peer discovery using libp2p


LeoHChen:
## Summary

use libp2p rendezvous mechanism for peer discovery

## Current Design

The current peer discovery is a local solution using ping/pong message and a naive beacon node.

libp2p has built-in peer discovery support using Kademilia DHT. 

## Problems

We don't have a mechanism to form a p2p network.

## Proposal

We have a spec on how the peer discovery should work harmony, using two overlays of p2p networks. This issue is open to keep track of the development progress.

LeoHChen:
- done, using rendezvous string.

PR #394, opened 2019-02-06
Enable libp2p peer discovery and test

LeoHChen:
## Issue

#373 

#### Test/Run Logs

the original test/deploy.sh works fine.  I just added a new parameter "-P" to enable libp2p peer discovery test.
Also, for harmony.go, I added a parameter -libp2p_pd for testing as well.

## TODO

- [ ] integrate 2nd stage peer discovery
- [ ] integrate leader and all pubkeys support

### DIFF
--- a/test/deploy.sh
+++ b/test/deploy.sh
@@ -19,7 +19,7 @@ function check_result() {
 }
 
 function cleanup() {
-   for pid in `/bin/ps -fu $USER| grep "harmony\|txgen\|soldier\|commander\|profiler\|beacon" | grep -v "grep" | grep -v "vi" | awk '{print $2}'`;
+   for pid in `/bin/ps -fu $USER| grep "harmony\|txgen\|soldier\|commander\|profiler\|beacon\|bootnode" | grep -v "grep" | grep -v "vi" | awk '{print $2}'`;
    do
        echo 'Killed process: '$pid
        $DRYRUN kill -9 $pid 2> /dev/null
@@ -57,6 +57,7 @@ USAGE: $ME [OPTIONS] config_file_name
    -k nodeport    kill the node with specified port number (default: $KILLPORT)
    -n             dryrun mode (default: $DRYRUN)
    -S             enable sync test (default: $SYNC)
+   -P             enable libp2p peer discovery test (default: $P2P)
 
 This script will build all the binaries and start harmony and txgen based on the configuration file.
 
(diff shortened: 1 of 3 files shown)


test/deploy.sh
```diff
@@ -131,8 +135,23 @@ $DRYRUN $ROOT/bin/beacon -numShards $SHARDS > $log_folder/beacon.log 2>&1 | tee
 sleep 1 #waiting for beaconchain
 MA=$(grep "Beacon Chain Started" $log_folder/beacon.log | awk -F: ' { print $2 } ')
 
+echo "launching boot node ..."
+$DRYRUN $ROOT/bin/bootnode > $log_folder/bootnode.log 2>&1 | tee -a $LOG_FILE &
+sleep 1
+BN_MA=$(grep "BN_MA" $log_folder/bootnode.log | awk -F\= ' { print $2 } ')
```
harmony-ek: This type of information valuable to outside users should be exposed via a more suitable means than just log extraction.  Log parsing shouldn't be part of a deployment pipeline.  I'd say write a one-liner file and then read from the script.  Like a PID file.

test/deploy.sh
```diff
@@ -131,8 +135,23 @@ $DRYRUN $ROOT/bin/beacon -numShards $SHARDS > $log_folder/beacon.log 2>&1 | tee
 sleep 1 #waiting for beaconchain
 MA=$(grep "Beacon Chain Started" $log_folder/beacon.log | awk -F: ' { print $2 } ')
```
harmony-ek: MA → BC_MA maybe?

test/deploy.sh
```diff
@@ -131,8 +135,23 @@ $DRYRUN $ROOT/bin/beacon -numShards $SHARDS > $log_folder/beacon.log 2>&1 | tee
...
 
+echo "launching boot node ..."
+$DRYRUN $ROOT/bin/bootnode > $log_folder/bootnode.log 2>&1 | tee -a $LOG_FILE &
+sleep 1
+BN_MA=$(grep "BN_MA" $log_folder/bootnode.log | awk -F\= ' { print $2 } ')
+
+HMY_OPT=
+HMY_OPT2=
```
harmony-ek: Maybe use `HMY_OPT+=...` and do away with this extra variable?

LeoHChen: HMY_OPT2 is not applicable to txgen, that's why I have to use two variables here.

test/deploy.sh
LeoHChen: sure, I can change that.

test/deploy.sh
LeoHChen: agree. this is an ugly hack. I have opened an issue for a better design.
https://github.com/harmony-one/harmony/issues/396

PR #401, opened 2019-02-07
add isBeacon flag to specify the node is a beacon chain node

LeoHChen:
add networkinfo/discovery service to beacon chain roles

## Issue

During the bootstrap process, we need to indicate the node is a beacon chain node.
Thus, we introduce a flag as an indicator.

#### Test/Run Logs

local travis_checker, build, and test passed.

### DIFF
--- a/api/service/discovery/service.go
+++ b/api/service/discovery/service.go
@@ -85,5 +85,7 @@ func (s *Service) pingPeer(peer p2p.Peer) {
 	content := host.ConstructP2pMessage(byte(0), buffer)
 	s.Host.SendMessage(peer, content)
 	log.Debug("Sent Ping Message to", "peer", peer)
-	s.stakingChan <- peer
+	if s.stakingChan != nil {
+		s.stakingChan <- peer
+	}
 }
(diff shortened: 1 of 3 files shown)


api/service/discovery/service.go
```diff
@@ -85,5 +85,7 @@ func (s *Service) pingPeer(peer p2p.Peer) {
 	content := host.ConstructP2pMessage(byte(0), buffer)
 	s.Host.SendMessage(peer, content)
 	log.Debug("Sent Ping Message to", "peer", peer)
-	s.stakingChan <- peer
+	if s.stakingChan != nil {
+		s.stakingChan <- peer
```
alajko: What does pingPeer do? My understanding is `peerDiscovery` returns few nodes I can connect to. Is the `peer` in the `pingPeer` function the the `bootnode` I am pinging? or it is one of peer(s) that has already been returned to me by `bootnode` (Or possibly just a test function).

LeoHChen: pingPeer is to establish the link between A and B.  A got the info of B from bootnode, but A need to send a ping message to B, to establish the link.  Otherwise, B won't know A as well.

PR #404, opened 2019-02-07
add resharding code

chaosma:
I have added resharding code as well as store the sharding state into the blockchain. The remaining part is to add shardState verification before write to blockchain so I can test the correctness of the code.


chaosma:
In current code, every N (N=5 for testing) block will do resharding and save the new shard state and randomness into the beacon chain block.

core/blockchain.go
```diff
@@ -1626,3 +1637,90 @@ func (bc *BlockChain) SubscribeChainSideEvent(ch chan<- ChainSideEvent) event.Su
...
 }
+
+// GetShardState retrives sharding state given block hash and block number
+func (bc *BlockChain) GetShardState(hash common.Hash, number uint64) types.ShardState {
+	if cached, ok := bc.shardStateCache.Get(hash); ok {
+		shardState := cached.(types.ShardState)
+		return shardState
+	}
+	shardState := rawdb.ReadShardState(bc.db, hash, number)
+	if shardState == nil {
```
LeoHChen: do you need any log here?

core/blockchain.go
```diff
@@ -1626,3 +1637,90 @@ func (bc *BlockChain) SubscribeChainSideEvent(ch chan<- ChainSideEvent) event.Su
...
+		epoch := GetEpochFromBlockNumber(number)
+		shardState = CalculateNewShardState(bc, epoch)
+		bc.shardStateCache.Add(hash, shardState)
+	}
+	return shardState
+}
+
+// ValidateNewShardState validate whether the new shard state root matches
+func (bc *BlockChain) ValidateNewShardState(block *types.Block) error {
```
LeoHChen: the return value is not clear to me?  return nil means okay?  please be more explicit.

core/rawdb/accessors_chain.go
```diff
@@ -373,3 +373,28 @@ func FindCommonAncestor(db DatabaseReader, a, b *types.Header) *types.Header {
 	}
 	return a
 }
+
+// ReadShardState retrieves sharding state
+func ReadShardState(db DatabaseReader, hash common.Hash, number uint64) types.ShardState {
+	data, _ := db.Get(shardStateKey(number, hash))
+	if len(data) == 0 {
```
LeoHChen: do you need any log here?

core/rawdb/accessors_chain.go
```diff
@@ -373,3 +373,28 @@ func FindCommonAncestor(db DatabaseReader, a, b *types.Header) *types.Header {
...
+	shardState := types.ShardState{}
+	if err := rlp.DecodeBytes(data, &shardState); err != nil {
+		log.Error("Fail to decode sharding state", "hash", hash, "number", number, "err", err)
+		return nil
+	}
+	return shardState
+}
+
+// WriteShardState stores sharding state into database
+func WriteShardState(db DatabaseWriter, hash common.Hash, number uint64, shardState types.ShardState) {
```
LeoHChen: why not return error?

core/resharding.go
```diff
@@ -0,0 +1,173 @@
...
+	"math/rand"
+	"sort"
+	"strconv"
+
+	"github.com/harmony-one/harmony/core/types"
+)
+
+const (
+	// InitialSeed is the initial random seed, a magic number to answer everything, remove later
+	InitialSeed int64 = 42
```
LeoHChen: can you use a random number here?

core/resharding.go
```diff
@@ -0,0 +1,173 @@
...
+// Suppose there are N shards, the first N/2 larger shards are called active committees
+// the rest N/2 smaller committees are called inactive committees
+// actually they are all just normal shards
+func (ss *ShardingState) sortCommitteeBySize() {
+	sort.Slice(ss.shardState, func(i, j int) bool {
+		return len(ss.shardState[i].NodeList) > len(ss.shardState[j].NodeList)
+	})
+}
+
+func (ss *ShardingState) assignNewNodes(newNodeList []types.NodeID) {
```
LeoHChen: please add some comments to elaborate on this algorithm.

core/resharding.go
```diff
@@ -0,0 +1,173 @@
...
+		ss.shardState[id].NodeList = append(ss.shardState[id].NodeList, nid)
+	}
+}
+
+// UpdateShardState will first add new nodes into shards, then use cuckoo rule to reshard to get new shard state
+func (ss *ShardingState) UpdateShardState(newNodeList []types.NodeID, percent float64) {
+	rand.Seed(ss.rnd)
+	ss.cuckooResharding(percent)
+	ss.assignNewNodes(newNodeList)
+	//ss.shardStateRoot = ss.shardState.Hash()
```
LeoHChen: please remove the unused statement here.

core/resharding_test.go
```diff
@@ -0,0 +1,18 @@
+package core
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestFakeGetInitShardState(t *testing.T) {
```
LeoHChen: not enough test functions.

core/resharding.md
```diff
@@ -0,0 +1,12 @@
...
+
+The main function of resharding is CalculcateNewShardState. It will take 3 inputs: newNodeList, oldShardState, randomSeed and output newShardState.
+The newNodeList will be retrieved from BeaconChain staking transaction during the previous epoch. The randomSeed and oldShardState is stored in previous epoch block. It should be noticed that the randomSeed generation currently is mocked. After the verifiable random function service is ready, the VRF service will generate the random seed for resharding. 
+
+The resharding process is as follows: we first get newNodeList from staking transactions from previous epoch and assign the new nodes evenly into the N/2 active committees. Then, we kick out X% of nodes from each active committees and put these kicked out nodes into inactive committees evenly. The percentage X roughly equals to the percentage of new nodes into active committee in order to balance the committee size.
```
LeoHChen: is this the resharding algorithm in our WP?

LeoHChen:
Travis build/test is also failed.

core/blockchain.go
chaosma: not need here

core/blockchain.go
chaosma: not all block contains shard value, in this case return nil. If the block contains shard value, return error if root hash does not match. only epoch block contains shard state hash.

core/rawdb/accessors_chain.go
chaosma: since not all blocks contains the shard state value, in this case, we just ignore it.

core/rawdb/accessors_chain.go
chaosma: I follow the patterns for other Write function, all of the Write functions here not return error.  In my opinion, return an error might be better.

core/resharding.go
chaosma: Yes, we can use a random number here. Since this is only for testing purpose, so I just fix a number to see whether each run get the same state for each epoch

core/resharding.md
chaosma: The basic idea is the same as the RapidChain paper. The percentage of kickout is I defined. I guess this detail is not in WP

core/resharding.go
chaosma: removed

chaosma:
The test failed is TestProcessMessageValidatorPrepared and TestProcessMessageValidatorCommit. It's unrelated to this pull request.

core/resharding.go
chaosma: I added the comments for this function.

core/blockchain.go
```diff
@@ -62,6 +63,12 @@ const (
...
 	badBlockLimit       = 10
 	triesInMemory       = 128
+	shardCacheLimit     = 2
+
+	// BlocksPerEpoch is the number of blocks in one epoch
+	// currently set to small number for testing
+	// in future, this need to be adjusted dynamically instead of constant
+	BlocksPerEpoch = 5
 
 	// BlockChainVersion ensures that an incompatible database forces a resync from scratch.
```
alajko: We should make this a global config param, along with number of shards.

chaosma:
travis build/test is fixed

core/blockchain.go
chaosma: In future, we should use a function to dynamically adjust the epoch size. It's to me more like a code level security parameter instead of a configuration parameter.

core/resharding.go
```diff
@@ -0,0 +1,173 @@
...
+	shardState types.ShardState
+}
+
+// sortedCommitteeBySize will sort shards by size
+// Suppose there are N shards, the first N/2 larger shards are called active committees
+// the rest N/2 smaller committees are called inactive committees
+// actually they are all just normal shards
+func (ss *ShardingState) sortCommitteeBySize() {
+	sort.Slice(ss.shardState, func(i, j int) bool {
+		return len(ss.shardState[i].NodeList) > len(ss.shardState[j].NodeList)
```
rlan35: For now, we can say each node has the same staking. But later, we need to switch to sort by amount of stakes. Please add a todo here.

core/resharding.go
```diff
@@ -0,0 +1,173 @@
...
+		id := numActiveShards + i%(ss.numShards-numActiveShards)
+		ss.shardState[id].NodeList = append(ss.shardState[id].NodeList, nid)
+	}
+}
+
+// UpdateShardState will first add new nodes into shards, then use cuckoo rule to reshard to get new shard state
+func (ss *ShardingState) UpdateShardState(newNodeList []types.NodeID, percent float64) {
+	rand.Seed(ss.rnd)
+	ss.cuckooResharding(percent)
+	ss.assignNewNodes(newNodeList)
```
rlan35: I think the algorithm start with assigning new node to active shards, and then do cuckoo resharding (with newly added nodes). So new nodes may up end in the inactive shards.

core/resharding.md
```diff
@@ -0,0 +1,12 @@
+## Resharding
+
+In current design, the epoch is defined to be fixed length, the epoch length is a constant parameter BlocksPerEpoch. In future, it will be dynamically adjustable according to security parameter. During the epoch transition, suppose there are N shards, we sort the shards according to the size of active nodes (that had staking for next epoch). The first N/2 larger shards will be called active committees, and the last N/2 smaller shards will be called inactive committees. Don't be confused by
+the name, they are all normal shards with same functional.
```
rlan35: typo: functional -> functions

core/resharding.md
```diff
@@ -0,0 +1,12 @@
...
+
+All the information about sharding will be stored in BeaconChain. A sharding state is defined as a map which maps each NodeID to the ShardID the node belongs to. Every node will have a unique NodeID and be mapped to one ShardID. At the beginning of a new epoch, the BeaconChain leader will propose a new block containing the new sharding state, the new sharding state is uniquely determined by the randomness generated by verifiable random function (VRF). During the consensus process, all the validators will perform the same calculation and verify the proposed sharding state is valid. After consensus reach, each node will write the new sharding state into the block. This block is called epoch block. In current code, it's the first block of each epoch in BeaconChain.
```
rlan35: generated by distributed randomness protocol (not just VRF)

rlan35: After consensus (is reached)

core/resharding.go
chaosma: I thought we want to the new nodes stay in active shards. In this case, I will change the order of above lines

core/resharding.go
chaosma: added

core/resharding.md
chaosma: thanks! corrected

PR #405, opened 2019-02-08
Integrate libp2p into harmony node

LeoHChen:
## Issue

integration of libp2p peer discovery into harmony node program

#### Test/Run Logs

local deploy.sh and travis_checker passed w/o enabling the libp2p_pd

test log is available on gist using -libp2p_pd mode.

https://gist.github.com/LeoHChen/a669d4e405443e91d0890e87bc0f3884

### DIFF
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -115,7 +115,18 @@ func (s *Service) DoService() {
 				return
 			}
 			if peer.ID != s.Host.GetP2PHost().ID() && len(peer.ID) > 0 {
-				utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs)
+				utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs, "my ID", s.Host.GetP2PHost().ID())
+				var wg sync.WaitGroup
+				wg.Add(1)
+				go func() {
+					defer wg.Done()
+					if err := s.Host.GetP2PHost().Connect(s.ctx, peer); err != nil {
+						utils.GetLogInstance().Warn("can't connect to peer node", "error", err)
+					} else {
+						utils.GetLogInstance().Info("connected to peer node", "peer", peer)
+					}
+				}()
+				wg.Wait()
 				ip := "127.0.0.1"
 				var port string
 				for _, addr := range peer.Addrs {
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -28,6 +28,7 @@ type Service struct {
 	peerChan    chan p2p.Peer
 	peerInfo    <-chan peerstore.PeerInfo
 	discovery   *libp2pdis.RoutingDiscovery
+	lock        sync.Mutex
 }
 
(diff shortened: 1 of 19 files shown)


api/service/networkinfo/service.go
```diff
@@ -115,7 +115,18 @@ func (s *Service) DoService() {
 				return
 			}
 			if peer.ID != s.Host.GetP2PHost().ID() && len(peer.ID) > 0 {
-				utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs)
+				utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs, "my ID", s.Host.GetP2PHost().ID())
+				var wg sync.WaitGroup
+				wg.Add(1)
```
mikedoan: let's do a Lock as group size is only 1 ?? it won't confuse developers. we can discuss if you disgree

api/service/networkinfo/service.go
```diff
@@ -115,7 +115,18 @@ func (s *Service) DoService() {
...
+				go func() {
+					defer wg.Done()
+					if err := s.Host.GetP2PHost().Connect(s.ctx, peer); err != nil {
+						utils.GetLogInstance().Warn("can't connect to peer node", "error", err)
+					} else {
+						utils.GetLogInstance().Info("connected to peer node", "peer", peer)
+					}
+				}()
+				wg.Wait()
 				ip := "127.0.0.1"
```
mikedoan: this does not belong to this PR but i want to discuss.

1/ should be set as constnat
2/ what's the logic behind that. why do we only add non-127.0.0.1 ip to peerChan?

p2p/host/hostv2/hostv2.go
```diff
@@ -210,3 +229,30 @@ func (host *HostV2) Close() error {
...
+		utils.GetLogInstance().Error("ConnectHostPeer", "new ma error", err, "peer", peer)
+		return
+	}
+	peerInfo, err := peerstore.InfoFromP2pAddr(peerAddr)
+	if err != nil {
+		utils.GetLogInstance().Error("ConnectHostPeer", "new peerinfo error", err, "peer", peer)
+		return
+	}
+	var wg sync.WaitGroup
+	wg.Add(1)
```
mikedoan: Let's use lock here?

node/node.go
```diff
@@ -587,6 +593,13 @@ func (node *Node) setupForShardValidator() {
 func (node *Node) setupForBeaconLeader() {
 	chanPeer := make(chan p2p.Peer)
 
+	var err error
+	node.groupReceiver, err = node.host.GroupReceiver(p2p.GroupIDBeacon)
```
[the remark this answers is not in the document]
mikedoan: this is generic group receiver. REsolved

api/service/networkinfo/service.go
LeoHChen: will do

p2p/host/hostv2/hostv2.go
LeoHChen: ok

PR #406, opened 2019-02-08
Add drand package for randomness

rlan35:
go test succeeded.
Running golint...
golint passed.
Running goimports...
goimports passed.
All Checks Passed!!! :-)


rlan35:
This is just initial commit. More need to be done.

mikedoan:
though drand was described in the whitepaper that will be easier to review if you write a quick doc describing context and logic of drand, drand_leader and the plan of integrating this logic into the current code.

rlan35:
The detailed steps of distributed randomness is here:

1. A leader sends an init message with the hash of the last block H(B ) to all the validators. n−1
2. For each validator i , after receiving the init message, a VRF is computed to create a
random number r and a proof : , where is the i pi
(r , p ) RF(sk , H(B ), v)
i i = V i n−1
ski
secret key of validator i and v is the current view number of consensus. Then, each
validator sends back (r , p ) to the leader.
i i
3. The leader waits until it receives at least f + 1 valid random numbers and combines them
with an XOR operation to get the preimage of the final randomness pRnd .
4. The leader runs BFT (discussed in §2) among all the validators to reach consensus on the
pRnd and commit it in block Bn .
5. After pRnd is committed, the leader starts computing the actual randomness
Rnd = V DF(pRnd, T) , where T is the VDF difficulty and is set algorithmically such that the
randomness can only be computed after k blocks.
6. Once Rnd is computed, the leader initiates a BFT among all validators to agree on the
validity of Rnd and finally commit the randomness into the blockchain.

It involves the broadcasting of an init message from the consensus leader to the validators and validators needs to return back their generated vrf values. After that, leader needs to include the pRand resulting from f+1 vrfs into the next block and run consensus on it. 

So pRand will need to communicate consensus service when it have the pRand value ready. For now, this PR is creating the basic message protocol for the init message broadcasting and vrf message returning part.

PR #413, opened 2019-02-09
Faucet and Stake Contract Testing [WIP]

alajko:
#### Test Coverage Data

* Before
* After

### DIFF
--- a/test/stake/main.go
+++ b/test/stake/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/ecdsa"
 	"fmt"
 	"log"
 	"math/big"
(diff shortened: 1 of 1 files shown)


rlan35:
Just in case. The testing for staking contract should be added into test/chain/main.go, instead of copying it as a new file and adding the extra bit of logic. Basically, test/chain/main.go should be the only testing file for all chain state/contract related testing code.

alajko:
>> The testing for staking contract should be added into test/chain/main.go, instead of copying it as a >> new file and adding the extra bit of logic. Basically, test/chain/main.go should be the only testing 
>> file for all chain state/contract related testing code.

Yes, that is the intention.

chaosma:
should we use withdraw(uint256) or withdraw(uint) in DepositContract

Issue #400, opened 2019-02-07
Change faucet smart contract to make it more generic


mikedoan:
## Summary

Change faucet smart contract to make it more generic. With it, we can send specific amount to an address.

## Current Design

See contracts/Faucet.sol. Currently it sends fixed quota 0.5 ether to a request address.

## Problems

Not convenient for staking development.

## Proposal

Change faucet smart contract to make it more generic. With it, we can send specific amount to an address.

PR #414, opened 2019-02-09
call a request to beacon to get nonce & create staking transaction

mikedoan:
## Issue

call a request to beacon to get nonce & create staking transaction

#### Test Coverage Data

NA 

#### Test/Run Logs
     14968 Done                    | tee -a $LOG_FILE
+ rm -rf ./db/harmony_127.0.0.1_9000 ./db/harmony_127.0.0.1_9001 ./db/harmony_127.0.0.1_9002 ./db/harmony_127.0.0.1_9003 ./db/harmony_127.0.0.1_9004 ./db/harmony_127.0.0.1_9005 ./db/harmony_127.0.0.1_9006 ./db/harmony_127.0.0.1_9007 ./db/harmony_127.0.0.1_9008 ./db/harmony_127.0.0.1_9009 ./db/harmony_127.0.0.1_9010 ./db/harmony_127.0.0.1_9011 ./db/harmony_127.0.0.1_9012
./test/deploy.sh: line 21: 14970 Killed: 9               $DRYRUN $ROOT/bin/harmony -ip $ip -port $port -log_folder $log_folder $DB -min_peers $MIN $HMY_OPT $HMY_OPT2 -key /tmp/$ip-$port.key 2>&1
     14971 Done                    | tee -a $LOG_FILE
+ check_result
+ find tmp_log/log-20190208-230443 -name 'leader-*.log'
+ find tmp_log/log-20190208-230443 -name 'validator-*.log'
+ echo ====== RESULTS ======
====== RESULTS ======
++ ./test/../test/cal_tps.sh tmp_log/log-20190208-230443/all-leaders.txt tmp_log/log-20190208-230443/all-validators.txt
+ results='2 shards, 118 consensus, 359 total TPS, 11 nodes
127.0.0.1, 59, 181.308'
+ echo 2 shards, 118 consensus, 359 total TPS, 11 nodes 127.0.0.1, 59, 181.308
+ tee -a tmp_log/log-20190208-230443/r.log
2 shards, 118 consensus, 359 total TPS, 11 nodes 127.0.0.1, 59, 181.308
+ echo 2 shards, 118 consensus, 359 total TPS, 11 nodes 127.0.0.1, 59, 181.308

### DIFF
--- a/api/service/staking/service.go
+++ b/api/service/staking/service.go
@@ -4,6 +4,8 @@ import (
 	"crypto/ecdsa"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/params"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/log"
@@ -81,15 +83,13 @@ func (s *Service) getAccountState(beaconPeer p2p.Peer) *proto.FetchAccountStateR
 func (s *Service) createStakingMessage(beaconPeer p2p.Peer) *message.Message {
 	accountState := s.getAccountState(beaconPeer)
 	toAddress := common.HexToAddress("0x4592d8f8d7b001e72cb26a73e4fa1806a51ac79d")
-	gasLimit := uint64(0)
-	gasPrice := big.NewInt(0)
 	tx := types.NewTransaction(
 		accountState.Nonce,
 		toAddress,
 		0, // beacon chain.
 		new(big.Int).SetBytes(accountState.Balance),
-		gasLimit,
-		gasPrice,
+		params.CallValueTransferGas*2,           // hard-code
+		big.NewInt(int64(params.Sha256BaseGas)), // pick some predefined gas price.
 		nil)
 
 	if signedTx, err := types.SignTx(tx, types.HomesteadSigner{}, s.accountKey); err == nil {
--- a/api/service/staking/service.go
+++ b/api/service/staking/service.go
@@ -21,19 +21,21 @@ import (
 // Service requires private key here which is not a right design.
 // In stead in the right design, the end-user who runs mining needs to provide signed tx to this service.
 type Service struct {
-	stopChan    chan struct{}
-	stoppedChan chan struct{}
-	peerChan    <-chan p2p.Peer
-	accountKey  *ecdsa.PrivateKey
+	stopChan      chan struct{}
+	stoppedChan   chan struct{}
+	peerChan      <-chan p2p.Peer
+	accountKey    *ecdsa.PrivateKey
+	stakingAmount int64
 }
 
 // New returns staking service.
-func New(accountKey *ecdsa.PrivateKey, peerChan <-chan p2p.Peer) *Service {
+func New(accountKey *ecdsa.PrivateKey, stakingAmount int64, peerChan <-chan p2p.Peer) *Service {
 	return &Service{
-		stopChan:    make(chan struct{}),
-		stoppedChan: make(chan struct{}),
-		peerChan:    peerChan,
-		accountKey:  accountKey,
+		stopChan:      make(chan struct{}),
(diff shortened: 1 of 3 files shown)


api/service/staking/service.go
```diff
@@ -1,25 +1,37 @@
...
+	"github.com/harmony-one/harmony/core/types"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/p2p"
 )
 
 // Service is the staking service.
+// Service requires private key here which is not a right design.
+// In stead in the right design, the end-user who runs mining needs to provide signed tx to this service.
 type Service struct {
 	stopChan    chan struct{}
```
alajko: I think there maybe away to go around the need for exposing private key, if we make it 2 steps. 
We first ask for `nonce` (via publicKey/Address) then we send `txn`. So we need not expose `privateKey`. What do you think?

alajko: One way to do that is: getting accountState could be a independent service? It might be required elsewhere as well.

api/service/staking/service.go
```diff
@@ -53,25 +65,46 @@ func (s *Service) Run() {
...
+	client := client.NewClient(beaconPeer.IP, beaconPeer.Port)
+	defer client.Close()
+	return client.GetBalance(crypto.PubkeyToAddress(s.accountKey.PublicKey))
+}
+
+func (s *Service) createStakingMessage(beaconPeer p2p.Peer) *message.Message {
+	accountState := s.getAccountState(beaconPeer)
+	toAddress := common.HexToAddress("0x4592d8f8d7b001e72cb26a73e4fa1806a51ac79d")
+	gasLimit := uint64(0)
+	gasPrice := big.NewInt(0)
```
alajko: Don't understand this address.

alajko:
As we discussed, a good way to deal with this to to have a `signing` library, that reads a transaction and sends back the signed transaction. This way we don't have to carry the `privateKey` around.
Approving this PR based, but gating on the library to be implemented.

api/service/staking/service.go
```diff
@@ -53,25 +65,46 @@ func (s *Service) Run() {
...
+	tx := types.NewTransaction(
+		accountState.Nonce,
+		toAddress,
+		0, // beacon chain.
+		new(big.Int).SetBytes(accountState.Balance),
+		gasLimit,
```
alajko: We should not stake the entire account balance :) It needs to take in a staking amount parameter.

api/service/staking/service.go
mikedoan: i will add a library for this.

api/service/staking/service.go
mikedoan: yeah, agree and i knew it. like i said the right design is to let miner user created signed tx in bytes and provide it to the new node for staking. it's a bit awkward to specify the amount of the staking, private key, public key though even the faucet contract does not allow us the capability to set how much we want to send free money/token to the testing/fake accounts.

That's why i created the issue https://github.com/harmony-one/harmony/issues/400

Let's not block things as we have a lot more work after PRs. You can remember how i wanted to approve your PR right away though i think it could need to be fixed.

mikedoan: See here for the faucet contract https://github.com/harmony-one/harmony/blob/master/contracts/Faucet.sol

alajko: Yes, but the staking contract allows for stake.
The PR needs signing library as we discussed?

alajko:
Approved but needs  1. Signing Txn Library 2. Stake as a param. If we send whole account balance as, we might not have enough gas for execution.

api/service/staking/service.go
mikedoan: created yesterday https://github.com/harmony-one/harmony/issues/415

mikedoan: And actually the library was against with the usage of constant fakeAccounts. it's not urgent now.

PR #422, opened 2019-02-11
BeaconLeader updating staked list.

alajko:
## Issue

The beaconleader to accept transaction and add the newly staked nodes to a list for resharding.

#### Test Coverage Data

* Before
* After

## TODO

- [x] Integrate with receiving transactions (with @mikedoan)
- [x] Integrate with sending list for resharding (with @chaosma)
- [x] Add Unit tests.


node/node.go
```diff
@@ -570,6 +570,44 @@ func (node *Node) RemovePeersHandler() {
...
+func (node *Node) getNewStakedNodesFromBlockNumToBlockNum (prevBlockNum, toCurrentBlock uint64)  ([]common.Address) {
+	Blockchain := node.Blockchain()
+	signerType  := types.HomesteadSigner{}
+	newNodesMap = make(map[common.Address]int)
+	var newNodes []common.Address
+	for i := prevBlockNum,; i < toCurrentBlock + 1; i++ {
+		block = Blockchain.GetBlockByNumber(from)
+		txns = block.Transactions(),
+		for txn := range txns {
+			if txn.Value() > 0 {  //If value >0 means its a staking deposit transaction
```
chaosma: Does each txn has threshold value?  A voting share has some minimum value threshold.

alajko: Yes, we can add that into smart contract later, so that the transaction is not accepted if its below the threshold.

node/contract.go
```diff
@@ -36,12 +36,12 @@ func (node *Node) AddStakingContractToPendingTransactions() {
 	dataEnc := common.FromHex(StakingContractBinary)
 	// Unsigned transaction to avoid the case of transaction address.
 	mycontracttx, _ := types.SignTx(types.NewContractCreation(uint64(0), node.Consensus.ShardID, contractFunds, params.TxGasContractCreation*10, nil, dataEnc), types.HomesteadSigner{}, priKey)
-	node.ContractAddresses = append(node.ContractAddresses, crypto.CreateAddress(contractAddress, uint64(0)))
+	node.StakingContractAddress = crypto.CreateAddress(contractAddress, uint64(0))
```
mikedoan: this is not right

mikedoan: you need to get deployed contract address.

mikedoan: Sorry, this hack logic is correct by this https://github.com/harmony-one/go-ethereum/blob/38cce9ac333d674616047be14c270d7cfbd43641/core/state_processor.go#L119

but we had 2 assumptions here
1/ nonce is 0
2/ you assumed  the address only creates one contract.

let put a separated function for this.

alajko: Done.

node/node.go
```diff
@@ -570,6 +598,51 @@ func (node *Node) RemovePeersHandler() {
...
+//UpdateStakingList updates the stakes of every node.
+func (node *Node) UpdateStakingList(block *types.Block) error {
+	signerType := types.HomesteadSigner{}
+	txns := block.Transactions()
+	for i := range txns {
+		txn := txns[i]
+		value := txn.Value().Int64()
+		currentSender, _ := types.Sender(signerType, txn)
+		_, isPresent := node.CurrentStakes[currentSender]
+		toAddress := txn.To()
```
mikedoan: move lines [610..613] to line 607.

node/node.go
```diff
@@ -257,6 +262,9 @@ func New(host p2p.Host, consensus *bft.Consensus, db ethdb.Database) *Node {
 			node.AddStakingContractToPendingTransactions()
 			node.DepositToFakeAccounts()
 		}
+		if node.Role == BeaconLeader || node.Role == BeaconValidator {
```
mikedoan: i'm planning to refactor the node.New

can you write a comment that we create this map to keep track of current stakes.

actually this data should be saved in somewhere in blockchain (block header?)

node/node.go
```diff
@@ -570,6 +598,51 @@ func (node *Node) RemovePeersHandler() {
...
+	for i := range txns {
+		txn := txns[i]
+		value := txn.Value().Int64()
+		currentSender, _ := types.Sender(signerType, txn)
+		_, isPresent := node.CurrentStakes[currentSender]
+		toAddress := txn.To()
+		if *toAddress != node.StakingContractAddress { //Not a address aimed at the staking contract.
+			continue
+		}
+		if value > int64(0) { //If value >0 means its a staking deposit transaction
```
mikedoan: let's make a switch case 

need some logic to exact the method call

node/contract.go
```diff
@@ -50,9 +50,21 @@ func (node *Node) CreateStakingWithdrawTransaction(stake int) (*types.Transactio
 		log.Error("Failed to get chain state", "Error", err)
 	}
 	nonce := state.GetNonce(crypto.PubkeyToAddress(DepositContractPriKey.PublicKey))
-	callingFunction := "0x2e1a7d4d"
-	contractData := callingFunction + hex.EncodeToString([]byte(strconv.Itoa(stake)))
-	dataEnc := common.FromHex(contractData)
+	//callingFunction := "0x2e1a7d4d"
```
mikedoan: remove it

node/node.go
alajko: This data is saved in the contract, but chao will explicitly put it in blockchain. Added a comment.

node/contract.go
alajko: Done

node/node.go
alajko: Added a issue: https://github.com/harmony-one/harmony/issues/430

PR #424, opened 2019-02-12
Log connections when -log_conn is given

harmony-ek:
This is for troubleshooting low-level network operations.

@LeoHChen, this should help us diagnose exactly when a connection is established/closed with regard to other events.


harmony-ek:
> Sorry, forget one comment, can you move the connLogger struct to utils?

You got it.  :-)

PR #427, opened 2019-02-12
Add client support feature to get staking info and modify the logic in staking service.

mikedoan:
## Issue

As the beacon chain deploys a smart contract and the staking transaction needs to have the deployed address to be created. For the testing development, we can not manually deploy the beacon chain and wait for its deploying staking contract and query in terminal for the deployed address.

We need to create a service in client support which runs in any node. The client support in beacon will return necessary info for staking including the deployed address.

#### Test Coverage Data
NA

#### Test/Run Logs

NA


alajko:
With respect to your comment on https://github.com/harmony-one/harmony/pull/422. In order to get the deployed contract address of a contract, we need to find the nonce of the address that created it. (Refer: https://solidity.readthedocs.io/en/v0.5.3/introduction-to-smart-contracts.html#index-8) Then we can (re)create the deployed address. Trivially, this is 0. The deployed contract address can just be public.

PR #428, opened 2019-02-12
Fix ping/pong messages did not reach all peers issue

LeoHChen:
## Issue

We use groupsend / pubsub functions to send messages to all peers, but the ping message didn't reach all peers. I have changed the way to send ping message to make sure it reaches all peers.

#### Test Coverage Data

* Before
* After

#### Test/Run Logs

```
test/deploy.sh -P test/configs/oneshard.txt
```
I have verified all the logs to make sure all ping messages are received and pong message is received by all validators.

## TODO

- [x] make sure pong message is received by all peers.


api/service/discovery/service.go
```diff
@@ -74,19 +79,21 @@ func (s *Service) Init() {
...
 }
 
-func (s *Service) pingPeer(peer p2p.Peer) {
+func (s *Service) pingPeer() {
+	tick := time.NewTicker(5 * time.Second)
 	ping := proto_discovery.NewPingMessage(s.host.GetSelfPeer())
 	buffer := ping.ConstructPingMessage()
 	content := host.ConstructP2pMessage(byte(0), buffer)
+
+	for {
```
astralblue: Infinite loop (goroutine leak)?  We need to exit the loop on something….  How about until we receive a matching pong?

LeoHChen: periodically ping is to make sure the public key of consensus is received by everyone in the shard.  It's not a bad idea to stop sending until we received a pong.  I'll add it in later PR.

mikedoan:
I'm not sure if it is the right design/setup but it's working now as you said.
Created this https://github.com/harmony-one/harmony/issues/432 for me to look at later.

mikedoan:
Please submit

api/service/networkinfo/service.go
```diff
@@ -104,50 +101,42 @@ func (s *Service) Run() {
...
 // DoService does network info.
 func (s *Service) DoService() {
-	for {
-		select {
-		case peer, ok := <-s.peerInfo:
-			if !ok {
-				utils.GetLogInstance().Debug("no more peer info", "peer", peer.ID)
-				return
+	for peer := range s.peerInfo {
+		if peer.ID != s.Host.GetP2PHost().ID() && len(peer.ID) > 0 {
```
mikedoan: we should add logic to receive stopChan channel to be stopped here.

api/service/networkinfo/service.go
```diff
@@ -104,50 +101,42 @@ func (s *Service) Run() {
...
+			utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs, "my ID", s.Host.GetP2PHost().ID())
+			s.lock.Lock()
```
mikedoan: why do we need lock here?

test/configs/oneshard.txt
```diff
@@ -0,0 +1,32 @@
+127.0.0.1 9000 leader 0
```
mikedoan: as you introduced new config file.

If this file will be used regularly, can you add a new deploy/test command in README.md; otherwise ignore this message.

PR #429, opened 2019-02-12
DYLD_LIBRARY_PATH → DYLD_FALLBACK_LIBRARY_PATH

harmony-ek:
## Issue

`DYLD_LIBRARY_PATH` has a “look here first before usual places” semantics, in order to let users use an alternative version of a shared library.  On the other side, `DYLD_FALLBACK_LIBRARY_PATH` has a “look here if not found in usual places” semantics, in order to let users specify additional locations where a shared library is found.

In our case, if a binary refers to a common shared library found in usual places, we want to use it; we simply want to specify additional locations where BLS/MCL/OpenSSL libraries are found.  So it is more appropriate to use `DYLD_FALLBACK_LIBRARY_PATH`.

This also fixes a nasty symbol table mismatch, which is the main reason for this change: Certain commonly used shared libraries - such as `libJPEG.dylib`, co-located within the ImageIO framework and expected by the main ImageIO library – are often also found in Homebrew or MacPorts root (`/opt/local/lib` for MacPort).  If `DYLD_LIBRARY_PATH` included `/opt/local/lib` for OpenSSL, many system programs linked against ImageIO will erroneously pick up `libJPEG.dylib` in `/opt/local/lib`, which lacks certain internal symbols that the framework version has.  Using `DYLD_FALLBACK_LIBRARY_PATH` makes `dyld` honor the full path to the framework version of `libJPEG`, which is hardcoded in the main ImageIO framework library.

## Test

Build succeeds on my MacPorts-based installation.  Prior to this change it would fail with the libJPEG errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ export CGO_CFLAGS="-I$GOPATH/src/github.com/harmony-one/bls/include -I$GOPATH/sr
 export CGO_LDFLAGS="-L$GOPATH/src/github.com/harmony-one/bls/lib -L/usr/local/opt/openssl/lib"
 export LD_LIBRARY_PATH=$GOPATH/src/github.com/harmony-one/bls/lib:$GOPATH/src/github.com/harmony-one/mcl/lib:/usr/local/opt/openssl/lib
 export LIBRARY_PATH=$LD_LIBRARY_PATH
-export DYLD_LIBRARY_PATH=$LD_LIBRARY_PATH
+export DYLD_FALLBACK_LIBRARY_PATH=$LD_LIBRARY_PATH
 
 mkdir -p $HOME/<path_of_your_choice>/src/github.com/harmony-one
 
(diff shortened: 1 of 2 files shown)


astralblue:
> LGTM, as long as you have tested on someone else's macbook.

Thanks, tested on @chaosma's MacBook and it works fine :-)

harmony-ek:
> It could be safer to introduce DYLD_LIBRARY_PATH with the same value?

The whole point of this PR is *not* to use `DYLD_LIBRARY_PATH` and use `DYLD_FALLBACK_LIBRARY_PATH` instead :-p

To walk through the problem:

1. OpenSSL lives in MacPorts root `/opt/local` (so I have `OPENSSL_DIR=/opt/local` in my environment).
2. `scripts/setup_bls_build_flags.sh` sees this, and `/opt/local/lib` into `DYLD_LIBRARY_PATH`.
3. `go` runs some internal tool that uses ImageIO framework.
4. ImageIO depends on /System/Library/Frameworks/ImageIO.framework/Versions/A/Resources/libJPEG.dylib`.
5. `dyld` starts searching for `libJPEG.dylib`.
6. `dyld` sees that `DYLD_LIBRARY_PATH` is defined, and tries to find `libJPEG.dylib` in directories listed there, including `/opt/local/lib`, before it tries the location hint (`/System/Library/Frameworks/ImageIO.framework/Versions/A/Resources`) hardcoded in ImageIO.
7. `dyld` finds `/opt/local/lib/libjpeg.dylib`, and loads it instead of the version in the ImageIO framework.
8. The `/opt/local/lib` version lacks certain symbols present in the ImageIO version.
9. But ImageIO expects those symbols.
10. `dyld` fails to resolve those missing symbols, and program execution fails with an abort trap.

This patch changes step 6 and onward, because `dyld` honors the hardcoded location hint and tries to find the library there first (and succeeds), before falling back to `DYLD_FALLBACK_LIBRARY_PATH`.

rlan35:
lgtm. Thanks for the fix.

PR #431, opened 2019-02-12
Fix pong message not reaching everyone

LeoHChen:
## Issue

Pong message needs to send to everyone in the same shard, as a way to consolidate the list of public key and peers. It is needed by our consensus algorithm.

#### Test Coverage Data

* Before
* After

#### Test/Run Logs

Local test on 30 peers in one shard and verified the log.  All validators received the pong message from leader.


LeoHChen:
this PR is actually not needed. merge it with #428

PR #433, opened 2019-02-13
Make drand as a service and make message passing work

rlan35:
go test succeeded.
Running golint...
golint passed.
Running goimports...
goimports passed.
All Checks Passed!!! :-)


node/node_handler.go
```diff
@@ -248,6 +259,13 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block) {
 		node.BroadcastNewBlock(newBlock)
 	}
 	node.AddNewBlock(newBlock)
+
+	// TODO: enable drand only for beacon chain
```
mikedoan: add comments here for your logic for the future review to understand your code & the future refactor

node/node_test.go
```diff
@@ -97,8 +99,10 @@ func TestAddPeers(t *testing.T) {
 		t.Fatalf("newhost failure: %v", err)
 	}
 	consensus := consensus.New(host, "0", []p2p.Peer{leader, validator}, leader)
+	dRand := drand.New(host, "0", []p2p.Peer{leader, validator}, leader, nil)
```
mikedoan: there are many logic in the main right now. eventually we'll clean up the main file like this https://github.com/ethereum/go-ethereum/blob/master/cmd/geth/main.go#L268

i don't know a right place but probably put this one into node package temporarily.

rlan35: ok, will do.

node/node_handler.go
rlan35: sure.

Issue #430, opened 2019-02-12
Switch Case for calling functions of smart contract.


alajko:
**Describe the bug**

Currently the update staking list functionality to deposit or withdraw stake (function `node.UpdateStakingList`)  is based on whether there is value transmitted in the transaction https://github.com/harmony-one/harmony/blob/3e4a8ca33bd58ab1af9d2d5dbec3b24b27c45f33/node/node.go#L614
The update staking list function should actually be based on a `switch` `case` where the different cases correspond to the deposit, withdraw or other function.signatures of the Staking Contract.

**To Reproduce**

Its in the code.

**Expected behavior**

switch case to handle calling deposit, withdraw or other functions in StakingContract.

alajko:
This is the fix: https://github.com/harmony-one/harmony/pull/437

PR #434, opened 2019-02-13
Updating Staking and Testing Staking and Faucet

alajko:
--------- Funding addresses for Faucet Contract Call ---------

Balances before call of faucet contract
contract balance:
7000000000000000000
user address balance
100037
staking address balance
720000

--------- Funding addresses for Faucet Contract Call DONE ---------

--------- Now Setting up Faucet Contract Call ---------

Balances AFTER CALL of faucet contract
contract balance:
6500000000000000000
user address balance
500000000000100037
staking address balance
720000

--------- Faucet Contract Call DONE ---------

--------- ************************** ---------

--------- Now Setting up Staking Contract ---------

Before Staking Balances
user address balance
500000000000100037
The balances for 2 more users:
200037
300037
stake contract balance :
0
stake address balance :
720000

--------- Staking Contract added to txns ---------

-- Now Staking with stake: 100000 --

After Staking Balances (should be less by 100000)
user address balance
500000000000000037
The balances for 2 more users:
100037
200037
faucet contract balance (unchanged):
6500000000000000000
stake contract balance :
300000
stake address balance :
720000

--------- Now Setting up Withdrawing Stakes ---------

-- Withdrawing Stake by amount: 5000 --

Withdraw Staking Balances (should be up by 5000)
500000000000005037
105037
205037
faucet contract balance (unchanged):
6500000000000000000
stake contract balance (should downup by 5000)
285000
stake address balance :
720000


alajko:
- [ ] Change `withdraw` and `deposit` call in previous PRs.
- [ ] Resolve https://github.com/harmony-one/harmony/issues/430 now that staking is tested.

test/stake/main.go
```diff
@@ -0,0 +1,325 @@
+package main
```
rlan35: Could you fit this into the single test/chain/main.go? These kind of code actually shouldn't exist in master repo, it's more like experimental code. So let's minimize the footprint of them.

contracts/DepositContract.sol
```diff
@@ -1,62 +1,27 @@
 pragma solidity >=0.4.22;
 
 contract DepositContract {
-    mapping(address => uint) private Stakes;
-    mapping(address => uint) private Indices;
```
rlan35: Can you explain the reason for these big change on the contract?

test/stake/main.go
alajko: Yes, I agree. I have it as my todo.

contracts/DepositContract.sol
alajko: Those extra data structures were added to enable us to count number of stakers and order in which they staked (e,g `accountList`). This functionality is not required currently. We have this information at block level and keep a running memory of staked accounts. Removed it from the contract to simplify it.

test/stake/main.go
alajko: Done.

contracts/DepositContract.sol
rlan35: It doesn't hurt to keep the information in smart contract. But I won't block you now.

alajko: Thanks have a issue for myself, if it ever becomes very relevant again: https://github.com/harmony-one/harmony/issues/438

PR #435, opened 2019-02-13
send staking transaction to beacon chain

mikedoan:
## Issue

send staking transaction to beacon chain. Just follow discovery example and it needs help to verify the logic. it seems to be wrong.

### DIFF
--- a/api/proto/common.go
+++ b/api/proto/common.go
@@ -32,6 +32,7 @@ const (
 	Client
 	Identity
 	DRand
+	Staking
 	// TODO: add more types
 )
 
@@ -94,6 +95,13 @@ func ConstructConsensusMessage(payload []byte) []byte {
 // ConstructDRandMessage creates a message with the payload and returns as byte array.
 func ConstructDRandMessage(payload []byte) []byte {
 	byteBuffer := bytes.NewBuffer([]byte{byte(DRand)})
+	byteBuffer.Write(payload)
+	return byteBuffer.Bytes()
+}
+
+// ConstructStakingMessage creates a message with the payload and returns as byte array.
+func ConstructStakingMessage(payload []byte) []byte {
+	byteBuffer := bytes.NewBuffer([]byte{byte(Staking)})
 	byteBuffer.Write(payload)
 	return byteBuffer.Bytes()
 }
--- a/api/service/staking/service.go
+++ b/api/service/staking/service.go
@@ -3,11 +3,13 @@ package staking
 import (
 	"crypto/ecdsa"
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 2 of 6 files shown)


api/service/staking/service.go
```diff
@@ -57,7 +76,7 @@ func (s *Service) Run() {
 		for {
 			select {
 			case peer := <-s.peerChan:
```
LeoHChen: the input channel is not correct I think. what's the proper way to trigger the staking service?

api/service/staking/service.go
```diff
@@ -57,7 +76,7 @@ func (s *Service) Run() {
...
-				utils.GetLogInstance().Info("Running role conversion")
+				utils.GetLogInstance().Info("Running staking service")
 				// TODO: Write some logic here.
 				s.DoService(peer)
```
LeoHChen: this one has to be a go routine and you may start it in the New function.

api/service/staking/service.go
```diff
@@ -71,8 +90,49 @@ func (s *Service) Run() {
 func (s *Service) DoService(peer p2p.Peer) {
```
LeoHChen: again, what's the proper mechanism to start sending the stake transaction?

api/service/staking/service.go
```diff
@@ -3,86 +3,148 @@ package staking
...
 	proto "github.com/harmony-one/harmony/api/client/service/proto"
 	"github.com/harmony-one/harmony/api/proto/message"
+	"github.com/harmony-one/harmony/core"
 	"github.com/harmony-one/harmony/core/types"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/p2p"
 )
 
+const (
+	// WaitTime is the delay time for resending staking transaction if the previous transaction did not get approved.
```
chaosma: in what condition the staking transaction will not get approved?

mikedoan: staking transaction. in this PR, the new node will try forever to send staking transaction. I would add the logic of the newnode exitting after X attempts.

api/proto/common.go
```diff
@@ -97,3 +98,10 @@ func ConstructDRandMessage(payload []byte) []byte {
 	byteBuffer.Write(payload)
 	return byteBuffer.Bytes()
 }
+
+// ConstructStakingMessage creates a message with the payload and returns as byte array.
+func ConstructStakingMessage(payload []byte) []byte {
```
LeoHChen: the raw message buffer has to be wrapped in constructp2pmessage function call to add the 5 bytes header. otherwise, the receiver won't be able to interpret the message.  please fix it.

mikedoan:
I'm just submitting this because this constant was 1 before. if it is still breaking it should be not because of this PR

api/proto/common.go
mikedoan: Forgot it. Thanks. Fixed

PR #436, opened 2019-02-13
propose using *Peer instead of Peer

mikedoan:
## Issue

Replace p2p.Peer by *p2p.Peer

## Problem

1/ Avoid allocation of Peer.
2/ There is no consistency among different places. at some place, we used *Peer, and some we used Peer


harmony-ek:
Let's defer this.  p2p.Peer is to be thinned (`IP`, `Port`, and `Addrs` removed; `PubKey` and `ValidatorID` factored out to a node-level data structure) down into a peer ID.

PR #437, opened 2019-02-13
Update Binary and Staking Calls based on new staking contract

alajko:
## Issue

With updated contract https://github.com/harmony-one/harmony/pull/434 , we need to fix the binary, deposit and withdraw calls. Will also fix https://github.com/harmony-one/harmony/issues/430.

## TODO

- [x] https://github.com/harmony-one/harmony/issues/430


mikedoan:
For me, i don't approve this PR - the logic is not very much - we should be a systematic way rather than compile, copy & paste. You can check in but i will clean it up later

node/node.go
```diff
@@ -631,27 +634,27 @@ func (node *Node) UpdateStakingList(block *types.Block) error {
...
 			continue
 		}
-		//This should be based on a switch case on function signature.
-		//TODO (ak) https://github.com/harmony-one/harmony/issues/430
-		if value > int64(0) { //If value >0 means its a staking deposit transaction
+		currentSender, _ := types.Sender(signerType, txn)
+		_, isPresent := node.CurrentStakes[currentSender]
+		data := txn.Data()
+		switch funcSignature := decodeFuncSign(data); funcSignature {
+		case "0xd0e30db0": //deposit
```
mikedoan: constant.

I made easy for you to check in the code a few times but i could not let it happen again.

node/node.go
```diff
@@ -631,27 +634,27 @@ func (node *Node) UpdateStakingList(block *types.Block) error {
...
+				//This means the node has increased its stake.
 				node.CurrentStakes[currentSender] += value
 			} else {
+				//This means its a new node that is staking the first time.
 				node.CurrentStakes[currentSender] = value
 			}
-		} else { //This means node has withdrawn stake.
-			getData := txn.Data()
-			value := decodeStakeCall(getData) //Value being withdrawn
+		case "0x2e1a7d4d": //withdaw
```
mikedoan: constant - please comment where did you get the constant.

node/contract.go
```diff
@@ -20,8 +20,7 @@ const (
...
...631905090565b60008060003373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002054821115156102c757816000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020600082825403925050819055503373ffffffffffffffffffffffffffffffffffffffff166108fc839081150290604051600060405180830381858888f19350505050158015610280573d6000803e3d6000fd5b506000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1681526020019081526020016000205490506102c8565b5b919050565b600160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1681565b60008060003373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002054905090565b6000346000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020600082825401925050819055506000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1681526020019081526020016000205490509056fea165627a7a723058204acf95662eb95006df1e0b8ba32316211039c7872bc6eb99d12689c1624143d80029"
```
mikedoan: you can not keep change like this. where is the binary? how do get it? need to have comment about this.

mikedoan:
@alajko  you need to learn about rebase @LeoHChen told the team. i saw that practice many times but i could not let it happen again.

node/node.go
```diff
@@ -672,6 +677,12 @@ func decodeStakeCall(getData []byte) int64 {
 	value.SetBytes(getData[4:]) //Escape the method call.
 	return value.Int64()
 }
+
+func decodeFuncSign(data []byte) string {
+	funcSign := hexutil.Encode(data[:4])
```
mikedoan: can you please show me the reference of this logic?

node/node_test.go
```diff
@@ -269,7 +269,7 @@ func TestUpdateStakingWithdrawal(t *testing.T) {
 	if !ok {
 		t.Error("The correct address was not present")
 	}
-	if value != 10 {
+	if value != int64(1000) {
```
mikedoan: why is it? constant?

node/contract.go
alajko: @mikedoan Binaries are generated from the contract. The contract is stored in `contracts/`. This is pretty standard.

node/node.go
alajko: This is based on a smart contract call, first 4 bytes are function call and later bytes are the params. Please refer to how `data` is generated for smart contracts in ethereum.

node/node_test.go
alajko: Please read the test carefully. It withdaws a stake of `10` from a current stake of `1010`, and thus checks whether `1000` was left.

node/contract.go
alajko: Specifically `contracts/DepositContract.sol`. You were tagged on the PR https://github.com/harmony-one/harmony/pull/434

alajko:
> For me, i don't approve this PR - the logic is not very much - we should be a systematic way rather than compile, copy & paste. You can check in but i will clean it up later

@mikedoan  What does "logic is not very much" mean. There are clear switch cases for the function flow, both `deposit` and `withdraw` functionality have tests, the contract which generates the binary is in its proper place in `contracts` folder. Eager to hear how we can be more systematic.

node/node.go
mikedoan: that's why i need you to give the reference? why 4? you should have put 4 as a constant at the top and give the reason/reference why 4?

node/node_test.go
mikedoan: okay. this is test file - what i meant we need to write a better code because i saw you did not specify constant in other files.

node/contract.go
mikedoan: this is bad. 1/ everything should be combined in one PR - 2/ you should have written a compile code and link to this constant so that everything you changed the .sol code this constant changed automatically

node/node.go
alajko: Done.

node/node.go
alajko: Done.

node/node_test.go
alajko: Added more comments to make this clear.

node/contract.go
alajko: 1. The 2 PRs could be together but it had its own testing mechanism.
2. There is currently no setup to compile and update constants. I am not sure we need one either e.g. We did not do that procedure with `FaucetContract` either.  However, such a setup will be useful in the future. It was not in the scope of this PR to build that set up.

alajko:
> @alajko you need to learn about rebase @LeoHChen told the team. i saw that practice many times but i could not let it happen again.

Some cons of rebase to consider before accepting it as a workflow.
(From:  https://www.atlassian.com/git/articles/git-team-workflows-merge-or-rebase)

- Rebasing doesn't play well with pull requests, because you can't see what minor changes someone made if they rebased (incidentally, the consensus inside the Bitbucket Server development team is to never rebase during a pull request).
- Rebasing can be dangerous! Rewriting history of shared branches is prone to team work breakage. This can be mitigated by doing the rebase/squash on a copy of the feature branch, but rebase carries the implication that competence and carefulness must be employed.
- It's more work: Using rebase to keep your feature branch updated requires that you resolve similar conflicts again and again. Yes, you can reuse recorded resolutions (rerere) sometimes, but merges win here: Just solve the conflicts one time, and you're set.
- Another side effect of rebasing with remote branches is that you need to force push at some point. The biggest problem we've seen at Atlassian is that people force push – which is fine – but haven't set git push.default. This results in updates to all branches having the same name, both locally and remotely, and that is dreadful to deal with.

node/node.go
```diff
@@ -116,6 +118,13 @@ type syncConfig struct {
 	client    *downloader.Client
 }
 
+//constants related to staking
+const (
+	depositFuncSignature  = "0xd0e30db0"
+	withdrawFuncSignature = "0x2e1a7d4d"
+	funcSingatureBytes    = 4
```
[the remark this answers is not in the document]
alajko: Discussed Offline. The comments explained the right way to do it: https://github.com/harmony-one/harmony/blob/e3442b78b66f64c903976976f34085c7909e2fb7/node/node.go#L314

PR #442, opened 2019-02-13
Add channel to pass pRnd from DRG to consensus; Make consensus engine a separate package to break import cycle.

rlan35:
2 shards, 78 consensus, 1312 total TPS, 11 nodes 127.0.0.1, 1, 719.096

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -225,18 +225,12 @@ func (bc *BlockChain) ValidateNewBlock(block *types.Block, address common.Addres
 
 // IsEpochBlock returns whether this block is the first block of an epoch.
 func IsEpochBlock(block *types.Block) bool {
-	if block.NumberU64()%BlocksPerEpoch == 0 {
-		return true
-	}
-	return false
+	return block.NumberU64()%BlocksPerEpoch == 0
 }
 
 // IsEpochLastBlock returns whether this block is the last block of an epoch.
 func IsEpochLastBlock(block *types.Block) bool {
-	if block.NumberU64()%BlocksPerEpoch == BlocksPerEpoch-1 {
-		return true
-	}
-	return false
+	return block.NumberU64()%BlocksPerEpoch == BlocksPerEpoch-1
 }
 
 func (bc *BlockChain) getProcInterrupt() bool {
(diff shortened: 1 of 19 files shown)


core/blockchain.go
```diff
@@ -223,6 +223,22 @@ func (bc *BlockChain) ValidateNewBlock(block *types.Block, address common.Addres
 	return nil
 }
 
+// IsEpochBlock returns whether this block is the first block of an epoch.
```
chaosma: I have a similar function  CheckEpochBlock in core/resharding.go. So we don't need to duplicate it

core/blockchain.go
```diff
@@ -223,6 +223,22 @@ func (bc *BlockChain) ValidateNewBlock(block *types.Block, address common.Addres
...
+func IsEpochBlock(block *types.Block) bool {
+	if block.NumberU64()%BlocksPerEpoch == 0 {
+		return true
+	}
+	return false
+}
+
+// IsEpochLastBlock returns whether this block is the last block of an epoch.
```
chaosma: maybe move this to core/resharding.go ?

drand/drand.go
```diff
@@ -23,7 +23,8 @@ type DRand struct {
 	bitmap                *bls_cosi.Mask
 	pRand                 *[32]byte
 	rand                  *[32]byte
```
chaosma: are we going to change the number of bytes in future? If so, make it as a variable

core/chain_makers.go
```diff
@@ -251,7 +251,7 @@ func makeHeaderChain(parent *types.Header, n int, engine consensus.Engine, db et
 }
 
 // makeBlockChain creates a deterministic chain of blocks rooted at parent.
-func makeBlockChain(parent *types.Block, n int, engine consensus.Engine, db ethdb.Database, seed int) []*types.Block {
+func makeBlockChain(parent *types.Block, n int, engine consensus_engine.Engine, db ethdb.Database, seed int) []*types.Block {
 	blocks, _ := GenerateChain(params.TestChainConfig, parent, engine, db, n, func(i int, b *BlockGen) {
 		b.SetCoinbase(common.Address{0: byte(seed), 19: byte(i)})
```
chaosma: what's the meaning of 0 and 19 here?

drand/drand.go
rlan35: I will change it to the correct size after adding real vrf.

core/chain_makers.go
rlan35: This is ethereum code. It's probably some default address.

core/blockchain.go
rlan35: I think the checking is better to be on blockchain level. So I switched the usages of yours to mine.

core/blockchain.go
rlan35: I kept both here and switched yours to mine.

Issue #445, opened 2019-02-14
Move host/peer/group discovery into P2P layer


harmony-ek:
## Summary

Move the various discovery logic into the P2P layer, freeing the application layer from having to deal with it (layer violation, complex and brittle code).

## Current Design

Applications currently get involved in host discovery (peer ID → locator resolution), peer discovery (discovery of peers to connect), and group discovery (group ID → (member peer ID)* resolution).

## Problems

Discovery is part of the basic P2P function, so should be present for all applications.  However, most application logics do not need control over these procedures, and as such they should not be involved.

Moreover, having basic discovery logic in userland is a layering violation, leading to complex and brittle code.  Application writer's mistake should not affect basic discovery operation.  Even if all the relevant logic were to be concentrated into one service (such as `networkinfo`), 

## Proposal

Move discovery function into P2P layer (and out of `networkinfo` and `discovery` services).

If some applications require network peer list and/or group peer list, avail such information via query and/or notification mech.

Since group membership is open, the number of nodes receiving group messages may be potentially huge.  For this reason, “get all group members” is scoped out of P2P.  Application-level logic requiring the nominal members of a group (such as the leader and all validators of a shard, no matter how many other client nodes are receiving group messages).

PR #447, opened 2019-02-14
[WIP] P2P automatic host/peer/group discovery

harmony-ek:
## Issue

See #445

# Below are TBD


harmony-ek:
> I also suggest you should try to merge your clean and cosmetic changes before the real functional changes. That's easier for you to break down the patch set.

Good idea, will do this now before I forget :-)

PR #449, opened 2019-02-14
Add real vrf into drand with validator generating vrf and leader verifying them; Also reduce manager_test.go's delay

rlan35:
go test succeeded.
Running golint...
golint passed.
Running goimports...
goimports passed.
All Checks Passed!!! :-)


chaosma:
LGTM

PR #456, opened 2019-02-15
Clean up p2p

harmony-ek:
This is a spin-off of #447, and contains various improvements not directly related to the host/peer/group discovery that #447 is addressing.


harmony-ek:
Since this is a collection of nebulous, potentially unrelated enhancements, you might find it easier to review individual commits.

PR #457, opened 2019-02-15
Check if all generated files are up to date in Travis-CI

harmony-ek:
I forgot to regenerate files yet again (and this time the oversight actually hit master, although it didn't break anything), so with my pointy hat on, here's a belated safeguard that I had promised when I worked on `//go:generate` the first time.

Beyond this commit, Travis-CI will regenerate all generated and checked-in source code to see if any of them has changed.  If so, it means that some of the generated files are outdated, so Travis-CI will fail.  This happens mostly when someone touches mocked interfaces or protobuf source files.


scripts/gogenerate.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
LeoHChen: no comment?  it's hard to maintain later.

scripts/gogenerate.sh
```diff
@@ -0,0 +1,33 @@
...
+esac
+git grep -l '^//go:generate ' | sort -t/ | (
+	gogenerate() {
+		case $# in 0) return;; esac
+		go generate "$@"
+	}
+	unset -v prevdir dir file
+	prevdir=
+	set --
+	while read -r file
```
LeoHChen: please add some comments to elaborate the block.

scripts/install_protoc.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+set -eu
+unset -v opt OPTIND OPTARG version platform destdir
+version=
+platform=
+destdir=/usr/local
+while getopts :V:P:d: opt
```
LeoHChen: please provide a -h option for usage.

scripts/gogenerate.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+unset -v progdir
+case "${0}" in
+*/*) progdir="${0%/*}";;
+*) progdir=.;;
+esac
+git grep -l '^//go:generate ' | sort -t/ | (
```
LeoHChen: pipe through the complicated function makes the code hard to read and debug.  recommend using some temporary file to store the output of "git grep  | sort",

scripts/gogenerate.sh
```diff
@@ -0,0 +1,33 @@
...
+	unset -v prevdir dir file
+	prevdir=
+	set --
+	while read -r file
+	do
+		case "${file}" in
+		*/*) dir="${file%/*}";;
+		*) dir=".";;
+		esac
+		case "${prevdir}" in
```
LeoHChen: I am a bit confused on what's the intention of this chunk of code.  Is there another way to write this code that is more readable?

harmony-ek: Factored the hard part into a separate utility script with comments.  Let me know if it's any better. :-)

scripts/gogenerate.sh
harmony-ek: I agree the three-stage pipe was probably too much.  Split that into two stages in the new utility script.

scripts/gogenerate.sh
harmony-ek: Done (in the utility script).

scripts/gogenerate.sh
harmony-ek: Added some.

scripts/install_protoc.sh
harmony-ek: Done!

PR #462, opened 2019-02-16
genesis block binary creation

chaosma:
add code for converting json file into binary code for genesis block


chaosma:
@mikedoan  Yes, I am also not sure where to put the file. Let's discuss together

PR #467, opened 2019-02-19
Txgen using Libp2p gossip to send TX to shard

LeoHChen:
## Issue

When we integrate libp2p, txgen also has to work with the libp2p based gossiping.

All new nodes started as a beacon chain client, in order to send a message and recv blocks.

#### Test Coverage Data

* Before
* After

#### Test/Run Logs

Test manually, using -libp2p_pd flag, started 3 nodes for beacon chain, and started txgen using libp2p_pd flag as well.  Saw HOORAY was reached for a few times.

## TODO

- [x] fix the DRG using libp2p gossiping

### DIFF
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -133,7 +133,9 @@ func (s *Service) DoService() {
 				}
 				p := p2p.Peer{IP: ip, Port: port, PeerID: peer.ID, Addrs: peer.Addrs}
 				utils.GetLogInstance().Info("Notify peerChan", "peer", p)
-				s.peerChan <- p
+				if s.peerChan != nil {
+					s.peerChan <- p
+				}
 			}
 		case <-s.stopChan:
 			return
--- a/cmd/client/txgen/main.go
+++ b/cmd/client/txgen/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/node"
 	"github.com/harmony-one/harmony/p2p"
+	p2p_host "github.com/harmony-one/harmony/p2p/host"
 	"github.com/harmony-one/harmony/p2p/p2pimpl"
 	peerstore "github.com/libp2p/go-libp2p-peerstore"
 	multiaddr "github.com/multiformats/go-multiaddr"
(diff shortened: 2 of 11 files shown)


consensus/consensus_leader.go
```diff
@@ -31,13 +31,17 @@ var (
 
 // WaitForNewBlock waits for the next new block to run consensus on
 func (consensus *Consensus) WaitForNewBlock(blockChannel chan *types.Block, stopChan chan struct{}, stoppedChan chan struct{}, startChannel chan struct{}) {
+	first := true
```
rlan35: Add a comment for why adding this boolean?

node/node.go
```diff
@@ -828,6 +845,12 @@ func (node *Node) setupForBeaconValidator() {
 		utils.GetLogInstance().Error("create group receiver error", "msg", err)
 		return
 	}
+	node.clientReceiver, err = node.host.GroupReceiver(p2p.GroupIDBeaconClient)
```
rlan35: Duplicated code. This and the above can be combined into a method.

api/service/networkinfo/service.go
```diff
@@ -133,7 +133,9 @@ func (s *Service) DoService() {
 				}
 				p := p2p.Peer{IP: ip, Port: port, PeerID: peer.ID, Addrs: peer.Addrs}
 				utils.GetLogInstance().Info("Notify peerChan", "peer", p)
-				s.peerChan <- p
+				if s.peerChan != nil {
+					s.peerChan <- p
+				}
```
mikedoan: else {
  utils..... Error("some error message because should not be nil")
}

consensus/consensus_leader.go
```diff
@@ -31,13 +31,17 @@ var (
...
 	go func() {
 		defer close(stoppedChan)
 		for {
 			select {
 			default:
-				// got the signal to start consensus
-				_ = <-startChannel
+				if first && startChannel != nil {
```
mikedoan: consolidate this when first = true and startChannel == nil

if you expect startChannel should be not nil then log error. Otherwise

consensus/consensus_leader.go
LeoHChen: will do.

node/node.go
LeoHChen: sure.

api/service/networkinfo/service.go
LeoHChen: it's possible to be nil if there is no need to do peer discovery. and no need to send data to a nil channel. This is just a protection.

consensus/consensus_leader.go
LeoHChen: startChannel != nil is for protection of program stuck.

node/node.go
```diff
@@ -313,6 +319,7 @@ func New(host p2p.Host, consensus *bft.Consensus, db ethdb.Database) *Node {
 
 	if consensus != nil && consensus.IsLeader {
 		node.State = NodeLeader
+		go node.ReceiveClientGroupMessage()
```
mikedoan: try to move this go routine and the go node.ReceiveGroupMessage() in to node_handler.go because they are related to networking

cmd/client/txgen/main.go
```diff
@@ -70,40 +71,64 @@ func main() {
...
+		candidateNode.AddPeer(bcPeer)
+		candidateNode.ContactBeaconChain(*bcPeer)
+		selfPeer := candidateNode.GetSelfPeer()
+		selfPeer.PubKey = candidateNode.PubK
+
+		shardIDLeaderMap = candidateNode.Leaders
+
+		debugPrintShardIDLeaderMap(shardIDLeaderMap)
+	} else {
+		shardIDLeaderMap = make(map[uint32]p2p.Peer)
```
mikedoan: commenet this is the new way

node/node.go
LeoHChen: sure. will look into that.

cmd/client/txgen/main.go
LeoHChen: ok

consensus/consensus_leader.go
```diff
@@ -31,13 +31,24 @@ var (
...
 func (consensus *Consensus) WaitForNewBlock(blockChannel chan *types.Block, stopChan chan struct{}, stoppedChan chan struct{}, startChannel chan struct{}) {
+	// gensis block is the first block to be processed.
+	// But we shouldn't start consensus yet, as we need to wait for all validators
+	// received the leader's pub key which will be propogated via Pong message.
+	// After we started the first consensus, we will go back to normal case to wait
+	// for new blocks.
+	// The signal to start the first consensus right now is the sending of Pong message (SendPongMessage function in node/node_handler.go
+	// but it can be changed to other conditions later
+	first := true
 	go func() {
```
harmony-ek: [Nit] This background function is the entirety of this function; can we convert this to foreground and instead call:
```
go s.consensus.WaitForNewBlock(s.blockChannel, s.stopChan, s.stoppedChan)
```
from `api/service/consensus/service.go`?  Two reasons:

1. Less indentation.
2. The function name sounds like a synchronous, blocking function but it is asynchronous/background.

consensus/consensus_leader.go
```diff
@@ -31,13 +31,24 @@ var (
...
 		defer close(stoppedChan)
 		for {
 			select {
 			default:
-				// got the signal to start consensus
-				_ = <-startChannel
+				if first && startChannel != nil {
+					// got the signal to start consensus
+					_ = <-startChannel
+					first = false
```
harmony-ek: The purpose of this `first` variable is to receive from (block on) `startChannel` only on the first iteration, correct?  A more idiomatic way would be to do away with `first` boolean variable and instead nullify `startChannel`, that is:
```suggestion
					startChannel = nil
```

This way, we wouldn't enter this `if startChannel != nil { … }` block on the second and further iterations.

node/node_handler.go
```diff
@@ -441,7 +460,12 @@ func (node *Node) pongMessageHandler(msgPayload []byte) int {
 	node.Consensus.Leader.PubKey = &bls.PublicKey{}
 	err = node.Consensus.Leader.PubKey.Deserialize(pong.LeaderPubKey)
 	if err != nil {
-		utils.GetLogInstance().Error("Unmarshal Leader PubKey Failed", "error", err)
+		utils.GetLogInstance().Error("Unmarshal Consensus Leader PubKey Failed", "error", err)
+	}
+	node.DRand.Leader.PubKey = &bls.PublicKey{}
```
harmony-ek: Why don't we introduce and use `func (drand *DRand) SetLeaderKey(key *bls.PublicKey)`?  This way, we don't have to flip the private `leader` to public, and if needed, the `SetLeaderKey()` would be a point of sanity check, e.g. resetting an already set leader key could be logged (since it should not happen normally IINM).

consensus/consensus_leader.go
LeoHChen: this is the old code and I don't feel I should touch it in my PR.  My PR is focusing on getting consensus working with libp2p.

node/node_handler.go
LeoHChen: I'll do it in the following PR.

consensus/consensus_leader.go
LeoHChen: different logic here.  startChannel is initialized to nil for the old networking code. Once we have libp2p fully working and tested, I'll clean up this code.

LeoHChen:
merge it for now, as this PR is getting bigger.  Will do a cleanup PR as @harmony-ek suggested.

PR #469, opened 2019-02-19
add resharding unit test 

chaosma:
added resharding unit tests

### DIFF
--- a/core/resharding_test.go
+++ b/core/resharding_test.go
@@ -3,16 +3,47 @@ package core
 import (
 	"fmt"
 	"testing"
-)
 
-func TestFakeGetInitShardState(t *testing.T) {
-	ss := fakeGetInitShardState()
-	for i := range ss {
-		fmt.Printf("ShardID: %v, NodeList: %v\n", ss[i].ShardID, ss[i].NodeList)
-	}
-}
+	"github.com/harmony-one/harmony/core/types"
+	"github.com/stretchr/testify/assert"
+)
 
 func TestFakeNewNodeList(t *testing.T) {
 	nodeList := fakeNewNodeList(42)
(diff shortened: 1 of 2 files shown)


core/resharding_test.go
```diff
@@ -3,16 +3,47 @@ package core
...
-		fmt.Printf("ShardID: %v, NodeList: %v\n", ss[i].ShardID, ss[i].NodeList)
-	}
-}
+	"github.com/harmony-one/harmony/core/types"
+	"github.com/stretchr/testify/assert"
+)
 
 func TestFakeNewNodeList(t *testing.T) {
 	nodeList := fakeNewNodeList(42)
 	fmt.Println("newNodeList: ", nodeList)
```
mikedoan: why is it a test?

core/resharding_test.go
```diff
@@ -3,16 +3,47 @@ package core
...
+	cpList = append(cpList, nodeList...)
+	Shuffle(nodeList)
+	cnt := 0
+	for i := 0; i < 10; i++ {
+		if cpList[i] == nodeList[i] {
+			cnt++
+		}
+	}
+	if cnt == 10 {
+		t.Error("Shuffle list is the same as original list")
```
mikedoan: can you clarify what you are trying to test here

core/resharding_test.go
chaosma: Since it's random, I just by looking at the result, can be removed

core/resharding_test.go
chaosma: When shuffle a list, we have a very low probability that the list stay the same

mikedoan: maybe should do another type of checking

PR #471, opened 2019-02-20
Add a simple POC VDF and integrate into the drand protocol

rlan35:
go test succeeded.
Running golint...
golint passed.
Running goimports...
goimports passed.
Running go generate...
go generate succeeded; all generated files seem up to date.
All Checks Passed!!! :-)


consensus/consensus.go
```diff
@@ -140,6 +143,29 @@ func (consensus *Consensus) UpdateConsensusID(consensusID uint32) {
 	}
 }
 
+// WaitForNewRandomness listens to the RndChannel to receive new VDF randomness.
+func (consensus *Consensus) WaitForNewRandomness() {
```
mikedoan: can we add randomness as a service? should be decoupled from calling from consensus service?

api/service/consensus/service.go
```diff
@@ -25,6 +25,7 @@ func (s *Service) StartService() {
 	s.stopChan = make(chan struct{})
 	s.stoppedChan = make(chan struct{})
 	s.consensus.WaitForNewBlock(s.blockChannel, s.stopChan, s.stoppedChan, s.startChan)
+	s.consensus.WaitForNewRandomness()
```
mikedoan: should we decouple randomness as a separated service?

consensus/consensus_leader.go
```diff
@@ -67,6 +64,15 @@ func (consensus *Consensus) WaitForNewBlock(blockChannel chan *types.Block, stop
 					// TODO: check validity of pRnd
 					newBlock.AddRandPreimage(binary.BigEndian.Uint32(pRnd))
 				}
+				rnd, blockHash, err := consensus.GetNextRnd()
+				if err != nil {
```
chaosma: err == nil ?

crypto/vdf/vdf.go
```diff
@@ -0,0 +1,54 @@
...
+func (vdf *VDF) GetOutputChannel() chan [32]byte {
+	return vdf.outputChan
+}
+
+// Execute runs the VDF until it's finished and put the result into output channel.
+func (vdf *VDF) Execute() {
+	vdf.finished = false
+	tempResult := vdf.input
+	for i := 0; i < vdf.difficulty; i++ {
+		tempResult = sha3.Sum256(tempResult[:])
```
chaosma: how long it takes to run one iteration of sha3 sum?

crypto/vdf/vdf.go
```diff
@@ -0,0 +1,54 @@
...
+// GetOutputChannel returns the vdf output channel.
+func (vdf *VDF) GetOutputChannel() chan [32]byte {
+	return vdf.outputChan
+}
+
+// Execute runs the VDF until it's finished and put the result into output channel.
+func (vdf *VDF) Execute() {
+	vdf.finished = false
+	tempResult := vdf.input
+	for i := 0; i < vdf.difficulty; i++ {
```
chaosma: will this use all of the cpu?

consensus/consensus.go
rlan35: Randomness is already a separate service. But consensus needs to wait for output from randomness service

consensus/consensus_leader.go
rlan35: opps, Done.

crypto/vdf/vdf.go
rlan35: It will use one core I think.

PR #476, opened 2019-02-20
Change rnds in block to 32 bytes

rlan35:


core/blockchain.go
```diff
@@ -1685,19 +1685,19 @@ func (bc *BlockChain) GetShardStateByHash(hash common.Hash) types.ShardState {
 }
 
 // GetRandSeedByNumber retrieves the rand seed given the block number, return 0 if not exist
-func (bc *BlockChain) GetRandSeedByNumber(number uint64) uint32 {
+func (bc *BlockChain) GetRandSeedByNumber(number uint64) [32]byte {
```
chaosma: why we need 32 byte data instead of 4 byte

PR #477, opened 2019-02-21
set back default threshold to 1

mikedoan:


mikedoan:
I'm just submitting this because this constant was 1 before. if it is still breaking it should be not because of this PR

PR #480, opened 2019-02-21
add beaconchain database support for normal shard nodes

chaosma:
all tests passed

### DIFF
--- a/cmd/harmony.go
+++ b/cmd/harmony.go
@@ -50,9 +50,14 @@ func attackDetermination(attackedMode int) bool {
 	return false
 }
 
-// InitLDBDatabase initializes a LDBDatabase.
-func InitLDBDatabase(ip string, port string, freshDB bool) (*ethdb.LDBDatabase, error) {
-	dbFileName := fmt.Sprintf("./db/harmony_%s_%s", ip, port)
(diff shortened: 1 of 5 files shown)


node/node.go
```diff
@@ -701,6 +701,21 @@ func (node *Node) setupForClientNode() {
 	node.serviceManager.RegisterService(service_manager.NetworkInfo, networkinfo.New(node.host, p2p.GroupIDBeacon, chanPeer))
 }
 
+// AddBeaconChainDatabase added database support for beaconchain blocks on normal sharding nodes (not BeaconChain node)
```
rlan35: s/added/adds

cmd/harmony.go
```diff
@@ -250,6 +255,11 @@ func main() {
 	currentNode.Role = node.NewNode
 
 	if *isBeacon {
+		var beacondb ethdb.Database
+		if *dbSupported {
+			beacondb, _ = InitLDBDatabase(*ip, *port, *freshDB, true)
```
mikedoan: this is not correct. this logic will happen when it is not beaconchain

chaosma: oops, fixed

node/node.go
chaosma: fixed it

PR #487, opened 2019-02-21
add logic to receive staking txns

mikedoan:
## Issue

add logic to receive staking txns

### DIFF
--- a/api/proto/common.go
+++ b/api/proto/common.go
@@ -85,6 +85,14 @@ func GetDRandMessagePayload(message []byte) ([]byte, error) {
 	return message[MessageCategoryBytes:], nil
 }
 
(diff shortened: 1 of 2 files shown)


rlan35:
Why do you need this extra logic to add staking tx to pending list? Using the normal Node/Transaction handler can accept any transactions.

mikedoan:
> Why do you need this extra logic to add staking tx to pending list? Using the normal Node/Transaction handler can accept any transactions.

That would work. The main reason is that this PR is using message.Message instead of the type of message using Node/Transaction.

Another benefit is that I would be able to test if the staking transaction is sent from new node by writing some logging - using the normal approach wouldn't let us know that-

PR #497, opened 2019-02-23
Node package code reorg

rlan35:


rlan35:
@LeoHChen  It's required to commit go.sum https://github.com/golang/go/wiki/Modules#should-i-commit-my-gosum-file-as-well-as-my-gomod-file

PR #502, opened 2019-02-26
add beaconchain node state syncing

chaosma:
beaconchain syncing for non-beaconchain nodes


core/types/gen_tx_json.go
```diff
@@ -7,8 +7,8 @@ import (
 	"errors"
 	"math/big"
 
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/harmony-one/harmony/vendor/github.com/ethereum/go-ethereum/common"
```
daywednes: can you please confirm that we have to specify the code from vendor? @chaosma is the right import @astralblue ?

node/node.go
```diff
@@ -104,6 +104,7 @@ type Node struct {
 	Consensus              *consensus.Consensus // Consensus object containing all Consensus related data (e.g. committee members, signatures, commits)
 	BlockChannel           chan *types.Block    // The channel to send newly proposed blocks
 	ConfirmedBlockChannel  chan *types.Block    // The channel to send confirmed blocks
+	BeaconBlockChannel     chan *types.Block    // The channel to send beacon blocks for non-beaconchain nodes
```
daywednes: why do we need this? We do not propose new beacon blocks for regular nodes

node/node.go
```diff
@@ -123,15 +124,17 @@ type Node struct {
 	State      State      // State of the Node
 	stateMutex sync.Mutex // mutex for change node state
 
-	TxPool *core.TxPool
-	Worker *worker.Worker
+	TxPool       *core.TxPool
+	Worker       *worker.Worker
+	BeaconWorker *worker.Worker // worker for beacon chain
```
daywednes: I believe we don't need it

node/node.go
chaosma: this is the channel to send the received block to beacon syncing program; it's not the same as the channel in consensus process

node/node.go
chaosma: Let me check if we can use one worker for multiple chains

rlan35: I think it's good to use two workers separately.

core/types/gen_tx_json.go
chaosma: fixed

PR #506, opened 2019-02-26
[cleanup] remove libp2p unicast support

LeoHChen:

## Issue

Since we have enabled libp2p/pubsub based message communication to reach consensus, there is no need to keep the unicast support of libp2p in our code base.

#### Test Coverage Data

* Before
* After


cmd/harmony.go
```diff
@@ -294,16 +237,8 @@ func main() {
 	consensus.OnConsensusDone = currentNode.PostConsensusProcessing
 	currentNode.State = node.NodeWaitToJoin
 
-	if !*libp2pPD {
-		if consensus.IsLeader {
-			currentNode.State = node.NodeLeader
```
mikedoan: do we still need this? may you elaborate?

mikedoan: address it and check in the code asap

LeoHChen: libp2ppd is just a flag to enable libp2p based peer discovery and gossiping. Since we enable it by default, there is no need to have a flag anymore.

mikedoan: node.State is not related to libp2p

LeoHChen:
1 shards, 66 consensus, 35 total TPS, 165 nodes
54.189.78.117, 66, 34.819

PR #507, opened 2019-02-26
setup for having staking service working

mikedoan:
## Issue

setup for having staking service working

new node will run with ip 127.0.0.1 and port 9100

### DIFF
--- a/cmd/harmony.go
+++ b/cmd/harmony.go
@@ -15,7 +15,6 @@ import (
 	"github.com/harmony-one/harmony/drand"
 	"github.com/harmony-one/harmony/internal/profiler"
 	"github.com/harmony-one/harmony/internal/utils"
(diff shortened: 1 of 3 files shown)


cmd/harmony.go
```diff
@@ -253,6 +202,7 @@ func main() {
 	currentNode := node.New(host, consensus, ldb)
 	currentNode.Consensus.OfflinePeers = currentNode.OfflinePeers
 	currentNode.Role = node.NewNode
+	currentNode.AccountKey = contract_constants.GenesisBeaconAccountPriKey
```
rlan35: This should be the new node's own key.

mikedoan: Resolved.

mikedoan:
Please review again

PR #512, opened 2019-02-28
Add faucet call before staking tx

rlan35:
Tested locally and the staking transaction is able to be included in the block


test/deploy_newnode.sh
```diff
@@ -115,7 +115,7 @@ mkdir -p $log_folder
 LOG_FILE=$log_folder/r.log
 
 HMY_OPT=
-HMY_OPT2=
+HMY_OPT2="-bootnodes /ip4/127.0.0.1/tcp/19876/p2p/QmWGSpah5QiBB1aKHKoJpM4VqJxcefc8kVxEbzXD59M8kP"
```
mikedoan: how do you generate this?

api/service/staking/service.go
```diff
@@ -30,6 +30,8 @@ const (
 	WaitTime = 5 * time.Second
 	// StakingContractAddress is the staking deployed contract address
 	StakingContractAddress = "TODO(minhdoan): Create a PR to generate staking contract address"
+	// StakingAmount is the amount of stake to put
+	StakingAmount = 10 ^ 18
```
mikedoan: this is xor operation. 10^18 .= 24.

if you want 10**18, you should use uint64(10e18)

node/node_handler.go
```diff
@@ -201,14 +201,23 @@ func (node *Node) messageHandler(content []byte, sender string) {
...
-func (node *Node) addStakingTxnIntoPendingTxns(msgPayload []byte) {
+func (node *Node) processStakingMessage(msgPayload []byte) {
 	msg := &message.Message{}
 	err := pb.Unmarshal(msgPayload, msg)
 	if err == nil {
 		stakingRequest := msg.GetStaking()
 		txs := types.Transactions{}
 		if err = rlp.DecodeBytes(stakingRequest.Transaction, &txs); err == nil {
-			utils.GetLogInstance().Error("Successfully added staking transaction to pending list.")
+			for _, tx := range txs {
```
mikedoan: we prepared sending money in the genesis https://github.com/harmony-one/harmony/blob/master/node/contract.go#L134

i forgot that yesterday. probably let me change the logic of taking account in harmony from one of the fake accounts which got tokens minted in genesis (constants.go)

test/deploy_newnode.sh
LeoHChen: this shouldn't be hardcoded.  Please use $BN_MA.

rlan35: This is copied from the deploy.sh beacon chain output. I added a comment for it. And deleted my copy.

node/node_handler.go
rlan35: Don't do it. I've already making the flow working.

api/service/staking/service.go
rlan35: Already changed to simply 10. And the flow is now working.

node/node_handler.go
mikedoan: but eventually we won't do it. it may be working now but eventually for the test-net and devnet we will have to do that way again. We already have the staking preparation here https://github.com/harmony-one/harmony/blob/master/node/contract.go#L137

mikedoan: your code won't work for testnet - we will have to switch back to this way again

mikedoan: I wrote this PR for you https://github.com/harmony-one/harmony/pull/514/

can you please enable this logic back and test with this PR?

node/node.go
```diff
@@ -257,12 +275,16 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, db ethdb.Database) *N
...
-		if node.Role == BeaconLeader || node.Role == BeaconValidator {
-			node.CurrentStakes = make(map[common.Address]int64)
-		}
+
 		utils.GetLogInstance().Debug("Created Genesis Block", "blockHash", chain.GetBlockByNumber(0).Hash().Hex())
 		node.Consensus.ConsensusBlock = make(chan *consensus.BFTBlockInfo)
 		node.Consensus.VerifiedNewBlock = make(chan *types.Block)
+
+		// Setup one time smart contracts
+		node.AddFaucetContractToPendingTransactions()
```
mikedoan: why did we move this logic in node_genesis and put it back here?

rlan35: This is not part of genesis. This is added as new transaction which goes into second block.

node/node_handler.go
rlan35: Your way is basically batch calling faucet while mine is dynamically calling faucet. There is no major pros or cons on both approach. Mine is just easier for testing and development. And I don't think your approach is significantly better than mine for testnet.

PR #513, opened 2019-02-28
Add beacon peer for every node

LeoHChen:
## Issue

The new node needs to get a list of beacon peers for state syncing on beacon chain blockchain.

#### Test Coverage Data

Added test program for the AddBeaconPeer function.

#### Test/Run Logs

+ results='1 shards, 166 consensus, 269 total TPS, 10 nodes, 11 total nodes, 8 total signatures
127.0.0.1, 166, 269.298

### DIFF
--- a/node/service_setup.go
+++ b/node/service_setup.go
@@ -25,7 +25,7 @@ func (node *Node) setupForShardLeader() {
 	nodeConfig, chanPeer := node.initNodeConfiguration()
 
 	// Register peer discovery service. No need to do staking for beacon chain node.
-	node.serviceManager.RegisterService(service.PeerDiscovery, discovery.New(node.host, nodeConfig, chanPeer))
+	node.serviceManager.RegisterService(service.PeerDiscovery, discovery.New(node.host, nodeConfig, chanPeer, node.AddBeaconPeer))
 	// Register networkinfo service. "0" is the beacon shard ID
 	node.serviceManager.RegisterService(service.NetworkInfo, networkinfo.New(node.host, p2p.GroupIDBeacon, chanPeer))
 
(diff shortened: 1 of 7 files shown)


node/service_setup.go
```diff
@@ -54,7 +54,7 @@ func (node *Node) setupForBeaconLeader() {
 	nodeConfig, chanPeer := node.initBeaconNodeConfiguration()
 
 	// Register peer discovery service. No need to do staking for beacon chain node.
-	node.serviceManager.RegisterService(service.PeerDiscovery, discovery.New(node.host, nodeConfig, chanPeer))
+	node.serviceManager.RegisterService(service.PeerDiscovery, discovery.New(node.host, nodeConfig, chanPeer, node.AddBeaconPeer))
```
chaosma: BeaconLeader doesn't need add beacon peers

node/service_setup.go
```diff
@@ -71,7 +71,7 @@ func (node *Node) setupForBeaconValidator() {
...
```
chaosma: BeaconValidator doesn't need add beacon peers

node/service_setup.go
```diff
@@ -93,7 +93,7 @@ func (node *Node) setupForClientNode() {
 	nodeConfig, chanPeer := node.initNodeConfiguration()
 
 	// Register peer discovery service.
-	node.serviceManager.RegisterService(service.PeerDiscovery, discovery.New(node.host, nodeConfig, chanPeer))
+	node.serviceManager.RegisterService(service.PeerDiscovery, discovery.New(node.host, nodeConfig, chanPeer, node.AddBeaconPeer))
```
chaosma: ClientNode doesn't need doesn't need add beacon peers, since we already have function to receive beacon block. But maybe we will change this later for client node.

node/service_setup.go
LeoHChen: this is a generic peer discovery service.  It is needed by everyone. I checked if the node is a beacon node or not within service.

node/service_setup.go
LeoHChen: I'll passing in a nil pointer here.

node/service_setup.go
LeoHChen: will pass in a nil pointer here.

PR #515, opened 2019-02-28
send a ping message before the timer tick

LeoHChen:
this shall speed up the bootstrap process by sending a ping message
immediately entering the goroutine

## Issue

The original service sends the 1st ping message after the timer tick (5sec).
This patch speeds up the process by sending one ping message immediately.

#### Test/Run Logs

+ results='1 shards, 30 consensus, 62 total TPS, 10 nodes, 11 total nodes, 8 total signatures
127.0.0.1, 30, 62.2404

### DIFF
--- a/api/service/discovery/service.go
+++ b/api/service/discovery/service.go
@@ -81,6 +81,7 @@ func (s *Service) contactP2pPeers() {
 	pingMsg.Node.Role = proto_node.ClientRole
 	clientMsgBuf := host.ConstructP2pMessage(byte(0), pingMsg.ConstructPingMessage())
 
+	s.sentPingMessage(p2p.GroupIDBeacon, regMsgBuf, clientMsgBuf)
 	for {
 		select {
 		case peer, ok := <-s.peerChan:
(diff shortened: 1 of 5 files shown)


api/service/discovery/service.go
```diff
@@ -104,7 +105,6 @@ func (s *Service) contactP2pPeers() {
 		case action := <-s.actionChan:
 			s.config.Actions[action.Name] = action.Action
 		case <-tick.C:
-			var err error
 			for g, a := range s.config.Actions {
 				if a == p2p.ActionPause {
 					// Recived Pause Message, to reduce the frequency of ping message to every 1 minute
```
mikedoan: if you remove the logic below, this means there won't be any logic for <-tick.C; why do we keep it and why do we change to tick to a new ticker?

harmony-ek: There *is* logic though: The `sentPingMessage()` call.

mikedoan: got it

LeoHChen:
Tested on AWS.

1 shards, 26 consensus, 42 total TPS, 159 nodes
54.184.111.147, 26, 41.9821

PR #516, opened 2019-03-01
Add logic for resharding

mikedoan:
## Issue

Add logic for resharding. I have to stop in the middle because i need to modify resharding calculation logic which currently does not calculate node role (leader or validator) but instead it only calculates new committees.

### DIFF
--- a/api/service/resharding/service.go
+++ b/api/service/resharding/service.go
@@ -1,8 +1,16 @@
-package rconversion
+package resharding
 
 import (
+	"time"
+
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
+	"github.com/harmony-one/harmony/core"
 	"github.com/harmony-one/harmony/internal/utils"
+)
+
+// Constants for resharding service.
+const (
+	ReshardingCheckTime = time.Second
 )
 
 // Service is the role conversion service.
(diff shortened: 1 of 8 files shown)


LeoHChen:
still fail travis checker.

harmony-ek:
> still fail travis checker.

Yep.  Two issues:

1. goimports ended up tidying one file; to fix, run scripts/goimports.sh on your side to format the file,  then commit/push it.
2. For some reason, gomock isn't installed on Travis, so mockgen is failing (and mock freshness check is skipped as a result).  I'll fix this.

node/service_setup.go
```diff
@@ -42,6 +35,16 @@ func (node *Node) setupForShardLeader() {
 	node.serviceManager.RegisterService(service.Randomness, randomness.New(node.DRand))
 }
 
+func (node *Node) stopShardLeaderServices() {
```
rlan35: These stop funcs seems pretty verbose. Is there a way to loop through the services and and stop signal iteratively?

mikedoan: yep. this is better. was thinking about this.

api/service/resharding/service.go
```diff
@@ -49,6 +58,18 @@ func (s *Service) Run(stopChan chan struct{}, stoppedChan chan struct{}) {
 
 // DoService does role conversion.
 func (s *Service) DoService() {
+	tick := time.NewTicker(ReshardingCheckTime)
```
chaosma: Resharding is block number based instead of time based. Do we really need a service?

mikedoan: Regarding my PR, the resharding service keeps checking the new  status from beacon blockchain - currently i designs it checking every single second.

Is there any suggestionof not having a service?

mikedoan: FYI, this is still an ongoing work

PR #519, opened 2019-03-01
Support multiple newnodes

rlan35:


cmd/harmony.go
```diff
@@ -106,6 +106,7 @@ func main() {
 
 	// isNewNode indicates this node is a new node
 	isNewNode := flag.Bool("is_newnode", false, "true means this node is a new node")
```
LeoHChen: why do you need is_newnode?  by default, every node is a new node if you don't specify is_beacon.  this is a redundant option.

rlan35: @mikedoan can answer this. He introduced this flag.

rlan35: @mikedoan can answer this. He introduced this flag.

PR #522, opened 2019-03-02
Modify shard state calculation

mikedoan:
## Issue

Modify shard state calculation

### DIFF
--- a/core/resharding.go
+++ b/core/resharding.go
@@ -13,6 +13,8 @@ import (
 const (
 	// InitialSeed is the initial random seed, a magic number to answer everything, remove later
 	InitialSeed uint32 = 42
+	// FirstEpoch is the number of the first epoch.
+	FirstEpoch = 0
 )
 
 // ShardingState is data structure hold the sharding state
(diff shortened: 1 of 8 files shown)


core/resharding.go
```diff
@@ -47,38 +49,61 @@ func (ss *ShardingState) assignNewNodes(newNodeList []types.NodeID) {
...
 }
 
 // UpdateShardState will first add new nodes into shards, then use cuckoo rule to reshard to get new shard state
-func (ss *ShardingState) UpdateShardState(newNodeList []types.NodeID, percent float64) {
+func (ss *ShardingState) assignLeaders() {
+	for i := 0; i < ss.numShards; i++ {
+		Shuffle(ss.shardState[i].NodeList)
+		for j := range ss.shardState[i].NodeList {
+			if j == 0 {
+				ss.shardState[i].NodeList[j].IsLeader = true
```
chaosma: need to check if coincident the older leader is selected again

core/types/shard_state.go
```diff
@@ -7,28 +7,31 @@ import (
 	"golang.org/x/crypto/sha3"
 )
 
-// NodeID is a unique ID represent a node
-type NodeID string
+// NodeInfo represents a node info
+type NodeInfo struct {
+	NodeID   string
+	IsLeader bool
```
chaosma: If we define data structure like this:  shardID->[leaderID, list_of_nodes_in_this_shardID], it will be more efficient. But I am OK with the current design

rlan35: What is the point of adding a IsLeader boolean here? I don't see how it's affecting the logic? Basically the first node after random shuffle is automatically the leader. what's the reason we mark them out explicitly?

chaosma: good point

mikedoan: i just modified the NodeInfo to NodeID which is a string. I still felt that we need to have indication to know who is the leader rather than using the first element of NodeList as the leader. it is because 
1/ if we do some testing for shuffle func - without any indication, we can not know who was the leader.
2/ hard to explain, but it's weird to consider the first element is the leader even I was the one who modified our resharding calcuation and designed the first element as the leader. with that, i just introduced one field to indicate the leader.

rlan35: ok, fair point

core/resharding.go
```diff
@@ -47,36 +49,50 @@ func (ss *ShardingState) assignNewNodes(newNodeList []types.NodeID) {
...
 	Shuffle(kickedNodes)
+	numInactiveShards := ss.numShards - numActiveShards
 	for i, nid := range kickedNodes {
-		id := numActiveShards + i%(ss.numShards-numActiveShards)
+		id := numActiveShards + i%numInactiveShards
 		ss.shardState[id].NodeList = append(ss.shardState[id].NodeList, nid)
 	}
 }
 
+// UpdateShardState will first add new nodes into shards, then use cuckoo rule to reshard to get new shard state
```
chaosma: wrong comment

mikedoan: thanks Chao. I'll fix it another ongoing PR soon to avoid rerun the travis tests.

core/resharding.go
mikedoan: yep, it depends on we want that happen or not - now let's make it happen randomly - if we want to avoid that we can use Leader to check and then reshuffe

PR #527, opened 2019-03-05
Add StakeLockContract with basic locking/unlocking based on epoch

rlan35:
We need a staking contract that support more functionalities such as locking based on epochs.  Will have another PR to add batch query func.

### DIFF
--- a/contracts/StakeLockContract.sol
+++ b/contracts/StakeLockContract.sol
@@ -1,0 +1,71 @@
+pragma solidity >=0.4.22;
+
+contract StakeLockContract {
+    /**
+     * @dev Error messages for require statements
+     */
+    string internal constant ALREADY_LOCKED = 'Tokens already locked';
+    string internal constant NOT_LOCKED = 'No tokens locked';
+    string internal constant AMOUNT_ZERO = 'Amount can not be 0';
+
+    /**
+     * @dev locked token structure
+     */
+    struct lockedToken {
+        uint256 _amount;
+        uint256 _blockNum;
+        uint256 _continueCount;
+    }
+
+    /**
+     * @dev Holds number & validity of tokens locked for a given reason for
+     *      a specified address
+     */
+    mapping(address => lockedToken) private locked;
+    mapping(address => uint) private indices;
+    address[] private accountList;
+
+    event Locked(address indexed _of, uint _amount, uint256 _epoch);
+    event Unlocked(address indexed account, uint index);
+
+	/**
+     * @dev Returns total tokens held by an address (locked + transferable)
+     * @param _of The address to query the total balance of
+     */
+    function lockedBalanceOf(address _of)
+        public
+        view
+        returns (uint256 amount)
+    {
+        amount = locked[_of]._amount;
+    }
+
+    /**
+     * @dev Locks a specified amount of tokens against an address
+     *      starting at the specific epoch
+     * @param _epoch The epoch where the lock happens
+     */
+    function lock(uint256 _epoch)
+        public
+        payable
+        returns (bool)
+    {
+        // If tokens are already locked, then functions extendLock or
+        // increaseLockAmount should be used to make any changes
+        require(lockedBalanceOf(msg.sender) == 0, ALREADY_LOCKED);
+        require(msg.value != 0, AMOUNT_ZERO);
+
+        locked[msg.sender] = lockedToken(msg.value, _epoch, 1);
+
+        emit Locked(msg.sender, msg.value, _epoch);
+        return true;
+    }
+
+    function getAccountCount() public view returns(uint count) {
+        return accountList.length;
+    }
+
+    function getStakedAccountList() public view returns(address[] memory stakedAccountList) {
+        return accountList;
+    }
+}
--- a/contracts/StakeLockContract.sol
+++ b/contracts/StakeLockContract.sol
@@ -8,12 +8,17 @@ contract StakeLockContract {
     string internal constant NOT_LOCKED = 'No tokens locked';
     string internal constant AMOUNT_ZERO = 'Amount can not be 0';
 
+    uint256 internal constant LOCK_PERIOD_IN_EPOCHS = 3;
+
+    uint256 internal numBlocksPerEpoch = 5;
+
     /**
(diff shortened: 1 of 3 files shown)


contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @dev Error messages for require statements
+     */
+    string internal constant ALREADY_LOCKED = 'Tokens already locked';
+    string internal constant NOT_LOCKED = 'No tokens locked';
+    string internal constant AMOUNT_ZERO = 'Amount can not be 0';
+
+    uint256 internal constant LOCK_PERIOD_IN_EPOCHS = 3;
+
+    uint256 internal numBlocksPerEpoch = 5;
```
LeoHChen: this is for testing only, please add a comment.

contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+
+contract StakeLockContract {
+    /**
+     * @dev Error messages for require statements
+     */
+    string internal constant ALREADY_LOCKED = 'Tokens already locked';
+    string internal constant NOT_LOCKED = 'No tokens locked';
+    string internal constant AMOUNT_ZERO = 'Amount can not be 0';
+
+    uint256 internal constant LOCK_PERIOD_IN_EPOCHS = 3;
```
LeoHChen: why 3 here? any comment?

contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+     *      a specified address
+     */
+    mapping(address => lockedToken) private locked;
+    mapping(address => uint) private indices;
+    address[] private accountList;
+
+    event Locked(address indexed _of, uint _amount, uint256 _epoch);
+    event Unlocked(address indexed account, uint index);
+
+	/**
```
LeoHChen: nit: alignment

contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+     */
+    function lock(uint256 _epoch)
+        public
+        payable
+        returns (bool)
+    {
+        // If tokens are already locked, then functions extendLock or
+        // increaseLockAmount should be used to make any changes
+        require(lockedBalanceOf(msg.sender) == 0, ALREADY_LOCKED);
+        require(msg.value != 0, AMOUNT_ZERO);
```
LeoHChen: != or > ?

contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+     * @dev Unlocks the unlockable tokens of a specified address
+     * @param _requestor Address of user, claiming back unlockable tokens
+     */
+    function unlock(address payable _requestor)
+        public
+        returns (uint256 unlockableTokens)
+    {
+        unlockableTokens = getUnlockableTokens(_requestor);
+
+        if (unlockableTokens != 0) {
```
LeoHChen: != 0 or >= 0?

contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * @dev Gets the epoch number of the specified block number.
+     */
+    function currentEpoch()
+        public
+        view
+        returns (uint256)
+    {
+        return block.number / numBlocksPerEpoch;
```
LeoHChen: any risk of overflow?

contracts/StakeLockContract.sol
```diff
@@ -0,0 +1,109 @@
...
+    {
+        // If tokens are already locked, then functions extendLock or
+        // increaseLockAmount should be used to make any changes
+        require(lockedBalanceOf(msg.sender) == 0, ALREADY_LOCKED);
+        require(msg.value != 0, AMOUNT_ZERO);
+
+        locked[msg.sender] = lockedToken(msg.value, block.number, currentEpoch(), 1);
+
+        emit Locked(msg.sender, msg.value, _epoch);
+        return true;
```
LeoHChen: in which condition this function will return false?

contracts/StakeLockContract.sol
rlan35: Other projects like Eth also lock for something like 3 epoch. It's just a good number to lock. Not too short not too long. We will decide the final number later (added comment)

contracts/StakeLockContract.sol
rlan35: Done.

contracts/StakeLockContract.sol
rlan35: Done.

contracts/StakeLockContract.sol
rlan35: Changed to > to be safer.

contracts/StakeLockContract.sol
rlan35: the value is uint

contracts/StakeLockContract.sol
rlan35: the two number are all uint256. As long as numBlocksPerEpoch is not 0, it should be fine.

contracts/StakeLockContract.sol
rlan35: This is a previous common practice to return true when execution succeeded, tho there is no false case.

LeoHChen:
Please add some testing log to the PR. Thanks.

contracts/StakeLockContract.sol
```diff
@@ -8,9 +8,9 @@ contract StakeLockContract {
     string internal constant NOT_LOCKED = 'No tokens locked';
     string internal constant AMOUNT_ZERO = 'Amount can not be 0';
 
-    uint256 internal constant LOCK_PERIOD_IN_EPOCHS = 3;
+    uint256 internal constant LOCK_PERIOD_IN_EPOCHS = 3;  // Final locking period TBD.
 
-    uint256 internal numBlocksPerEpoch = 5;
+    uint256 internal numBlocksPerEpoch = 5;  // This value is for testing only
```
nicolasburtey: maybe use FIXME keyword to be able to parse it later easily?

PR #529, opened 2019-03-05
Add node config module

LeoHChen:
## Issue

Add a nodeconfig module to group node specific configuration, so that it be included in both node and services modules.

#### Test Coverage Data

* Before
* After

### DIFF
--- a/internal/configs/node/config.go
+++ b/internal/configs/node/config.go
@@ -1,0 +1,165 @@
+// Package nodeconfig includes all the configuration variables for a node.
+// It is a global configuration for node and other services.
+// It will be included in node module, and other modules.
+package nodeconfig
+
+import (
+	"fmt"
+	"sync"
+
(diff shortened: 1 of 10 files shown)


internal/configs/node/config.go
```diff
@@ -0,0 +1,164 @@
...
+// configs is a list of node configuration.
+// It has at least one configuration.
+// The first one is the default, global node configuration
+var configs []ConfigType
+var onceForConfigs sync.Once
+
+// GetConfigs return the indexed ConfigType variable
+func GetConfigs(index int) *ConfigType {
+	onceForConfigs.Do(func() {
+		configs = make([]ConfigType, index+1)
```
mikedoan: add constant

mikedoan: configs = make([]ConfigType, MaxConfig)

LeoHChen: fixed.

PR #530, opened 2019-03-06
 Add listLockedAddresses and add contract abi

rlan35:
output of listLockedAddresses:

0: address[]: lockedAddresses 0xCA35b7d915458EF540aDe6068dFe2F44E8fa733c,0x14723A09ACff6D2A60DcdF7aA4AFf308FDDC160C,0x4B0897b0513fdC7C541B6d9D7E929C4e5364D2dB
1: uint256[]: blockNums 1150071,1150072,1150073
2: uint256[]: lockPeriodCounts 1,1,1

ABI:
"{Constructor:function () returns() Methods:map[balanceOf:function balanceOf(address _of) constant returns(uint256 balance) currentEpoch:function currentEpoch() constant returns(uint256) unlock:function unlock() returns(uint256 unlockableTokens) getUnlockableTokens:function getUnlockableTokens(address _of) constant returns(uint256 unlockableTokens) lock:function lock() returns(bool) listLockedAddresses:function listLockedAddresses() constant returns(address[] lockedAddresses, uint256[] blockNums, uint256[] lockPeriodCounts)] Events:map[Locked:event Locked(address indexed _of, uint256 _amount, uint256 _epoch) Unlocked:event Unlocked(address indexed account, uint256 index)]}"


contracts/StakeLockContract.sol
```diff
@@ -19,7 +19,8 @@ contract StakeLockContract {
         uint256 _amount;
         uint256 _blockNum;
         uint256 _epochNum;
-        uint256 _continueCount;
+        uint256 _lockPeriodCount;  // The number of locking period the token will be locked.
+        uint256 _index;  // The index in the addressList
```
LeoHChen: nit: alignment.
btw: better to add some comments to all the fields in this struct.

contracts/StakeLockContract.sol
```diff
@@ -28,74 +29,78 @@ contract StakeLockContract {
...
     {
         // If tokens are already locked, then functions extendLock or
         // increaseLockAmount should be used to make any changes
-        require(lockedBalanceOf(msg.sender) == 0, ALREADY_LOCKED);
+        require(balanceOf(msg.sender) == 0, ALREADY_LOCKED);
         require(msg.value != 0, AMOUNT_ZERO);
 
-        locked[msg.sender] = lockedToken(msg.value, block.number, currentEpoch(), 1);
+        // By default, the tokens can only be locked for one locking period.
+        locked[msg.sender] = lockedToken(msg.value, block.number, currentEpoch(), 1, addressList.push(msg.sender) - 1);
```
LeoHChen: what if the same address calls the lock function twice?

contracts/StakeLockContract.sol
```diff
@@ -28,74 +29,78 @@ contract StakeLockContract {
...
+        unlockableTokens = getUnlockableTokens(msg.sender);  // For now the unlockableTokens is all the tokens for a address
+        require(unlockableTokens != 0, NO_TOKEN_UNLOCKABLE);
 
-        if (unlockableTokens != 0) {
-            _requestor.transfer(unlockableTokens);
-        }
+        uint indexToRemove = locked[msg.sender]._index;
+        delete locked[msg.sender];
+        addressList[indexToRemove] = addressList[addressList.length - 1];
+        locked[addressList[addressList.length - 1]]._index = indexToRemove;
```
LeoHChen: why do you need two structure, locked, and addressList to save the data?  why don't just put the address in the locked list?

contracts/StakeLockContract.sol
rlan35: Done.

contracts/StakeLockContract.sol
rlan35: it will throw ALREADY_LOCKED error

contracts/StakeLockContract.sol
rlan35: Solidity's mapping only store values keyed by the hash of the keys. The original keys are not stored so there is no way to enumerate a map. Have to add extra array to store the keys: https://ethereum.stackexchange.com/questions/15337/can-we-get-all-elements-stored-in-a-mapping-in-the-contract

contracts/StakeLockContract.json
```diff
@@ -0,0 +1,143 @@
+[
```
alajko: I am wondering whether it makes sense to check-in the `json` file. We can always generate the `json` file from: `solc StakeLockContract.sol --abi`.

rlan35: We are checking in the generated files in general. But I will use the version from solc in next PR. I will remove this now.

alajko: Sorry, are you planning to removing it now or in next PR?

alajko: Offline @rlan35 re-iterated earlier consensus on keeping generated files. 
@harmony-ek points that `go get` breaks if there are no generated files. As `go get` skips `go generate` as it can run arbitrary command embedded in a potentially untrusted source repo on the target machine. e.g. we don't want to `go get` a repo having `//go:generate sudo rm -rf /`

PR #533, opened 2019-03-07
set up nodeconfig and refactor harmony.go

mikedoan:
## Issue

set up nodeconfig and refactor harmony.go. Now consensus and Node get set up based on nodeconfig


cmd/harmony/main.go
```diff
@@ -140,6 +139,11 @@ func initSetup() {
 
 func createGlobalConfig() *nodeconfig.ConfigType {
 	nodeConfig := nodeconfig.GetGlobalConfig()
+	nodeConfig.StakingPriKey = node.LoadStakingKeyFromFile(*stakingKeyFile, *accountIndex)
+
+	// Initialize leveldb if dbSupported.
+	nodeConfig.MainDB, _ = InitLDBDatabase(*ip, *port, *freshDB, false)
```
harmony-ek: Let's not ignore errors.

cmd/harmony/main.go
```diff
@@ -226,10 +223,7 @@ func main() {
 		}
 		currentNode.NodeConfig.SetShardGroupID(p2p.GroupIDBeacon)
 	} else {
-		var beacondb ethdb.Database
-		if *dbSupported {
-			beacondb, _ = InitLDBDatabase(*ip, *port, *freshDB, true)
-		}
+		beacondb, _ := InitLDBDatabase(*ip, *port, *freshDB, true)
```
harmony-ek: Let's not ignore errors (2).

cmd/harmony/main.go
harmony-ek: (I know you are just refactoring existing code, but let's DTRT when we touch something and leave it cleaner than it was. 😄)

harmony-ek: Also, technically `MainDB` is a live object that holds reference to system resource (on-disk file) and not simply a data-only configuration parameter.  (`ip`, `port`, and `freshDB` are.)  I'd suggest inlining this and eliminating it from config object.

cmd/harmony/main.go
```diff
@@ -138,11 +113,34 @@ func initSetup() {
...
 	nodeConfig.StakingPriKey = node.LoadStakingKeyFromFile(*stakingKeyFile, *accountIndex)
 
-	// Initialize leveldb if dbSupported.
+	nodeConfig.P2pPriKey, _, err = utils.LoadKeyFromFile(*keyFile)
+	if err != nil {
+		panic(err)
+	}
+
+	// Setup Bls keys
+	nodeConfig.BlsPriKey, nodeConfig.BlsPubKey = utils.GenKey(*ip, *port)
```
harmony-ek: A consistent naming convention would be helpful?  `P2pPriKey` refers to the key usage (P2P), whereas `BlsPriKey` refers to the key algorithm (BLS).  I'd prefer usage-based naming convention; how about `NodePriKey`?  (Plus, we don't want to change the variable name to something else in the future simply because we ditch BLS and use some other signature algorithm. 😄)

cmd/harmony/main.go
```diff
@@ -138,11 +113,34 @@ func initSetup() {
...
+	if nodeConfig.BlsPriKey == nil || nodeConfig.BlsPubKey == nil {
+		panic(fmt.Errorf("generate key error"))
+	}
+
+	// Initialize leveldb for main blockchain and beacon.
 	nodeConfig.MainDB, _ = InitLDBDatabase(*ip, *port, *freshDB, false)
+	if !*isBeacon {
+		nodeConfig.BeaconDB, _ = InitLDBDatabase(*ip, *port, *freshDB, true)
```
harmony-ek: Ditto here.  `BeaconDB` is a live object, not a configuration parameter.

internal/configs/node/config.go
```diff
@@ -57,16 +57,18 @@ const (
...
+	client   p2p.GroupID // the client group ID of the shard
+	isClient bool        // whether this node is a client node, such as wallet/txgen
+	isBeacon bool        // whether this node is a beacon node or not
+	isLeader bool        // whether this node is a leader or not
+	shardID  uint32      // shardID of this node
+	role     Role        // Role of the node
+
+	ShardIDString string
+	StringRole    string
+	Host          p2p.Host
```
harmony-ek: Ditto.  `Host` is a live object.

cmd/harmony/main.go
mikedoan: Fixing the error.

As for the MainDB i agree it maybe not the best setup. I may need to rework on this but the goal of this PR is to generate consensus and node struct based on single node config. I'm even still not happy with the node config yet but it's the starter effort.

cmd/harmony/main.go
mikedoan: done

cmd/harmony/main.go
mikedoan: let's discuss offline. NodePriKey is consusing to me. basically there are 3 set of keys with clear function.

- p2p keys
- account key (staking accounts)
- bls keys used to do multi-sig

NodePriKey does not imply anything with Bls

cmd/harmony/main.go
mikedoan: done.

internal/configs/node/config.go
mikedoan: Host is actually an interface.

mikedoan: let's discuss offline

cmd/harmony/main.go
harmony-ek: Discussed offline.  FTR: `ConsensusPriKey`

internal/configs/node/config.go
```diff
@@ -61,6 +65,21 @@ type ConfigType struct {
 	isLeader bool        // whether this node is a leader or not
 	shardID  uint32      // shardID of this node
 	role     Role        // Role of the node
+
+	// TODO(minhdoan): Clean up later.
+	ShardIDString   string
+	StringRole      string
```
LeoHChen: This is redundant. please clean it up for sure.

LeoHChen:
Please add a link to the Jenkins job shown this change is actually working fine on AWS.

LeoHChen:
You may need to rebase your code and test on AWS again.  I tried to test your branch and got only 2 consensus.
http://jenkins.harmony.one/job/branch_build_test/37/console

internal/configs/node/config.go
mikedoan: right now, i'm moved all code around. in the original harmony.go role is used as a string and it's easier for me to keep the same type because refactoring more will make people more difficult to review code. let's wait for the next PR

PR #535, opened 2019-03-07
Use solc generated abi and update CurrentStake by querying stake contract.

rlan35:
Tested locally and the result of listLockedAddresses is fetched successfully.


node/node_handler.go
```diff
@@ -306,6 +301,12 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block) {
 			node.ConfirmedBlockChannel <- newBlock
 		}()
 	}
+
+	if node.NodeConfig.Role() == nodeconfig.BeaconLeader {
+		utils.GetLogInstance().Info("Updating staking list")
+		node.UpdateStakingList(node.QueryStakeInfo())
+		node.printStakingList()
+	}
 }
```
alajko: Does it make logical difference that `UpdateStakingList` should happen after Broadcast of new block or after DRG (in case it is epoch block).

node/staking.go
```diff
@@ -20,56 +22,38 @@ import (
...
-				//This means its a new node that is staking the first time.
-				node.CurrentStakes[currentSender] = value
+// StakeInfo is the struct for the return value of listLockedAddresses func in stake contract.
+type StakeInfo struct {
+	LockedAddresses  []common.Address
+	BlockNums        []*big.Int
+	LockPeriodCounts []*big.Int
+	Amounts          []*big.Int
+}
+
```
alajko: It will be helpful to add comments to what `LockPeriodCounts` means. Could we change `Amounts` to `Stakes`?

node/node_handler.go
alajko: If not, then this file doesn't need to be checked-in

rlan35: UpdateStakingList should happen after the block is added to the chain.

node/staking.go
```diff
@@ -20,56 +22,38 @@ import (
...
-					node.CurrentStakes[currentSender] -= value
-				} else if node.CurrentStakes[currentSender] == value {
-					delete(node.CurrentStakes, currentSender)
-				} else {
-					continue //Overdraft protection.
-				}
-			} else {
-				continue //no-op: a node that is not staked cannot withdraw stake.
+			if curEpoch-startEpoch <= lockPeriodCount.Uint64()*lockPeriodInEpochs {
+				node.CurrentStakes[addr] = amount
```
alajko: Is lockPeriodCount a unit of lockPeriod? e.g. If node will have a lockPeriod = "2".  Then it will be locked for 2*3(=lockPeriodInEpochs) = 6 epochs? If so, this calculation makes sense.

node/staking.go
alajko: I know its there in the contract itself ;) But a comment will help.

alajko:
We can skip the binary file that got added (Its 4.7 MB)?

node/staking.go
rlan35: Done.

node/staking.go
rlan35: yes.

node/staking.go
```diff
@@ -20,55 +22,38 @@ import (
...
+type StakeInfo struct {
+	LockedAddresses  []common.Address
+	BlockNums        []*big.Int
+	LockPeriodCounts []*big.Int // The number of locking period the token will be locked.
+	Amounts          []*big.Int
+}
+
+// UpdateStakingList updates staking information by querying the staking smart contract.
+func (node *Node) UpdateStakingList(stakeInfo *StakeInfo) {
+	node.CurrentStakes = make(map[common.Address]*big.Int)
```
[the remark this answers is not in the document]
rlan35: Update staking list should been called every epoch. Not every block.

rlan35:
> We can skip the binary file that got added (Its 4.7 MB)?

which binary file?

alajko:
Screen Shot 2019-03-08 at 12 51 54 PM

PR #536, opened 2019-03-07
set up nodeconfig and refactor harmony.go

mikedoan:
## Issue

set up nodeconfig and refactor harmony.go. Now consensus and Node get set up based on nodeconfig


mikedoan:
http://jenkins.harmony.one/job/branch_build_test/46/

PR #542, opened 2019-03-08
Migrate faucet contract to abi

rlan35:


alajko:
LGTM. Can you add a readme to generate the abi binding.

alajko:
Since you are cleaning up, can you remove the binary file that I indicated in PR #535.

rlan35:
> Since you are cleaning up, can you remove the binary file that I indicated in PR #535.

Thanks, just removed.

Issue #540, opened 2019-03-08
peer discovery return private ip addresses


chaosma:
**Describe the bug**
When a new node/client connected to bootnode and asked for a list of peers, the bootnode will return a list of peers including both public and private ip address

**To Reproduce**
Run the current wallet code [#525](https://github.com/harmony-one/harmony/pull/525)  ./wallet balances. 

**Expected behavior**
Only the public ip addresses should be returned. All the 172.x.x.x should not be returned. 
<img width="1657" alt="screen shot 2019-03-07 at 5 46 52 pm" src="https://user-images.githubusercontent.com/3767044/54001799-5ea2a180-4101-11e9-953d-2083635e2ae9.png">

harmony-ek:
Ranges to filter.

Localhost

* 127.0.0.0/8 (127.x.y.z)

RFC 1918

* 10.0.0.0/8 (10.x.y.z)
* 172.16.0.0/12 (172.[16-31].y.z)
* 192.168.0.0/16 (192.168.y.z)

CGN

* 100.64.0.0/10 (100.[64-127].y.z)

Multicast & reserved

* 224.0.0.0/5 ([224-255].x.y.z)

LeoHChen:
fixed

PR #545, opened 2019-03-08
#540 filter out private/multicast addresses

LeoHChen:

## Issue

#540 

#### Test Coverage Data

* Before
* After

### DIFF
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


api/service/networkinfo/service.go
```diff
@@ -118,7 +117,7 @@ func (s *Service) DoService() {
 		select {
 		case peer := <-s.peerInfo:
 			if peer.ID != s.Host.GetP2PHost().ID() && len(peer.ID) > 0 {
-				//				utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs, "my ID", s.Host.GetP2PHost().ID())
+				//	utils.GetLogInstance().Info("Found Peer", "peer", peer.ID, "addr", peer.Addrs, "my ID", s.Host.GetP2PHost().ID())
```
chaosma: same format change as my last commit

LeoHChen: yes, I changed it as it is the file I touched and I did the clean up btw.  In your case, you don't have any other changes in this file, so it's unnecessary for you to touch the file just for cleanup.

LeoHChen:
http://jenkins.harmony.one/job/branch_build_test/54/console

1 shards, 30 consensus, 50 total TPS, 109 nodes

api/service/networkinfo/service.go
```diff
@@ -132,9 +131,11 @@ func (s *Service) DoService() {
 					if err != nil {
 						continue
 					}
-					nip := netaddr.(*net.TCPAddr).IP.String()
-					if strings.Compare(nip, "127.0.0.1") != 0 {
-						ip = nip
+					nip := netaddr.(*net.TCPAddr).IP
+					if nip.IsInterfaceLocalMulticast() || nip.IsLinkLocalMulticast() || nip.IsLoopback() || nip.IsMulticast() {
```
harmony-ek: Maybe:
```suggestion
					if nip.GlobalUnicast() || CGNPrefix.HasIP(nip) {
```

and

```go
// CGNPrefix is the carrier-grade NAT address space (RFC 6598).
var CGNPrefix = netaddr.MustNewIPNetwork("100.64.0.0/10")
```

This covers all of the 4 cases plus a few more: Unspecified (0.0.0.0) and global broadcast (255.255.255.255), and CGN.  It is also closer to our original intention: We want nodes to publish only globally-routable unicast address.

LeoHChen: sure.

LeoHChen:
http://jenkins.harmony.one/job/branch_build_test/58/console 

1 shards, 61 consensus, 50 total TPS, 110 nodes
@harmony-ek , please approve this PR.

api/service/networkinfo/service.go
```diff
@@ -114,27 +113,32 @@ func (s *Service) Run() {
 
 // DoService does network info.
 func (s *Service) DoService() {
+	_, ipv4Net, err := net.ParseCIDR("100.64.0.0/10")
```
harmony-ek: Also for the future reference, [`netaddr.MustNewIPNetwork()`](https://godoc.org/github.com/xgfone/netaddr#MustNewIPNetwork) is useful when we parse a string literal into an IP prefix: Since the input is a literal, it will always succeed or always fail, and if it always fails, might as well panic early (especially if the variable is declared at package level) because that's most likely a bug (like a typo).

PR #551, opened 2019-03-11
Proposal PR

alajko:
Proposal PR to demonstrate how a variable node able to configure services might look like.

TODO
- [ ] Remove service <> node cyclic imports.

### DIFF
--- a/internal/configs/node/config.go
+++ b/internal/configs/node/config.go
@@ -27,6 +27,7 @@ const (
 	NewNode
 	ClientNode
 	WalletNode
+	VariableNode
 )
 
 func (role Role) String() string {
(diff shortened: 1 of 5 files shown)


internal/configs/node/config.go
```diff
@@ -66,10 +69,12 @@ type ConfigType struct {
 	isClient bool        // whether this node is a client node, such as wallet/txgen
 	isBeacon bool        // whether this node is a beacon node or not
 	isLeader bool        // whether this node is a leader or not
+	isWallet bool        // whether this node is a leader or not
```
LeoHChen: ?

internal/configs/node/config.go
```diff
@@ -66,10 +69,12 @@ type ConfigType struct {
...
 	shardID  uint32      // shardID of this node
 	role     Role        // Role of the node
 
-	ShardIDString   string
+	ShardIDString string
```
LeoHChen: why change it here?

PR #554, opened 2019-03-11
[build] fix build error after update of bls

LeoHChen:

## Issue

update go.mod to fix build issue after bls update.

#### Test Coverage Data

* Before
* After

### DIFF
--- a/api/proto/node/node_test.go
+++ b/api/proto/node/node_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/harmony-one/harmony/core/state"
 	"github.com/harmony-one/harmony/core/types"
 
-	//	"fmt"
 	"math/big"
 	"reflect"
 	"testing"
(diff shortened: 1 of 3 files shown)


api/proto/node/node_test.go
```diff
@@ -131,7 +131,7 @@ func TestInfoToString(t *testing.T) {
 		ValidatorID: 1,
 		PeerID:      "peer",
 	}
-	if strings.Compare(info.String(), "Info:127.0.0.1/81=>1/<peer.ID 3sdfvR>") != 0 {
-		t.Error("Info string mismatch")
+	if !strings.Contains(info.String(), "Info:127.0.0.1/81=>1/<peer.ID") {
+		t.Errorf("Info string mismatch: %v", info.String())
```
rlan35: the test still fails here. Seems the result is flaky depending on platform

LeoHChen: okay, I'll change it to get test pass.

PR #556, opened 2019-03-12
Calculate resharding based on real stake info

rlan35:


core/resharding.go
```diff
@@ -127,17 +133,48 @@ func GetShardingStateFromBlockChain(bc *BlockChain, epoch uint64) *ShardingState
...
+		for _, nodeID := range shard.NodeList {
+			addr := common.Address{}
+			addrBytes, err := hex.DecodeString(string(nodeID))
+			if err != nil {
+				utils.GetLogInstance().Error("Failed to decode address hex")
+			}
+			addr.SetBytes(addrBytes)
+			oldAddresses[addr] = true
+			_, ok := (*stakeInfo)[addr]
+			if !ok {
```
chaosma: it should be changed to "if ok" if it means the addr is still in stake?

rlan35: Good catch. Fixed.

PR #560, opened 2019-03-12
Setup genesis shard state, and fix other issues

rlan35:


core/resharding.go
```diff
@@ -190,15 +197,15 @@ func (ss *ShardingState) calculateKickoutRate(newNodeList []types.NodeID) float6
 	return math.Max(0.1, math.Min(rate, 1.0))
 }
 
-// getInitShardState returns the initial shard state at genesis.
-func getInitShardState(numberOfShards, numNodesPerShard int) types.ShardState {
+// GetInitShardState returns the initial shard state at genesis.
+func GetInitShardState() types.ShardState {
 	shardState := types.ShardState{}
-	for i := 0; i < numberOfShards; i++ {
+	for i := 0; i < GenesisShardNum; i++ {
```
chaosma: At the very beginning, there are not 3 shards (GenesisShardNum = 3), then the resharding will broke again

core/resharding.go
```diff
@@ -134,12 +140,13 @@ func GetShardingStateFromBlockChain(bc *BlockChain, epoch uint64) *ShardingState
 // CalculateNewShardState get sharding state from previous epoch and calculate sharding state for new epoch
 // TODO: currently, we just mock everything
 func CalculateNewShardState(bc *BlockChain, epoch uint64, stakeInfo *map[common.Address]*structs.StakeInfo) types.ShardState {
-	if epoch == FirstEpoch {
-		return getInitShardState(3, 10)
+	if epoch == GenesisEpoch {
+		return GetInitShardState()
```
mikedoan: how do you consolidate this hard-code config together with the params fed to deploy.sh?

if the config def to deploy.sh has 1 beacon chain with 20 nodes this may be not correct

rlan35: Yea, for now, I am keeping them in sync.

core/resharding.go
rlan35: I added the check to fix the break, but conceptually, there are 3 shards, with 1 shard populated and 2 empty.

mikedoan: that's fair enough. please add TODO

core/resharding.go
mikedoan: that's fair enough. please add TODO

PR #561, opened 2019-03-13
convert drand proto to message.proto

mikedoan:
## Issue

convert drand proto to message.proto


mikedoan:
I'm also working on convert txgen message into message protocol

mikedoan:
> LGTM, can you re-trigger a Jenkins build or do a branch build so that we know this patch won't impact the current test?

I did rebuild but it still failed http://jenkins.harmony.one/job/build_test_on_pr/43/console

PR #562, opened 2019-03-13
add loterry contract

mikedoan:
## Issue

add loterry contract


contracts/Lottery.sol
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    modifier restricted() {
+        require(msg.sender == manager, RESTRICTED_MESSAGE);
+        _;
+    }
+
+    function getPlayers() public view returns (address payable[] memory) {
+        return players;
+    }
```
alajko: We don't use `getPlayers()`?

contracts/Lottery.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    function pickWinner() public payable restricted {
+        uint index = random() % players.length;
+        players[index].transfer(address(this).balance);
+        players = new address payable[](0);
+    }
+
+    modifier restricted() {
+        require(msg.sender == manager, RESTRICTED_MESSAGE);
+        _;
```
alajko: not required.

contracts/Lottery.sol
alajko: Do we plan to use `getPlayers()` later?

alajko:
Could we test this as done here for faucet and staking: `test/chain/main.go`.

contracts/Lottery.sol
```diff
@@ -0,0 +1,38 @@
...
+
+    address public manager;
+    address payable[] public players;
+
+    constructor() public {
+        manager = msg.sender;
+    }
+
+    function enter() public payable {
+        require(msg.value > .01 ether, ENTER_MESSAGE);
```
LeoHChen: nit: the .01 ether is not consistent with the message 0.1 ether

contracts/Lottery.sol
```diff
@@ -0,0 +1,38 @@
...
+
+        players.push(msg.sender);
+    }
+
+    function random() private view returns (uint) {
+        return uint(keccak256(abi.encodePacked(now, msg.sender, this)));
```
LeoHChen: what's the range of the random number?

contracts/Lottery.sol
mikedoan: yes, i'm actually also returning balances

contracts/gen.sh
```diff
@@ -0,0 +1,3 @@
+abigen -sol Lottery.sol  -out Lottery.go --pkg contracts
+# abigen -sol Faucet.sol  -out Faucet.go --pkg contracts
+# abigen -sol StakeLockContract.sol  -out StakeLockContract.go --pkg contracts
```
alajko: Any reason to comment out Faucet and StakeLockContract?

node/contract_test.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+	consensus := consensus.New(host, "0", []p2p.Peer{leader, validator}, leader)
+	return New(host, consensus, nil)
+
+}
+
+func TestAddLotteryContract(t *testing.T) {
+	node := prepareNode(t)
+	node.AddLotteryContract()
+	if len(node.DemoContractAddress) == 0 {
```
alajko: This would test whether the DemoContractAddress has been set. This would be testing L133 `node/contract.go` However we would not know whether the contract gets correctly deployed and whether we are able to call functions within the contract e.g. How to test whether `getPlayers()` works. This would require us to simulate block formation and deployment of contract, you can see how this done here: `test/chain/main.go`.

node/node.go
```diff
@@ -258,6 +259,10 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, db ethdb.Database) *N
 		node.AddFaucetContractToPendingTransactions()
 		node.CurrentStakes = make(map[common.Address]*structs.StakeInfo)
 		node.AddStakingContractToPendingTransactions() //This will save the latest information about staked nodes in current staked
+
+		// TODO(minhdoan): Think of a better approach to deploy smart contract.
+		// This is temporary for demo purpose.
+		node.AddLotteryContract()
```
alajko: This is indeed a problem. (Not to be fixed in this PR) but thanks for calling it out.  My earlier investigation had brought me to this: https://gist.github.com/skabbass1/f37e35998723cd2bb8e21a01fefbb0d7#file-deploy-go But I am not sure whether it can work with what we are building.

node/contract_test.go
mikedoan: i'll fix this

contracts/Lottery.sol
mikedoan: _ is used to replace _ by all the code of the function using modifier

node/contract_test.go
mikedoan: i already tested in local but i didn't write test out. I agree it's not fully tested but it's in WIP - i would add more tests later but at least there wasn't any tests before this PR.

I'm looking to move all the code test/main.go as unit test here

node/node.go
mikedoan: Offline discussed

contracts/gen.sh
mikedoan: if i run gen for Faucet and StateLockContract if it will generate the diff between the current go files and the new ones

mikedoan:
> Test whether smart contract works, so we have greater confidence once its deployed to the network.

Of course, this is one of items i'm working on. given the tight timeline this discouraged all effort i'm trying

alajko:
@mikedoan is working on a testing setup. Approving.

PR #568, opened 2019-03-13
Fix and tidy go mod

harmony-ek:
## Issue

go-datastore, like other IPFS/libp2p projects already did, switched to module-style versioning, and old tags (v3.y.z) are now prefixed with gx/ and not available for Go module imports.

While we are here, run `go mod tidy` to populate test and other dependencies not normally captured via simple builds.  It also marks check.v1 as a direct dependency, which it is.

#### Test Coverage Data

No code change; test coverage shouldn't be different.

#### Test/Run Logs

Awaiting PR builds.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,7 @@ require (
 	github.com/gorilla/mux v1.7.0
 	github.com/harmony-one/bls v0.0.0-20190309234102-1bd75ac96c09
 	github.com/hashicorp/golang-lru v0.5.1
-	github.com/ipfs/go-datastore v3.2.0+incompatible
+	github.com/ipfs/go-datastore v0.0.1
 	github.com/libp2p/go-libp2p v0.0.2
 	github.com/libp2p/go-libp2p-crypto v0.0.1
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -3,18 +3,25 @@ module github.com/harmony-one/harmony
...
+	github.com/fjl/memsize v0.0.0-20180929194037-2a09253e352a // indirect
+	github.com/go-ole/go-ole v1.2.4 // indirect
 	github.com/go-stack/stack v1.8.0 // indirect
 	github.com/golang/mock v1.2.0
 	github.com/golang/protobuf v1.3.0
 	github.com/gorilla/mux v1.7.0
 	github.com/harmony-one/bls v0.0.0-20190309234102-1bd75ac96c09
 	github.com/hashicorp/golang-lru v0.5.1
-	github.com/ipfs/go-datastore v3.2.0+incompatible
+	github.com/ipfs/go-datastore v0.0.1
```
LeoHChen: this will break Jenkins build.

harmony-ek: Explain?

PR #573, opened 2019-03-13
[networkinfo] retry for bootnode connection

LeoHChen:
We experienced a few cases that node failed to connect to bootnode and
the node program will not be useable. So, we add a retry mechanism to
increase the connection times to bootnode.


api/service/networkinfo/service.go
```diff
@@ -76,15 +76,22 @@ func (s *Service) Init() error {
 		s.bootnodes = utils.BootNodes
 	}
 
+	// retry the connection to bootnode in case the initial connection is failed
+	connectionRetry := 3
```
mikedoan: move it to constant

api/service/networkinfo/service.go
```diff
@@ -76,15 +76,22 @@ func (s *Service) Init() error {
...
+	waitInRetry := 2 * time.Second
```
mikedoan: ditto

LeoHChen:
> Since we know that there cannot possibly be good outcome ever if this initial discovery fails, I'd say 3 tries is way too low. I'd say try at least 15-20 times and fail-exit if still unsuccessful, with a (long-term) option to retry indefinitely until it succeeds, which is useful for daemon mode where there is no expectation for the process to exit.

make sense. I'll increase the number.

PR #575, opened 2019-03-14
Assign new nodes to empty shards at first epoch; Make kickout percent constant; Make NodeID a struct

rlan35:


core/resharding.go
```diff
@@ -67,9 +70,12 @@ func (ss *ShardingState) cuckooResharding(percent float64) {
 		}
 		numKicked := int(percent * float64(len(ss.shardState[i].NodeList)))
 		if numKicked == 0 {
-			numKicked++
+			numKicked++ // At least kick one node out
 		}
 		length := len(ss.shardState[i].NodeList)
+		if length-numKicked <= 0 {
```
chaosma: In this case, we can kick out part of numKicked for example first 10%

api/service/staking/service.go
```diff
@@ -87,7 +87,7 @@ func (s *Service) Run() {
 				if s.IsStaked() {
 					return
 				}
-				s.DoService()
+				//s.DoService()
```
chaosma: No need to DoService?

core/resharding.go
```diff
@@ -19,12 +18,16 @@ import (
...
 	GenesisEpoch = 0
+	// FirstEpoch is the number of the first epoch.
+	FirstEpoch = 1
 	// GenesisShardNum is the number of shard at genesis
-	GenesisShardNum = 3
+	GenesisShardNum = 4
 	// GenesisShardSize is the size of each shard at genesis
 	GenesisShardSize = 10
+	// CuckooRate is the percentage of nodes getting reshuffled in the second step of cuckoo resharding.
+	CuckooRate = 0.1
```
chaosma: If we use constant rate, how to make sure it's safe when number of new nodes is greater than 10% ?

core/resharding.go
```diff
@@ -138,33 +144,40 @@ func GetShardingStateFromBlockChain(bc *BlockChain, epoch uint64) *ShardingState
...
-// TODO: currently, we just mock everything
 func CalculateNewShardState(bc *BlockChain, epoch uint64, stakeInfo *map[common.Address]*structs.StakeInfo) types.ShardState {
 	if epoch == GenesisEpoch {
 		return GetInitShardState()
 	}
 	ss := GetShardingStateFromBlockChain(bc, epoch-1)
+	if epoch == FirstEpoch {
+		newNodes := []types.NodeID{}
+		for addr, stakeInfo := range *stakeInfo {
+			newNodes = append(newNodes, types.NodeID{addr.Hex(), hex.EncodeToString(stakeInfo.BlsAddress[:])})
```
chaosma: what's the difference of these two addresses? I thought we can blsAddr as NodeID

api/service/staking/service.go
rlan35: Temporarily disable repeated DoService. It's done only once now.

core/resharding.go
rlan35: It will get stable after a few epochs. The thinking is that with previous kickout calculation, if a lot of nodes joins at once, it's a risk to disturb the network a lot.

core/resharding.go
rlan35: each node have two addresses. It's better to keep track of both of them.

core/resharding.go
rlan35: It shouldn't matter if we kick out a constant of 10%.  This condition only happen when numKicked++ happens, and is only to make sure we don't kick out all nodes.

PR #577, opened 2019-03-14
Teach wallet to include optional tx input data

harmony-ek:
Usage: `./wallet transfer <other flags> --inputData=<base64-encoded input data>`

Example: `./wallet transfer --from=0 --to=1 --amount=1 --shardID=1 --inputData='0mGw+F=='`

## Test

Local test passed.


alajko:
@harmony-ek  Would it make sense to investigate sending smart contracts by extending wallets. (FYI @mikedoan).

PR #584, opened 2019-03-16
Update README.md

LeoHChen:
Update the git clone command.


LeoHChen:
the original instruction still work for me. ?

PR #588, opened 2019-03-17
Node script update

LeoHChen:
Update the node.sh script for external node runner.

* check running as root
* provide some more help message

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -60,12 +60,28 @@ function setup_env
    echo "session required pam_limits.so" | sudo tee -a /etc/pam.d/common-session
 }
 
+function find_harmony_process
+{
+   unset -v pidfile pid
+   pidfile="harmony-${PUB_IP}.pid"
+   pid=$!
+   echo "${pid}" > "${pidfile}"
+   ps -f -p "${pid}"
+}
+
+######## main #########
+if [[ $EUID -ne 0 ]]; then
+   echo "This script must be run as root"
+   echo Please use \"sudo ./node.sh\"
+   exit 1
+fi
+
 killnode
 
 mkdir -p latest
 BUCKET=pub.harmony.one
 OS=$(uname -s)
-REL=20190216
+REL=banjo
 
 if [ "$OS" == "Darwin" ]; then
    FOLDER=release/$REL/darwin-x86_64/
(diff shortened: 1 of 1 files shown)


scripts/node.sh
```diff
@@ -101,17 +108,25 @@ fi
 # find my public ip address
 myip
 
-# public beacon node multiaddress
-BC_MA=/ip4/54.183.5.66/tcp/9999/ipfs/QmdQVypu6NSm7m8bNZj5EJCnjPhXR8QyRmDnDBidxGaHWi
+# public boot node multiaddress
+BN_MA=/ip4/100.26.90.187/tcp/9876/p2p/QmZJJx6AdaoEkGLrYG4JeLCKeCKDjnFz2wfHNHxAqFSGA9,/ip4/54.213.43.194/tcp/9876/p2p/QmQayinFSgMMw5cSpDUiD9pQ2WeP6WNmGxpZ6ou3mdVFJX
```
harmony-ek: Good enough for now, but hardcoding these addresses/port numbers/node IDs is not such a good idea.  Opened #591 to track this.

scripts/node.sh
```diff
@@ -101,17 +108,25 @@ fi
...
 
 if [ "$OS" == "Linux" ]; then
 # Run Harmony Node
-   nohup ./harmony -bc_addr $BC_MA -ip $PUB_IP -port $NODE_PORT > harmony-${PUB_IP}.log 2>&1 &
+   LD_LIBRARY_PATH=$(pwd) nohup ./harmony -bootnodes $BN_MA -ip $PUB_IP -port $NODE_PORT -is_beacon > harmony-${PUB_IP}.log 2>&1 &
```
harmony-ek: A [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)-refactor is due here.  How about adding a reusable script `run_with_libpath.sh`:

```
#!/bin/sh
set -eu
unset -v libpath_var libpath
libpath_var=LD_LIBRARY_PATH  # most platforms use this
case "$(uname -s)" in
Darwin) libpath_var=DYLD_FALLBACK_LIBRARY_PATH;;
esac
libpath="${1-}"
shift 1 || { echo "run_with_libpath: libpath not given" >&2; exit 64; }
exec env "${libpath_var}=${libpath}" "$@"
```

and then using it from `node.sh`:

```
# Add this around the beginning of the script
unset -v progname progdir
progname="${0##*/}"
case "${0}" in
*/*) progdir="${0%/*}";;
*) progdir=.;;
esac
```

```
# Substitute this for the if-then-else-fi
"${progdir}/run_with_libpath.sh" "$(pwd)" nohup ./harmony -bootnodes $BN_MA -ip $PUB_IP -port $NODE_PORT -is_beacon > harmony-${PUB_IP}.log 2>&1 &
```

scripts/node.sh
```diff
@@ -101,17 +108,25 @@ fi
...
 else
-   ./harmony -bc_addr $BC_MA -ip $PUB_IP -port $NODE_PORT > harmony-${PUB_IP}.log 2>&1 &
+   DYLD_FALLBACK_LIBRARY_PATH=$(pwd) ./harmony -bootnodes $BN_MA -ip $PUB_IP -port $NODE_PORT -is_beacon > harmony-${PUB_IP}.log 2>&1 &
 fi
 
+echo "############### Running Harmony Process ###############"
+ps -ef | grep harmony | grep -v grep
```
harmony-ek: Let's record the running PID:

```
unset -v pidfile pid
pidfile="harmony-${PUB_IP}.pid"
pid=$!
echo "${pid}" > "${pidfile}"
ps -f -p "${pid}"
```

scripts/node.sh
```diff
@@ -101,17 +108,25 @@ fi
...
+echo
+echo
+
 echo Please run the following command to inspect the log
 echo "tail -f harmony-${PUB_IP}.log"
 
+echo
+echo You may use \"sudo kill -9 harmony\" to terminate running harmony node program.
```
harmony-ek: Now that we are about to persist DB contents, let's not suggest `kill -9` but simply `kill`.

`kill -9` will prevent whatever cleanup routine (typically bound to `SIGTERM` handler) from running, and the next time the node starts up, the DB opening procedure may find it necessary to replay the intent logs.

harmony-ek:
Merged from `master` to unblock builds.

PR #590, opened 2019-03-18
[networkinfo] regular advertise to bootnode

LeoHChen:
DHT will clear out the entries every 24 hours.
The node needs to regularly announce to bootnode so that other peers can
find the peer.

We do regular advertise every 12 hours.

### DIFF
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -42,6 +42,9 @@ var (
 const (
 	waitInRetry       = 2 * time.Second
 	connectionTimeout = 3 * time.Minute
+
+	// register to bootnode every ticker
+	dhtTicker = 6 * time.Hour
 )
 
 // New returns role conversion service.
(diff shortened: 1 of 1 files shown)


api/service/networkinfo/service.go
```diff
@@ -42,6 +42,9 @@ var (
 const (
 	waitInRetry       = 2 * time.Second
 	connectionTimeout = 3 * time.Minute
+
+	// register to bootnode every ticker
+	dhtTicker = 12 * time.Hour
```
rlan35: How about every 6 hours. For 12 hours, if the first announce fails, then the DHT will be cleaned too.

harmony-ek:
Merged from fixed `master` to unblock builds.  `*fingers crossed*`

api/service/networkinfo/service.go
LeoHChen: okay. changed.

PR #595, opened 2019-03-19
Banjo

LeoHChen:
This PR reverted the BLS key change that broke the consensus.

It also fixed the wallet by adding sleep after p2p communication.


LeoHChen:
http://jenkins.harmony.one/job/branch_build_test/90/console

That test actually pass.
1 shards, 99 consensus, 65 total TPS, 110 nodes

PR #596, opened 2019-03-19
[build] fix jenkins build issue

LeoHChen:
This will fix the Jenkins build issue after the removal of $GOPATH in Jenkins configuration.

This is needed for go_executable_build.sh script to locate the MCL and BLS libraries.

### DIFF
--- a/scripts/setup_bls_build_flags.sh
+++ b/scripts/setup_bls_build_flags.sh
@@ -1,8 +1,18 @@
 # no shebang; to be sourced from other scripts
 
+unset -v progdir
+case "${0}" in
+*/*) progdir="${0%/*}";;
+*) progdir=.;;
+esac
+
 unset -v gopath
 gopath=$(go env GOPATH)
+# HMY_PATH is the common root directory of all harmony repos
 HMY_PATH="${gopath%%:*}/src/github.com/harmony-one"
+if [ ! -d $HMY_PATH ]; then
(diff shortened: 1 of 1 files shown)


scripts/setup_bls_build_flags.sh
```diff
@@ -1,8 +1,17 @@
...
+case "${0}" in
+*/*) progdir="${0%/*}";;
+*) progdir=.;;
+esac
+
 unset -v gopath
 gopath=$(go env GOPATH)
 HMY_PATH="${gopath%%:*}/src/github.com/harmony-one"
+if [ ! -d $HMY_PATH ]; then
+   HMY_PATH=$(realpath $progdir/../..)
```
harmony-ek: I think we can just leave this line and remove gopath-based logic as we moved to modules and shouldn't pull in GOPATH anymore.  That is, we can replace lines 9–14 with just this one line.  Existing dev environments already conform to this standard—which essentially says the `bls` and `mcl` repo directories are in the same directory as the `harmony` repo directory—so I think it will be backward compatible.

mikedoan:
with this, is the deploy.sh working now? my comment does not affect your check-in.

scripts/setup_bls_build_flags.sh
LeoHChen: okay.

LeoHChen:
> with this, is the deploy.sh working now? my comment does not affect your check-in.

yes, it still works.

PR #598, opened 2019-03-19
add stop/restart support;add archival config flag;

chaosma:


core/blockchain.go
```diff
@@ -1521,7 +1521,6 @@ func (bc *BlockChain) InsertHeaderChain(chain []*types.Header, checkFreq int) (i
 	if i, err := bc.hc.ValidateHeaderChain(chain, checkFreq); err != nil {
 		return i, err
 	}
-
```
LeoHChen: ?

node/node.go
```diff
@@ -228,16 +233,21 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, db ethdb.Database) *N
 	if host != nil && consensusObj != nil {
 		// Consensus and associated channel to communicate blocks
 		node.Consensus = consensusObj
-
-		// Init db.
```
LeoHChen: ?

node/node.go
```diff
@@ -228,16 +233,21 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, db ethdb.Database) *N
...
 		database := db
 		if database == nil {
 			database = ethdb.NewMemDatabase()
+			chain, err = node.GenesisBlockSetup(database)
+			isFirstTime = true
+		} else {
+			chain, err = node.InitBlockChainFromDB(db, node.Consensus)
+			isFirstTime = false
+			if err != nil || chain == nil || chain.CurrentBlock().NumberU64() <= 0 {
+				chain, err = node.GenesisBlockSetup(database)
```
LeoHChen: do you need a log line so that we know what's happening?

node/node.go
```diff
@@ -461,3 +472,13 @@ func (node *Node) AddBeaconChainDatabase(db ethdb.Database) {
 	node.beaconChain = chain
 	node.BeaconWorker = worker.New(params.TestChainConfig, chain, node.Consensus, pki.GetAddressFromPublicKey(node.SelfPeer.ConsensusPubKey), node.Consensus.ShardID)
 }
+
+func (node *Node) InitBlockChainFromDB(db ethdb.Database, consensus *consensus.Consensus) (*core.BlockChain, error) {
```
LeoHChen: you need comment on public function.

node/node.go
```diff
@@ -461,3 +472,13 @@ func (node *Node) AddBeaconChainDatabase(db ethdb.Database) {
...
+	chainConfig := params.TestChainConfig
+	if consensus != nil {
+		chainConfig.ChainID = big.NewInt(int64(consensus.ShardID)) // Use ChainID as piggybacked ShardID
+	}
+	cacheConfig := core.CacheConfig{Disabled: false, TrieNodeLimit: 256 * 1024 * 1024, TrieTimeLimit: 5 * time.Minute}
```
LeoHChen: please use some constants here.

core/blockchain.go
chaosma: I made several debug info in the code, not cleaned up fully

node/node.go
chaosma: added because of the travis check

node/node.go
chaosma: If the local database is available, we will create the genesis block. it's not an error

node/node.go
chaosma: This is the default value from Ethereum code. They didn't use constant either. This is a placeholder. After we add the archival node role, I will remove it and add config for archival node.

core/blockchain.go
chaosma: add space back

node/node.go
chaosma: add back

node/node.go
LeoHChen: Yes, I understand.  Please just add an informational log.

chaosma:
Yes, it will read the backup file by default.

PR #599, opened 2019-03-19
Add  verbose to build process

LeoHChen:
Default is no verbose output.  That helps identify the build errors.

### DIFF
--- a/scripts/go_executable_build.sh
+++ b/scripts/go_executable_build.sh
@@ -13,6 +13,7 @@ GOOS=linux
 GOARCH=amd64
 FOLDER=/${WHOAMI:-$USER}
 RACE=
+VERBOSE=
 
 unset -v progdir
 case "${0}" in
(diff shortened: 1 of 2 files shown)


scripts/go_executable_build.sh
```diff
@@ -50,6 +51,7 @@ OPTIONS:
    -b bucket      set the upload bucket name (default: $BUCKET)
    -f folder      set the upload folder name in the bucket (default: $FOLDER)
    -r             enable -race build option (default: $RACE)
+   -v             verbose build process (default: $VERBOSE)
```
harmony-ek: `VERBOSE` defaults to an empty string on line 16 above, so this line would be confusing:

```
   -v             verbose build process (default: )
```

LeoHChen: yes, it is default empty as to simplify the code.

PR #605, opened 2019-03-20
Bump up genesis beaconchain size to 50

rlan35:
Tested locally and it reached repeated consensus.


test/deploy.sh
```diff
@@ -136,26 +136,28 @@ NUM_NN=0
 sleep 2
```
mikedoan: is this change related to resharding work? we don't want to effect one beacon chain

PR #606, opened 2019-03-21
refactor state sync; decouple from consensus

chaosma:
Current refactor decouple the state sync with consensus. But it will not join consensus because of the consensusID not match.
Remaining to do:
1. Replace beacon sync with this pr's state sync
2. Add logic to handle join consensus (with/without old logic)


api/service/syncing/syncing.go
```diff
@@ -28,6 +26,7 @@ const (
 	TimesToFail                           = 5
 	RegistrationNumber                    = 3
 	SyncingPortDifference                 = 3000
+	inSyncThreshold                       = 1
```
rlan35: Let's have some comments on all these numbers.

api/service/syncing/syncing.go
```diff
@@ -562,3 +554,30 @@ func (ss *StateSync) RegisterNodeInfo() int {
...
 	return count
 }
+
+// getMaxPeerHeight gets the maximum blockchain heights from peers
+func (ss *StateSync) getMaxPeerHeight() uint64 {
+	ss.CleanUpNilPeers()
+	maxHeight := uint64(0)
+	for id := range ss.syncConfig.peers {
+		peerConfig := ss.syncConfig.peers[id]
+		response := peerConfig.client.GetBlockChainHeight()
```
rlan35: Better make these calls go routine. Otherwise, it will be slow.

api/service/syncing/syncing.go
```diff
@@ -562,3 +554,30 @@ func (ss *StateSync) RegisterNodeInfo() int {
...
+		}
+	}
+	return maxHeight
+}
+
+// SyncLoop will keep syncing with peers until catches up
+func (ss *StateSync) SyncLoop(bc *core.BlockChain, worker *worker.Worker) {
+	for {
+		otherHeight := ss.getMaxPeerHeight()
+		if ss.currentHeight+inSyncThreshold >= otherHeight {
```
rlan35: what's the purpose of inSyncThreshold?

node/node_syncing.go
```diff
@@ -103,6 +107,16 @@ func (node *Node) IsOutOfSync(consensusBlockInfo *consensus.BFTBlockInfo) bool {
 	return true
 }
 
+// DoSync syncs with peers until catchup, this function is not coupled with consensus
+func (node *Node) DoSync() {
+	<-node.peerReadyChan
+	ss := syncing.CreateStateSync(node.SelfPeer.IP, node.SelfPeer.Port, node.Consensus.PubKey.GetAddress())
```
rlan35: If this is not related to consensus, so the node.Consensus.PubKey.GetAddress() shouldn't be needed, right?

api/service/syncing/syncing.go
chaosma: This means the state syncing is almost ready to join consensus. Here it is just one block behind the consensus

node/node_syncing.go
chaosma: If the public key is used to identify the peer ip and address, we still need it. Since the synced node will broadcast verified block to unsynced node. Public key is a way to identify which unsynced node to send

api/service/syncing/syncing.go
chaosma: comments added

api/service/syncing/syncing.go
chaosma: I changed to goroutine. Please take a look at the changes

api/service/syncing/syncing.go
```diff
@@ -158,9 +148,12 @@ func (peerConfig *SyncPeerConfig) GetBlocks(hashes [][]byte) ([][]byte, error) {
 }
 
 // CreateSyncConfig creates SyncConfig for StateSync object.
-func (ss *StateSync) CreateSyncConfig(peers []p2p.Peer) {
+func (ss *StateSync) CreateSyncConfig(peers []p2p.Peer) bool {
 	utils.GetLogInstance().Debug("CreateSyncConfig: len of peers", "len", len(peers))
-	utils.GetLogInstance().Debug("CreateSyncConfig: len of peers", "peers", peers)
+	if peers == nil || len(peers) == 0 {
```
alajko: nit: peers == nil means len(peers) == 0 in golang

alajko: so enoguh to check `if len(peers) == 0`

api/service/syncing/syncing.go
```diff
@@ -170,9 +163,10 @@ func (ss *StateSync) CreateSyncConfig(peers []p2p.Peer) {
 			ip:   peers[id].IP,
 			port: peers[id].Port,
 		}
-		utils.GetLogInstance().Debug("[SYNC] CreateSyncConfig: peer port to connect", "port", peers[id].Port)
+		//	utils.GetLogInstance().Debug("[SYNC] CreateSyncConfig: peer port to connect", "port", peers[id].Port)
```
alajko: either uncomment or delete?

node/node_syncing.go
alajko: Node needs to have a public key outside of the `consensusObj`  It can then set the consensus pubkey to be same. This will help us decouple consensus from node identity.

api/service/syncing/syncing.go
chaosma: changed it

api/service/syncing/syncing.go
chaosma: think of possible debug later, but sure I just removed it

alajko:
Can you demonstrate it is working locally? Maybe attach a gist.

chaosma:
Local test result: 

Node 9010 start 30 seconds later and catches up with "IN SYNC" status at 27th block. But since it cannot join consensus at this pr, it just stops there

====== RESULTS ======
++ ./../test/cal_tps.sh tmp_log/log-20190320-234733/all-leaders.txt tmp_log/log-20190320-234733/all-validators.txt
+ results='1 shards, 244 consensus, 1142 total TPS, 6 nodes, 7 total nodes, 5 total signatures
127.0.0.1, 244, 1142.14
peerID: 9002, numOfConsensus: 244
peerID: 9003, numOfConsensus: 244
peerID: 9001, numOfConsensus: 244
peerID: 9010, numOfConsensus: 27
peerID: 9004, numOfConsensus: 244
peerID: 9005, numOfConsensus: 244'

api/service/syncing/downloader/client.go
```diff
@@ -22,7 +22,7 @@ func ClientSetup(ip, port string) *Client {
 	client := Client{}
 	client.opts = append(client.opts, grpc.WithInsecure())
 	var err error
-	client.conn, err = grpc.Dial(fmt.Sprintf("%s:%s", ip, port), client.opts...)
+	client.conn, err = grpc.Dial(fmt.Sprintf(ip+":"+port), client.opts...)
 	if err != nil {
 		utils.GetLogInstance().Info("client.go:ClientSetup fail to dial: ", "error", err)
```
alajko: nit: change back.

PR #616, opened 2019-03-22
unify state sync, beacon sync and archival sync

chaosma:
contains three sub commits:
1. unify state sync and beacon sync (i.e. two cases depend on whether needs join consensus)
2. move archival node as validator role for security concern; archival node fully works
3. the network can resume after restart again (but one node restart has issue caused by leader pubkey not found, since the discovery service only run once)

two issues left:
1. last mile issue need to be solved for node do state sync and then join consensus
2. modify p2p discovery logic

### DIFF
--- a/api/service/discovery/service.go
+++ b/api/service/discovery/service.go
@@ -89,7 +89,6 @@ func (s *Service) contactP2pPeers() {
 			// It is just a temporary hack. When we work on re-sharding to regular shard, this has to be changed.
 			if !s.config.IsBeacon {
 				if s.addBeaconPeerFunc != nil {
-					utils.GetLogInstance().Debug("hehe", "peer", peer)
 					s.addBeaconPeerFunc(&peer)
 				}
 			}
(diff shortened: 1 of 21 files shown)


contracts/Faucet.go
```diff
@@ -31,7 +31,7 @@ var (
...
+const FaucetBin = `0x6080604052678ac7230489e80000600055600180546001600160a01b03191633179055610105806100316000396000f3fe6080604052348015600f57600080fd5b506004361060325760003560e01c806327c78c421460375780634ddd108a14605c575b600080fd5b605a60048036036020811015604b57600080fd5b50356001600160a01b03166074565b005b606260d4565b60408051918252519081900360200190f35b6001546001600160a01b03163314608a57600080fd5b60005430311015609957600080fd5b600080546040516001600160a01b0384169282156108fc02929190818181858888f1935050505015801560d0573d6000803e3d6000fd5b5050565b30319056fea165627a7a7230582074db1331b7f2d21b5bd86ef9604bbc95f33d7bc301e7616b5b12bc206d226f350029`
```
mikedoan: remove it

chaosma: removed

mikedoan:
can you test it with txgen too. ./test/deploy.sh -P ./test/conf...

chaosma:
I tested with txgen, it works fine. But txgen cannot do the stop/restart test because of the nonce not match for txgen transactions. I test the stop/restart using wallet which can pick the correct nonce for given accounts.

PR #619, opened 2019-03-22
Flat-rate block reward

harmony-ek:
Start paying a fixed block reward, currently set to 1/1000 per block per signature.

### Test Coverage Data

TBD

### Test/Run Logs

```
{"destLen":2,"ip":"127.0.0.1","lvl":"dbug","msg":"accumulateRewards: setting group sig mask","port":"9000","srcLen":2,"t":"2019-03-26T11:37:03.972277-07:00"}
{"ip":"127.0.0.1","lvl":"dbug","msg":"accumulateRewards: rewards have been paid out","numAccounts":9,"numSigs":9,"port":"9000","t":"2019-03-26T11:37:03.972631-07:00","totalAmount":"9000000000000000"}
```


harmony-ek:
Publishing PR for the overall idea.  Debugging an issue where blocks have empty commit bitmap header.

node/staking.go
```diff
@@ -29,6 +29,7 @@ const (
 // UpdateStakingList updates staking information by querying the staking smart contract.
 func (node *Node) UpdateStakingList(stakeInfoReturnValue *structs.StakeInfoReturnValue) {
 	node.CurrentStakes = make(map[common.Address]*structs.StakeInfo)
+	node.CurrentStakesByNode = make(map[[20]byte]*structs.StakeInfo)
```
LeoHChen: why hardcode the [20]byte ?

node/node.go
```diff
@@ -269,6 +294,7 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, db ethdb.Database) *N
 			// Setup one time smart contracts
 			node.AddFaucetContractToPendingTransactions()
 			node.CurrentStakes = make(map[common.Address]*structs.StakeInfo)
+			node.CurrentStakesByNode = make(map[[20]byte]*structs.StakeInfo)
```
LeoHChen: why hardcode the [20]byte

consensus/consensus.go
```diff
@@ -113,6 +122,34 @@ type Consensus struct {
...
+	return consensus.stakeInfoFinder
+}
+
+// SetStakeInfoFinder sets the stake information finder instance this
+// consensus uses, e.g. for block reward distribution.
+func (consensus *Consensus) SetStakeInfoFinder(stakeInfoFinder StakeInfoFinder) {
+	consensus.stakeInfoFinder = stakeInfoFinder
+}
+
+// StakeInfoFinder finds the staking account for the given consensus key.
```
LeoHChen: good interface design for reverse reference. :-D 👍

LeoHChen:
too many commits in the PR.  Should consider merge the small commits together.

LeoHChen:
0 consensus on Jenkins test.  Are you using the latest base?

harmony-ek:
> 0 consensus on Jenkins test. Are you using the latest base?

Yes.  I have actually seen the same issue, where consensus stops after a few blocks.  This is an intermittent failure, and I wonder if the master has the same issue.  Will try and test the master branch multiple times in a row.

harmony-ek:
@rlan35 @mikedoan @chaosma @LeoHChen please take a look.

node/staking.go
```diff
@@ -32,25 +32,79 @@ func (node *Node) UpdateStakingList(stakeInfoReturnValue *structs.StakeInfoRetur
...
+			logger.Info("expired stake; skipping")
+		}
+	}
+}
+
+// UpdateStakingListWithInitShardState updates the current stakes list with
+// initial hardcoded stakes.
+//
+// Initial stakers are assumed to be active since ever and until forever,
+// but with zero stake.
```
chaosma: what's the purpose of until forever with zero stake?

harmony-ek: They are the stakers hardcoded in the genesis block.  Eventually they will go away and be replaced by real ones.

harmony-ek:
Rebased and retargeted toward Banjo.  Travis CI build in progress and must pass.  Discounting Jenkins failures due to txgen not being able to process block rewards as a known issue.  Cc: @LeoHChen

cmd/client/txgen/main.go
```diff
@@ -123,6 +123,8 @@ func main() {
 	consensusObj := consensus.New(host, 0, nil, p2p.Peer{}, nil)
 	clientNode := node.New(host, consensusObj, nil, false)
 	clientNode.Client = client.NewClient(clientNode.GetHost(), shardIDs)
+	consensusObj.SetStakeInfoFinder(clientNode)
```
LeoHChen: It won't work with txgen, why do you need to change it here?

consensus/consensus.go
```diff
@@ -585,11 +622,68 @@ func (consensus *Consensus) Prepare(chain consensus_engine.ChainReader, header *
...
+		// without a parent block whose signer to reward.
+		return
+	}
+	if consensus.ChainReader == nil {
+		utils.GetLogInstance().Error("accumulateRewards: ChainReader is nil")
+		return
+	}
+	parent := consensus.ChainReader.GetBlockByHash(header.ParentHash)
+	if parent == nil {
+		utils.GetLogInstance().Error(
```
LeoHChen: please just return error in this function

consensus/consensus.go
```diff
@@ -585,11 +622,68 @@ func (consensus *Consensus) Prepare(chain consensus_engine.ChainReader, header *
...
+			"accumulateRewards: cannot retrieve parent block",
+			"curHash", header.Hash().Hex(),
+			"parentHash", header.ParentHash.Hex(),
+		)
+		return
+	}
+	mask, err := bls_cosi.NewMask(consensus.PublicKeys, consensus.PubKey)
+	if err != nil {
+		utils.GetLogInstance().Error(
```
LeoHChen: same as above, so many errors, please just return error.

consensus/consensus.go
```diff
@@ -585,11 +622,68 @@ func (consensus *Consensus) Prepare(chain consensus_engine.ChainReader, header *
...
+			"accumulateRewards: cannot set group sig mask bits", "error", err)
+		return
+	}
+	numSigs := 0
+	numAccounts := 0
+	totalAmount := big.NewInt(0)
+	for _, key := range mask.GetPubKeyFromMask(true) {
+		numSigs++
+		stakeInfos := consensus.stakeInfoFinder.FindStakeInfoByNodeKey(key)
+		if len(stakeInfos) == 0 {
```
LeoHChen: don't understand, if length is error, why you still set nodeAddr here?

consensus/consensus.go
```diff
@@ -585,11 +622,68 @@ func (consensus *Consensus) Prepare(chain consensus_engine.ChainReader, header *
...
+			nodeAddr := key.GetAddress()
+			utils.GetLogInstance().Error(
```
LeoHChen: It's a warning, right?  You just warn and continue.

consensus/consensus.go
```diff
@@ -585,11 +622,68 @@ func (consensus *Consensus) Prepare(chain consensus_engine.ChainReader, header *
...
+	if err := mask.SetMask(parentSigners); err != nil {
+		utils.GetLogInstance().Error(
+			"accumulateRewards: cannot set group sig mask bits", "error", err)
+		return
+	}
+	numSigs := 0
+	numAccounts := 0
+	totalAmount := big.NewInt(0)
+	for _, key := range mask.GetPubKeyFromMask(true) {
+		numSigs++
```
LeoHChen: numSigs is just the length of mask.GetPubKeyFromMask, right? it is not used, why do we need a variable here?

node/staking.go
harmony-ek: It's a convention copied from elsewhere where we store BLS addresses.  Proper fix requires replacing all occurrences of `[20]byte` for BLS addresses throughout the code.  Will track this in #670.

consensus/consensus.go
harmony-ek: <3

consensus/consensus_validator_test.go
```diff
@@ -48,6 +48,10 @@ func (MockChainReader) GetBlock(hash common.Hash, number uint64) *types.Block {
 	return &types.Block{}
 }
 
+func (MockChainReader) GetBlockByHash(hash common.Hash) *types.Block {
```
LeoHChen: Comment for a public function, please. If it is empty now, please add a TODO in the comment.

node/node.go
harmony-ek: Ditto.  See #670.

node/staking.go
```diff
@@ -32,25 +31,68 @@ func (node *Node) UpdateStakingList(stakeInfoReturnValue *structs.StakeInfoRetur
...
 	for i, addr := range stakeInfoReturnValue.LockedAddresses {
 		blockNum := stakeInfoReturnValue.BlockNums[i]
+		blsAddr := stakeInfoReturnValue.BlsAddresses[i]
 		lockPeriodCount := stakeInfoReturnValue.LockPeriodCounts[i]
-
 		startEpoch := core.GetEpochFromBlockNumber(blockNum.Uint64())
 		curEpoch := core.GetEpochFromBlockNumber(node.blockchain.CurrentBlock().NumberU64())
 
-		if startEpoch == curEpoch {
+		if startEpoch >= curEpoch {
```
LeoHChen: any comment on this change?

consensus/consensus.go
harmony-ek: This isn't that frequent.  Perhaps will happen once for genesis block, but shouldn't happen anymore after that.  If we see this message, it means that the current block is on chain but the previous block is not, which is a serious invariant violation: We deserve to be loud about that failure.

consensus/consensus.go
harmony-ek: Same; infrequent but serious.  If our consensus key is not found in the local consensus key list, it means we are in a shard where we shouldn't be.

consensus/consensus.go
harmony-ek: Good point, will fix.

consensus/consensus.go
harmony-ek: A signer's staking information not being found is a logical error which warrants a deeper look than just a warning, because it means that the BLS key somehow managed to sign a block while not having a valid stake—a chain security breach.

consensus/consensus_validator_test.go
harmony-ek: Will do.

node/staking.go
harmony-ek: This is for logical consistency/futureproofing.  `startEpoch` is the epoch in which the staking transaction was received.  The stake is valid from `(startEpoch + 1)` through `(startEpoch + lockPeriod)`; if `startEpoch` is equal to or greater than `curEpoch`, the stake is not yet active.

The current logic only captures `startEpoch == curEpoch` case, which is fine for now because a stake always begins to take effect in the immediately following epoch (`curEpoch + 1`), so `startEpoch` is always `curEpoch` when a stake is received.  However, if we relax the rule so that a stake can start taking effect farther down the road, the equality check will fail to catch such future stakes.  `startEpoch >= curEpoch` captures both cases.  Cc: @rlan35

consensus/consensus.go
harmony-ek: Fixed in commit 715a8ffc6d64b4d621bbfcf9d492cb78073511da.

consensus/consensus_validator_test.go
harmony-ek: Fixed in commit a8048ca0f923562b48478ce532e2086aa5ddff9b.

cmd/client/txgen/main.go
harmony-ek: Once txgen is fixed to deal with staking, the call will become necessary here.  But for now I'll back it out so it doesn't get into others' merges.

harmony-ek: Fixed in b7df7eec4de725f935b1cc495c228ad9d59670f2.

PR #620, opened 2019-03-22
Other commits for demo dapp

mikedoan:
## Issue

Other commits for demo dapp


api/proto/message/server.go
```diff
@@ -5,6 +5,7 @@ package message
 // TODO(minhdoan): Refactor and clean up the other client service.
 import (
 	"context"
+	fmt "fmt"
```
chaosma: no need to rename the same

node/service_setup.go
```diff
@@ -60,6 +61,10 @@ func (node *Node) setupForBeaconLeader() {
 	node.serviceManager.RegisterService(service.BlockProposal, blockproposal.New(node.Consensus.ReadySignal, node.WaitForConsensusReady))
 	// Register client support service.
 	node.serviceManager.RegisterService(service.ClientSupport, clientsupport.New(node.blockchain.State, node.CallFaucetContract, node.getDeployedStakingContract, node.SelfPeer.IP, node.SelfPeer.Port))
+	// TODO(minhdoan): We will remove the old client support and use the new client support which uses new message protocol.
+	// Register client new support service.
+	node.serviceManager.RegisterService(service.RestClientSupport, restclientsupport.New(
```
LeoHChen: so, it only works for beacon leader?

api/proto/message/server.go
```diff
@@ -34,8 +35,8 @@ func (s *Server) Process(ctx context.Context, message *Message) (*Response, erro
 	if message.GetType() != MessageType_LOTTERY_REQUEST {
 		return &Response{}, ErrWrongMessage
 	}
-	utils.GetLogInstance().Info("Recieved:", "message", message)
 	lotteryRequest := message.GetLotteryRequest()
+	fmt.Println("lottery request", lotteryRequest)
```
LeoHChen: log doesn't work here?

node/service_setup.go
mikedoan: as of now - i can add todo later. 

the reason is that for demo we only deploy the lottery contract in beacon chain. in the future we should provide developers with capability to deploy contract in any shard and we should have client support for each shard.

api/proto/message/server.go
mikedoan: removed

Issue #611, opened 2019-03-21
new node join testnet isssues


LeoHChen:
**Describe the bug**

I launched a testnet and run txgen/wallet for consensus for some time.  When I started a new node, I can see the new node had sent ping message to network and leader received it.

However, there are two issues here.

1) I didn't see state syncing working.  
2) Also, new node didn't get pong message, that's it failed to verify the leader messages.

chaosma:
The pong message didn't fully works. 
Test 1:  use only one new node and join the network 30s later.  Pass test
Test 2: use two new nodes, one join network after 30s and the other new node join network after 60s. The second one didn't receive pong message with leader's public key.

LeoHChen:
The problem is solved with changes from me and chao.

PR #621, opened 2019-03-22
[pong] actively sending pong message

LeoHChen:


LeoHChen:
This is to fix the first issue of new node. Now, new node received Pong message and can join the consensus.
The consensus is not fully working yet, due to state sync not happening.

https://github.com/harmony-one/harmony/issues/611

PR #628, opened 2019-03-23
Store group sigs in the real header, not in a copy

harmony-ek:
(*Block) Header() returns a (shallow) copy of the header, where we were
storing group signatures and signer bitmaps.  This had no effect, and
the on-chain block was still missing these signatures.


core/types/block.go
```diff
@@ -159,6 +161,28 @@ type Block struct {
 	ReceivedFrom interface{}
 }
 
+// SetPrepareSig sets the block's prepare group signature.
+func (b *Block) SetPrepareSig(sig []byte, signers []byte) {
```
LeoHChen: please add unit test.

core/types/block.go
```diff
@@ -159,6 +161,28 @@ type Block struct {
...
+		utils.GetLogInstance().Warn("SetPrepareSig: sig size mismatch",
+			"srcLen", len(sig),
+			"dstLen", len(b.header.PrepareSignature))
+	}
+	copy(b.header.PrepareSignature[:], sig[:])
+	b.header.PrepareBitmap = append(signers[:0:0], signers...)
+}
+
+// SetCommitSig sets the block's commit group signature.
+func (b *Block) SetCommitSig(sig []byte, signers []byte) {
```
LeoHChen: please add unit test.

core/types/block.go
harmony-ek: Added.

core/types/block.go
harmony-ek: Added.

harmony-ek:
@LeoHChen @rlan35 UT added, please let me know if this LGTY.

PR #640, opened 2019-03-24
[test only] update bootnode/ip for the devnet

LeoHChen:
This PR is for reference only.  If you want to test wallet on our devnet, you may need to update the bootnode/rpc server list.


LeoHChen:
no longer needed. I have added an environment variable of RPC nodes support in wallet.

PR #642, opened 2019-03-24
[wallet] support rpc server using environment variable

LeoHChen:
use an external environment variable to specify the rpc servers
of the wallet. this will greatly simplify the testing.

example of the environment variable:

export RpcNodes="127.0.0.1:8989,192.168.129.1:5555"


cmd/client/wallet/main.go
```diff
@@ -134,10 +142,29 @@ func main() {
...
+		default:
+			break ARG
+		}
+	}
+
+	if len(os.Getenv("RpcNodes")) > 0 {
+		rpcServers = utils.StringsToPeers(os.Getenv("RpcNodes"))
+	}
+	if len(rpcServers) == 0 {
+		fmt.Println("Error: please set environment variable RpcNodes")
```
lzl124631x: nit: have a const for env var string, so that we can refer to it and keep consistant.

PR #644, opened 2019-03-25
[pong] send regular pong message to new peers

LeoHChen:
This shall solve the issue of new node didn't get the pong message issue.

### DIFF
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -396,6 +396,8 @@ func (node *Node) pingMessageHandler(msgPayload []byte, sender string) int {
 // SendPongMessage is the a goroutine to periodcally send pong message to all peers
 func (node *Node) SendPongMessage() {
 	tick := time.NewTicker(2 * time.Second)
+	tick2 := time.NewTicker(120 * time.Second)
+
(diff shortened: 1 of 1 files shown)


node/node_handler.go
```diff
@@ -431,19 +433,33 @@ func (node *Node) SendPongMessage() {
...
 					sentMessage = true
-					// stop sending ping message
-					node.serviceManager.TakeAction(&service.Action{Action: service.Stop, ServiceType: service.PeerDiscovery})
 					// wait a bit until all validators received pong message
 					time.Sleep(200 * time.Millisecond)
+
 					// only need to notify consensus leader once to start the consensus
 					if firstTime {
+						// Leader stops sending ping message
+						node.serviceManager.TakeAction(&service.Action{Action: service.Stop, ServiceType: service.PeerDiscovery})
```
rlan35: Why the leader stops sending ping here?

LeoHChen: main purpose of ping message is to send pubkey.  Leader's pubkey will be sent in pong message.  I think there is no need for leader to keep sending ping message then.

PR #645, opened 2019-03-25
[DO NOT REVIEW]

rlan35:
Before our network only support one shard of beacon chain and our genesis sharding state only have beacon chain populated. Now we have sharded p2p network and also 1+3 gensis shards.

Tested locally with 5 node each shard and 10 node each shard.


LeoHChen:
can you squash your commits into a big commit?

rlan35:
> can you squash your commits into a big commit?

PR #653, opened 2019-03-27
[log] print a message in validator if consensus reached

LeoHChen:
This will help the external node runner to check on their log.

### DIFF
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -281,7 +281,10 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block) {
 	utils.GetLogInstance().Info("PostConsensusProcessing")
 	if node.Consensus.IsLeader {
 		node.BroadcastNewBlock(newBlock)
+	} else {
+		utils.GetLogInstance().Info("BINGO !!! Reached Consensus", "ConsensusID", node.Consensus.GetConsensusID())
 	}
+
 	node.AddNewBlock(newBlock)
 
 	// TODO: enable drand only for beacon chain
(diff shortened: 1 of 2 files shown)


LeoHChen:
Jenkins, test this please.

node/node_handler.go
```diff
@@ -281,7 +281,10 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block) {
 	utils.GetLogInstance().Info("PostConsensusProcessing")
 	if node.Consensus.IsLeader {
 		node.BroadcastNewBlock(newBlock)
+	} else {
+		utils.GetLogInstance().Info("!!! BINGO !!!", "Reached Consensus", node.Consensus.GetConsensusID())
```
lzl124631x: nit: "reached consensus" is used as the key to `consensusID` which is not as straightforward as "ConsensusID" and moving "Reached Consensus" to after "Bingo"

LeoHChen: sure. changed.

node/node_handler.go
```diff
@@ -281,7 +281,10 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block) {
 	utils.GetLogInstance().Info("PostConsensusProcessing")
 	if node.Consensus.IsLeader {
 		node.BroadcastNewBlock(newBlock)
+	} else {
+		utils.GetLogInstance().Info("BINGO !!! Reached Consensus", "ConsensusID", node.Consensus.GetConsensusID())
 	}
```
alajko: Is there a validator related ID we can include here. Else we will have a flood of messages with same `consensusID`? consensusID is always `0`?

LeoHChen: consensusID change after each consensus.  This is per validator message, it won't flood.

LeoHChen:
Jenkins, test this please.

PR #659, opened 2019-03-28
[log] disable log from rpc server

LeoHChen:
The excessive log on RPC server shall be disabled.


LeoHChen:
unless it goes to log file. For now, I just disable it.

PR #664, opened 2019-03-29
Initialize committee pubkey with genesis shard state; Change bls address to pubkey in staking contract.

rlan35:
Our previous setup is to populate consensus committee keys on the fly based on ping pong message, which was a temporary solution to make testnet work. This PR added the necessary code to setup the initial committee based on genesis shard state. This PR should be backward compatible as the ping/pong logic are not completely removed yet.


node/node.go
```diff
@@ -256,8 +258,6 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, db ethdb.Database, is
 			os.Exit(1)
 		}
 		node.blockchain = chain
-		// Store the genesis shard state into db.
-		node.blockchain.StoreNewShardState(node.blockchain.CurrentBlock(), nil)
```
LeoHChen: why this is removed?

node/staking.go
```diff
@@ -44,8 +46,12 @@ func (node *Node) UpdateStakingList(stakeInfoReturnValue *structs.StakeInfoRetur
 		}
 		// True if the token is still staked within the locking period.
 		if curEpoch-startEpoch <= lockPeriodCount.Uint64()*lockPeriodInEpochs {
+			blsPubKey := types.BlsPublicKey{}
+			copy(blsPubKey[:32], stakeInfoReturnValue.BlsPubicKeys1[i][:])
```
LeoHChen: why copy 3 part?

consensus/consensus.go
```diff
@@ -499,6 +513,12 @@ func (consensus *Consensus) DebugPrintValidators() {
 func (consensus *Consensus) UpdatePublicKeys(pubKeys []*bls.PublicKey) int {
 	consensus.pubKeyLock.Lock()
 	consensus.PublicKeys = append(pubKeys[:0:0], pubKeys...)
+	consensus.CommitteeAddresses = map[common.Address]bool{}
+	for _, pubKey := range consensus.PublicKeys {
+		consensus.CommitteeAddresses[utils.GetBlsAddress(pubKey)] = true
```
chaosma: if a validator not in the committee, does its key not exist here or value is false?

core/types/shard_state.go
```diff
@@ -11,10 +12,13 @@ import (
 // ShardState is the collection of all committees
 type ShardState []Committee
 
+// BlsPublicKey defines the bls public key
+type BlsPublicKey [96]byte
+
 // NodeID represents node id (BLS address).
 type NodeID struct {
 	EcdsaAddress string
```
chaosma: may use Address type instead of string as well

node/node.go
rlan35: It's not removed. it's moved to the other method InitGenesisShardState

node/staking.go
rlan35: because public key is 96 bytes and solidity doesn't support long byte array very well

consensus/consensus.go
rlan35: Just not exist, there won't be false value. The map is used as a set

core/types/shard_state.go
rlan35: using fixed type is better than using string. Easier to debug.

rlan35:
> Should add the flag to disable the original ping/pong message, to make sure the new mechanism works. How do you solve the problem of the new node then?

New node should still work. As I didn't change the previous pingpong logic.

Issue #675, opened 2019-04-03
integer divide by zero crash


LeoHChen:
**Describe the bug**

On latest banjo branch, launched test net and git the following errors on the validators.

```
2019/04/02 04:29:02 Command running ./harmony [-ip 52.208.78.188 -port 9000 -log_folder ../tmp_log/log-20190402.042758 -bootnodes /ip4/54.213.43.194/tcp/9876/p2p/QmQayinFSgMMw5cSpDUiD9pQ2WeP6WNmGxpZ6ou3mdVFJX,/ip4/100.26.90.187/tcp/9876/p2p/QmZJJx6AdaoEkGLrYG4JeLCKeCKDjnFz2wfHNHxAqFSGA9 -min_peers 36 -is_beacon -account_index 48]
2019/04/02 04:47:42 Stderr 2019/04/02 04:29:02 No priviate key can be loaded from filekeyfile./.hmykey
2019/04/02 04:29:02 Using random private key
panic: runtime error: integer divide by zero

goroutine 27944 [running]:
github.com/harmony-one/harmony/core.(*ShardingState).assignNewNodes(0xc0048a9dd0, 0xc00516a800, 0x32, 0x40)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/core/resharding.go:57 +0x28b
github.com/harmony-one/harmony/core.(*ShardingState).Reshard(0xc0048a9dd0, 0xc00516a800, 0x32, 0x40, 0x3fb999999999999a)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/core/resharding.go:108 +0x88
github.com/harmony-one/harmony/core.CalculateNewShardState(0xc002ce6000, 0x7, 0xc00003c1c8, 0x1a64b96163ecf822, 0x1215050672bb47fa, 0x46)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/core/resharding.go:169 +0x7e8
github.com/harmony-one/harmony/core.(*BlockChain).GetNewShardState(0xc002ce6000, 0xc002cda090, 0xc00003c1c8, 0xc003c318d8, 0x411b28, 0xc001daa1e0)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/core/blockchain.go:1716 +0x15b
github.com/harmony-one/harmony/core.(*BlockChain).StoreNewShardState(0xc002ce6000, 0xc002cda090, 0xc00003c1c8, 0x0, 0x0, 0x0)
[...]

PR #679, opened 2019-04-03
Log and handle shard state retrieval errors

harmony-ek:
## Issue

This is a partial WIP for #675.  Among other things, this PR eliminates the use of nil pointers returned upon failure.

### DIFF
--- a/core/resharding.go
+++ b/core/resharding.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/harmony-one/bls/ffi/go/bls"
 	"github.com/harmony-one/harmony/contracts/structs"
+	"github.com/harmony-one/harmony/internal/ctxerror"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/internal/utils/contract"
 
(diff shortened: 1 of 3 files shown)


core/resharding.go
```diff
@@ -132,22 +133,47 @@ func GetEpochFromBlockNumber(blockNumber uint64) uint64 {
...
+		return nil, ctxerror.New("cannot get shard state for epoch",
+			"epoch", epoch).WithCause(err)
+	}
 
 	rndSeedBytes := bc.GetRandSeedByNumber(number)
+	if rndSeedBytes == [32]byte{} {
+		utils.GetLogInstance().Warn("cannot get randomness from block",
+			"blockNumber", number)
+		// TODO ek – should we fail here?  Continuing here would end up using
+		//  all-zeros for randomness, which seems like a security breach.
```
harmony-ek: @rlan35 what do you think?

core/resharding.go
```diff
@@ -132,22 +133,47 @@ func GetEpochFromBlockNumber(blockNumber uint64) uint64 {
...
 	}
-	ss := GetShardingStateFromBlockChain(bc, epoch-1)
+	ss, err := GetShardingStateFromBlockChain(bc, epoch-1)
+	if err != nil {
+		utils.GetLogInstance().Crit("CANNOT CALCULATE SHARDING STATE",
+			"epoch", epoch)
+		// TODO ek – what should we do here?  We are processing a new epoch
+		//  block, so we better have the new sharding state before advancing
+		//  to the next block (which would require the new sharding state),
+		//  but the sharding state is not available here.
```
harmony-ek: @chaosma @rlan35, what do you think?

core/resharding.go
```diff
@@ -132,22 +133,47 @@ func GetEpochFromBlockNumber(blockNumber uint64) uint64 {
 }
 
 // GetShardingStateFromBlockChain will retrieve random seed and shard map from beacon chain for given a epoch
-func GetShardingStateFromBlockChain(bc *BlockChain, epoch uint64) *ShardingState {
+func GetShardingStateFromBlockChain(bc *BlockChain, epoch uint64,
+) (*ShardingState, error) {
 	number := GetBlockNumberFromEpoch(epoch)
-	shardState := bc.GetShardStateByNumber(number)
+	shardState, err := bc.GetShardStateByNumber(number)
```
harmony-ek: BTW, we have both `ShardingState` and `ShardState` (confusing naming).  The former is a more comprehensive collection of sharding information including randomness, and the latter is just typedef-ed as a list of `Committee`-s.  Maybe we should rename `ShardState` to `Committees` to better reflect this.  Cc: @chaosma

core/resharding.go
harmony-ek: BTW this question is essential to resolving this issue.

core/resharding.go
rlan35: sure. sounds good. Leave it as TODO for now.

core/resharding.go
rlan35: Leave it as todo for now. The resharding code needs some final tweaking. We still need more work to finalize the code.

PR #686, opened 2019-04-06
Fix out-of-range index

### LINKED ISSUE #676: index out of range
**Describe the bug**

On the recently launched testnet using Banjo branch, the following bug was found that crashed the node.

```
panic: runtime error: index out of range

goroutine 474038 [running]:
github.com/harmony-one/harmony/api/service/syncing.(*StateSync).RegisterNodeInfo(0xc0014b2200, 0xc003d720a0)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/api/service/syncing/syncing.go:547 +0xa5a
github.com/harmony-one/harmony/node.(*Node).CalculateResponse(0xc0000fe000, 0xc005e587e0, 0xdd4340, 0xc003d72000, 0x11efb80)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/node/node_syncing.go:243 +0xb0e
github.com/harmony-one/harmony/api/service/syncing/downloader.(*Server).Query(0xc0001f6070, 0x12068e0, 0xc001d9fef0, 0xc005e587e0, 0xc0001f6070, 0xc001d9fef0, 0xc00585abd0)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/api/service/syncing/downloader/server.go:24 +0x3d
github.com/harmony-one/harmony/api/service/syncing/downloader/proto._Downloader_Query_Handler(0xdd4340, 0xc0001f6070, 0x12068e0, 0xc001d9fef0, 0xc003d72000, 0x0, 0x12068e0, 0xc001d9fef0, 0xc00375e480, 0x41c)
        /home/ec2-user/go/src/github.com/harmony-one/harmony/api/service/syncing/downloader/proto/downloader.pb.go:318 +0x23e
google.golang.org/grpc.(*Server).processUnaryRPC(0xc001180480, 0x1210960, 0xc005c5d200, 0xc0024c4200, 0xc00267c450, 0x1addd90, 0x0, 0x0, 0x0)
        /home/ec2-user/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:971 +0x4b5
google.golang.org/grpc.(*Server).handleStream(0xc001180480, 0x1210960, 0xc005c5d200, 0xc0024c4200, 0x0)
        /home/ec2-user/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:1250 +0xd25
google.golang.org/grpc.(*Server).serveStreams.func1.1(0xc005b295e0, 0xc001180480, 0x1210960, 0xc005c5d200, 0xc0024c4200)
        /home/ec2-user/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:690 +0x9f
created by google.golang.org/grpc.(*Server).serveStreams.func1
        /home/ec2-user/go/pkg/mod/google.golang.org/grpc@v1.19.0/server.go:688 +0xa1
```

harmony-ek:
## Issue

There were a few places where `StateSync.syncConfig.peers` were iterated over without proper locking.  Introduce read/write mutex (`SyncConfig.mtx`) and hold it appropriately.

In order to simplify and faciliate this locking, this PR also includes elimination of sync peers whose `client` is `nil`.  This enables folding a few methods down one level (from `StateSync` to `SyncConfig`) so that sync config mutex need not be acquired externally.

Fixes #676.

## Test

Local test succeeds (proceeds past initial syncing).

## TODO

Test this in testnet-like (no txgen) setup.


node/node_syncing.go
```diff
@@ -61,10 +62,9 @@ SyncingLoop:
 				node.stateSync = syncing.CreateStateSync(node.SelfPeer.IP, node.SelfPeer.Port, node.GetSyncID())
 			}
 			if node.stateSync.GetActivePeerNumber() == 0 {
-				if node.stateSync.CreateSyncConfig(getPeers()) {
-					node.stateSync.MakeConnectionToPeers()
-				} else {
-					utils.GetLogInstance().Debug("[SYNC] no active peers, continue SyncingLoop")
+				peers := getPeers()
+				if err := node.stateSync.CreateSyncConfig(peers); err != nil {
+					ctxerror.Log15(utils.GetLogInstance().Debug, err)
```
LeoHChen: what's this?

harmony-ek:
I disagree that this is reinventing the wheel. 
 sync.Map is a specialized type for one-member data structure, and does not eliminate the need for an independent mutex that guards against inter-member invariants.  [Even its GoDoc recognizes this](https://golang.org/pkg/sync/#Map): "The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content."

node/node_syncing.go
harmony-ek: ctxerror is the structured (context-aware) error support added in response to your comment about #675.  It enables a function to return an error with named context objects, much like how log15 loggers take message along with named context objects.

`Log15` takes a log15-style error logging function and an error instance, and simply logs the error message if the error instance is a genetic error type, or logs the message with context parameters if the error instance is a ctxerror.  I agree that the syntax is a bit backward/functional-programming-y/feels like a Russian reversal joke; suggestions are welcome on how to improve this.

LeoHChen: why only here use this Log15 function?

LeoHChen:
> I disagree that this is reinventing the wheel.
> sync.Map is a specialized type for one-member data structure, and does not eliminate the need for an independent mutex that guards against inter-member invariants. [Even its GoDoc recognizes this](https://golang.org/pkg/sync/#Map): "The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content."

You are using a slice, and most use case of the slice is like a map.  Why don't you use a map directly?
If you use a map and add a lock, can you convenience me that there is a special requirement to use our own lock to protect the map other than the standard sync.Map.

The code should be written for readability and maintainability.
If there is no strong reason, I'd use the standard library instead of our own implementation.

harmony-ek:
Map-level safety (a guarantee that concurrent access to the map won't crash the program) isn't the only thing that we are trying to preserve here.  We are trying to guarantee is something stronger than just map-level safety: We need to ensure that members aren't changing while the set of members as a whole.

Take `cleanupPeers()` as an example.  It tries to delete peers not sharing the same block hash list as the given peer, as part of a bigger task—`GetBlockHashesConsensusAndCleanUp()`—whose purpose is to analyze the peer list and leave only the largest group of peers in terms of their agreement in block hash list.  This task requires an unchanging view of the entire list while the analysis and cleanup is in progress, i.e. no one should be allowed to update the map concurrently.  If you could show me a way of doing this using `sync.Map` without introducing a concurrent update race, I'd be all ears.

harmony-ek:
Retargeting for the master branch.

harmony-ek:
Build failed for `ctxerror` was missing; cherry-picked them from banjo.

node/node_syncing.go
harmony-ek: I did this specific error refactoring because it was blocking a clean reduction of layering interdependencies (and also, returning only a true/false boolean indicator to the caller is not an idiomatic Go pattern).  Feel free to use `ctxerror` elsewhere you see fit though.

internal/ctxerror/ctxerror.go
```diff
@@ -0,0 +1,124 @@
...
+// WithCause chains an error after the receiver.
+// It returns the merged/chained instance,
+// where the message is “<receiver.Message>: <c.Message>”,
+// and with contexts merged (ones in c takes precedence).
+func (e *ctxError) WithCause(c error) CtxError {
+	r := &ctxError{msg: e.msg + ": ", ctx: make(map[string]interface{})}
+	for k, v := range e.ctx {
+		r.ctx[k] = v
+	}
+	switch c := c.(type) {
```
mikedoan: lgtm overall, i'm just curious c is already an error - why is it here?

harmony-ek: @mikedoan, that's a [variable-shadowing type switch idiom](https://golang.org/doc/effective_go.html#type_switch).  The left-hand-side variable `c` is local to the switch block scope, and shadows the original `c` on the right-hand side, which is in the enclosing scope.  Since it is a type switch, the LHS gets the same value as the RHS, so this has the effect of dynamically typecasting `c` down to more specific types:

```go
var v interface{}
v = /* something */
switch v := v.(type) {
// In all of the cases below,
// the shadowing (LHS) v has the same value as the original (RHS) v.
case TypeA:
    // v is downcasted to TypeA in this case
    /* ... */
case TypeB:
    // v is downcasted to TypeB in this case
    /* ... */
default:
    // v is still interface{} in this case
    /* ... */
}
```

PR #687, opened 2019-04-06
add -race for building binary by default

mikedoan:
## Issue

As we are working on stability. add -race for building binary by default.


mikedoan:
Jenkins, test this please.

PR #688, opened 2019-04-06
change to libp2p.ID for nodeID in /shard

lzl124631x:
Currently the map `validators` is a private field. I think exposing the mapping between peerID and blsPubKey might not be desirable? (WDYT). So I just changed them to peerIDs.


LeoHChen:
Looks good in general.  Please provide an URL that we can look,  Or attach a screenshot to the PR.

lzl124631x:
@LeoHChen  
image

PR #695, opened 2019-04-08
Do not try to add self key to mask

harmony-ek:
This may fail the newnode case because the new node may still be trying to sync up before joining the consensus, hence its key is not in `consensus.PublicKeys` yet.

When this happens, the new node will not be able to sync blocks and won't join consensus.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -647,7 +647,7 @@ func (consensus *Consensus) accumulateRewards(
 			"parentHash", header.ParentHash.Hex(),
 		)
 	}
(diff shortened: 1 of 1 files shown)


harmony-ek:
@chaosma, yes, this addresses one of the issues we are seeing.  Needs to go in though, because this brings the code one step closer in the right direction to the resolution of the issue.

harmony-ek:
IOW, please evaluate this fix in its own merit, not just in the bigger frame.

chaosma:
Ship it!

PR #704, opened 2019-04-10
[debug] add debug option in build

LeoHChen:

This PR is to enable debug option for go build by disabling optimizations so that it is friendlier to the debugger.


LeoHChen:
Jenkins, test this please.

Issue #714, opened 2019-04-12
[Feature] display the real block number in explorer


LeoHChen:
# Problem/limitation At Hand

I launched a new testnet with DB persisted from the previous launch, which means the blockchain has 2000+ blocks already mined and all nodes have restored the same blockchain already.  The blockchain runs fine and can keep doing consensus with the new transaction.  However, on the explorer, I can see the block height starts from 0 again, even though on the log file, we knew the block height starts from 2000 already. 

**Which component?**

- [ ] consensus
- [ ] deploy script
- [ ] wallet
- [x] explorer
- [ ] node script
- [ ] test automation
- [ ] demo app
- [ ] others ...

# Proposed Solution

The explorer should display the real block height number from the blockchain when it display the data.

# Alternatives Considered

# Additional Context

lzl124631x:
I suspect it's because the explorer storage (see explorer/storage.go) doesn't pick up from where it stopped. + @mikedoan to take a look. Thanks.

LeoHChen:
A workaround fix: https://github.com/harmony-one/harmony/pull/717

PR #717, opened 2019-04-12
Set init consensus ID

LeoHChen:
The consensusID always starts from 0, which is used to display as the block height in explorer in our current design. This can be confusing as in Issue (https://github.com/harmony-one/harmony/issues/714)

So, the solution is when we launch a blockchain with persisted data from the previous launch, we set the init consensus ID to be the block height restored from persisted data. So, it at least displays the right block height.

### DIFF
--- a/test/deploy.sh
+++ b/test/deploy.sh
@@ -8,6 +8,8 @@ USER=$(whoami)
 set -x
 set -eo pipefail
 
+export GO111MODULE=on
+
 function check_result() {
    find $log_folder -name leader-*.log > $log_folder/all-leaders.txt
    find $log_folder -name validator-*.log > $log_folder/all-validators.txt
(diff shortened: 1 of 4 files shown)


test/deploy.sh
```diff
@@ -8,6 +8,8 @@ USER=$(whoami)
 set -x
 set -eo pipefail
 
+export GO111MODULE=on
+
```
alajko: Thanks.

lzl124631x:
So I can see height starting from 100 but I can't see block data of blocks 0~99. Right?

LeoHChen:
> So I can see height starting from 100 but I can't see block data of blocks 0~99. Right?

yes, that's the missing part of our current API.

LeoHChen:
Jenkins, test this please.

LeoHChen:
Jenkins, test this please.

LeoHChen:
Jenkins, test this please.

PR #728, opened 2019-04-16
Fix multi shard client communication

LeoHChen:
## Issue

The current 1+3 shards p2p communication is using beacon client topic for all client communication, which is wrong.  Each shard should have its own client topic for communication with their own clients.

This patch set fixed the client side multi-shard communication by using shard specific topic for communication. Each shard has a unique topic to talk to shard client.

## Test

Tested locally by:

* launch a blockchain with 4 shards
* use local wallet.ini to talk to four shards
* getFreeToken from all shards
* do the token transfer on different shards with different tx amount
* all tx went through

#### Test/Run Logs

The full test log is available at https://gist.github.com/LeoHChen/d7161b16f5f8ecf3c28eb05c23ccc11b


LeoHChen:
Jenkins, test this please.

LeoHChen:
Jenkins, test this please.

Issue #749, opened 2019-04-20
invalid memory access caused crash


LeoHChen:
**Describe the bug**

Get a recent wallet crash with the following errors.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xcc84b6]

goroutine 29 [running]:
github.com/harmony-one/harmony/p2p/host/hostv2.(*GroupReceiverImpl).Receive(0x0, 0x1229380, 0xc0000b6050, 0x1213280, 0xc0000105a0, 0x0, 0x0, 0x0, 0x0, 0x0)
   /home/leochen/go/src/github.com/harmony-one/harmony/p2p/host/hostv2/hostv2.go:90 +0x26
github.com/harmony-one/harmony/node.(*Node).ReceiveGlobalMessage(0xc000328000)
   /home/leochen/go/src/github.com/harmony-one/harmony/node/node_handler.go:45 +0x83
created by github.com/harmony-one/harmony/node.New
   /home/leochen/go/src/github.com/harmony-one/harmony/node/node.go:310 +0x5c1
```

PR #750, opened 2019-04-20
Fix null pointer error #749

LeoHChen:
## Issue

Fix issue https://github.com/harmony-one/harmony/issues/749

### DIFF
--- a/p2p/host/hostv2/hostv2.go
+++ b/p2p/host/hostv2/hostv2.go
@@ -88,7 +88,7 @@ func (r *GroupReceiverImpl) Receive(ctx context.Context) (
 	msg []byte, sender libp2p_peer.ID, err error,
 ) {
 	if r.sub == nil {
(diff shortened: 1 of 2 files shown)


p2p/host/hostv2/hostv2.go
```diff
@@ -87,6 +87,9 @@ func (r *GroupReceiverImpl) Close() error {
 func (r *GroupReceiverImpl) Receive(ctx context.Context) (
 	msg []byte, sender libp2p_peer.ID, err error,
 ) {
+	if r.sub == nil {
+		return nil, libp2p_peer.ID(""), fmt.Errorf("Receive: r.sub == nil")
```
harmony-ek: Thanks for catching this.  `r.sub == nil` means that the receiver has been closed (see `Close()` above), so a more descriptive error message would be:
```suggestion
		return nil, libp2p_peer.ID(""), fmt.Errorf("GroupReceiver has been closed")
```

PR #762, opened 2019-04-22
[docker] update docker-node.sh

LeoHChen:
update valid range.

### DIFF
--- a/scripts/docker/docker-node.sh
+++ b/scripts/docker/docker-node.sh
@@ -32,7 +32,7 @@ shift $(($OPTIND-1))
 account_id=$1
 
 if [ -z "$account_id" ]; then
-  echo "Please provide account id. Valid ranges are 46-49, 96-99, 146-149, 196-199."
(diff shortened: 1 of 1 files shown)


scripts/docker/docker-node.sh
```diff
@@ -32,7 +32,7 @@ shift $(($OPTIND-1))
 account_id=$1
 
 if [ -z "$account_id" ]; then
-  echo "Please provide account id"
+  echo "Please provide account id. Valid ranges are 46-49, 96-99, 146-149, 196-199."
```
chaosma: why other id not valid?

harmony-ek: @chaosma, they are used by Harmony-operated genesis nodes, and are not available to the end users (node operators), for whom this message is intended.

BTW @LeoHChen, here's a snippet:

```suggestion
  echo "Please provide account index.  Valid ranges are 46-49, 96-99, 146-149, 196-199.  Please contact Harmony if not sure which one to use."
```

chaosma:
> One suggestion below.

@harmony-ek  In this case, we could check user input to make sure their inputs are valid

PR #773, opened 2019-04-24
Add gomock matcher utilities

harmony-ek:
SSIA.  These matchers can be passed as gomock expected call arguments.  I use them in my other UTs for the upcoming block rewards code.

```go
mock.EXPECT().Method(matchers.Slice{3, gomock.Not(4), 5})
```

ensures that the method is called with a three-member slice or array, whose first/third element must be 3 and 5 respectively, and the second element must not be 4.

```go
mock.EXPECT().Method(matchers.Struct{"A": 10, "B": 20})
```

ensures that the method is called with a struct or a pointer to a struct with two data members named `A` and `B`, and with values 10 and 20 respectively.

```go
mock.EXPECT().Method(matchers.Path("core/types/block.go"))
```

ensures that the method is called with a pathname whose trailing part is `core/types/block.go`, such as:

* `core/types/block.go`
* `github.com/harmony-one/harmony/core/types/block.go`
* `/Users/ek/h/core/types/block.go`

(but not `hardcore/types/block.go`).

See godoc for details.

### Test Coverage Data

```
ok  	github.com/harmony-one/harmony/gomock_matchers	0.008s	coverage: 100.0% of statements in ../../harmony/...
```

### DIFF
--- a/gomock_matchers/path.go
+++ b/gomock_matchers/path.go
@@ -1,0 +1,35 @@
+package matchers
+
+import (
+	"fmt"
+	"path"
+	"strings"
+)
+
+// Path is a pathname matcher.
+//
+// A value matches if it is the same as the matcher pattern or has the matcher
+// pattern as a trailing component.  For example,
+// a pattern "abc/def" matches "abc/def" itself, "omg/abc/def",
+// but not "abc/def/wtf", "abc/omg/def", or "xabc/def".
+//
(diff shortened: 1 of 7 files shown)


harmony-ek:
> lgtm - assume you already tested

Haha yep, this is both unit-tested 100% and also battle-tested in the block reward code.

PR #774, opened 2019-04-24
Log call sites

harmony-ek:
## Summary

This PR introduces `GetLogger`, `WithCaller`, and `WithCallerSkip`, which add call site information (function name, filename, line number) as extra contexts `funcName`, `funcFile`, `funcLine` to a logger.  It helps track down which function emitted the log and also quickly locate logs emitted by the target function.  (This could also be used as a groundwork for dynamic logging.)

* `funcName` is a fully qualified function name, with package and receiver (if applicable) information.
* `funcFile` is the basename of the source file in which the call site resides.
* `funcLine` is the line number of the call site in the source file.

A sample triple:

* `funcName`: `"github.com/harmony-one/harmony/core/types".(*Block).Hash`
* `funcFile`: `block.go`
* `funcLine`: 1234

## Usage

Use `GetLogger()` for direct one-shot logging calls:

```go
utils.GetLogger().Debug("omg")
```

Wrap reusable loggers with `WithCaller()`:

```go
logger := utils.GetLogInstance().New("blockHash", b.Hash())

/* ... */

utils.WithCaller(logger).Debug("invalid block header")
```

Or use a local `getLogger()` idiom:

```go
logger := utils.GetLogInstance().New("blockHash", b.Hash())
getLogger := func() log.Logger { return WithCallerSkip(logger, 1) }

/* ... */

getLogger().Debug("invalid block header")
```

Since `getLogger()` uses closure, logger can later be augmented with more context with `logger = logger.New(…)` syntax.

## Test

image


chaosma:
Nice!  For usage, what is the pattern for logger. Previously, the logger is like  logger.Debug("message","key1","val1","key2","val2",...) . Are there any more patterns allowed?

harmony-ek:
> Nice! For usage, what is the pattern for logger. Previously, the logger is like logger.Debug("message","key1","val1","key2","val2",...) . Are there any more patterns allowed?

The existing `Level(msg, key1, val1, ...)` pattern does not change.  However, if you find yourself logging certain context(s) frequently esp. in the same function, I'd suggest create a sub-logger by using the `New(key1, val1, ...)` method of any logger.  Example:

```go
logger := utils.GetLogInstance().New("blockHash", b.Header().Hash())
```

Then whenever you use `logger`, it will automatically add the block hash as an extra context.

This can also be factored out as a utility function:

```go
// BlockLogger returns a sub-logger of the given logger,
// with various block information added as extra contexts.
func BlockLogger(logger log.Logger, b *types.Block) log.Logger {
	h := b.Header()
	return logger.New(
		"blockHash", h.Hash(),
		"blockShardID", h.ShardID,
		"blockEpoch", h.Epoch,
		"blockNumber", h.Number)
}
```
```go
logger := BlockLogger(utils.GetLogInstance(), b)
// use logger and it will add block hash/shard/epoch/number
```

PR #777, opened 2019-04-26
Banjo hack

mikedoan:
## Issue

avoid account without enough balance participating the lottery

### DIFF
--- a/node/demo_contract.go
+++ b/node/demo_contract.go
@@ -73,7 +73,7 @@ func (node *Node) CreateTransactionForEnterMethod(amount int64, priKey string) e
 	balance, err := node.GetBalanceOfAddress(address)
 	if err != nil {
 		return err
-	} else if balance.Cmp(Amount) == -1 {
+	} else if balance.Cmp(Amount) == -1 && amount <= 0 {
 		return ErrLotteryAppFailed
 	}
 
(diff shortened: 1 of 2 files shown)


node/demo_contract.go
```diff
@@ -67,9 +67,17 @@ func (node *Node) CreateTransactionForEnterMethod(amount int64, priKey string) e
...
 
 	key, err := crypto.HexToECDSA(priKey)
-	nonce := node.GetNonceOfAddress(crypto.PubkeyToAddress(key.PublicKey))
+	address := crypto.PubkeyToAddress(key.PublicKey)
 	Amount := big.NewInt(amount)
 	Amount = Amount.Mul(Amount, big.NewInt(params.Ether))
+	balance, err := node.GetBalanceOfAddress(address)
+	if err != nil {
+		return err
+	} else if balance.Cmp(Amount) == -1 {
```
LeoHChen: what if Amount == 0

mikedoan: our web server hard-cord amount == 1

the logic of our end point is /enter?key=xvvU&amount=dddd
so if the amount > = key's balance - we should accept that

LeoHChen: what if people found the API and hack it to set amount = -1?  we are building defensive code here, and can't make assumption on the user input.

PR #781, opened 2019-04-26
initial pbft view change commit

chaosma:
* Add minimum view change structures and refactoring consensus codes
* Add timeout class for view change
* Not touch main code, large refactoring will come later
* Add README for PBFT including view change


consensus/view_change.go
```diff
@@ -0,0 +1,61 @@
...
+
+// Mode return the current node mode
+func (pm *PbftMode) Mode() Mode {
+	return pm.mode
+}
+
+// SetMode set the node mode as required
+func (pm *PbftMode) SetMode(m Mode) {
+	pm.mux.Lock()
+	pm.mux.Unlock()
```
LeoHChen: defer

internal/utils/timer.go
```diff
@@ -0,0 +1,63 @@
+package utils
```
LeoHChen: no test on this timer package.

consensus/view_change.go
chaosma: done

internal/utils/timer.go
chaosma: done

PR #784, opened 2019-04-30
Foundation node block reward for cello

harmony-ek:
## Issue

This is Banjo's foundational reward massaged to deal with only static set of nodes (foundational nodes) without new nodes joining.

## Test

Local test passed with a few thousand blocks.  Travis-CI succeeded.  Jenkins testing passed (see checks below).

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -34,6 +34,12 @@ import (
 	"github.com/harmony-one/harmony/p2p/host"
 )
 
+// Block reward per block signature.
+// TODO ek – per sig per stake
+var (
+	BlockReward = big.NewInt(params.Ether / 1000)
+)
+
 // Consensus is the main struct with all states and data related to consensus process.
 type Consensus struct {
 	// The current state of the consensus
@@ -598,7 +604,10 @@ func (consensus *Consensus) VerifySeal(chain consensus_engine.ChainReader, heade
 func (consensus *Consensus) Finalize(chain consensus_engine.ChainReader, header *types.Header, state *state.DB, txs []*types.Transaction, receipts []*types.Receipt) (*types.Block, error) {
 	// Accumulate any block and uncle rewards and commit the final state root
 	// Header seems complete, assemble into a block and return
-	accumulateRewards(chain.Config(), state, header)
+	err := consensus.accumulateRewards(chain.Config(), state, header)
+	if err != nil {
+		ctxerror.Log15(utils.GetLogInstance().Error, err)
+	}
 	header.Root = state.IntermediateRoot(false)
 	return types.NewBlock(header, txs, receipts), nil
 }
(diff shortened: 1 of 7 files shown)


consensus/consensus.go
```diff
@@ -7,25 +7,37 @@ import (
...
 	"github.com/harmony-one/harmony/p2p"
 	"github.com/harmony-one/harmony/p2p/host"
-	libp2p_peer "github.com/libp2p/go-libp2p-peer"
-	"golang.org/x/crypto/sha3"
+)
+
+// Block reward per block signature.
+// TODO ek – per sig per stake
+var (
+	BlockReward = big.NewInt(params.Ether / 1000)
```
chaosma: what does 1000 mean here, just some constant?

consensus/consensus.go
```diff
@@ -674,3 +770,81 @@ func (consensus *Consensus) GetNodeIDs() []libp2p_peer.ID {
 func (consensus *Consensus) GetConsensusID() uint32 {
 	return consensus.consensusID
 }
+
+// GenesisStakeInfoFinder is a stake info finder implementation using only
+// genesis accounts.
+// When used for block reward, it rewards only foundational nodes.
+type GenesisStakeInfoFinder struct {
+	byNodeKey map[types.BlsPublicKey][]*structs.StakeInfo
+	byAccount map[common.Address][]*structs.StakeInfo
```
chaosma: why one address has a list of stakeinfo?

rlan35: Good question.

consensus/consensus.go
harmony-ek: Yep.  0.001 HRX per signature.  This is potentially subject to change after reward calibration later this week.

consensus/consensus.go
harmony-ek: This is for the case where one node represents multiple ECDSA accounts, e.g. proxy staking.  The design came earlier from the staking discussion with you and Nicolas.  Right now we don’t have proxy staking so one BLS key always represents one account and vice versa, but it could change in the future.

harmony-ek: This is for the case where one node represents multiple ECDSA accounts, e.g. proxy staking.  The design came earlier from the staking discussion with you and Nicolas.  Right now we don’t have proxy staking so one BLS key always represents one account and vice versa, but it could change in the future.

harmony-ek:
@chaosma, no, this version is a specific optimization for the current testnet where no resharding occurs.  The full resharding-ready version is much bigger in scope with on-chain ShardInfo.

harmony-ek:
@LeoHChen, I believe the version is good to go.  The only thing remaining is your test, after which you could merge this into cello.  I’m giving you a (co-)assignee bit.

harmony-ek:
Tested on all 4 shards on @LeoHChen  devnet.  Changing block reward to 0.1 HMY/sig per his suggestion.

consensus/consensus.go
harmony-ek:
```suggestion
	BlockReward = big.NewInt(params.Ether / 1000)
```

PR #794, opened 2019-04-30
Fix 0 balance issue on explorer; increase consensus delay

rlan35:


api/service/explorer/service.go
```diff
@@ -255,6 +260,16 @@ func (s *Service) GetExplorerAddress(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	data.Address = address
+
+	// Check the balance from blockchain rather than local DB dump
+	if s.GetAccountBalance != nil {
+		address := common.HexToAddress(id)
+		balance, err := s.GetAccountBalance(address)
```
mikedoan: this is dangeous| i don't consider it as fix because the balance will be changing by timing of explorer querying.

rlan35: Why? Blockchain is the canonical data. We should return blockchain's data no matter what.

mikedoan: richard frontend will call it if requested. the balance of an address after a certain tx should be certain at that time (let's say, yesterday). 

now today i'm looking at the same tx, it could be different

PR #795, opened 2019-04-30
Display pingpong message's effect

rlan35:


LeoHChen:
only disable the handling logic of ping/pong, didn't disable the ping/pong messages.

PR #827, opened 2019-05-09
basic framework for view change

chaosma:
add version to distinguish existing (v1) and working on (v2) consensus


consensus/pbft_log.go
```diff
@@ -0,0 +1,76 @@
+package consensus
+
+import (
+	mapset "github.com/deckarep/golang-set"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/harmony-one/bls/ffi/go/bls"
+	"github.com/harmony-one/harmony/core/types"
+)
+
+// PbftLog represents the log stored by a node during PBFT process
```
LeoHChen: why use a new log package?

LeoHChen: shouldn't it be in utils/ module?

core/resharding.go
```diff
@@ -24,7 +24,7 @@ const (
 	// GenesisShardNum is the number of shard at genesis
 	GenesisShardNum = 4
 	// GenesisShardSize is the size of each shard at genesis
-	GenesisShardSize = 50
+	GenesisShardSize = 5
```
LeoHChen: don't check in debug configuration, please.

hehe.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
LeoHChen: hehe? :-D

consensus/consensus.go
```diff
@@ -445,3 +401,118 @@ func NewGenesisStakeInfoFinder() (*GenesisStakeInfoFinder, error) {
...
+			}
+			if consensus.viewChangeTimeout.CheckExpire() {
+				if consensus.mode.Mode() == Normal {
+					continue
+				}
+				consensusID := consensus.mode.ConsensusID()
+				consensus.startViewChange(consensusID + 1)
+			}
+		case <-tick.C:
+			consensus.tryPublishBlock()
```
LeoHChen: shouldn't you start the idleTimeout after the block published? otherwise, you may lose some time for consensus.

consensus/consensus.go
```diff
@@ -31,8 +38,29 @@ var (
 
 // Consensus is the main struct with all states and data related to consensus process.
 type Consensus struct {
-	phase PbftPhase
-	mode  PbftMode
+	ConsensusVersion string
+	phase            PbftPhase
```
rlan35: Please add comment for phase, mode, and seqNum

consensus/pbft_log.go
```diff
@@ -0,0 +1,76 @@
...
+			it.Stop()
+		}
+	}
+	return found
+}
+
+// GetBlockByNumber returns teh block matches the given block number
+func (log *PbftLog) GetBlockByNumber(id uint64) *types.Block {
+	var found *types.Block
+	it := log.Blocks().Iterator()
```
rlan35: Can we use directly lookup instead of iterator?

consensus/pbft_log.go
```diff
@@ -0,0 +1,76 @@
...
+
+// AddBlock add a new block into the log
+func (log *PbftLog) AddBlock(block *types.Block) {
+	log.blocks.Add(block)
+}
+
+// GetBlockByHash returns the block matches the given block hash
+func (log *PbftLog) GetBlockByHash(hash common.Hash) *types.Block {
+	var found *types.Block
+	it := log.Blocks().Iterator()
```
rlan35: similar comment as below.

core/resharding.go
```diff
@@ -24,7 +24,7 @@ const (
 	// GenesisShardNum is the number of shard at genesis
 	GenesisShardNum = 4
 	// GenesisShardSize is the size of each shard at genesis
-	GenesisShardSize = 50
```
rlan35: Mistakenly committed, I do it all the time :). Please remove.

hehe.sh
rlan35: hehe, please remove :)

consensus/consensus.go
```diff
@@ -445,3 +401,118 @@ func NewGenesisStakeInfoFinder() (*GenesisStakeInfoFinder, error) {
 	}
 	return f, nil
 }
+
+// Start is the entry point and main loop for consensus
+func (consensus *Consensus) Start(stopChan chan struct{}, stoppedChan chan struct{}) {
```
rlan35: Better put all the v2 code into a separate file like consensus_v2.go

consensus/config.go
```diff
@@ -6,4 +6,5 @@ const (
 	// blockDuration is the period a node try to publish a new block if it's leader
 	blockDuration  time.Duration = 5 * time.Second
 	receiveTimeout time.Duration = 5 * time.Second
+	maxLogSize     uint32        = 1000
```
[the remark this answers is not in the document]
chaosma: this is the pbft message log, which will be used to determine the logic in the consensus protocol. e.g. if we receive a new message with same view number but different block has comparing to some previous message, then we just ignore the new message.

api/service/consensus/service.go
```diff
@@ -27,8 +27,11 @@ func (s *Service) StartService() {
 	utils.GetLogInstance().Info("Starting consensus service.")
 	s.stopChan = make(chan struct{})
 	s.stoppedChan = make(chan struct{})
-	// after implementation finished, replace WaitForNewBlock by: s.consensus.Start(s.stopChan, s.stoppedChan)
-	s.consensus.WaitForNewBlock(s.blockChannel, s.stopChan, s.stoppedChan, s.startChan)
+	if s.consensus.ConsensusVersion == "v1" {
```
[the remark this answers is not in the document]
chaosma: the ConsensusVersion is initialized with const value. The reason to do this is to avoid cyclic import when I tried to directly refer to consensus.ConsensusVersion (where consensus is the package).

consensus/consensus.go
```diff
@@ -31,8 +38,29 @@ var (
...
-	mode  PbftMode
+	ConsensusVersion string
+	phase            PbftPhase
+	mode             PbftMode
+	seqNum           uint32
+	// channel to receive consensus message
+	MsgChan chan []byte
+	// timer to make sure leader publishes block in a timely manner; if not
+	// then this node will propose view change
+	idleTimeout utils.Timeout
```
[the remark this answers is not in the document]
chaosma: yes, I think it's a good idea

consensus/pbft_log.go
chaosma: sorry for naming confusing. it's not the log for debugging, it's the storage/log for consensus messages.

consensus/pbft_log.go
chaosma: The issue here is that for the same block number, we might receive multiple messages but with different consensusID. Another reason is sometime we might want to look up by block number sometimes might want to look up by block hash. So just put them into a list

consensus/consensus.go
chaosma: added

Issue #612, opened 2019-03-21
keep track of all open ports


LeoHChen:
Checkout details about our Technical bounty program here:
https://github.com/harmony-one/harmony-open

## Summary

Checkout details about our Technical bounty program here:
https://github.com/harmony-one/harmony-open
We need to have a central place to keep track of all open ports.

## Current Design

currently, we have many open ports for different services, and we are losing control.  Please add a central configuration for all open ports.  This will help beef up the security and audit.  We shall also consider synergizing all the ports into a few open ports.  That will be much easier for our node operator as well.

## Problems

Too many open ports, and configuration scattered everywhere.

## Proposal

use the internal/configs package to keep track of all open ports.

we can create a package like just configuration, just to record all the open ports and their usages.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 15000.0 HARBTY   attached to it.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/harmony-one/harmony/612/2920).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/harmony-one/harmony/612/2920).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $97,941.27 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [lzxm160](https://gitcoin.co/profile/lzxm160) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/harmony-one/harmony/612/2920?mutate_worker_action=approve&worker=lzxm160) | [reject worker](https://gitcoin.co/issue/harmony-one/harmony/612/2920?mutate_worker_action=reject&worker=lzxm160))_.**

fix many blockchain project's issues,hope i can help with this

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/harmony-one/harmony/612/2920).

gitcoinbot:
[...]

PR #828, opened 2019-05-10
Update RPC framework

lvyelin:
## Issue

Update RPC framework

## Changes:

1. Remove eth rpc code and unnecessary changes from
7a0f18f. Import the package instead.
2. Move HmyAPIBackend into core package so it has more access to core.
3. Add API interface for services.
4. Start RPC for all roles, for both HTTP (baseport+10) and WS (baseport+20).
3. Keep implemented APIs but move to internal/hmyapi package.

## Test
$ curl -H "Content-Type: application/json" -d '{"method":"hmy_getBlockByNumber","params":["latest",true],"id":1}' http://127.0.0.1:9010/
{"jsonrpc":"2.0","id":1,"result":{"difficulty":"0x0","extraData":"0x","gasLimit":"0x35b7ffdfe","gasUsed":"0x17f00","hash":"0x4ac90cba6c09496855a46139dfe36df1c9e54e90e420eb5944d69953a4bf3aa7","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","miner":"0x0000000000000000000000000000000000000000","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","number":"0x177","parentHash":"0x18a8a5e6f10d3e368e16c0de6ad287247c45eb3e0cb405a78f3352c66000b3d2","receiptsRoot":"0x9a4f1f141492f2870a8bf8fa6e8646fcb6007bf66a71c8e7687d4dda7ac5c057","size":"0x3ab","stateRoot":"0x86bf5ac1f6d8385640901968ebc3893f2500cbb9edf454b4fbb9537c96772d95","timestamp":1557723465,"transactionsRoot":"0x9a24d322060f456d1f146dcb064dd3c2b0715305b30c9560e838911023c704b7","uncles":[]}}


LeoHChen:
The PR failed travis-ci, you need to run golint to fix the format issue.

lvyelin:
@LeoHChen @lzl124631x 
Update the patch and keep APIs previously implemented.
Please review.

internal/hmyapi/backend.go
```diff
@@ -1,22 +1,21 @@
...
-const namespace = "hmy"
-
 // GetAPIs returns all the APIs.
-func GetAPIs(b *rpc.HmyAPIBackend) []rpc.API {
+func GetAPIs(b *core.HmyAPIBackend) []rpc.API {
 	nonceLock := new(AddrLocker)
 	return []rpc.API{
 		{
-			Namespace: namespace,
+			Namespace: "hmy",
```
LeoHChen: why not use const?

node/rpc.go
```diff
@@ -0,0 +1,133 @@
...
+	httpListener net.Listener
+	httpHandler  *rpc.Server
+
+	wsListener net.Listener
+	wsHandler  *rpc.Server
+
+	httpEndpoint = ""
+	wsEndpoint   = ""
+
+	httpModules      = []string{"hmy"}
```
LeoHChen: const?

node/rpc.go
```diff
@@ -0,0 +1,133 @@
...
+	apiBackend = core.NewBackend(node.blockchain, node.TxPool)
+
+	apis := hmyapi.GetAPIs(apiBackend)
+	for _, service := range node.serviceManager.GetServices() {
+		apis = append(apis, service.APIs()...)
+	}
+
+	port, _ := strconv.Atoi(nodePort)
+
+	httpEndpoint = fmt.Sprintf("localhost:%v", port+rpcHTTPPortOffset)
```
LeoHChen: use net.JoinHostPort, and why localhost?

node/rpc.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := node.startHTTP(httpEndpoint, apis, httpModules, nil, httpVirtualHosts, httpTimeouts); err != nil {
+		return err
+	}
+
+	wsEndpoint = fmt.Sprintf("localhost:%v", port+rpcWSPortOffset)
```
LeoHChen: same as above, why localhost?

node/rpc.go
```diff
@@ -0,0 +1,133 @@
+package node
+
+import (
+	"fmt"
+	"net"
+	"strconv"
+	"strings"
+
+	"github.com/ethereum/go-ethereum/log"
```
LeoHChen: use our internal log?

api/service/restclientsupport/service.go
```diff
@@ -259,3 +260,8 @@ func (s *Service) NotifyService(params map[string]interface{}) {
 func (s *Service) SetMessageChan(messageChan chan *msg_pb.Message) {
 	s.messageChan = messageChan
 }
+
+// APIs for the services.
+func (s *Service) APIs() []rpc.API {
```
lzl124631x: All the APIs are nil. I don't think it's necessary to add this interface.

core/api_backend.go
```diff
@@ -1,30 +1,30 @@
-package rpc
+package core
```
lzl124631x: I feel `api_backend` is only serving rpc, so putting it into rpc folder makes the code more cohesive.

lzl124631x: I feel `api_backend` is only serving/relevant to rpc (not shared/used by anything else), so putting it into rpc folder makes the code more cohesive.

api/service/restclientsupport/service.go
LeoHChen: my understanding is all services can provide RPC service, via this interface.  sounds like a good framework.

internal/hmyapi/backend.go
lvyelin: may add some other API with different namespace. so didn't use const.

node/rpc.go
lvyelin: we may able to modify enabled modules at run time and restart RPC to provide more modules. so not using const.

api/service/restclientsupport/service.go
lvyelin: Yes, it is just a framework, so the services can provide API.

core/api_backend.go
lvyelin: Moving to core package can access more in the package.
It is the functions that core provided to the RPC, so it should be put into the provider (core package).

lzl124631x:
Thanks for the change. My bad that I didn't realize that we don't at all need to copy & paste eth-rpc code. Good catch.

It looks good overall.

lvyelin:
@LeoHChen @lzl124631x Do you agree to use +10 and +20 ports?

node/rpc.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/harmony-one/harmony/core"
+	"github.com/harmony-one/harmony/internal/hmyapi"
+)
+
+const (
+	rpcHTTPPortOffset = 10
+	rpcWSPortOffset   = 20
```
lzl124631x: Move all port offsets to a single file so that we won't step on each other.

lzl124631x:
> @LeoHChen @lzl124631x Do you agree to use +10 and +20 ports?

Sure. Nit: Move all port offsets to a single file so that we won't step on each other and we can know the ports at a glance without needing to jump between files.

node/rpc.go
lvyelin: I see our code also use eth log. which log do you prefer?

api/service/restclientsupport/service.go
lzl124631x: I can see your point. But I don't think it's needed in this PR -- kinda premature to add it.
* If we found we never add APIs in this way these interfaces are dead code. 
* If we need to add APIs to some service, why not add it by then.

internal/hmyapi/backend.go
lzl124631x: But for this hmy namespace I still think we should make it a const.

core/api_backend.go
lzl124631x: Ah. You're right I confused this with the internal public.go. This one matches better with `eth/api_backend.go`

node/rpc.go
LeoHChen: there is a ticket for this task.  https://github.com/harmony-one/harmony/issues/612, no need to block this PR.

node/rpc.go
LeoHChen: try https://github.com/harmony-one/harmony/commit/83bd41fb930dd3822f2ceb9d3b22bbde7dbc375f

PR #835, opened 2019-05-13
Simple Fix to reduce invalid txns

alajko:
Fixing "nonce is low" Issue.

TODO
- [ ] Need to fix for shard 1.


LeoHChen:
will it break shard0 or not?

alajko:
Doesn't break shard 0.  @LeoHChen

PR #849, opened 2019-05-16
Fixing TXGen

alajko:
## Issue

The TXGen sends stop messages and kills all nodes in the network.


alajko:
Tests checked

- [ ] Local Test: Does TXGen stop after duration specified.
- [ ] Local Test: TXGen doesn't kill the nodes after it is done sending txns.
- [ ] Remote Test: Does TXGen stop in syncLoop if duration is less for it to completely sync. (t=30s)

alajko:
TODO
- [ ] Remote Test: TXGen stop after right duration
- [  ] Remote Test: TXGen doesn't kill the nodes after it is done sending txns.

alajko:
shard0txns

PR #850, opened 2019-05-16
sendTx rpc

lzl124631x:
* Set `GenesisShardSize = 5`
* Run `./test/deploy.sh ./test/configs/ten-oneshard.txt`
* `http://localhost:9015` is the endpoint.

Search for `TODO(ricl): placeholder`. I left two stubs waiting for you to add.


core/api_backend.go
```diff
@@ -82,3 +85,23 @@ func (b *HmyAPIBackend) HeaderByNumber(ctx context.Context, blockNr rpc.BlockNum
 func (b *HmyAPIBackend) GetPoolNonce(ctx context.Context, addr common.Address) (uint64, error) {
 	return b.txPool.State().GetNonce(addr), nil
 }
+
+// SendTx ...
+func (b *HmyAPIBackend) SendTx(ctx context.Context, signedTx *types.Transaction) error {
+	return b.txPool.Add(ctx, signedTx)
```
mikedoan: txPool is not used in our blockchain - by not blocking  it you an add TODO to fix this logic later.

internal/hmyapi/util.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/harmony-one/harmony/core"
+	"github.com/harmony-one/harmony/core/types"
+)
+
+// SubmitTransaction is a helper function that submits tx to txPool and logs a message.
+func SubmitTransaction(ctx context.Context, b *core.HmyAPIBackend, tx *types.Transaction) (common.Hash, error) {
+	if err := b.SendTx(ctx, tx); err != nil {
+		return common.Hash{}, err
+	}
+	if tx.To() == nil {
```
mikedoan: i think this logic should be handled by blockchain transaction processing - our blockchain will handle this case. we don't want to have the same logic and different places handle different logic.

lzl124631x:
Abandon in favor of https://github.com/harmony-one/harmony/pull/852

PR #854, opened 2019-05-16
Remove STOP message

harmony-ek:
Currently, all nodes—*even ones in production*—honor the `STOP` message which causes them to drop dead.  This was meant only for testing, and now that its only user—txgen—does not send it anymore and we calculate TPS in another way without relying upon the blockchain report that the `STOP` handler furnishes, this is just a useless backdoor waiting to be exploited.

So… remove it.

### DIFF
--- a/api/proto/node/node.go
+++ b/api/proto/node/node.go
@@ -24,7 +24,7 @@ const (
 	Transaction MessageType = iota
 	Block
 	Client
-	Control
+	_    // used to be Control
 	PING // node send ip/pki to register with leader
 	PONG // node broadcast pubK
 	ShardState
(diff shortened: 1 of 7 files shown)


alajko:
Thanks for this crucial change @harmony-ek

alajko:
merged on green.

harmony-ek:
It seems that Jenkins failed with too low of TPS, so TPS gathering logic of Jenkins pipeline seems to be working.  Which means…

Jenkins, test this please.

harmony-ek:
Repurposed PR to just remove `STOP` handler altogether.  @alajko removed `STOP` from `txgen`, and we don't need the blockchain report (we gather TPS another way), so `STOP` is just a backdoor waiting to be exploited.  We don't need none of that.

PR #864, opened 2019-05-19
Missing err check after calling consensus.New() in setUpTxGen()

tsuwu:
There are the following code pieces in setUpTxGen():
	consensusObj, err := consensus.New(myhost, uint32(shardID), p2p.Peer{}, nil)
	txGen := node.New(myhost, consensusObj, nil, false) //Changed it : no longer archival node.

If err != nil and consensusObj is nil, then "node.Consensus = consensusObj" would end up nil after a new node has been created. 

This might be the root cause of #727

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -153,7 +153,7 @@ type Node struct {
 	ContractAddresses            []common.Address
 
 	// For puzzle contracts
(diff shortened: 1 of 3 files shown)


harmony-ek:

PR #868, opened 2019-05-22
p2p peerdiscovery fix

chaosma:
not fully work


LeoHChen:
fixed in https://github.com/harmony-one/harmony/pull/934

PR #869, opened 2019-05-22
use single source of truth instead having is_leader truth in 2 objects

mikedoan:
1/ use a single source of truth instead of having is_leader in 2 objects. that would cause inconsistency in case you missed to update it at both places.
2/ no more quickly adding code in main file.


rlan35:
It's a good change, but I am wondering is there any bug for it? As once we switch to rotated leader setup, there should be no fixed leader configuration.

PR #870, opened 2019-05-23
use single source of truth instead having is_leader truth in 2 objects

mikedoan:
1/ use a single source of truth instead of having is_leader in 2 objects. that would cause inconsistency in case you missed to update it at both places.
2/ Clean up code in main file


mikedoan:
No need to take a look at this one.

PR #876, opened 2019-05-24
PBFT view change 

chaosma:
Local test passed. HOORAY!


consensus/config.go
```diff
@@ -2,9 +2,20 @@ package consensus
...
-	// blockDuration is the period a node try to publish a new block if it's leader
-	blockDuration  time.Duration = 5 * time.Second
 	receiveTimeout time.Duration = 5 * time.Second
-	maxLogSize     uint32        = 1000
+	// The duration of viewChangeTimeout; when a view change is initialized with v+1
+	// timeout will be equal to viewChangeDuration; if view change failed and start v+2
+	// timeout will be 2*viewChangeDuration; timeout of view change v+n is n*viewChangeDuration
+	viewChangeDuration time.Duration = 30 * time.Second
+
+	bootstrapDuration time.Duration = 60 * time.Second
```
rlan35: what is bootstrap duration? Please add a comment.

api/proto/message/message.proto
```diff
@@ -102,3 +92,18 @@ message DrandRequest {
...
+message ViewChangeRequest {
+    uint32 consensus_id = 1;
+    uint64 seq_num = 2;
+    bytes sender_pubkey = 3;
+    bytes leader_pubkey = 4;
+    bytes payload = 5; // message payload: either m1 type or m2 type 
+    bytes viewchange_sig = 6; // signature on payload
+                        
+    // below is for newview message only
+    bytes m1_aggsigs = 7; // m1: |block_hash|bhp_sigs|bhp_bitmap|
```
rlan35: what is m1 and m2? The <view change> and <new view> message?

consensus/consensus.go
```diff
@@ -76,6 +61,16 @@ type Consensus struct {
 	prepareBitmap        *bls_cosi.Mask
 	commitBitmap         *bls_cosi.Mask
 
+	// Commits collected from view change
+	bhpSigs          map[common.Address]*bls.Sign // signature on m1 type message
```
rlan35: what are bhp and nil？

consensus/config.go
chaosma: When network start, it takes a longer time for all the nodes to be ready. So I make the bootstrap time longer, otherwise, the nodes might start view change too soon. The exact time period might need some adjustment.

api/proto/message/message.proto
chaosma: m1 type message payload records the prepared message a validator received and then start a view change because of timeout.  m2 type message is just the NIL message which means the validator not receive any prepared message yet and start a view change. These are the only two possible two situations.

consensus/consensus.go
chaosma: bhp: blockHashPrepared  (the signature for m1 type message)
nil: the signature for m2 type message.
I am not very good at naming :(

consensus/config.go
```diff
@@ -2,9 +2,20 @@ package consensus
...
+
+	// timeout duration for announce/prepare/commit
+	phaseDuration time.Duration = 20 * time.Second
+	maxLogSize    uint32        = 1000
 )
+
+// NIL is the m2 type message
+var NIL = []byte{0x01}
```
chaosma: NIL rename to emptyPrepare

consensus/consensus.go
chaosma: TODO: add description

consensus/consensus_v2.go
```diff
@@ -683,19 +709,27 @@ func (consensus *Consensus) Start(blockChannel chan *types.Block, stopChan chan
...
-					consensus.startViewChange(consensus.consensusID + 1)
-				}
-				if consensus.commitTimeout.CheckExpire() {
-					consensus.startViewChange(consensus.consensusID + 1)
-				}
-				if consensus.viewChangeTimeout.CheckExpire() {
-					if consensus.mode.Mode() == Normal {
-						continue
+
+			case <-ticker.C:
```
chaosma: move to first case

consensus/consensus_v2.go
```diff
@@ -683,19 +709,27 @@ func (consensus *Consensus) Start(blockChannel chan *types.Block, stopChan chan
...
+				if !consensus.PubKey.IsEqual(consensus.LeaderPubKey) {
+					for k, v := range consensus.consensusTimeout {
+						if !v.CheckExpire() {
+							continue
+						}
+						if k != "viewchange" {
```
chaosma: change to enum or const

consensus/consensus_viewchange_msg.go
```diff
@@ -0,0 +1,94 @@
...
+		ServiceType: msg_pb.ServiceType_CONSENSUS,
+		Type:        msg_pb.MessageType_VIEWCHANGE,
+		Request: &msg_pb.Message_Viewchange{
+			Viewchange: &msg_pb.ViewChangeRequest{},
+		},
+	}
+
+	vcMsg := message.GetViewchange()
+	vcMsg.ConsensusId = consensus.mode.GetConsensusID()
+	vcMsg.SeqNum = consensus.seqNum
```
chaosma: seqNum to blockNum

consensus/view_change.go
```diff
@@ -93,10 +98,342 @@ func (consensus *Consensus) GetNextLeaderKey() *bls.PublicKey {
...
+	}
+	consensus.mode.SetMode(ViewChanging)
+	consensus.mode.SetConsensusID(consensusID)
+	nextLeaderKey := consensus.GetNextLeaderKey()
+	consensus.LeaderPubKey = consensus.GetNextLeaderKey()
+	if nextLeaderKey.IsEqual(consensus.PubKey) {
+		return
+	}
+
+	diff := consensusID - consensus.consensusID
```
chaosma: consensusID -> viewID

api/proto/message/message.proto
```diff
@@ -45,25 +36,24 @@ enum MessageType {
 //
 // The request field will be either one of the structure corresponding to the MessageType type.
 message Message {
-  ReceiverType receiver_type = 1;
```
mikedoan: why removing ReceiverType? I think it's still beneficial

consensus/config.go
chaosma: I added comments to explain the naming

api/proto/message/message.proto
chaosma: it's not used anywhere since we can determine the receiver by their bls public key. we always need to check the key validity instead of the message type because the message type can be construct by the validator arbitrarily

consensus/consensus_v2.go
chaosma: I will reduce the number of timeout keys and test it in later pr

chaosma:
Jenkins, test this please.

PR #879, opened 2019-05-25
clean up and fix correct error messages. Now we should watch out of a…

mikedoan:
1/ Please watch out of all error messages. If there were error we should find out why
2/ This PR is to avoid confusing of having many error messages when debuging.


mikedoan:
Jenkins, test it please.

PR #888, opened 2019-05-28
Add foundational nodes

LeoHChen:


LeoHChen:
Jenkins, test this please.

LeoHChen:
> you need to add separated file foundation_node_constants.go for foundation node instead of mixing up with our code.
> 
> the addresses in constant files are just for testing and created by us.

I won't be able to change it right now.  The account index is also bounded to deployment.  I can make some change in next PR to fix it.

PR #889, opened 2019-05-28
Do not panic upon out-of-range account index

harmony-ek:
Out-of-range account index (currently >= 200) translates into a shard ID
(>= 4) not found in the genesis table, so its genesis cannot be
initialized.  Since this can be induced by misconfiguration, do not
panic but just crit out.

This error was seen during drum testnet launch.


harmony-ek:
Jenkins, test this please.

PR #892, opened 2019-05-28
Update Li's public key and add mine

harmony-ek:
His old keypair was generated using an old wallet (on MBP) and isn't
in the right Ethereum keystore format.

While I'm here add mine too at account index 199.


LeoHChen:
Jenkins, test this please.

PR #906, opened 2019-05-30
Add get_balance api for RPC

mikedoan:
Add get_balance api for RPC

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/bloombits"
 	"github.com/ethereum/go-ethereum/ethdb"
+	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/params"
 	"github.com/ethereum/go-ethereum/rpc"
@@ -184,4 +185,9 @@ func (b *APIBackend) GetPoolTransactions() (types.Transactions, error) {
 		txs = append(txs, batch...)
 	}
 	return txs, nil
+}
+
+// GetBalance ...
+func (b *APIBackend) GetBalance(address common.Address) (*hexutil.Big, error) {
+	return nil, nil
 }
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -112,9 +112,11 @@ func (b *APIBackend) GetReceipts(ctx context.Context, hash common.Hash) (types.R
 }
 
 // EventMux ...
+// TODO: this is not implemented or verified yet for harmony.
 func (b *APIBackend) EventMux() *event.TypeMux { return b.hmy.eventMux }
 
 // BloomStatus ...
+// TODO: this is not implemented or verified yet for harmony.
 func (b *APIBackend) BloomStatus() (uint64, uint64) {
 	sections, _, _ := b.hmy.bloomIndexer.Sections()
 	return params.BloomBitsBlocks, sections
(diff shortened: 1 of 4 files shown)


hmy/api_backend.go
```diff
@@ -185,3 +198,9 @@ func (b *APIBackend) GetPoolTransactions() (types.Transactions, error) {
 	}
 	return txs, nil
 }
+
+// GetBalance returns balance of an given address.
+func (b *APIBackend) GetBalance(address common.Address) (*hexutil.Big, error) {
+	balance, err := b.hmy.nodeAPI.GetBalanceOfAddress(address)
+	return (*hexutil.Big)(balance), err
```
LeoHChen: is there any risk (balance) be nil?

mikedoan: Get balance is handled by ethereum GetBalance. I look around ethereum code if nil it returns Big0 etc... For extremely conservative, we can have that check. But if we do that we would have to handle that many other places.
```
// GetBalance retrieves the balance from the given address or 0 if object not found
func (stateDB *DB) GetBalance(addr common.Address) *big.Int {
	stateObject := stateDB.getStateObject(addr)
	if stateObject != nil {
		return stateObject.Balance()
	}
	return common.Big0
}
```

LeoHChen: fine

hmy/api_backend.go
```diff
@@ -127,53 +130,63 @@ func (b *APIBackend) ProtocolVersion() int {
...
-// SubscribeChainEvent ...
+// SubscribeChainEvent subcribes chain event.
+// TODO: this is not implemented or verified yet for harmony.
 func (b *APIBackend) SubscribeChainEvent(ch chan<- core.ChainEvent) event.Subscription {
 	return b.hmy.BlockChain().SubscribeChainEvent(ch)
 }
 
-// SubscribeChainHeadEvent ...
+// SubscribeChainHeadEvent subcribes chain head event.
+// TODO: this is not implemented or verified yet for harmony.
```
lzl124631x: I think some of these are already working now, as I can listen to filterChange of newBlockFilter

hmy/api_backend.go
```diff
@@ -127,53 +130,63 @@ func (b *APIBackend) ProtocolVersion() int {
...
 
 // HeaderByHash ...
+// TODO: this is not implemented or verified yet for harmony.
 func (b *APIBackend) HeaderByHash(ctx context.Context, blockHash common.Hash) (*types.Header, error) {
 	// TODO(ricl): implement
 	return nil, nil
 }
 
 // ServiceFilter ...
+// TODO: this is not implemented or verified yet for harmony.
```
lzl124631x: For the APIs that I've already added TODO, no need to add the TODO twice.

internal/hmyapi/blockchain.go
```diff
@@ -72,9 +72,5 @@ func (s *PublicBlockChainAPI) GetStorageAt(ctx context.Context, address common.A
 // given block number. The rpc.LatestBlockNumber and rpc.PendingBlockNumber meta
 // block numbers are also allowed.
 func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (*hexutil.Big, error) {
-	state, _, err := s.b.StateAndHeaderByNumber(ctx, blockNr)
-	if state == nil || err != nil {
-		return nil, err
-	}
-	return (*hexutil.Big)(state.GetBalance(address)), state.Error()
+	return s.b.GetBalance(address)
```
lzl124631x: Why you just ignored the blockNr? That's an important input parameter.

lzl124631x: Have you tested the API of the old code and new code? What are the behaviors?

mikedoan: agree. right now for testing we will try to get the latest balance first. We will add complete behaviour later. Let me add todo

hmy/api_backend.go
mikedoan: done

lzl124631x:
I tested locally that the getBalance is working but this error I guess is because we don't have that account in blockchain.

image

internal/hmyapi/blockchain.go
lzl124631x: I suspect the old code works.

mikedoan:
Richard please approve.

mikedoan:
Screen Shot 2019-05-30 at 11 49 26 AM

PR #911, opened 2019-05-30
tx args

lzl124631x:
fix tx args issue.

image


mikedoan:
i don't understand the issue?

lzl124631x:
@mikedoan this was for partial support of `sendTransaction` API. Now it can recognize the shardID.

PR #918, opened 2019-05-30
add consensus testplan

chaosma:


test/consensus_testplan.md
```diff
@@ -0,0 +1,21 @@
+## testplan for consensus
+
+### incorrect data
+- wrong signature 
+- construct incorrect consensus messages
+
+
+### timeout
+- leader timeout in announce phase
```
rlan35: We are having a single timeout for the consensus, right? timeout on each stage is not robust.

chaosma: right, we will have a single timeout, code not changed yet. let me remove duplicates

PR #919, opened 2019-05-30
reduce number of timeouts; make timeout more stable

chaosma:


consensus/config.go
```diff
@@ -11,13 +11,15 @@ const (
...
 	maxLogSize    uint32        = 1000
 )
 
+type timeoutType int
+
 const (
-	timeoutNormal     = "normal"
-	timeoutViewChange = "viewchange"
+	timeoutNormal timeoutType = iota
+	timeoutViewChange
```
rlan35: const should be have first letter upper case.

consensus/config.go
```diff
@@ -11,13 +11,15 @@ const (
...
 	// timeout duration for announce/prepare/commit
-	phaseDuration time.Duration = 20 * time.Second
+	phaseDuration time.Duration = 60 * time.Second
 	maxLogSize    uint32        = 1000
 )
 
+type timeoutType int
+
 const (
-	timeoutNormal     = "normal"
```
rlan35: I think change "Normal" to "Consensus" will be more informative.

consensus/config.go
```diff
@@ -11,13 +11,15 @@ const (
 	viewChangeDuration time.Duration = 30 * time.Second
 
 	// timeout duration for announce/prepare/commit
-	phaseDuration time.Duration = 20 * time.Second
+	phaseDuration time.Duration = 60 * time.Second
 	maxLogSize    uint32        = 1000
 )
 
+type timeoutType int
```
rlan35: TimeoutType

chaosma: why upper case? we only use it internally

consensus/config.go
chaosma: Okay

consensus/config.go
chaosma: OK

PR #922, opened 2019-05-31
Add signature grace period

harmony-ek:
The current scheme accepts and rewards only the first `2f+1` signers.  Topologically disadvantaged nodes, i.e. far away from the leader, typically loses out the commit race and earn no block rewards.  This commit allows a 1-second grace period after reaching the `2f+1` quorum.


harmony-ek:
The TPS reduction (52 < 80) is expected due to the 1-second delay.  Awaiting approval for this exception.

PR #924, opened 2019-06-01
add raw tx gen & fix signing for sdk

mikedoan:
1/ add raw tx gen
2/ fix signing for sdk: add shardID field for signer hashing.


mikedoan:
Jenkins, test this please.

PR #933, opened 2019-06-02
consensus logic fixes

chaosma:
Added the following:

- in view change mode, node accepts normal message and change state at committed message; 
- accept newview message regardless of leaderkey matching;
- In consensus SYNCING mode add normal consensus message in the log;
- leader also has timeout mechanism in consensus process

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -33,8 +33,9 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
 		return
 	}
 
-	// stop receiving normal message when is in viewchanging
-	if consensus.mode.Mode() == ViewChanging && msg.Type != msg_pb.MessageType_VIEWCHANGE && msg.Type != msg_pb.MessageType_NEWVIEW {
+	// when node is in ViewChanging mode, it still accepts normal message into PbftLog to avoid possible trap forever
+	// but drop PREPARE and COMMIT which are message types for leader
+	if consensus.mode.Mode() == ViewChanging && (msg.Type == msg_pb.MessageType_PREPARE || msg.Type == msg_pb.MessageType_COMMIT) {
 		return
 	}
 
@@ -112,7 +113,7 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
 		utils.GetLogInstance().Debug("onAnnounce verifySenderKey failed", "error", err)
 		return
 	}
-	if !senderKey.IsEqual(consensus.LeaderPubKey) {
+	if !senderKey.IsEqual(consensus.LeaderPubKey) && consensus.mode.Mode() != Syncing {
 		utils.GetLogInstance().Warn("onAnnounce senderKey not match leader PubKey", "senderKey", senderKey.GetHexString(), "leaderKey", consensus.LeaderPubKey.GetHexString())
 		return
 	}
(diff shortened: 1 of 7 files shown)


consensus/consensus_v2.go
```diff
@@ -33,8 +33,8 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
 		return
 	}
 
-	// stop receiving normal message when is in viewchanging
-	if consensus.mode.Mode() == ViewChanging && msg.Type != msg_pb.MessageType_VIEWCHANGE && msg.Type != msg_pb.MessageType_NEWVIEW {
+	// stop receiving message send to leader when is in viewchanging but accept other normal messages to avoid view change trap
```
rlan35: “stop receiving message send to leader”？ need to reword so it's more understandable.

rlan35: why only ignore PREPARE and COMMIT message in viewchange mode now?

consensus/consensus_v2.go
```diff
@@ -136,14 +136,19 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
 
+	if blockObj.NumberU64() != recvMsg.BlockNum || recvMsg.BlockNum < consensus.blockNum {
+		utils.GetLogger().Warn("blockNum not match", "recvBlockNum", recvMsg.BlockNum, "blockObjNum", blockObj.NumberU64(), "myBlockNum", consensus.blockNum)
+		return
+	}
+
 	// check block data transactions
-	if err := consensus.VerifyHeader(consensus.ChainReader, blockObj.Header(), false); err != nil {
-		utils.GetLogInstance().Warn("onAnnounce block content is not verified successfully", "error", err)
+	if err := consensus.VerifyHeader(consensus.ChainReader, blockObj.Header(), false); err != nil && consensus.mode.Mode() != Syncing {
```
rlan35: better just first check if it's in sync mode and then ignore the VerifyHeader check.

consensus/consensus_v2.go
```diff
@@ -136,14 +136,19 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
+		utils.GetLogInstance().Warn("onAnnounce block content is not verified successfully", "error", err, "inChain", consensus.ChainReader.CurrentHeader().Number, "have", blockObj.Header().ParentHash)
 		return
 	}
 	if consensus.BlockVerifier == nil {
 		// do nothing
-	} else if err := consensus.BlockVerifier(&blockObj); err != nil {
+	} else if err := consensus.BlockVerifier(&blockObj); err != nil && consensus.mode.Mode() != Syncing {
```
rlan35: same here. Logic will be cleaner that way.

consensus/consensus_v2.go
```diff
@@ -167,6 +172,10 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
 	consensus.pbftLog.AddMessage(recvMsg)
 	consensus.pbftLog.AddBlock(&blockObj)
 
+	if consensus.mode.Mode() == ViewChanging {
```
rlan35: add some log info and explain why returning.

consensus/consensus_v2.go
```diff
@@ -321,12 +330,6 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
 	}
 
 	blockHash := recvMsg.BlockHash
-	pubKey := recvMsg.SenderPubkey
```
rlan35: These was redundant code before?

consensus/consensus_v2.go
```diff
@@ -607,60 +628,64 @@ func (consensus *Consensus) tryCatchup() {
...
-
-				break
-			}
-		}
-		if cnt == 0 { // didn't find match block for current blockNum, return
+		msg := consensus.pbftLog.FindMessageByMaxViewID(preparedMsgs)
+		if msg == nil {
 			break
 		}
+		consensus.blockHash = [32]byte{}
```
rlan35: Why these are moved out of loop?

consensus/consensus_v2.go
chaosma: Since PREPARE and COMMIT are only the need receive, the validators will save the prepared and committed message in their log

consensus/consensus_v2.go
chaosma: right, let me change the order because verify header is heavy

consensus/consensus_v2.go
chaosma: sure

consensus/consensus_v2.go
chaosma: yes, it's redundant

consensus/consensus_v2.go
chaosma: In previous code, the loop is unnecessary, especially I add FindMessageByMaxViewID. i.e. we only choose the latest viewID for prepared message.

consensus/consensus_v2.go
chaosma: because we still want to add the message and block to catch up later if viewchanging stucked. I will add some comments

consensus/consensus_v2.go
rlan35: so should be “stop receiving message sent to leader”

chaosma:
Jenkins, test this please.

chaosma:
Jenkins, test this please.

PR #937, opened 2019-06-03
Increase shard size to 100

LeoHChen:


LeoHChen:
4 shards, 133 consensus, 0 total TPS, 353 nodes
54.149.153.40, 33, 0.0234906
54.91.99.134, 34, 0.0112401
54.191.95.74, 33, 0.0119632
34.234.93.150, 33, 0.0110841
Tue Jun 4 09:18:03 UTC 2019 : analysis took 0 minutes and 0 seconds
Tue Jun 4 09:18:03 UTC 2019 : sync log to s3
s3://harmony-benchmark/logs/2019/06/04/090723

consensus/config.go
```diff
@@ -11,8 +11,8 @@ const (
 	viewChangeDuration time.Duration = 30 * time.Second
 
 	// timeout duration for announce/prepare/commit
-	phaseDuration     time.Duration = 60 * time.Second
-	bootstrapDuration time.Duration = 60 * time.Second
+	phaseDuration     time.Duration = 90 * time.Second
```
chaosma: bootstrapDuration looks good. But phaseDuration is too long which means every time the view change need wait for 90 seconds. It can just be temp work around

PR #953, opened 2019-06-05
Remove ether

alajko:
## Issue

Replace Ether Denomination by ONE


core/denominations/denominations.go
```diff
@@ -0,0 +1,32 @@
...
+// Example: To get the wei value of an amount in 'gwei', use
+//
+//    new(big.Int).Mul(value, big.NewInt(params.GWei))
+// These are the multipliers for ether denominations.
+// Example: To get the wei value of an amount in 'gwei', use
+//
+//    new(big.Int).Mul(value, big.NewInt(params.GWei))
+//
+const (
+	Wei  = 1
```
harmony-ek: Can we go with `Atto` and `Nano` instead?

chaosma: I vote for quark = 1, atom = 1e9

PR #955, opened 2019-06-05
moving denominations under common

alajko:
## Issue

removing ether and moving denominations under common.


common/denominations/denominations.go
```diff
@@ -0,0 +1,8 @@
+package denominations
+
+//ONE Units
+const (
+	Wei  = 1
```
harmony-ek: `Atto`?

common/denominations/denominations.go
```diff
@@ -0,0 +1,8 @@
...
+	GWei = 1e9
```
harmony-ek: `Nano`?

alajko: done

common/denominations/denominations.go
alajko: done

PR #957, opened 2019-06-05
add limit for transaction pool. print out error for debugging

mikedoan:
add limit for transaction pool. print out error for debugging

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -48,6 +48,10 @@ const (
 	NodeLeader                         // Node is the leader of some shard.
 )
 
+const (
+	TxPoolLimit = 20000
+)
+
 func (state State) String() string {
 	switch state {
 	case NodeInit:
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -48,6 +48,11 @@ const (
 	NodeLeader                         // Node is the leader of some shard.
 )
 
+const (
+	// TxPoolLimit is the limit of transaction pool.
+	TxPoolLimit = 20000
```
LeoHChen: Can we make it bigger?  what's the proper number?

node/node.go
```diff
@@ -223,8 +228,14 @@ func (node *Node) Beaconchain() *core.BlockChain {
 func (node *Node) addPendingTransactions(newTxs types.Transactions) {
 	node.pendingTxMutex.Lock()
 	node.pendingTransactions = append(node.pendingTransactions, newTxs...)
+	// If length of pendingTransactions is greater than TxPoolLimit then by greedy take the TxPoolLimit recent transactions.
+	if len(node.pendingTransactions) > TxPoolLimit {
+		utils.GetLogInstance().Error("Got more transactions than expected and this could caused OOM", "num", len(newTxs), "totalPending", len(node.pendingTransactions))
```
LeoHChen: Is this an error or warning?

rlan35: should be warning.

node/node.go
rlan35: one tx is around 250-500 bytes. we should be able to do 200000

mikedoan: 1/ 20k TPS is a record for one shard. i don't think we would expect that.
2/ the goal of this is to find the culprit of OOM - If this really reaches 20k there is something wrong for example the finality is too long which, going forward, we need to look into.

mikedoan: discord discussed

node/node.go
mikedoan: fixed

PR #959, opened 2019-06-05
add bls gen

mikedoan:
1/ add bls gen in wallet
2/ integrate with harmony/main.go


mikedoan:
okay. so it's also more about keystore management too right?

LeoHChen:
> okay. so it's also more about keystore management too right?

yes, just make it a bit nicer and more secure to our FN runners.

internal/blsgen/lib.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func encryptFile(filename string, data []byte, passphrase string) {
+	err := ioutil.WriteFile(filename, data, 0777)
+	check(err)
+}
+
+func decryptFile(filename string, passphrase string) []byte {
+	data, _ := ioutil.ReadFile(filename)
```
LeoHChen: no error checking?

internal/blsgen/lib.go
```diff
@@ -0,0 +1,143 @@
...
+func check(err error) {
+	// handle this error
+	if err != nil {
+		// print it out
+		fmt.Println(err)
+	}
+}
+
+func encryptFile(filename string, data []byte, passphrase string) {
+	err := ioutil.WriteFile(filename, data, 0777)
```
LeoHChen: 0777 => 0600

internal/blsgen/lib.go
```diff
@@ -0,0 +1,143 @@
...
+
+func check(err error) {
+	// handle this error
+	if err != nil {
+		// print it out
+		fmt.Println(err)
+	}
+}
+
+func encryptFile(filename string, data []byte, passphrase string) {
```
LeoHChen: you didn't use passphrase?

test/blsgen/main.go
```diff
@@ -0,0 +1,110 @@
...
+	nonceSize := gcm.NonceSize()
+	nonce, ciphertext := data[:nonceSize], data[nonceSize:]
+	plaintext, err := gcm.Open(nil, nonce, ciphertext, nil)
+	if err != nil {
+		panic(err.Error())
+	}
+	return plaintext
+}
+
+func encryptFile(filename string, data []byte, passphrase string) {
```
LeoHChen: many duplications here.

internal/blsgen/lib.go
mikedoan: this func gets removed

test/blsgen/main.go
```diff
@@ -0,0 +1,110 @@
+package main
```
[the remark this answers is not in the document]
mikedoan: this is the playground where we can use in the future.

test/blsgen/main.go
mikedoan: playground code.

cmd/client/wallet/main.go
```diff
@@ -324,6 +326,26 @@ func processExportCommand() {
...
+		password = utils.AskForPassphrase("Passphrase: ")
+		password2 := utils.AskForPassphrase("Passphrase again: ")
+		if password != password2 {
+			fmt.Printf("Passphrase doesn't match. Please try again!\n")
+			os.Exit(3)
+		}
+	}
+
+	privateKey, fileName := blsgen.GenBlsKeyWithPassPhrase(password)
+	fmt.Printf("Bls private key: %s\n", privateKey.GetHexString())
```
LeoHChen: Please print out the public key here.  That's the key user needs to send to us so that we can add it into our genesis block. @rlan35 should know it.

PR #960, opened 2019-06-05
Fix garbled validator address errors

harmony-ek:
common.Address is itself a byte array, so the %s format specifier
outputs its (binary) bytes directly without any encoding.  Explicitly
Hex() it.


harmony-ek:
> Do we have other garbled output?

Not that I noticed.  This one is particularly bad because it is output quite frequently and in one case messed up Lukas's tmux session by shifting its codepage (after which one gets gibberish and graphemes instead of regular alphabets).

Issue #961, opened 2019-06-05
[crash] concurrent map write and iteration


LeoHChen:
Latest master branch code.

---

fatal error: concurrent map iteration and map write

goroutine 113 [running]:
runtime.throw(0xfc8d8a, 0x26)
        /home/ec2-user/.gimme/versions/go1.12.linux.amd64/src/runtime/panic.go:617 +0x72 fp=0xc007fe0e88 sp=0xc007fe0e58 pc=0x432812
runtime.mapiternext(0xc0036967e0)
        /home/ec2-user/.gimme/versions/go1.12.linux.amd64/src/runtime/map.go:860 +0x520 fp=0xc007fe0f10 sp=0xc007fe0e88 pc=0x414540
runtime.mapiterinit(0xe500a0, 0xc001d4f8f0, 0xc0036967e0)
        /home/ec2-user/.gimme/versions/go1.12.linux.amd64/src/runtime/map.go:850 +0x1c9 fp=0xc007fe0f30 sp=0xc007fe0f10 pc=0x413f29
reflect.mapiterinit(0xe500a0, 0xc001d4f8f0, 0xc000277ab2)
        /home/ec2-user/.gimme/versions/go1.12.linux.amd64/src/runtime/map.go:1343 +0x54 fp=0xc007fe0f60 sp=0xc007fe0f30 pc=0x415354
reflect.Value.MapKeys(0xe500a0, 0xc00014c6a8, 0x1b5, 0x4301, 0xe500a0, 0x2)
        /home/ec2-user/.gimme/versions/go1.12.linux.amd64/src/reflect/value.go:1138 +0xc4 fp=0xc007fe0fd8 sp=0xc007fe0f60 pc=0x491c94
github.com/fjl/memsize.(*context).scanMap(0xc007fe1de0, 0xe500a0, 0xc00014c6a8, 0x1b5, 0x9306ff)
        /home/ec2-user/go/pkg/mod/github.com/fjl/memsize@v0.0.0-20180929194037-2a09253e352a/memsize.go:226 +0x10f fp=0xc007fe1088 sp=0xc007fe0fd8 pc=0x9302bf
github.com/fjl/memsize.(*context).scanContent(0xc007fe1de0, 0xc00014c6a8, 0xe500a0, 0xc00014c6a8, 0x1b5, 0xc00014c6a8)
[...]

PR #964, opened 2019-06-05
fix memsize concurrent issue

mikedoan:
fix memsize concurrent issue


mikedoan:
https://github.com/harmony-one/harmony/issues/961

PR #965, opened 2019-06-05
VRF generation using BLS (base code and test case)

coolcottontail:
## Issue
Added VRF using BLS, implemented following the requirement and naming convention of 
https://tools.ietf.org/html/draft-irtf-cfrg-vrf-04 

## Test

unit level tests are in crypto/bls/bls_vrf_test.go, 

#### Test Coverage Data

Google's p256 tests are used in unit test


coolcottontail:
Jenkins, test this please.

crypto/vrf/bls/bls_vrf_test.go
```diff
@@ -0,0 +1,131 @@
...
+	_, proof := k.Evaluate(data)
+	for i := 0; i < len(proof)*8; i++ {
+		// Flip bit in position i.
+		if _, err := pk.ProofToHash(data, flipBit(proof, i)); err == nil {
+			t.Errorf("Verify unexpectedly succeeded after flipping bit %v of vrf", i)
+		}
+	}
+}
+
+func flipBit(a []byte, pos int) []byte {
```
rlan35: What is this for?

coolcottontail: This is just a utility function to flip bits in a byte stream. I copied from google's p256 test file so that I can run the same bit flipping test (flip a bit in proof so that verification should fail).

PR #970, opened 2019-06-06
fix node syncing stability issue

chaosma:


consensus/consensus_v2.go
```diff
@@ -618,11 +630,11 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
...
 	utils.GetLogInstance().Info("tryCatchup: commit new blocks", "blockNum", consensus.blockNum)
-	if consensus.phase != Commit && consensus.mode.Mode() == Normal {
-		return
-	}
+	//	if consensus.phase != Commit && consensus.mode.Mode() == Normal {
+	//		return
+	//	}
 	currentBlockNum := consensus.blockNum
-	consensus.switchPhase(Announce)
+	consensus.phase = Announce
```
rlan35: If i's a hard override, better to be in the switchPhase() too. phase switch on in one place. like "switchPhase(Announce, override: true)"

chaosma: will add it in another pr with 2f+1 signature checking

PR #971, opened 2019-06-06
VRF using BLS integration to harmony, replacing VRF p256

coolcottontail:
## Issue

Use the BLS signature instead of P256 signature for VRF

## Test

Tested the VRF verification across 10 epoches.

### DIFF
--- a/drand/drand.go
+++ b/drand/drand.go
@@ -12,7 +12,7 @@ import (
 	"github.com/harmony-one/harmony/core/types"
 	bls_cosi "github.com/harmony-one/harmony/crypto/bls"
 	"github.com/harmony-one/harmony/crypto/vrf"
-	"github.com/harmony-one/harmony/crypto/vrf/p256"
+	vrf_bls "github.com/harmony-one/harmony/crypto/vrf/bls"
 	"github.com/harmony-one/harmony/p2p"
 )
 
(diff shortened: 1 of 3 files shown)


drand/drand.go
```diff
@@ -122,9 +122,11 @@ func New(host p2p.Host, ShardID uint32, peers []p2p.Peer, leader p2p.Peer, confi
 	}
 
 	// VRF keys
-	priKey, pubKey := p256.GenerateKey()
-	dRand.vrfPriKey = &priKey
-	dRand.vrfPubKey = &pubKey
+	sk := vrf_bls.NewVRFSigner(blsPriKey)
+	pk := vrf_bls.NewVRFVerifier(blsPriKey.GetPublicKey())
+	dRand.vrfPriKey = &sk
```
rlan35: We should instead reuse the blsKey in the consensus object. Shouldn't generate a separate bls key anymore.

coolcottontail: The node consensus private key (blsPrivKey) is reused here. It's just been typecast-ed to the base rypto/vrf interface. 

func NewVRFSigner(seck *bls.SecretKey) vrf.PrivateKey {
	return &PrivateKey{seck}
}

coolcottontail:
replaced with a new PR
https://github.com/harmony-one/harmony/pull/1257

PR #982, opened 2019-06-08
New gensis module

LeoHChen:
Merge GenesisAccount with BLSAccount data structure, and create a new genesis module, with separate Harmony accounts and FN accounts.


rlan35:
Please sync to master so the jenkins can pass.

PR #983, opened 2019-06-08
Change shard state ECDSA address to use common address/byte format

rlan35:
It's better to have fixed-length bytes than string in the blockchain data structure.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -334,7 +334,7 @@ func accumulateRewards(
 		numAccounts++
 		account := member.EcdsaAddress
 		getLogger().Info("rewarding block signer",
-			"account", account,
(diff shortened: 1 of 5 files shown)


consensus/consensus.go
```diff
@@ -332,9 +332,9 @@ func accumulateRewards(
 			continue
 		}
 		numAccounts++
-		account := common.HexToAddress(member.EcdsaAddress)
+		account := member.EcdsaAddress
 		getLogger().Info("rewarding block signer",
-			"account", account,
+			"account", common.HexToAddress(account),
```
harmony-ek: Actually my bad, we don't need this change.  x_x

```suggestion
			"account", account,
```

PR #984, opened 2019-06-08
fix nil pointer crash in onNewView function

chaosma:


chaosma:
Jenkins, test this please

consensus/view_change.go
```diff
@@ -401,7 +401,8 @@ func (consensus *Consensus) onNewView(msg *msg_pb.Message) {
 	consensus.vcLock.Lock()
 	defer consensus.vcLock.Unlock()
 
-	if recvMsg.M3AggSig == nil {
+	if recvMsg.M3AggSig == nil || recvMsg.M3Bitmap == nil {
+		utils.GetLogInstance().Debug("onNewView M3AggSig or M3Bitmap is nil")
```
mikedoan: is it possibly expected? why not error?

chaosma: good point. I changed to error

PR #1001, opened 2019-06-10
blsgen lib and wallet bls generator

mikedoan:
blsgen lib and wallet bls generator

### DIFF
--- a/internal/blsgen/lib.go
+++ b/internal/blsgen/lib.go
@@ -1,0 +1,105 @@
+package blsgen
+
+import (
+	"crypto/aes"
+	"crypto/cipher"
+	"crypto/md5"
+	"crypto/rand"
+	"encoding/hex"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"time"
+
+	ffi_bls "github.com/harmony-one/bls/ffi/go/bls"
+	"github.com/harmony-one/harmony/crypto/bls"
+)
+
+func toISO8601(t time.Time) string {
+	var tz string
+	name, offset := t.Zone()
+	if name == "UTC" {
+		tz = "Z"
+	} else {
+		tz = fmt.Sprintf("%03d00", offset/3600)
+	}
+	return fmt.Sprintf("%04d-%02d-%02dT%02d-%02d-%02d.%09d%s",
+		t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), tz)
+}
+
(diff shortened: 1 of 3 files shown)


mikedoan:
an example of running the wallet

```Minhs-MacBook-Pro:harmony minhdoan$ ./bin/wallet blsgen
Passphrase: 
Passphrase again: 
Bls private key: 65b64e6efe3a3bd27ddf44f0cc8acb2e8ce81db5bf390cf386eced735a371b48
Bls public key: 1 14ff05ef290fcf3b06ee9008e3ec26af88d3ab858bb9ea33814dc8ffc28c2281f6156c8dcc2ed7a6922f5a4c5553f78e a7675b7debc39a15c108f5fff8da847588a82f92e9a1ddf312bcd8e4436528d9b1586288c14d730b18b9ba75777170c
File storing the ENCRYPTED private key with your passphrase: UTC--2019-06-10T18-56-32.008026000Z--bls_473971030ec6e35ade6241e84dc8ff7e949fe54e
```

internal/blsgen/lib.go
```diff
@@ -0,0 +1,105 @@
...
+	} else {
+		tz = fmt.Sprintf("%03d00", offset/3600)
+	}
+	return fmt.Sprintf("%04d-%02d-%02dT%02d-%02d-%02d.%09d%s",
+		t.Year(), t.Month(), t.Day(), t.Hour(), t.Minute(), t.Second(), t.Nanosecond(), tz)
+}
+
+func keyFileName(publicKey *ffi_bls.PublicKey) string {
+	ts := time.Now().UTC()
+	address := publicKey.GetAddress()
```
rlan35: Use public key instead.

mikedoan: public key is too long - just use bls

internal/blsgen/lib.go
```diff
@@ -0,0 +1,105 @@
...
+	return priKey
+}
+
+func createHash(key string) string {
+	hasher := md5.New()
+	hasher.Write([]byte(key))
+	return hex.EncodeToString(hasher.Sum(nil))
+}
+
+func encrypt(data []byte, passphrase string) []byte {
```
rlan35: Is this the standard process similar to ethereum's keystore? Can we instead adapt ethereum's keystore code for BLS?

mikedoan: keystore in ethereum is just a management box around Address common.Address and PrivateKey *ecdsa.PrivateKey - here we are dealing with different cryptography

PR #1003, opened 2019-06-10
use serialized public key instead of standard public key in wallet tool

mikedoan:
use serialized public key instead of standard public key in wallet tool


mikedoan:
Jenkins, test this please.

PR #1009, opened 2019-06-11
script used to find and upload libs to aws in the future.

mikedoan:
script used to find and upload libs to aws in the future.


mikedoan:
Jenkins, test this please.

mikedoan:
Jenkins is very unstable - this file does nothing with our code.

LeoHChen:
this script only works for macos, please add proper check for linux as well. also, this script shall be integrated into scripts/go_executable_build.sh.  No need to add another one.

PR #1011, opened 2019-06-11
quick fix for shard distro

mikedoan:
quick fix for shard distro


mikedoan:
Jenkins, test this please.

PR #1013, opened 2019-06-11
Use new bls key format.

mikedoan:
Use new bls key format.


cmd/harmony/main.go
```diff
@@ -210,6 +213,29 @@ func initSetup() {
 	hmykey.SetHmyPass(myPass)
 }
 
+func setUpConsensusKey(nodeConfig *nodeconfig.ConfigType) {
+	consensusPriKey := &bls.SecretKey{}
+
+	// If FN node running, they should either specify blsPrivateKey or the file with passphrase
+	if *blsPrivateKey != "" {
+		consensusPriKey.DeserializeHexStr(*blsPrivateKey)
```
rlan35: Put error checking and log the details so debugging is easier.

PR #1018, opened 2019-06-12
Blskey encrypted keys

mikedoan:
Blskey encrypted keys


mikedoan:
we don't want human readable - it's pretty bad that .hmy/keystore can be easy to decompress.

mikedoan:
@LeoHChen i'm submitting this now - we need to support bls encrypted first - we can change to readable format as you are thinking - but we can discuss about this later.

chaosma:
Human readable should be okay. they are encrypted, will be safe.

PR #1019, opened 2019-06-12
Update FN bls keys, the last version have some problems.

rlan35:


mikedoan:
https://github.com/harmony-one/harmony/pull/1013/ is going to clean up that logic

PR #1024, opened 2019-06-12
Fix wallet balance inconsistency issue

rlan35:


alajko:
It trusts the first node more than other nodes.

cmd/client/wallet/main.go
```diff
@@ -563,34 +571,46 @@ func convertBalanceIntoReadableFormat(balance *big.Int) string {
...
+				response, err := client.GetBalance(address)
+				if err != nil {
+					log.Info("failed to get balance, retrying ...")
+					time.Sleep(200 * time.Millisecond)
+					continue
+				}
+				log.Debug("FetchBalance", "response", response)
+				respBalance := big.NewInt(0)
+				respBalance.SetBytes(response.Balance)
+				if balance.Cmp(respBalance) < 0 {
```
chaosma: what if some server is slower, and the balance is greater than previous but not the latest? should we loop through the end of the loop and pick the maximum value as latest balance

alajko: I don't think picking maximum balance is correct. Maximum nonce, might be more correct to reflect more recent transactions.

alajko:
We just trust the first node as vs trusting a randomn node. So I don't understand this change.

cmd/client/wallet/main.go
chaosma: right, maximum nonce

alajko: Maximum value might not be the correct state as well.

rlan35: yea, we have to rely on consistent nodes to check. This is already working fine. No major problem so far.

alajko: Ideally the correct way is picking max of (nonce, balance). (In that order) per shard.

PR #1029, opened 2019-06-13
new node join stability tweak

chaosma:


chaosma:
Jenkins, test this please

consensus/consensus_service.go
```diff
@@ -551,6 +551,7 @@ func (consensus *Consensus) RegisterRndChannel(rndChannel chan [64]byte) {
 func (consensus *Consensus) checkViewID(msg *PbftMessage) error {
 	// just ignore consensus check for the first time when node join
 	if consensus.ignoreViewIDCheck {
+		consensus.mode.SetMode(Normal)
```
mikedoan: can you write a comment of reasoning why you set this? someone can pick up the code easily. also in future we could understand your code too.

chaosma: I test it locally. Both way works. But I believe this tweak will make it better. Because before set to normal, we will always accept new consensus messages. It means the later set to normal the better

PR #1036, opened 2019-06-13
Add new encrypted key and lib to find key from keystore.

mikedoan:
Add new encrypted key and lib to find key from keystore.


LeoHChen:
why Jenkins failed completely? @harmony-ek , @mikedoan

PR #1037, opened 2019-06-13
migration to bech32

alajko:
Migration to Bech32


alajko:
Subsumed https://github.com/harmony-one/harmony/pull/1032/files

PR #1039, opened 2019-06-14
remove address and use publickey in consensus; clean up consensus code

chaosma:
* remove the common.Address or Bech32 address usage in consensus to avoid confusion; use publickey instead
* clean up some codes


chaosma:
Jenkins, test this please

PR #1044, opened 2019-06-14
logging fix

mikedoan:
logging fix


test/blsprivate/main.go
```diff
@@ -0,0 +1,491 @@
+package main
```
LeoHChen: don't check in this blsprivate.

PR #1053, opened 2019-06-14
Fix consensus stability

rlan35:
1. node/node_newblock:  WaitForConsensusReady: add block proposal retry the same block 
2. node/node_handler: remove msgchan timeout channel
3. add shardID in consensus msg
4. tryAnnounce -> Announce
5. BAD BLOCK found, remove my blockchain for some blocks


consensus/consensus_v2.go
```diff
@@ -37,6 +37,11 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
 		return
 	}
 
+	if msg.GetConsensus().ShardId != consensus.ShardID {
```
chaosma: view change message has different format, msg.GetViewchange(). need separate checking

rlan35: resolved.

PR #1060, opened 2019-06-15
Simplify build process with Makefile

nyetwurk:
## Issue

Simplify building using makefile

## Test

Source code did not change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,19 @@
+TOP:=$(realpath ..)
+export CGO_CFLAGS:=-I$(TOP)/bls/include -I$(TOP)/mcl/include -I/usr/local/opt/openssl/include
+export CGO_LDFLAGS:=-L$(TOP)/bls/lib -L/usr/local/opt/openssl/lib
+export LD_LIBRARY_PATH:=$(TOP)/bls/lib:$(TOP)/mcl/lib:/usr/local/opt/openssl/lib
+export LIBRARY_PATH:=$(LD_LIBRARY_PATH)
+export DYLD_FALLBACK_LIBRARY_PATH:=$(LD_LIBRARY_PATH)
+export GO111MODULE:=on
+
+.PHONY: all libs exe
+
+all: libs
+	./scripts/go_executable_build.sh
+
+libs:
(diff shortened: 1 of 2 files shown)


harmony-ek:

nyetwurk:
The various random targets in the README (wallet, tx generator etc.) can also probably be added to the Makefile

nyetwurk:
Understood. The problem with relying on pkg-config is that it presumes the builder is going to actually install the mcl and bls libraries after building them, which the current build procedure (as documented by the readme) correctly (imo) does not require.

The question is whether the long term plans involve changing this. From a strict deb/brew packaging perspective, this would require also packaging mcl and bls. 

If the intent though is to allow a completely local build, IMO the procedure described in README.md should still be automated. The question then is if that automation belongs in the harmony repo itself or some sort of meta building system repo.

harmony-ek:
Right.  pkg-config is maybe a little too deep.  However, a local build should still be achievable using a throwaway/local installation prefix such as `$HOME/hmylocal` (inside which the usual `bin`, `lib`, `include`, etc. would be found), so what we need is a minimum build/install logic into a designated prefix.  (mcl/bls actually uses CMake as an alternative build system, so this shouldn't be too hard.)  pkg-config integration is one step further than this, and full deb/rpm/brew packaging a few steps more.

nyetwurk:
Looking into packaging for mcl/bls

If i submit a PR with a debian/ directory and update the build instructions to include building/installing deb packages will it be acceped? Brew and rpm would still have to be done separately but at least this would be a start to the larger packaging effort which imo should be used for updating rather than the fragile and unsecure current curl method.

LeoHChen:
the packaging is a step further. the current makefile is simple enough to be usable.  Maybe we don't make libs a dependency of exes, as we don't usually change libs.  But we may include a target of libs, so that it can be built on demand.  Thanks @nyetwurk

nyetwurk:
Due to makefile trickery, "make all" will not work because i would have to do `all: libs exes`, but dependencies are not ordered if they are in the same rule, so I can't force libs to come before exes w/o making a separate, third target. I'll add it anyway and tell me what you think.

PR #1064, opened 2019-06-16
Addbiz nodes

alajko:
New FN.


core/resharding.go
```diff
@@ -27,7 +27,7 @@ const (
 	// GenesisShardSize is the size of each shard at genesis
 	GenesisShardSize = 100
 	// GenesisShardHarmonyNodes is the number of harmony node at each shard
-	GenesisShardHarmonyNodes = 75
+	GenesisShardHarmonyNodes = 74
```
rlan35: Why change to 74?

alajko: Discussed offline.

rlan35: Please change back to 75.

alajko: Dicussed offline, Please approve.

PR #1073, opened 2019-06-16
Change BlsPriKey to DummyKey

rlan35:


core/resharding.go
```diff
@@ -23,7 +23,7 @@ const (
 	// FirstEpoch is the number of the first epoch.
 	FirstEpoch = 1
 	// GenesisShardNum is the number of shard at genesis
-	GenesisShardNum = 4
+	GenesisShardNum = 1
```
mikedoan: this is for debugging only?

cmd/harmony/main.go
```diff
@@ -253,9 +253,9 @@ func createGlobalConfig() *nodeconfig.ConfigType {
 	consensusPriKey := &bls.SecretKey{}
 
 	if *isGenesis {
-		err := consensusPriKey.DeserializeHexStr(genesisAccount.BlsPriKey)
+		err := consensusPriKey.DeserializeHexStr(genesisAccount.DummyKey)
```
mikedoan: why using this DymmyKey? BlsPriKey is also generated keys by us currently?

core/resharding.go
rlan35: Sorry my bad. Reverted this line.

cmd/harmony/main.go
rlan35: To avoid people's confusion.

PR #1079, opened 2019-06-16
move block broadcast to prepared phase

chaosma:
* move block broadcast to prepared phase
* move block verifier earlier from tryCatchup
* make finalizeCommits logic robust in case of the leader change


consensus/consensus_v2.go
```diff
@@ -136,83 +137,51 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
 
-	// check block header is valid
-	var blockObj types.Block
-	err = rlp.DecodeBytes(block, &blockObj)
-	if err != nil {
-		consensus.getLogger().Warn("[OnAnnounce] Unparseable block header data", "error", err, "MsgBlockNum", recvMsg.BlockNum, "MsgPayloadBlockNum", blockObj.NumberU64())
-		return
-	}
-
-	if blockObj.NumberU64() != recvMsg.BlockNum || recvMsg.BlockNum < consensus.blockNum {
```
rlan35: recvMsg.BlockNum < consensus.blockNum   should still apply.

consensus/consensus_leader_msg.go
```diff
@@ -21,8 +21,6 @@ func (consensus *Consensus) constructAnnounceMessage() []byte {
 	}
 	consensusMsg := message.GetConsensus()
 	consensus.populateMessageFields(consensusMsg)
-	// n byte of block header
-	consensusMsg.Payload = consensus.block // TODO: send only block header in the announce phase.
```
rlan35: Announce payload should still contain block header data. Otherwise, announce message is totally irrelevant to the new proposed block.

consensus/consensus_v2.go
```diff
@@ -136,83 +137,51 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
-		consensus.getLogger().Warn("[OnAnnounce] BlockNum not match", "MsgBlockNum", recvMsg.BlockNum, "blockNum", blockObj.NumberU64())
-		return
-	}
-
-	if consensus.mode.Mode() == Normal {
-		// skip verify header when node is in Syncing mode
-		if err := consensus.VerifyHeader(consensus.ChainReader, blockObj.Header(), false); err != nil {
```
rlan35: OnAnnounce should still verify the header info.

consensus/consensus_v2.go
```diff
@@ -357,32 +326,55 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
...
+	if blockObj.NumberU64() != recvMsg.BlockNum || recvMsg.BlockNum < consensus.blockNum {
+		consensus.getLogger().Warn("[OnPrepared] BlockNum not match", "MsgBlockNum", recvMsg.BlockNum, "blockNum", blockObj.NumberU64())
+		return
+	}
+	if blockObj.Header().Hash() != recvMsg.BlockHash {
+		consensus.getLogger().Warn("[OnPrepared] BlockHash not match", "MsgBlockNum", recvMsg.BlockNum, "MsgBlockHash", recvMsg.BlockHash, "blockObjHash", blockObj.Header().Hash())
+		return
+	}
+	if consensus.mode.Mode() == Normal {
+		if err := consensus.VerifyHeader(consensus.ChainReader, blockObj.Header(), false); err != nil {
```
rlan35: We should add BlockVerifier check too. I think we forgot to do that. (we basically did verify the block content all these time....)

api/proto/message/message.proto
```diff
@@ -86,6 +86,7 @@ message ConsensusRequest {
   bytes block_hash = 4;
   bytes sender_pubkey = 5;
   bytes payload = 6;
+  bytes block = 7;
```
rlan35: Move block under block_hash with identifier 5 "block = 5;".   we can still refresh the protobuf definition now. Let's do it.

consensus/consensus_v2.go
```diff
@@ -357,32 +326,55 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
...
+			consensus.getLogger().Warn("[OnPrepared] Block content is not verified successfully", "error", err, "inChain", consensus.ChainReader.CurrentHeader().Number, "MsgBlockNum", blockObj.Header().Number)
+			return
+		}
+	}
+
+	consensus.getLogger().Debug("[OnPrepared] Prepared message and block added", "MsgViewID", recvMsg.ViewID, "MsgBlockNum", recvMsg.BlockNum, "blockHash", recvMsg.BlockHash)
```
rlan35: Move this  log after they are truly added.

consensus/consensus_v2.go
```diff
@@ -648,13 +646,14 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 		return
 	}
 
+	consensus.getLogger().Debug("[OnCommitted] Committed message added", "MsgViewID", recvMsg.ViewID, "MsgBlockNum", recvMsg.BlockNum)
```
rlan35: Move After AddMessage.

consensus/consensus_leader_msg.go
chaosma: block header added

consensus/consensus_v2.go
chaosma: right, added

consensus/consensus_v2.go
chaosma: earlier check is better. I moved the blockverifier check here from tryCatchup, so we don't have to check it again in tryCatchup

consensus/consensus_v2.go
chaosma: add log should be not error, but I moved it for logical consistency

consensus/consensus_v2.go
chaosma: moved

consensus/consensus_v2.go
rlan35: I see. So we did have blockverifier. Just within tryCatchup. I thought we didn't verify the block content.

consensus/consensus.go
```diff
@@ -103,6 +103,8 @@ type Consensus struct {
 	blockHash [32]byte
 	// Block to run consensus on
 	block []byte
+	// BlockHeader to run consensus on
+	blockHeader []byte
```
rlan35: Should reset/clear this data when it's not valid anymore (e.g. ResetState). Otherwise, it may mess things up.

chaosma: reset; I also modify the logic of leader finalizeCommits to make it robust in case of leader change

PR #1083, opened 2019-06-18
add hard limit for node syncing request to avoid grpc connection

chaosma:
* add hard limit for node syncing request to avoid grpc connection
* clean cache after finish state syncing


api/service/syncing/syncing.go
```diff
@@ -106,6 +107,26 @@ type StateSync struct {
 	syncMux            sync.Mutex
 }
 
+// level = 0 only clean "old" blocks, level > 0 clean up everything
+func (ss *StateSync) cleanCache(level int) {
```
harmony-ek: Code smell is strong with this one.

1. Lines 119–127 are an almost exact duplicate of lines 114–118, with only one diffecence: `configPeer.newBlocks` is `nil`-ed out.
2. `level int` arguments looks like a magic constant from call sites (e.g. around line 683, where we call this twice, once with 0 then again with 1.  Reader of that code will not know what is going on.

Can you please refactor this?  After refactoring, the following should be met:

1. `cleanCache(0)` ⇒ `purgeOldBlocksFromCache()`
2. `cleanCache(1)` ⇒ `purgeAllBlocksFromCache()`
3. Looping is done only once.
4. If you need a `clear all` boolean flag, the flag should be of type `bool` and named appropriately, e.g. `purgeNewOnesToo bool`.

api/service/syncing/syncing.go
```diff
@@ -290,18 +311,22 @@ func (sc *SyncConfig) GetBlockHashesConsensusAndCleanUp() bool {
...
 		ss.syncConfig.ForEachPeer(func(peerConfig *SyncPeerConfig) (brk bool) {
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
-				response := peerConfig.client.GetBlockHashes(startHash)
+				response := peerConfig.client.GetBlockHashes(startHash, size)
 				if response == nil {
 					return
 				}
+				if len(response.Payload) > int(size+1) {
```
harmony-ek: Per robustness principle, be liberal in what you accept.  Just accept the first `size` and discard the rest.

node/node_syncing.go
```diff
@@ -200,18 +200,30 @@ func (node *Node) CalculateResponse(request *downloader_pb.DownloaderRequest) (*
...
+		startHeight := startBlock.NumberU64()
+		endHeight := node.Blockchain().CurrentBlock().NumberU64()
+		if startHeight >= endHeight {
+			return response, fmt.Errorf("[SYNC] GetBlockHashes Request failed. I am not higher than requested node, my Height %v, request node Height %v", endHeight, startHeight)
+		}
+		if endHeight-startHeight > size {
+			endHeight = startHeight + size
+		}
+		for blockNum := startHeight; blockNum <= endHeight; blockNum++ {
+			block := node.Blockchain().GetBlockByNumber(blockNum)
```
harmony-ek: Check `block == nil` here.  In fact, we do not have to loop to `endHeight`, but simply start at `startHeight` and keep going until `len(response.Payload) >= size` or `block == nil` (which indicates that the chain has ended at `blockNum-1`).

api/service/syncing/syncing.go
chaosma: added the logic

api/service/syncing/syncing.go
chaosma: refactor the cleanCache into two for readability

node/node_syncing.go
chaosma: good point!

PR #1085, opened 2019-06-18
Remove dummy

mikedoan:
Remove dummy and enable using public keys.


cmd/client/txgen/main.go
```diff
@@ -110,13 +110,12 @@ func setUpTXGen() *node.Node {
 	startIdx := 0
 	endIdx := startIdx + core.GenesisShardSize
 	for _, acct := range genesis.GenesisAccounts[startIdx:endIdx] {
-		secretKey := bls2.SecretKey{}
-		if err := secretKey.DeserializeHexStr(acct.DummyKey); err != nil {
-			_, _ = fmt.Fprintf(os.Stderr, "cannot parse secret key: %v\n",
-				err)
+		pub := &bls2.PublicKey{}
+		if err := pub.DeserializeHexStr(acct.BlsPublicKey); err != nil {
+			fmt.Println("Can not deserialize publickey")
```
rlan35: "publickey" => "public key"; also print out the error.

PR #1100, opened 2019-06-19
Call GetPassphraseFromSource only once

harmony-ek:
When `GetPassphraseFromSource` is used with `stdin` or `fd:xyz`, it exhausts the standard input or the given file descriptor `xyz`, so a second call to it returns an empty string.  Deal with this by ensuring it is called only once from `main`.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -102,6 +102,7 @@ var (
 	enableGC           = flag.Bool("enableGC", true, "Enable calling garbage collector manually .")
 	blsKeyFile         = flag.String("blskey_file", "", "The encrypted file of bls serialized private key by passphrase.")
 	blsPass            = flag.String("blspass", "", "The file containing passphrase to decrypt the encrypted bls file.")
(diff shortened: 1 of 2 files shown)


cmd/harmony/main.go
```diff
@@ -182,35 +183,24 @@ func initSetup() {
...
-			os.Exit(100)
-		}
+	consensusPriKey, err := blsgen.LoadBlsKeyWithPassPhrase(*blsKeyFile, blsPassphrase)
+	if err != nil {
+		fmt.Printf("error when loading bls key, err :%v\n", err)
+		os.Exit(100)
+	}
+	index, acc := genesis.IsBlsPublicKeyIndex(consensusPriKey.GetPublicKey().SerializeToHexStr())
+	if index < 0 {
+		fmt.Println("Can not found your bls key.")
```
rlan35: Can not find*

harmony-ek: LOL that was just moved from `setupConsensusAndNode`… will fix separately

PR #1101, opened 2019-06-19
Fix “multiple key files found” if pwd is $HOME

harmony-ek:
This is because the same file will be discovered twice.  Identify this case by using the same-file test(1) predicate (`X -ef Y`) and error out if the files are distinct.

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -143,7 +143,8 @@ case "${BLSKEYFILE}" in
          BLSKEYFILE="${f}"
          ;;
       *)
-         err 69 "multiple key files found (${f}, ${BLSKEYFILE}); please use -k to specify"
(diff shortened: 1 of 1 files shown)


scripts/node.sh
```diff
@@ -143,7 +143,8 @@ case "${BLSKEYFILE}" in
          BLSKEYFILE="${f}"
          ;;
       *)
-         err 69 "multiple key files found (${f}, ${BLSKEYFILE}); please use -k to specify"
+         [ "${f}" -ef "${BLSKEYFILE}" ] || \
```
LeoHChen: "-ef" is for testing hardlink. https://www.tldp.org/LDP/abs/html/fto.html . Is that what you want to test?

harmony-ek: Yes.

PR #1111, opened 2019-06-20
add import tool for bls

mikedoan:
add import tool for bls

-----
Passphrase:      
Passphrase again:           
Your encrypted bls file with the passphrased is written at 8fb9de02c7f9b8baad2905994c6d9f559b99a72fdd6aa20eab692bb1fcc25c712ba7410d4dae8dad182d714e80d7d787.key.                                                
We loaded the file and the loaded private key DOES match your private key!

### DIFF
--- a/cmd/client/wallet/main.go
+++ b/cmd/client/wallet/main.go
@@ -583,7 +583,7 @@ func importBls() {
 		}
 		fmt.Println("We loaded the file and the loaded private key DOES match your private key!")
 	} else {
-		fmt.Println("The private key should not be empty.")
+		fmt.Println("Please specify the hexadecimal private key string using --key")
 	}
 }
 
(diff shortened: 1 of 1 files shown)


cmd/client/wallet/main.go
```diff
@@ -537,6 +545,48 @@ func blsRecoveryCommand() {
...
+			fmt.Printf("Error when loading the private key with the passphrase. Err: %v", err)
+			os.Exit(101)
+		}
+		if !privateKey.IsEqual(privateKey2) {
+			fmt.Println("The loaded private key does not match your private key.")
+			os.Exit(101)
+		}
+		fmt.Println("We loaded the file and the loaded private key DOES match your private key!")
+	} else {
+		fmt.Println("The private key should not be empty.")
```
harmony-ek: I can see the user going “Huh?  Empty?  Did I specify some empty thing?  I didn't?”  Let's frame from the user's perspective:

```suggestion
		fmt.Println("Please specify the hexadecimal private key string using --key")
```

PR #1114, opened 2019-06-20
fix path of binaries for wallet.sh

mikedoan:
fix path of binaries for wallet.sh

### DIFF
--- a/scripts/wallet.sh
+++ b/scripts/wallet.sh
@@ -6,11 +6,11 @@ REL=drum
 
 case "$OS" in
     Darwin)
-        FOLDER=release/${REL}/darwin-x86_64/
+        FOLDER=release/darwin-x86_64/${REL}/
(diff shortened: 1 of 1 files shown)


scripts/wallet.sh
```diff
@@ -43,7 +43,19 @@ Commands:
         --amount         - The amount of token to transfer
         --shardID        - The shard Id for the transfer
         --inputData      - Base64-encoded input data to embed in the transaction
-    8. blsgen        - Generates bls keys with passphrase and store the private key locally
+    8. export        - Export account key to a new file")
```
LeoHChen: what's the ")?

mikedoan: my bad!

mikedoan:
tested in linux machine too.
```
Passphrase: 
Passphrase again: 
Your encrypted bls file with the passphrased is written at 8fb9de02c7f9b8baad2905994c6d9f559b99a72fdd6aa20eab692bb1fcc25c712ba7410d4dae8dad182d714e80d7d787.key.
We loaded the file and the loaded private key DOES match your private key!
[ec2-user@ip-172-31-28-31 ~]$ ls
```

PR #1115, opened 2019-06-20
[node.sh] run node.sh as non-root user is possible

LeoHChen:


LeoHChen:
> A few:
> 
> * Help says `-s` runs setup env only, but it does not.
> * If `-S` is given and we skip `setup_env`, we should at least check if the environment has been set correctly.

* fixed '-s' option, exit after the check

PR #1122, opened 2019-06-21
fix cpu issues at networkinfo service

mikedoan:
fix cpu issues at networkinfo service

### DIFF
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -48,6 +48,9 @@ const (
 
 	// register to bootnode every ticker
 	dhtTicker = 6 * time.Hour
+
+	// wait for peerinfo.
+	peerInfoWait = time.Second
 )
 
 // New returns role conversion service.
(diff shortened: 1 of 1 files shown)


core/resharding.go
```diff
@@ -24,11 +24,11 @@ const (
 	// FirstEpoch is the number of the first epoch.
 	FirstEpoch = 1
 	// GenesisShardNum is the number of shard at genesis
-	GenesisShardNum = 4
```
LeoHChen: don't check in debug change.

api/service/networkinfo/service.go
```diff
@@ -161,38 +164,46 @@ func (s *Service) DoService() {
...
-					if err != nil {
-						continue
-					}
-					nip := netaddr.(*net.TCPAddr).IP
-					if (nip.IsGlobalUnicast() && !utils.IsPrivateIP(nip)) || cgnPrefix.Contains(nip) {
-						ip = nip.String()
-						port = fmt.Sprintf("%d", netaddr.(*net.TCPAddr).Port)
+		case <-peerInfoTick.C:
+			select {
+			case peer := <-s.peerInfo:
```
rlan35: make a non-blocking channel.

PR #1124, opened 2019-06-21
Add VerifySeal; move commit sig to next block

chaosma:
* remove prepared/commit sig, add lastcommitsig in header field;
* modify accumulateReward according to lastcommitsig
* Add VerifySeal for PoS


consensus/consensus.go
```diff
@@ -340,7 +340,7 @@ func accumulateRewards(
 	if err != nil {
 		return ctxerror.New("cannot create group sig mask").WithCause(err)
 	}
-	if err := mask.SetMask(parentHeader.CommitBitmap); err != nil {
+	if err := mask.SetMask(header.LastCommitBitmap); err != nil {
```
rlan35: Add a todo to take care of epoch transition. Because right it's matching current bitmap with committee of last block, which only works if there is committee change.

consensus/consensus_service.go
```diff
@@ -227,9 +228,62 @@ func (consensus *Consensus) VerifyHeaders(chain consensus_engine.ChainReader, he
...
+	publicKeys, err := retrievePublicKeysFromLastBlock(chain, header)
+	if err != nil {
+		return ctxerror.New("[VerifySeal] Cannot retrieve publickeys from last block").WithCause(err)
+	}
+	payload := append(header.LastCommitSignature[:], header.LastCommitBitmap...)
+	aggSig, mask, err := readSignatureBitmapByPublicKeys(payload, publicKeys)
+	if err != nil {
+		return ctxerror.New("[VerifySeal] Unable to deserialize the LastCommitSignature and LastCommitBitmap in Block Header").WithCause(err)
+	}
+	if count := utils.CountOneBits(mask.Bitmap); count < consensus.Quorum() {
```
rlan35: consensus.Quorum() is based on committee size of current block, not last block. At least, add a todo here if not fix it now.

consensus/consensus_v2.go
```diff
@@ -736,6 +737,26 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 	return
 }
 
+// LastCommitSig returns the byte array of aggregated commit signature and bitmap of last block
+func (consensus *Consensus) LastCommitSig() ([]byte, []byte, error) {
+	if consensus.blockNum <= 1 {
+		return nil, nil, nil
+	}
+	msgs := consensus.PbftLog.GetMessagesByTypeSeq(msg_pb.MessageType_COMMITTED, consensus.blockNum-1)
```
rlan35: Add a todo to put commit sigs into db, instead of memory.

consensus/consensus.go
chaosma: added todo

consensus/consensus_v2.go
chaosma: This should be OK, since I periodically clean up the memory

consensus/consensus_service.go
chaosma: todo added

PR #1133, opened 2019-06-22
Mainnet preparation

rlan35:
Fix empty coinbase address during transaction validation; It caused mismatching receipt issue and not able to validate block on validators.
Populate coinbase with current leader's ECDSA address
  -remove BLSAddress utility to avoid confusion
Fix missing ViewID issue.
Update EVM to latest constantinople version to avoid invalid opcode issue.
Separate mainnet genesis allocation from testnet setup so mainnet have 12.6B only on shard 0.
Disable faceut/staking rpcs for mainnet.

Tested locally with both mainnet and testnet network. Tried wallet balances/transfer/getFreeToken too.

### DIFF
--- a/core/genesis.go
+++ b/core/genesis.go
@@ -259,9 +259,6 @@ func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
 		ShardStateHash: g.ShardStateHash,
 		ShardState:     shardStateBytes,
 	}
-	if g.GasLimit == 0 {
-		head.GasLimit = 10000000000 // TODO(RJ): figure out better solution. // params.GenesisGasLimit
-	}
 	statedb.Commit(false)
 	statedb.Database().TrieDB().Commit(root, true)
 
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -241,6 +241,7 @@ func NewBlock(header *Header, txs []*Transaction, receipts []*Receipt) *Block {
 	} else {
 		b.header.ReceiptHash = DeriveSha(Receipts(receipts))
 		b.header.Bloom = CreateBloom(receipts)
+		utils.GetLogInstance().Info("Receipt Root Proposed2", "root", b.header.ReceiptHash.Hex(), "len", len(receipts))
 	}
 
 	return b
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -89,13 +89,8 @@ type Header struct {
 	ShardID             uint32      `json:"shardID"          gencodec:"required"`
 	LastCommitSignature [96]byte    `json:"lastCommitSignature"  gencodec:"required"`
 	LastCommitBitmap    []byte      `json:"lastCommitBitmap"     gencodec:"required"` // Contains which validator signed
-	Vrf                 [32]byte    `json:"vrf"`
-	VrfProof            [96]byte    `json:"vrfProof"`
-	Vdf                 [258]byte   `json:"vdf"`
-	VdfProof            [258]byte   `json:"vdfProof"`
 	ShardStateHash      common.Hash `json:"shardStateRoot"`
 	ShardState          []byte      `json:"shardState"`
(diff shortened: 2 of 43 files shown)


chaosma:
Can you add some descriptions of changes?

node/node_newblock.go
```diff
@@ -71,10 +72,15 @@ func (node *Node) WaitForConsensusReadyv2(readySignal chan struct{}, stopChan ch
...
+					coinbase := node.Consensus.SelfAddress
 					// Normal tx block consensus
-					selectedTxs := node.getTransactionsForNewBlock(MaxNumberOfTransactionsPerBlock)
+					selectedTxs := types.Transactions{} // Empty transaction list
+					if node.NodeConfig.GetNetworkType() != nodeconfig.Mainnet {
+						selectedTxs = node.getTransactionsForNewBlock(MaxNumberOfTransactionsPerBlock, coinbase)
+					}
 					utils.GetLogInstance().Info("PROPOSING NEW BLOCK ------------------------------------------------", "blockNum", node.Blockchain().CurrentBlock().NumberU64()+1, "threshold", threshold, "selectedTxs", len(selectedTxs))
-					if err := node.Worker.CommitTransactions(selectedTxs); err != nil {
+					if err := node.Worker.CommitTransactions(selectedTxs, coinbase); err != nil {
```
chaosma: If set coinbase at node.getTransactionsForNewBlock, here we don't have to set coinbase again? Or we can set everything here instead of at getTransactionsForNewBlock

core/genesis.go
```diff
@@ -259,9 +259,6 @@ func (g *Genesis) ToBlock(db ethdb.Database) *types.Block {
 		ShardStateHash: g.ShardStateHash,
 		ShardState:     shardStateBytes,
 	}
-	if g.GasLimit == 0 {
```
chaosma: what is the default gas limit now?

core/types/block.go
```diff
@@ -275,10 +273,10 @@ func CopyHeader(h *Header) *Header {
 		cpy.ShardState = make([]byte, len(h.ShardState))
 		copy(cpy.ShardState, h.ShardState)
 	}
-	if len(h.CrossLinks) > 0 {
```
chaosma: If we comment out crosslinks, we have to do fork later to bring it back. probably just keep it as placeholder

node/node_genesis.go
```diff
@@ -70,10 +69,12 @@ func (node *Node) SetupGenesisBlock(db ethdb.Database, shardID uint32, myShardSt
 	switch node.NodeConfig.GetNetworkType() {
 	case nodeconfig.Mainnet:
 		chainConfig = *params.MainnetChainConfig
-		foundationAddress := common.HexToAddress("0xE25ABC3f7C3d5fB7FB81EAFd421FF1621A61107c")
-		genesisFunds := big.NewInt(GenesisFund)
-		genesisFunds = genesisFunds.Mul(genesisFunds, big.NewInt(denominations.One))
-		genesisAlloc[foundationAddress] = core.GenesisAccount{Balance: genesisFunds}
+		if shardID == 0 {
+			foundationAddress := common.HexToAddress("0xE25ABC3f7C3d5fB7FB81EAFd421FF1621A61107c")
```
chaosma: This is our real foundationAddress?

rlan35: so far, yes.

core/types/block.go
rlan35: We will have to do fork either way. Now, all these are just wasting spaces.

core/genesis.go
rlan35: Doesn't matter, as we don't process tx now.

node/node_newblock.go
rlan35: One is for update headers, another is for processing transactions. Both needed.

PR #1135, opened 2019-06-22
Print out bls pubkey

LeoHChen:


cmd/harmony/main.go
```diff
@@ -200,9 +200,10 @@ func setUpConsensusKeyAndReturnIndex(nodeConfig *nodeconfig.ConfigType) (int, *g
 		fmt.Printf("error when loading bls key, err :%v\n", err)
 		os.Exit(100)
 	}
-	index, acc := genesis.IsBlsPublicKeyIndex(consensusPriKey.GetPublicKey().SerializeToHexStr())
+	pubKey := consensusPriKey.GetPublicKey()
+	index, acc := genesis.IsBlsPublicKeyIndex(pubKey.SerializeToHexStr())
 	if index < 0 {
-		fmt.Println("Can not found your bls key.")
+		fmt.Printf("can not found your bls key: %s\n", pubKey.SerializeToHexStr())
```
harmony-ek:
```suggestion
		fmt.Printf("cannot find your BLS key in the genesis/FN tables: %s\n", pubKey.SerializeToHexStr())
```
Clarifies where the key is missing.  Otherwise people may think the key was not found on disk and start re-checking their `-k` (node.sh) option, only to find nothing wrong with it.

PR #1137, opened 2019-06-22
disable rpc for mainnet

mikedoan:
disable rpc for mainnet

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -139,12 +139,6 @@ var (
 		"Do not propose view change (testing only)")
 )
 
-const (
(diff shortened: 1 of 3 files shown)


cmd/harmony/main.go
```diff
@@ -139,6 +139,12 @@ var (
 		"Do not propose view change (testing only)")
 )
 
+const (
+	MainNet = "mainnet"
+	TestNet = "testnet"
+	DevNet  = "devnet"
```
rlan35: let's just convert the nodeconfig.Mainnet ....const to string intead of int.

PR #1139, opened 2019-06-22
get option to get public key based on encrypted bls file and passphrase

mikedoan:
get option to get public key based on encrypted bls file and passphrase

### DIFF
--- a/cmd/client/wallet/main.go
+++ b/cmd/client/wallet/main.go
@@ -112,6 +112,7 @@ var (
 
 	getBlsPublicCommand = flag.NewFlagSet("getBlsPublic", flag.ExitOnError)
 	blsKey2             = getBlsPublicCommand.String("key", "", "The raw private key.")
+	blsFile2            = getBlsPublicCommand.String("file", "", "The encrypted bls file.")
 )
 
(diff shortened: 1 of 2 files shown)


cmd/client/wallet/main.go
```diff
@@ -608,6 +610,19 @@ func getBlsPublic() {
...
+		if password != password2 {
+			fmt.Printf("Passphrase doesn't match. Please try again!\n")
+			os.Exit(100)
+		}
+		privateKey, err := blsgen.LoadBlsKeyWithPassPhrase(*blsFile2, password)
+		if err != nil {
+			fmt.Printf("error when loading bls key, err :%v\n", err)
+			os.Exit(100)
+		}
+		fmt.Printf("Your bls public key is: %s\n", privateKey.GetPublicKey().SerializeToHexStr())
```
alajko: After speaking with John, it seems he wants the user to check whether this public key matches what the user has with them.  Can you just change the printf statement to "Your generate bls public key is: ... It should match what you have".

mikedoan: discussed offline

PR #1141, opened 2019-06-23
Make fetching balances multi-thread

rlan35:


mikedoan:
this is just a nice to have for client point of view?

PR #1152, opened 2019-06-23
add peer info of incoming grpc request in state syncing

chaosma:


api/service/syncing/downloader/server.go
```diff
@@ -29,6 +30,8 @@ func (s *Server) Query(ctx context.Context, request *pb.DownloaderRequest) (*pb.
 	if err != nil {
 		return nil, err
 	}
+	p, _ := peer.FromContext(ctx)
+	utils.GetLogInstance().Debug("[SYNC] Server Received Request", "RequestType", request.Type, "PeerAddress", p.Addr)
```
rlan35: Change to Info so it will be shown.

chaosma: Sure

PR #1172, opened 2019-06-27
Fix explorer issue of not showing balance.

mikedoan:
1/ node syncing for explorer node.
2/ calculate block reward for explorer node.
3/ add setup for explorer node.

### DIFF
--- a/api/service/explorer/service.go
+++ b/api/service/explorer/service.go
@@ -267,30 +267,19 @@ func (s *Service) GetExplorerAddress(w http.ResponseWriter, r *http.Request) {
 	id := r.FormValue("id")
 	key := GetAddressKey(id)
 
-	var result interface{}
+	utils.GetLogInstance().Info("Querying address", "address", id)
+	data := &Data{}
 	defer func() {
-		if err := json.NewEncoder(w).Encode(result); err != nil {
+		if err := json.NewEncoder(w).Encode(data.Address); err != nil {
 			ctxerror.Warn(utils.WithCallerSkip(utils.GetLogInstance(), 1), err,
 				"cannot JSON-encode address")
 		}
 	}()
-	data := &Data{}
 	if id == "" {
 		return
 	}
-	db := s.storage.GetDB()
-	bytes, err := db.Get([]byte(key))
-	if err != nil {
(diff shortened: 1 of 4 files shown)


LeoHChen:
Jenkins passed.  @mikedoan , Can you elaborate a bit more on this PR on what kind of issue resolved and why we need this PR?  It's hard to understand the logic behind the "more logic". Thanks.

mikedoan:
Already did.

1/ node syncing for explorer node -> for explorer we do need to have state syncing but the explorer node does not need to send blocks for nodes like normal nodes(leader + validator)

2/ calculate block reward for explorer node. -> currently the blocks received from a explorer node do include reward info - we need to calculate the reward based on reward info attached to the blocks.

3/ add setup for explorer node.  -> fix some logic for the explorer service  /address end point.

api/service/explorer/service.go
```diff
@@ -300,7 +289,16 @@ func (s *Service) GetExplorerAddress(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	result = data.Address
+	db := s.storage.GetDB()
+	bytes, err := db.Get([]byte(key))
+	if err != nil {
+		ctxerror.Warn(utils.GetLogger(), err, "cannot read address from db", "id", id)
```
LeoHChen: why different usage here? and line 299?

node/node_explorer.go
```diff
@@ -81,6 +82,13 @@ func (node *Node) ExplorerMessageHandler(payload []byte) {
 	return
 }
 
+// AddNewBlockForExplorer add new block for explorer.
+func (node *Node) AddNewBlockForExplorer(block *types.Block) {
+	// TODO: Currently assumping blocks come in order.
```
LeoHChen: what's the impact if blocks didn't come in order?

api/service/explorer/service.go
mikedoan: i kept the same logic of the original code. just move the logic of getbalance at the front.

node/node_explorer.go
mikedoan: I'm writing another PR for this.
1/ i don't want to have a big PR
2/ i was telling another PR will come right after that - that's why i need to check in this one first.

Issue #1174, opened 2019-06-27
Unify application loggers with internal shared logger


hsienchiaolee:
## Summary

Default global logger are still used across the application instead of the custom logger in `utils`.

## Current Design

Use default global logger from `go-ethereum/log` without proper setup.

## Problems

Given that global logger does not get setup during application launches, it is not tracking some of the contexts we want to pass along through out the application logging. Also, since the file handler has not been added to the global logger, it would only output to console, but not write the logs to disk.

## Proposal

Unify loggers across the application to use shared logger that has been customized during application launch.

PR #1175, opened 2019-06-27
Unify loggers across application

hsienchiaolee:
## Issue
#1174 

Note:
* Purposely chunked changes into different commits per package.
• There are still some loggers that may need to be replaced. Skipped in this PR given lack of context and want to avoid breaking things unintentionally.

## Test
N/A, ran `./test/test_before_submit.sh`

## TODO
N/A

### DIFF
--- a/api/proto/node/node.go
+++ b/api/proto/node/node.go
@@ -6,14 +6,13 @@ import (
 	"fmt"
 	"log"
 
-	"github.com/harmony-one/harmony/internal/utils"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
-	"github.com/harmony-one/harmony/core/types"
+	peer "github.com/libp2p/go-libp2p-peer"
 
 	"github.com/harmony-one/harmony/api/proto"
-	peer "github.com/libp2p/go-libp2p-peer"
+	"github.com/harmony-one/harmony/core/types"
+	"github.com/harmony-one/harmony/internal/utils"
 )
 
 // MessageType is to indicate the specific type of message under Node category
@@ -105,7 +104,8 @@ func SerializeBlockchainSyncMessage(blockchainSyncMessage *BlockchainSyncMessage
 	encoder := gob.NewEncoder(&result)
 	err := encoder.Encode(blockchainSyncMessage)
 	if err != nil {
-		log.Panic(err)
+		utils.GetLogger().Crit("Error", err)
+		panic(err)
 	}
 	return result.Bytes()
 }
@@ -116,7 +116,8 @@ func DeserializeBlockchainSyncMessage(d []byte) (*BlockchainSyncMessage, error) 
 	decoder := gob.NewDecoder(bytes.NewReader(d))
 	err := decoder.Decode(&blockchainSyncMessage)
 	if err != nil {
-		log.Panic(err)
+		utils.GetLogger().Crit("Error", err)
+		panic(err)
 	}
 	return &blockchainSyncMessage, err
 }
--- a/api/proto/node/node.go
+++ b/api/proto/node/node.go
@@ -117,7 +117,6 @@ func DeserializeBlockchainSyncMessage(d []byte) (*BlockchainSyncMessage, error) 
 	err := decoder.Decode(&blockchainSyncMessage)
 	if err != nil {
 		utils.GetLogger().Crit("Error", err)
-		panic(err)
 	}
 	return &blockchainSyncMessage, err
 }
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -32,7 +32,6 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
-	"github.com/ethereum/go-ethereum/log"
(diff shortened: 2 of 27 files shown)


harmony-ek:

LeoHChen:
Jenkins, test this please.

api/proto/node/node.go
```diff
@@ -116,7 +116,8 @@ func DeserializeBlockchainSyncMessage(d []byte) (*BlockchainSyncMessage, error)
 	decoder := gob.NewDecoder(bytes.NewReader(d))
 	err := decoder.Decode(&blockchainSyncMessage)
 	if err != nil {
-		log.Panic(err)
+		utils.GetLogger().Crit("Error", err)
+		panic(err)
```
LeoHChen: is log.Panic will panic the process?  Or do we need to panic it here?  @mikedoan please review.

internal/utils/utils.go
```diff
@@ -64,7 +63,8 @@ func ConvertFixedDataIntoByteArray(data interface{}) []byte {
 	buff := new(bytes.Buffer)
 	err := binary.Write(buff, binary.BigEndian, data)
 	if err != nil {
-		log.Panic(err)
+		GetLogger().Crit("Failed to convert fixed data into byte array", "err", err)
+		panic(err)
```
LeoHChen: don't think we need to panic here.  @mikedoan

internal/utils/utils.go
```diff
@@ -73,7 +73,8 @@ func ConvertFixedDataIntoByteArray(data interface{}) []byte {
 func GetUniqueIDFromIPPort(ip, port string) uint32 {
 	reg, err := regexp.Compile("[^0-9]+")
 	if err != nil {
-		log.Panic("Regex Compilation Failed", "err", err)
+		GetLogger().Crit("Regex Compilation Failed", "err", err)
+		panic(err)
```
LeoHChen: same thing, no need to panic here.

internal/utils/utils.go
```diff
@@ -199,15 +200,16 @@ func LoadKeyFromFile(keyfile string) (key p2p_crypto.PrivKey, pk p2p_crypto.PubK
...
 		key, pk, err = GenKeyP2PRand()
 		if err != nil {
-			log.Panic("LoadKeyFromFile", "GenKeyP2PRand Error", err)
+			GetLogger().Crit("LoadKeyFromFile", "GenKeyP2PRand Error", err)
+			panic(err)
 		}
 		err = SaveKeyToFile(keyfile, key)
 		if err != nil {
-			log.Print("LoadKeyFromFile", "failed to save key to keyfile", err)
+			GetLogger().Info("failed to save key to keyfile", "keyfile", err)
```
LeoHChen: this is error

api/proto/node/node.go
hsienchiaolee: [GoDoc](https://golang.org/pkg/log/#Panic)
> Panic is equivalent to Print() followed by a call to panic().

internal/utils/utils.go
hsienchiaolee: If we are certain that this regex is always going to compile, can we skip the error check completely?

internal/utils/utils.go
hsienchiaolee: Not sure about the context here, but what's the damage if it fails to write?

node/contract.go
```diff
@@ -148,7 +147,7 @@ func (node *Node) GetNonceOfAddress(address common.Address) uint64 {
 func (node *Node) GetBalanceOfAddress(address common.Address) (*big.Int, error) {
 	state, err := node.Blockchain().State()
 	if err != nil {
-		log.Error("Failed to get chain state", "Error", err)
+		utils.GetLogger().Error("Failed to get chain state", "Error", err)
```
mikedoan: GetLogger is an expensive call - we don't have to it extensively

mikedoan: @harmony-ek it was EK opinion. Please talk with him.

api/proto/node/node.go
LeoHChen: @mikedoan , do we need to panic here? please comment.

internal/utils/utils.go
LeoHChen: agree, not need to check the error here with the simple regex.

api/proto/node/node.go
rlan35: Don't panic here please. We should avoid panic (exit the software) as much as possible

core/blockchain.go
```diff
@@ -1123,7 +1122,7 @@ func (bc *BlockChain) insertChain(chain types.Blocks) (int, []interface{}, []*ty
 	for i, block := range chain {
 		// If the chain is terminating, stop processing blocks
 		if atomic.LoadInt32(&bc.procInterrupt) == 1 {
-			log.Debug("Premature abort during blocks processing")
+			utils.GetLogger().Debug("Premature abort during blocks processing")
```
rlan35: These change on ethereum's log code will potentially add a lot more logs for us. Can you test the logs before and after this change and see how much difference in the size?

hsienchiaolee: Ran both versions for 10 minutes, barely any increase:
• Before: ~2.0M
• After: ~2.1M
Seems like most of these are not called or are `Error` log types (which shouldn't happen often).

PR #1176, opened 2019-06-27
retrieve leaderKey/viewID information before join consensus

chaosma:


consensus/consensus_service.go
```diff
@@ -603,3 +603,51 @@ func (consensus *Consensus) getLogger() log.Logger {
...
+func (consensus *Consensus) getLeaderPubKeyFromCoinbase(header *types.Header) (*bls.PublicKey, error) {
+	shardState, err := consensus.ChainReader.ReadShardState(header.Epoch)
+	if err != nil {
+		return nil, ctxerror.New("cannot read shard state",
+			"epoch", header.Epoch,
+			"coinbaseAddr", header.Coinbase,
+		).WithCause(err)
+	}
+
+	committee := shardState.FindCommitteeByID(header.ShardID)
```
rlan35: what if it can not find the leader here. Will it still joining consensus eventually?

chaosma: If cannot find  leader, fall back to our current way to join syncing, i.e. ignoreViewIDCheck=true and update information after receives the first consensus message

PR #1177, opened 2019-06-27
add blocks for explorer correctly and clean them up to avoid OOM

mikedoan:
add blocks for explorer correctly and clean them up to avoid OOM


LeoHChen:
Please test it with explorer.

PR #1178, opened 2019-06-27
[bootnode] add two more bootnodes for r3 release

LeoHChen:


LeoHChen:
I've updated https://harmony.one/node.sh to point to release folder, which is https://s3-us-west-1.amazonaws.com/pub.harmony.one/release/linux-x86_64/r3/node.sh

So, the node.sh script will be released together with the harmony binary, which is better than pointing to the head of the tree version.

PR #1203, opened 2019-07-05
read zerolog instead to calculation CONSENSUS

mikedoan:
read zerolog to calculate consensus as we are moving log to zerolog


mikedoan:
Jenkins, test this please.

hsienchiaolee:
Note: We should remove the zerolog prefix once we migrate everything over. Just have to make sure we remove them here as well.

mikedoan:
Agree Kai! I'll add a TODO

PR #1204, opened 2019-07-05
Add stat.sh into scripts folder

xChickens:
Adds a script without any conflict that allows users to check some statistics on their foundational nodes.

Basic functionality of the script:
1. Checks validation status by checking time the script is executed and compares it with the timestamp of the last bingo and sees whether or not it's within 60 seconds.
2. Prints out the balances in whichever shards the user has balance in. 
3. If run for the first time within 24 hours or if the node was reset, it saves the last bingo timestamp and last balance to the files "prevtime" and "prevbal". It also prints the difference in current balance and saved balance.
4. Calculates the estimated balance in the next 24 hours based on latest bingo timestamp and balance compared to the stored time and balance
5. Prints the timestamp of the latest bingo and the interval between the last two bingos.


harmony-ek:

scripts/stat.sh
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env bash
+
+# stat.sh - Check foundational node status/statistics (requires 'bc' installed)
+
+# Get timestamps
+lastbingo=`grep -i bingo latest/*.log |tail -n 1 |cut -f 2 -d 'r' |cut -c 16-34 | tr T \ `
```
LeoHChen: use $() to replace all the `` to make it more uniform.

scripts/stat.sh
```diff
@@ -0,0 +1,140 @@
...
+# stat.sh - Check foundational node status/statistics (requires 'bc' installed)
+
+# Get timestamps
+lastbingo=`grep -i bingo latest/*.log |tail -n 1 |cut -f 2 -d 'r' |cut -c 16-34 | tr T \ `
+time1=$(date -d "$lastbingo" +%s)
+rawtime=`date +%s`
+time2=$(($rawtime - 60))
+created=`grep -i -m 1 allocated latest/*.log |cut -f 7-9 -d ":" |cut -f 1 -d "." |cut -c 2- | tr T \ `
+latest=`grep -i bingo latest/*.log |tail -n 1 |cut -f 2 -d 'T' |cut -c 1-14`
+bingos=`grep -i bingo latest/*.log |tail -n 2 |cut -f 2 -d 'T' |cut -c 7-14`
```
LeoHChen: duplicated grep query, why don't you just use a temporary file to store all the 'grep -i bingo' ?

scripts/stat.sh
```diff
@@ -0,0 +1,140 @@
...
+if [[ $timediff -lt 7 ]]
+  then
+    sleep $((7 - timediff))
+    balances=`./wallet.sh balances`
+  else
+    balances=`./wallet.sh balances`
+fi
+
+# Sum balances
+bal0=`echo "$balances" |grep -i "shard 0" |cut -f 2 -d ":" |cut -f1 -d "," | rev | cut -c 5- | rev`
```
LeoHChen: duplicated code, use an array to simplify the code.

scripts/stat.sh
```diff
@@ -0,0 +1,140 @@
...
+reward=`echo "$curbal - $prevbal" | bc -l`
+if [[ $passed != 0 ]]
+  then
+    daily=`echo "scale=5; $reward * 86400 / $passed" | bc -l`
+  else
+    daily=0
+fi
+
+# Show results
+echo "$balances" |grep -i "address"
```
LeoHChen: duplicated code, use an array, and a function to simplify the duplicated code.

scripts/stat.sh
```diff
@@ -0,0 +1,140 @@
...
+echo $daily
+
+# Find and show the time of the latest bingo
+echo -e '    Latest Bingo      :  \c' && echo $latest
+
+# Fetch the time/seconds of the last two bingos
+bingo1=`echo $bingos |cut -c 1-8`
+bingo2=`echo $bingos |cut -c 10-17`
+
+# Delete leading zero
```
LeoHChen: why is it needed? any example on why?

xChickens:
Needs another review

scripts/stat.sh
```diff
@@ -0,0 +1,122 @@
...
+bingo2=$(echo $bingos |cut -c 10-17)
+
+# Ensure bingo2 is greater than bingo 1
+if (( $(echo "$bingo2 <= $bingo1" |bc -l) ))
+    then
+        bingo2=$(echo "$bingo2 + 60" | bc -l)
+fi
+
+# Calculate and show the time difference
+difference=$(echo $bingo2-$bingo1 | bc -l)
```
LeoHChen: why don't you just use bash math?  Is the difference a floating number?

scripts/stat.sh
```diff
@@ -0,0 +1,122 @@
...
+
+# Find and show the time of the latest bingo
+echo "    Latest Bingo      :  $latest"
+
+# Fetch the time/seconds of the last two bingos
+bingo1=$(echo $bingos |cut -c 1-8)
+bingo2=$(echo $bingos |cut -c 10-17)
+
+# Ensure bingo2 is greater than bingo 1
+if (( $(echo "$bingo2 <= $bingo1" |bc -l) ))
```
LeoHChen: why don't you compare the two number directly in bash?

scripts/stat.sh
xChickens: Yes, bash doesn't have the ability to do floating point calculations.

scripts/stat.sh
xChickens: Bash also can't compare two floating point numbers without use of bc.

xChickens:
> looks good in general. maybe you could put this script into harmony-one/harmony-ops repo.

Would it be preferably to just be in the harmony-ops repo vs the main repo at all?

LeoHChen:
Thanks for the comment.  Please put the script into harmony-ops repo then.  No other comments. @xChickens

PR #1208, opened 2019-07-09
mint demo accounts in testnet

mikedoan:
Mint demo accounts in testnet. The demo accounts will be used for SDK testing or other testing apps.


node/demo_constants.go
```diff
@@ -0,0 +1,27 @@
+package node
```
rlan35: Can you move these test code into /test?

LeoHChen:
I don't know why do we need demo accounts.  With faucet, can you get token for any account?

PR #1215, opened 2019-07-12
disable toShardID as it was not used yet

mikedoan:
disable toShardID as it was not used yet

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -46,14 +46,16 @@ type Transaction struct {
 }
 
 type txdata struct {
-	AccountNonce uint64          `json:"nonce"    gencodec:"required"`
-	Price        *big.Int        `json:"gasPrice" gencodec:"required"`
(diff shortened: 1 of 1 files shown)


LeoHChen:
why can't we add it in SDK?  I don't think we should keep adding testing and demo code in the code base.  We are building a production system.

mikedoan:
Search ToShardID in our code. it's not used anywhere. I guess it will be used for cross shard transaction.

neeboo:
Please make decision on that. If already added, dont remove then.

LeoHChen:
we won't remove it.  For now, this parameter is not used yet.

neeboo:
> we won't remove it. For now, this parameter is not used yet.
> […](#)

Currently I have added it to latest sdk version. It is a breaking change for Transaction rlp signing. So please be careful adding or removing some field here.

PR #1221, opened 2019-07-16
add hmy_call

lzl124631x:
Add `hmy_call` support

image

---

Also verified it supports input parameter.

image

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -3,9 +3,11 @@ package hmy
 import (
 	"context"
 	"errors"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
+	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/ethereum/go-ethereum/core/bloombits"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/event"
@@ -16,6 +18,7 @@ import (
 	"github.com/harmony-one/harmony/core"
 	"github.com/harmony-one/harmony/core/state"
 	"github.com/harmony-one/harmony/core/types"
+	"github.com/harmony-one/harmony/core/vm"
 )
 
 // APIBackend An implementation of internal/hmyapi/Backend. Full client.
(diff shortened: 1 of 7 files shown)


hmy/api_backend.go
```diff
@@ -206,3 +209,17 @@ func (b *APIBackend) GetBalance(address common.Address) (*hexutil.Big, error) {
 func (b *APIBackend) NetVersion() uint64 {
 	return b.hmy.NetVersion()
 }
+
+// GetEVM returns a new EVM entity
+func (b *APIBackend) GetEVM(ctx context.Context, msg core.Message, state *state.DB, header *types.Header) (*vm.EVM, func() error, error) {
+	state.SetBalance(msg.From(), math.MaxBig256)
```
mikedoan: is it from ethereum code ? why do we need to set balance in state for From ?

internal/hmyapi/blockchain.go
```diff
@@ -81,3 +95,95 @@ func (s *PublicBlockChainAPI) BlockNumber() hexutil.Uint64 {
...
+	}(time.Now())
+
+	state, header, err := b.StateAndHeaderByNumber(ctx, blockNr)
+	if state == nil || err != nil {
+		return nil, 0, false, err
+	}
+	// Set sender address or use a default if none specified
+	var addr common.Address
+	if args.From == nil {
+		if wallets := b.AccountManager().Wallets(); len(wallets) > 0 {
```
mikedoan: if from == nil why do we get the first one from default wallet?

hmy/api_backend.go
lzl124631x: Yes. I'm not sure though. https://github.com/ethereum/go-ethereum/blob/40cdcf8c47ff094775aca08fd5d94051f9cf1dbb/les/api_backend.go#L114.

Posted a [question](https://ethereum.stackexchange.com/q/72977/54923)

internal/hmyapi/blockchain.go
```diff
@@ -81,3 +95,95 @@ func (s *PublicBlockChainAPI) BlockNumber() hexutil.Uint64 {
...
+
+	// Get a new instance of the EVM.
+	evm, vmError, err := b.GetEVM(ctx, msg, state, header)
+	if err != nil {
+		return nil, 0, false, err
+	}
+	// Wait for the context to be done and cancel the evm. Even if the
+	// EVM has finished, cancelling may be done (repeatedly)
+	go func() {
+		<-ctx.Done()
```
mikedoan: is it from ethereum code ? do you know when ctx.Done gets a signal? if it never gets signal it would be stuck forever.

lzl124631x: According to [context/Done](https://golang.org/src/context/context.go?s=2437:5895#L52), it depends. I'm using `WithTimeOut` in `hmyclient/main.go` so `Done` will gets the signal on timeout.

internal/hmyapi/blockchain.go
lzl124631x: Posted [question](https://ethereum.stackexchange.com/questions/72979/why-does-the-docall-function-use-the-first-account-by-default)

hmy/api_backend.go
lzl124631x: The code is borrowed from [go-ethereum](https://github.com/ethereum/go-ethereum/blob/40cdcf8c47ff094775aca08fd5d94051f9cf1dbb/les/api_backend.go#L114)

internal/hmyapi/blockchain.go
lzl124631x: The code is borrowed from [go-etherum](https://github.com/ethereum/go-ethereum/blob/f578d41ee6b3087f8021fd561a0b5665aea3dba6/internal/ethapi/api.go#L738)

PR #1229, opened 2019-07-20
VRF integration step 1 

coolcottontail:
Design doc: https://docs.google.com/document/d/1JCcA7pM84q_wI1QAeRjbBXbkTjwcU6In9xHZ15NJJlo/edit

Summary

1)VRF is issued by the current beacon shard leader when a new block is announced.
2) The leader gets the hash of previous block (block( newBlock.number() - 1). hash() ) , use it to generate VRF and Proof through BLS_VRF algorithm. The generated VRF/Proof is put to block header fields: Vrf[32], VrfProof[96]
3)The new block is announced and all validators check the new Vrf and VrfProof in the onAnnounce(...) phase. Validate that Vrf/VrfProof is correct using the consensus leader's public BLS key
4)The new VRF is added to a VRF list for the current consensus.
5)All nodes wait for leader change
6)Leader changed
7)The new leader puts a new VRF to the block chain same as in 1)
8)The new leader checks the current VRF list, if the total number of VRFs >= 2/3 shard nodes, it xor all VRF and start VDF process. If not, go to 4)

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -90,6 +90,10 @@ type Header struct {
 	LastCommitSignature [96]byte    `json:"lastCommitSignature"  gencodec:"required"`
 	LastCommitBitmap    []byte      `json:"lastCommitBitmap"     gencodec:"required"` // Contains which validator signed
 	ShardStateHash      common.Hash `json:"shardStateRoot"`
+	Vrf                 [32]byte    `json:"vrf"`
+	VrfProof            [96]byte    `json:"vrfProof"`
+	Vdf                 [258]byte   `json:"vdf"`
+	VdfProof            [258]byte   `json:"vdfProof"`
 	ShardState          []byte      `json:"shardState"`
 }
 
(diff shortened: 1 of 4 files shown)


core/types/block.go
```diff
@@ -90,6 +90,10 @@ type Header struct {
 	LastCommitSignature [96]byte    `json:"lastCommitSignature"  gencodec:"required"`
 	LastCommitBitmap    []byte      `json:"lastCommitBitmap"     gencodec:"required"` // Contains which validator signed
 	ShardStateHash      common.Hash `json:"shardStateRoot"`
+	Vrf                 [32]byte    `json:"vrf"`
```
LeoHChen: will this cause compatibility issue with the mainnet as the block structure changed?  how to make sure it is backward compatible?

coolcottontail:
replaced with a new PR
https://github.com/harmony-one/harmony/pull/1257

PR #1230, opened 2019-07-22
Configurable blocks per epoch

LeoHChen:


internal/configs/sharding/fixedschedule.go
```diff
@@ -14,6 +14,10 @@ func (s fixedSchedule) InstanceForEpoch(epoch *big.Int) Instance {
 	return s.instance
 }
 
+func (s fixedSchedule) BlocksPerEpoch() uint64 {
+	return 1000000000000
```
chaosma: Put this as constant is better to search and modify in future

PR #1236, opened 2019-07-24
fix network restart issue for mainnet shard0 down

chaosma:
local test passed

### DIFF
--- a/node/node_newblock.go
+++ b/node/node_newblock.go
@@ -7,6 +7,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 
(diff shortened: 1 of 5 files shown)


node/node_newblock.go
```diff
@@ -59,6 +63,16 @@ func (node *Node) WaitForConsensusReadyv2(readySignal chan struct{}, stopChan ch
 						continue
 					}
 
+					//TODO: remove it after shard0 fix
+					num := node.Blockchain().CurrentBlock().NumberU64()
+					utils.GetLogInstance().Info("hehe", "num", num)
```
rlan35: remove test log.

chaosma: removed

PR #1238, opened 2019-07-24
[consensus] recovery block when shard is stuck

LeoHChen:

### DIFF
--- a/consensus/config.go
+++ b/consensus/config.go
@@ -15,8 +15,10 @@ const (
 	maxLogSize        uint32        = 1000
 	// threshold between received consensus message blockNum and my blockNum
 	consensusBlockNumBuffer uint64 = 2
-	// ReProposeBlockNum: leader propose this block from it's database and validators accept them
-	ReProposeBlockNum uint64 = 270108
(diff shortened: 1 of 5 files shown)


consensus/config.go
```diff
@@ -15,8 +15,10 @@ const (
 	maxLogSize        uint32        = 1000
 	// threshold between received consensus message blockNum and my blockNum
 	consensusBlockNumBuffer uint64 = 2
-	// ReProposeBlockNum: leader propose this block from it's database and validators accept them
-	ReProposeBlockNum uint64 = 270108
+	// ReProposeBlockNumShard0: leader propose this block from it's database and validators accept them
+	ReProposeBlockNumShard0 uint64 = 270108
+	// ReProposeBlockNumShard2: leader propose this block from it's database and validators accept them
+	ReProposeBlockNumShard2 uint64 = 275058
```
chaosma: MsgBlock is the proposing block, then repropose block is 275057

LeoHChen: changed.

rlan35: Yes, we should repropose the previous block of the stuck block

PR #1243, opened 2019-07-24
Add Pagerduty validation status script

xChickens:
Script checks timestamp of last bingo in log and triggers an alert to Pagerduty if more than 60 seconds have passed since. The script should be used with crontab to be run every minute in order to constantly check for validation status.

NOTE: If a Pagerduty has been triggered and node somehow starts validating again, the script will automatically send a resolve event to Pagerduty. This means that if the alert on Pagerduty is manually resolved (not simply acknowledged), the script will hang because it will try to post a resolve for an alert that doesn't exist. Not 100% sure if this functionality is wanted.


harmony-ek:

xChickens:
wrong branch

PR #1244, opened 2019-07-24
Add pagerduty validation status script

xChickens:
Script checks timestamp of last bingo in log and triggers an alert to Pagerduty if more than 60 seconds have passed since. The script should be used with crontab to be run every minute in order to constantly check for validation status.

> NOTE: If a Pagerduty has been triggered and node somehow starts validating again, the script will automatically send a resolve event to Pagerduty. This means that if the alert on Pagerduty is manually resolved (not simply acknowledged), the script will hang because it will try to post a resolve for an alert that doesn't exist. Not 100% sure if this functionality is wanted.

EDIT: Fixed

### DIFF
--- a/scripts/pagerduty.sh
+++ b/scripts/pagerduty.sh
@@ -36,8 +36,8 @@ delay=60
 ### Pagerduty
 if [[ $(($curtime - $latest)) -gt $delay ]]; then
     # Trigger an alert
-    trigger=$(curl -X POST -H "$HEADER" -d "$TRIGGER" "$POSTURL")
+    trigger=$(curl -s -X POST -H "$HEADER" -d "$TRIGGER" "$POSTURL")
 else
     # Resolve an alert
-    resolve=$(curl -X POST -H "$HEADER" -d "$RESOLVE" "$POSTURL")
+    resolve=$(curl -s -X POST -H "$HEADER" -d "$RESOLVE" "$POSTURL")
 fi
--- a/scripts/pagerduty.sh
+++ b/scripts/pagerduty.sh
@@ -1,14 +1,14 @@
 #!/usr/bin/env bash
 
-# Pagerduty constants
+### Log variables
 LOGFILE="/home/tmp_log/log-20190628.153354/*.log"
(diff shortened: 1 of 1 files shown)


harmony-ek:

harmony-ek:
Jenkins, test this please…?  Pretty pwease?  (⊙.⊙)//~

scripts/pagerduty.sh
```diff
@@ -0,0 +1,57 @@
...
+RESOLVE="\
+{\
+  \"routing_key\": \"$KEY\",\
+  \"event_action\": \"resolve\",\
+  \"dedup_key\": \"$IP\"\
+}"
+JSON_HEADER="Content-Type: application/json"
+POSTURL="https://events.pagerduty.com/v2/enqueue"
+
+# Validation check constants
```
harmony-ek: Can you make this script not hang even if the incident has been manually resolved?

scripts/pagerduty.sh
```diff
@@ -0,0 +1,57 @@
...
+  \"dedup_key\": \"$IP\",\
+  \"payload\": {\
+    \"summary\": \"Node NOT validating: Shard-$SHARD_NUM: $IP\",\
+    \"source\": \"$IP\",\
+    \"severity\": \"critical\"\
+  }\
+}"
+RESOLVE="\
+{\
+  \"routing_key\": \"$KEY\",\
```
harmony-ek: (Nit) auth materials such as this usually belongs to the HTTP request header in REST APIs.  Consider passing this as a header (`curl -H "X-Routing-Key: $KEY" ...`).

scripts/pagerduty.sh
xChickens: I found out it doesn't actually stall, seems like something was wrong with the code before and I fixed it in one of the prior commits.

harmony-ek:
Jenkins… test… this… please… \*plop*

LeoHChen:
@harmony-ek , can you deploy this script to a few of our nodes?

PR #1245, opened 2019-07-25
Add leader consensus retry logic

rlan35:
Add logic to resent each consensus message from the leader if the next phase is not reached in retryInterval time. This is different from the last consensus retry logic which is on the whole consensus round level (which has many flaws). This one is one the message level and is more robust.


consensus/consensus.go
```diff
@@ -233,6 +235,7 @@ type StakeInfoFinder interface {
 func New(host p2p.Host, ShardID uint32, leader p2p.Peer, blsPriKey *bls.SecretKey) (*Consensus, error) {
 	consensus := Consensus{}
 	consensus.host = host
```
chaosma: If not used, we can remove consensus.host

consensus/consensus_msg_sender.go
```diff
@@ -0,0 +1,138 @@
...
+// Reset resets the sender's state for new block
+func (sender *MessageSender) Reset(blockNum uint64) {
+	sender.blockNumMutex.Lock()
+	sender.blockNum = blockNum
+	sender.blockNumMutex.Unlock()
+	sender.StopAllRetriesExceptCommitted()
+	sender.messagesToRetry.Range(func(key interface{}, value interface{}) bool {
+		if msgRetry, ok := value.(*MessageRetry); ok {
+			if msgRetry.msgType != msg_pb.MessageType_COMMITTED {
+				sender.messagesToRetry.Delete(key)
```
chaosma: when do we delete the committed message?

PR #1248, opened 2019-07-26
Initial shard setup using latest epoch information

chaosma:
initial setup of shard information use latest epoch info to replace genesis epoch info

cleanup unnecessary configs when init a node

add more logs in state syncing


api/service/syncing/downloader/client.go
```diff
@@ -41,10 +41,12 @@ func (client *Client) Close() {
 }
 
 // GetBlockHashes gets block hashes from all the peers by calling grpc request.
-func (client *Client) GetBlockHashes(startHash []byte, size uint32) *pb.DownloaderResponse {
+func (client *Client) GetBlockHashes(startHash []byte, size uint32, ip, port string) *pb.DownloaderResponse {
```
LeoHChen: why use ip and port in the function here?  It is weird. You'd wrap it up in the 'request' or other object.

api/service/syncing/downloader/server.go
```diff
@@ -24,7 +25,14 @@ type Server struct {
 
 // Query returns the feature at the given point.
 func (s *Server) Query(ctx context.Context, request *pb.DownloaderRequest) (*pb.DownloaderResponse, error) {
-	response, err := s.downloadInterface.CalculateResponse(request)
+	var pinfo string
```
LeoHChen: better add a comment on why you are doing this.

api/service/syncing/syncing.go
```diff
@@ -334,8 +334,12 @@ func (ss *StateSync) GetConsensusHashes(startHash []byte, size uint32) bool {
 			wg.Add(1)
 			go func() {
 				defer wg.Done()
-				response := peerConfig.client.GetBlockHashes(startHash, size)
+				response := peerConfig.client.GetBlockHashes(startHash, size, ss.selfip, ss.selfport)
```
LeoHChen: selfip, and selfport, why?

cmd/harmony/main.go
```diff
@@ -272,22 +263,22 @@ func createGlobalConfig(isLeader bool) *nodeconfig.ConfigType {
 		panic("unable to new host in harmony")
 	}
 
-	if err := nodeConfig.Host.AddPeer(&nodeConfig.Leader); err != nil {
-		ctxerror.Warn(utils.GetLogger(), err, "(*p2p.Host).AddPeer failed",
-			"peer", &nodeConfig.Leader)
-	}
+	//	if err := nodeConfig.Host.AddPeer(&nodeConfig.Leader); err != nil {
```
LeoHChen: remove the code if they are not needed.

node/node.go
```diff
@@ -391,46 +388,37 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, chainDBFactory shardc
...
 
 	// Get genesis epoch shard state from chain
-	genesisEpoch := big.NewInt(core.GenesisEpoch)
-	shardState, err := node.Beaconchain().GetShardState(genesisEpoch, nil)
-	if err != nil {
-		return ctxerror.New("cannot read genesis shard state").WithCause(err)
+	blockNum := node.Blockchain().CurrentBlock().NumberU64()
+	node.Consensus.SetMode(consensus.Listening)
+	epoch := core.ShardingSchedule.CalcEpochNumber(blockNum)
+	getLogger().Info("[InitShardState] Try To Get PublicKeys from database", "blockNum", blockNum, "shardID", shardID)
```
LeoHChen: can you use zerolog?

test/configs/local-resharding.txt
```diff
@@ -13,13 +13,13 @@
...
-127.0.0.1 9104 newnode one1z05g55zamqzfw9qs432n33gycdmyvs38xjemyl 95117937cd8c09acd2dfae847d74041a67834ea88662a7cbed1e170350bc329e53db151e5a0ef3e712e35287ae954818
-127.0.0.1 9105 newnode one1ljznytjyn269azvszjlcqvpcj6hjm822yrcp2e 68ae289d73332872ec8d04ac256ca0f5453c88ad392730c5741b6055bc3ec3d086ab03637713a29f459177aaa8340615
-127.0.0.1 9107 newnode one1uyshu2jgv8w465yc8kkny36thlt2wvel89tcmg 1c1fb28d2de96e82c3d9b4917eb54412517e2763112a3164862a6ed627ac62e87ce274bb4ea36e6a61fb66a15c263a06
-127.0.0.1 9108 newnode one103q7qe5t2505lypvltkqtddaef5tzfxwsse4z7 b179c4fdc0bee7bd0b6698b792837dd13404d3f985b59d4a9b1cd0641a76651e271518b61abbb6fbebd4acf963358604
-127.0.0.1 9109 newnode one1658znfwf40epvy7e46cqrmzyy54h4n0qa73nep 576d3c48294e00d6be4a22b07b66a870ddee03052fe48a5abbd180222e5d5a1f8946a78d55b025de21635fd743bbad90
-127.0.0.1 9110 newnode one1d2rngmem4x2c6zxsjjz29dlah0jzkr0k2n88wc 16513c487a6bb76f37219f3c2927a4f281f9dd3fd6ed2e3a64e500de6545cf391dd973cc228d24f9bd01efe94912e714
+127.0.0.1 9100 validator one1ghkz3frhske7emk79p7v2afmj4a5t0kmjyt4s5 eca09c1808b729ca56f1b5a6a287c6e1c3ae09e29ccf7efa35453471fcab07d9f73cee249e2b91f5ee44eb9618be3904
```
LeoHChen: how do you test newnode join if you removed newnode?

node/node.go
LeoHChen: maybe print out epoch as well.

node/node.go
```diff
@@ -391,46 +388,37 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, chainDBFactory shardc
...
-		pubKey := &bls.PublicKey{}
-		pubKeyBytes := node.BlsPublicKey[:]
-		err = pubKey.Deserialize(pubKeyBytes)
-		if err != nil {
-			return ctxerror.New("cannot deserialize BLS public key",
-				"shardID", shardID,
-				"pubKeyBytes", hex.EncodeToString(pubKeyBytes),
-			).WithCause(err)
+	for _, key := range pubKeys {
+		if key.IsEqual(node.Consensus.PubKey) {
```
LeoHChen: don't we have a function to check isLeader?

chaosma: This is to check if your key is in the record, not check whether it's leader or not

test/configs/local-resharding.txt
chaosma: a node is a new node if it's not in the current committee, the newnode role is rather very arbitrary and don't have much difference from validator roles.

cmd/harmony/main.go
chaosma: removed

api/service/syncing/syncing.go
chaosma: This is to add ip and port (port is used for local debug) in the request, so that the server side know who is requesting. It is used for debugging purpose.

api/service/syncing/downloader/server.go
chaosma: this is for debugging purpose. it's kind of overlap with selfip and selfport in your comment below since it will not contain the correct port number in local test

api/service/syncing/downloader/client.go
chaosma: the request is created inside this function, so I have to pass it as parameter. Let me know if you have better suggestion.

test/configs/local-resharding.txt
chaosma: First, every new joined node will do state syncing. Second, it will join consensus only if it finds it is in the committee.

node/node.go
chaosma: change to use zerolog

api/service/syncing/downloader/client.go
```diff
@@ -41,10 +41,12 @@ func (client *Client) Close() {
...
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	request := &pb.DownloaderRequest{Type: pb.DownloaderRequest_HEADER, BlockHash: startHash, Size: size}
+	request.Ip = ip
+	request.Port = port
```
LeoHChen: where are those two fields being used?  I didn't see the IP/Port being used anyway?

chaosma: It's used in the CalculateResponse function in node/node_syncing.go

api/service/syncing/downloader/server.go
chaosma: comment added

PR #1257, opened 2019-07-27
DRG generation (VRF/VDF integration)

coolcottontail:
## DRG generation
Spec: https://docs.google.com/document/d/1JCcA7pM84q_wI1QAeRjbBXbkTjwcU6In9xHZ15NJJlo/edit

Use db cache to keep list of VRF block numbers in the current epoch and VDF block number in the current epoch.  

Leaders update db cache when VRF/VDF is generated 
Validators update db cache during conensus onAnnounce 

VDF generation won't across epochs (vdf will finish within current epoch)

## Test
1. tested basic function flow with  using 1,2,3, 4 leader changes per epoch (VRF generation, VRF validation, VDF generation, VDF validation)  by leaders and validators
2. tested with different shard config, different size of consensus, different VDF delay
3. tested various wrong VDF/VRF and make sure it won't cause crash/panic
4. manually kill nodes to trigger view change, 

## TODO
1. integrate with real leader change code
2. TEST WITH REAL LEADER CHANGE!!!


consensus/consensus_v2.go
```diff
@@ -15,10 +15,16 @@ import (
...
+	vrf_bls "github.com/harmony-one/harmony/crypto/vrf/bls"
 	"github.com/harmony-one/harmony/internal/ctxerror"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/p2p"
 	"github.com/harmony-one/harmony/p2p/host"
+	"github.com/harmony-one/vdf/src/vdf_go"
+)
+
+const (
+	vdfDifficulty = 5000 // This takes about 20s to finish the vdf
```
rlan35: Can make it more than 2 minutes will be better. The current view change timeout for leader is 1 minutes and the leader stays as leader for about 3 minutes

coolcottontail: ok. this 20s is only for localnet testing purpose. I will update to real delay based on aws instance calibration.

consensus/consensus_v2.go
```diff
@@ -205,6 +211,102 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
 				Msg("[OnAnnounce] Block content is not verified successfully")
 			return
 		}
+
+		if consensus.ShardID == 0 {
+			//TODO: HB fix this
+			if recvMsg.BlockNum%core.ShardingSchedule.BlocksPerEpoch() == 0 {
```
rlan35: This condition seems not right.

consensus/consensus_v2.go
```diff
@@ -205,6 +211,102 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
+				}
+
+				if hash != headerObj.Vrf {
+					consensus.getLogger().Warn().
+						Str("MsgBlockNum", headerObj.Number.String()).
+						Msg("[OnAnnounce] VRF proof is not valid")
+					return
+				}
+
+				consensus.vrfBlockNumbers, _ = consensus.ChainReader.ReadEpochVrfBlockNums(headerObj.Epoch)
```
rlan35: If we always read from disk. We can make "consensus.vrfBlockNumbers" a local variable.

consensus/consensus_v2.go
```diff
@@ -205,6 +211,102 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
+						Msg("[OnAnnounce] failed to write VRF block number to local db")
+				}
+
+				consensus.getLogger().Info().
+					Str("MsgBlockNum", headerObj.Number.String()).
+					Int("Number of VRF in this epoch", len(consensus.vrfBlockNumbers)).
+					Msg("[OnAnnounce] validated a new VRF")
+			}
+
+			zeroVdfBytes := make([]byte, 258)
```
rlan35: I thought vdf is 516 bytes?

consensus/consensus_v2.go
```diff
@@ -205,6 +211,102 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
 				Msg("[OnAnnounce] Block content is not verified successfully")
 			return
 		}
+
+		if consensus.ShardID == 0 {
+			//TODO: HB fix this
+			if recvMsg.BlockNum%core.ShardingSchedule.BlocksPerEpoch() == 0 {
+				consensus.generateNewVrf = true
```
rlan35: The validators don't need to have these flags (generateNewVdf/Vrf) because they don't generate them, they only verify them.

consensus/consensus_v2.go
coolcottontail: 258 is VDF and another 258 is proof for the VDF.

consensus/consensus_v2.go
coolcottontail: My initial implementation is making it a local variable but the values will lost in case  the node can is killed and later restarted.   So I used local cache/db for these values.  The values also need to be updated during node sync.

consensus/consensus_v2.go
coolcottontail: ok!

LeoHChen:
The commit message history is too messy.  Please re-organize them, using "git rebase -i"
[git rebase -i]( https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

rlan35:
Generally randomness logic looks good to me. Have one last comment regarding block header field.

LeoHChen:
also, please don't push to branch in master repo.  Please use your fork to push the PR.
https://github.com/harmony-one/harmony-open/tree/master/git-cheat-sheet

consensus/consensus_v2.go
coolcottontail: removed these flags.

consensus/consensus_v2.go
coolcottontail: use local variables and db/cache for them

consensus/consensus_v2.go
coolcottontail: used different vdfDifficulty for testnet/mainnet/localnet

internal/configs/sharding/mainnet.go
```diff
@@ -10,6 +10,8 @@ const (
 	mainnetEpochBlock1 = 327680 // 20 * 2^14
 	blocksPerShard     = 16384  // 2^14
 	mainnetV1Epoch     = 1
+
+	mainnetVdfDifficulty = 100000 // This takes about 200s to finish the vdf
```
rlan35: In the future our leader rotate every 40 blocks which is around 200s with 5 second block time. So better to make the vdf difficulty around 100s or 2 minutes

consensus/consensus.go
```diff
@@ -270,6 +274,9 @@ func New(host p2p.Host, ShardID uint32, leader p2p.Peer, blsPriKey *bls.SecretKe
 
 	consensus.ReadySignal = make(chan struct{})
 
+	// channel for receiving newly generated VDF
+	consensus.RndChannel = make(chan [548]byte)
```
LeoHChen: use a constant for 548?

consensus/consensus_service.go
```diff
@@ -39,20 +39,21 @@ func (consensus *Consensus) WaitForNewRandomness() {
 }
 
 // GetNextRnd returns the oldest available randomness along with the hash of the block there randomness preimage is committed.
-func (consensus *Consensus) GetNextRnd() ([32]byte, [32]byte, error) {
+func (consensus *Consensus) GetNextRnd() ([516]byte, [32]byte, error) {
 	if len(consensus.pendingRnds) == 0 {
-		return [32]byte{}, [32]byte{}, errors.New("No available randomness")
+		return [516]byte{}, [32]byte{}, errors.New("No available randomness")
```
LeoHChen: use constant to replace 516

core/types/block.go
```diff
@@ -90,6 +90,10 @@ type Header struct {
 	LastCommitSignature [96]byte    `json:"lastCommitSignature"  gencodec:"required"`
 	LastCommitBitmap    []byte      `json:"lastCommitBitmap"     gencodec:"required"` // Contains which validator signed
 	ShardStateHash      common.Hash `json:"shardStateRoot"`
+	Vrf                 [32]byte    `json:"vrf"`
```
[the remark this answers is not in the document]
rlan35: We should test whether the addition of empty new []bytes change the hash of the block. If not, I think we may not need extra versioning. Otherwise, we need to use different version of the struct for the hard fork.

consensus/consensus_v2.go
```diff
@@ -1073,33 +1094,79 @@ func (consensus *Consensus) Start(blockChannel chan *types.Block, stopChan chan
...
-					//}
-
-					rnd, blockHash, err := consensus.GetNextRnd()
+					// generate VRF if the current block has a new leader
+					if !consensus.ChainReader.IsSameLeaderAsPreviousBlock(newBlock) {
+						vrfBlockNumbers, err := consensus.ChainReader.ReadEpochVrfBlockNums(newBlock.Header().Epoch)
+						if err != nil {
+							consensus.getLogger().Error().Err(err).
+								Uint64("MsgBlockNum", newBlock.NumberU64()).
+								Msg("[ConsensusMainLoop] failed to read VRF block number from local db")
```
chaosma: return here if err != nil ?

consensus/consensus_service.go
coolcottontail: about the constant vs numbers. I thought it is purposely done this way (golang?) because most code in harmony use numbers for things like size of pub key,  size of signature etc... so I follow their style. 

if using constant, should I we also change those signatures/pub key?

consensus/consensus_v2.go
coolcottontail: here the err just means there's no VRF numbers cached for this epoch.  so we need to generate new ones by continuing the execution below.  should I use info instead of error for the message ?

chaosma: Got it, in this case probably info is better.

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/gorilla/mux v1.7.2
 	github.com/harmony-ek/gencodec v0.0.0-20190215044613-e6740dbdd846
 	github.com/harmony-one/bls v0.0.5
+	github.com/harmony-one/vdf v0.0.0-20190727000855-54d976869062
```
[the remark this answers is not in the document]
coolcottontail: done

consensus/consensus_service.go
coolcottontail: used constants to replace numbers

core/types/block.go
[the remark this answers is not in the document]
harmony-ek: This comes admittedly a bit too late, but I must point out that this does change the hash: Because this adds 32 bytes of data.  Even if we use a byte slice, an empty slice consists of one header+length indicator byte, so it is not completely backward compatible.

We may need to revert this PR until we have a proper data structure versioning strategy.

PR #1258, opened 2019-07-27
Implement is epoch block

LeoHChen:
added unit tests as well.


coolcottontail:
LGTM

BTW, do we assume uint64 is enough for blockchain numbers ?   It seems many functions regarding blocknum convert from big.int to uint64.

PR #1261, opened 2019-07-29
switch back to normal node after update viewID/leader information fro…

chaosma:


chaosma:
Close PR, duplicate with the same change in r3 branch which will be merged into master later

PR #1263, opened 2019-07-29
patch to create test accounts[DO NOT REVIEW]

mikedoan:
patch to create test accounts


fxfactorial:
@mikedoan can we close this?

PR #1269, opened 2019-07-31
[monitoring] Use zerolog instead of log-15

xChickens:


harmony-ek:

PR #1293, opened 2019-08-06
cross shard transaction

chaosma:
- add crosslink data structure and leveldb read/write
- modify transaction structure and method for cross shard tx
- add cxReceipt data type; add cxReceipt encode/decode; 
- cross shard evm call support
- cross shard wallet support
- destination shard receipts receive and process
- cross link verification


core/blockchain.go
```diff
@@ -1959,6 +1959,26 @@ func (bc *BlockChain) WriteEpochVdfBlockNum(epoch *big.Int, blockNum *big.Int) e
 	return nil
 }
 
+// WriteCrossLinks saves the hashes of crosslinks by shardID and blockNum combination key
+func (bc *BlockChain) WriteCrossLinks(cls []types.CrossLink) error {
+	var err error
+	for _, cl := range cls {
+		err = rawdb.WriteCrossLinkShardBlock(bc.db, cl.ShardID(), cl.BlockNum(), cl.Bytes())
```
rlan35: cross-link should only happen on shard 0

chaosma: shardID here is from other shardID, only shard0 will call this function to write into datebase

core/types/crosslink.go
```diff
@@ -0,0 +1,44 @@
...
+	"sort"
+
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// CrossLink is only used on beacon chain to store the hash links from other shards
+type CrossLink struct {
+	shardID  uint32
+	blockNum uint64
+	hash     common.Hash
```
rlan35: TODO: Add state root and cxreceipt root.

api/proto/node/node.go
```diff
@@ -36,6 +35,12 @@ type BlockchainSyncMessage struct {
 	BlockHashes []common.Hash
 }
 
+// CXReceiptsMessage carrys the cross shard receipts and merkle proof
+type CXReceiptsMessage struct {
+	CXS types.CXReceipts
```
rlan35: CXS => Receipts; MKP => MerkleProof (Don't use abbreviation for naming)

api/proto/node/node.go
```diff
@@ -186,3 +194,20 @@ func DeserializeEpochShardStateFromMessage(payload []byte) (*types.EpochShardSta
...
+func ConstructCXReceiptsMessage(cxs types.CXReceipts, mkp *types.CXMerkleProof) []byte {
+	msg := &CXReceiptsMessage{CXS: cxs, MKP: mkp}
+
+	byteBuffer := bytes.NewBuffer([]byte{byte(proto.Node)})
+	byteBuffer.WriteByte(byte(Block))
+	byteBuffer.WriteByte(byte(Receipt))
+	by, err := rlp.EncodeToBytes(msg)
+
+	if err != nil {
+		log.Fatal(err)
```
rlan35: Don't use fatal (don't exit if not necessary)

cmd/client/wallet/main.go
```diff
@@ -92,6 +92,7 @@ var (
 	transferReceiverPtr   = transferCommand.String("to", "", "Specify the receiver account")
 	transferAmountPtr     = transferCommand.Float64("amount", 0, "Specify the amount to transfer")
 	transferShardIDPtr    = transferCommand.Int("shardID", 0, "Specify the shard ID for the transfer")
+	transferToShardIDPtr  = transferCommand.Int("toShardID", 0, "Specify the shard ID for the transfer")
```
rlan35: Specify the destination shard ID for the cross shard transfer

core/state_processor.go
```diff
@@ -53,47 +53,47 @@ func NewStateProcessor(config *params.ChainConfig, bc *BlockChain, engine consen
...
 // ApplyTransaction attempts to apply a transaction to the given state database
 // and uses the input parameters for its environment. It returns the receipt
 // for the transaction, gas used and an error if the transaction failed,
 // indicating the block was invalid.
-func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.DB, header *types.Header, tx *types.Transaction, usedGas *uint64, cfg vm.Config) (*types.Receipt, uint64, error) {
+func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *common.Address, gp *GasPool, statedb *state.DB, header *types.Header, tx *types.Transaction, usedGas *uint64, cfg vm.Config) (*types.Receipt, *types.CXReceipt, uint64, error) {
 	msg, err := tx.AsMessage(types.MakeSigner(config, header.Number))
-	if err != nil {
-		return nil, 0, err
+	// skip signer err for additiononly tx
```
rlan35: Can we avoid the error inside tx.AsMessage? The problem is, Here the error may not be signer error and we are blindly skipping the error.

core/state_transition.go
```diff
@@ -74,6 +76,8 @@ type Message interface {
 	Nonce() uint64
 	CheckNonce() bool
 	Data() []byte
+
+	TxType() types.TransactionType
```
rlan35: so we are actually not adding new field into transaction struct itself?

core/state_transition.go
```diff
@@ -211,15 +216,20 @@ func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bo
 	} else {
 		// Increment the nonce for the next transaction
 		st.state.SetNonce(msg.From(), st.state.GetNonce(sender.Address())+1)
-		ret, st.gas, vmerr = evm.Call(sender, st.to(), st.data, st.gas, st.value)
+		ret, st.gas, vmerr = evm.Call(sender, st.to(), st.data, st.gas, st.value, st.txType)
```
rlan35: Again, can we not report the insufficient fund error, by changing the code in evm.Call?

core/types/transaction.go
```diff
@@ -55,6 +77,8 @@ type txdata struct {
 	Amount       *big.Int        `json:"value"      gencodec:"required"`
 	Payload      []byte          `json:"input"      gencodec:"required"`
 
+	TxType TransactionType `json:"transactionType" gencodec:"required"`
```
rlan35: Why do we need to add this extra field? The same information is derived from "ShardID" and "ToShardID". I think this field is redundant inside the transaction data.

chaosma: Yes, I am refactoring the code and removing the txType.

core/types/crosslink.go
chaosma: since you are going to use header instead, I will just ignore this comment.

api/proto/node/node.go
chaosma: changed

cmd/client/wallet/main.go
chaosma: thanks for checking!

core/state_transition.go
chaosma: yes

core/state_transition.go
chaosma: now we are removing txType for addition only

core/state_processor.go
```diff
@@ -126,5 +156,34 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
...
+			utils.Logger().Warn().Msg("ApplyIncomingReceipts: Invalid incoming receipt!!")
+			continue
+		}
+		utils.Logger().Info().Msgf("ApplyIncomingReceipts: ADDING BALANCE %d", cx.Amount)
+
+		if !db.Exist(*cx.To) {
+			db.CreateAccount(*cx.To)
+		}
+		db.AddBalance(*cx.To, cx.Amount)
+		db.IntermediateRoot(config.IsEIP158(header.Number)).Bytes()
```
chaosma: remove Bytes()

PR #1294, opened 2019-08-06
Added wrapper for DRG generation based on epoch number

coolcottontail:
## Issue
Need to adjust mainnet DRG feature on EPOCH

mainnetRandomnessStartingEpoch = 10

The code should be removed after DRG is on mainnet. 

#### Test Coverage Data

* Before
* After

### DIFF
--- a/internal/configs/sharding/mainnet.go
+++ b/internal/configs/sharding/mainnet.go
@@ -14,6 +14,9 @@ const (
 
 	mainnetVdfDifficulty  = 50000 // This takes about 100s to finish the vdf
 	mainnetConsensusRatio = float64(0.66)
+
+	// TODO: remove it after randomness feature turned on mainnet
+	mainnetRandomnessStartingEpoch = 10
 )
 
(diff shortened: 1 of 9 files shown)


LeoHChen:
The data structure change has to be wrapped. @harmony-ek @rlan35 , please chime in.  I don't know how much we should test on the data structure change.  Will it break existing mainnet if it is not wrapped?

internal/configs/sharding/mainnet.go
```diff
@@ -14,6 +14,9 @@ const (
 
 	mainnetVdfDifficulty  = 50000 // This takes about 100s to finish the vdf
 	mainnetConsensusRatio = float64(0.66)
+
+	// TODO: remove it after randomness feature turned on mainnet
+	mainnetRandomnessStartingEpoch = 10
```
LeoHChen: for mainnet, let's put a huge number initially here.

LeoHChen:
Jenkins, test this please.

PR #1296, opened 2019-08-07
Metrics service update

flicker-harmony:
Added Tx Pool size metric (gauge type), added -metrics flag to node.sh (false by default - doesn't collect metrics), fixed storage (dumps metrics if failed to push to prometheus).

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -463,6 +463,11 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	pool.promoteExecutables(nil)
 }
 
+// GetTxPoolSize returns tx pool size.
+func (pool *TxPool) GetTxPoolSize() uint64 {
+	return uint64(len(pool.pending)) + uint64(len(pool.queue))
+}
+
 // Stop terminates the transaction pool.
 func (pool *TxPool) Stop() {
 	// Unsubscribe all subscriptions registered from txpool
--- a/internal/configs/node/config.go
+++ b/internal/configs/node/config.go
@@ -78,6 +78,7 @@ type ConfigType struct {
 	Port     string      // Port of the node.
 	IP       string      // IP of the node.
 
+	MetricsFlag     bool   // collect and upload metrics flag
 	PushgatewayIP   string // metrics pushgateway prometheus ip
 	PushgatewayPort string // metrics pushgateway prometheus port
 	StringRole      string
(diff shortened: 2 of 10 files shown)


harmony-ek:

LeoHChen:
Jenkins, test this please.

core/tx_pool.go
```diff
@@ -463,6 +463,11 @@ func (pool *TxPool) reset(oldHead, newHead *types.Header) {
 	pool.promoteExecutables(nil)
 }
 
+// GetTxPoolSize returns tx pool size.
+func (pool *TxPool) GetTxPoolSize() uint64 {
```
denniswon: is this class actually used? I thought this was not being used yet until txPool integration happens.

internal/configs/node/config.go
```diff
@@ -178,6 +179,16 @@ func (conf *ConfigType) SetPushgatewayPort(port string) {
 	conf.PushgatewayPort = port
 }
 
+// SetMetricsFlag set the metrics flag
```
denniswon: @LeoHChen I'm a bit confused whether nodeConfig is being deprecated or not.
Maybe I misheard, but thought shardingconfig is where new configurations should be added (which also made it confusing if configs unrelated to sharding specifically should also be put there)

Could you clarify?

core/tx_pool.go
flicker-harmony: It is used for tx pool size metric.

denniswon: but we are not using TxPool yet. cc: @rlan35 @mikedoan ?

internal/configs/node/config.go
LeoHChen: nodeConfig is mostly used when we have command-line flags.  shardingconfig is not for command-line, those are built-in parameters that shouldn't be changed in command line.

denniswon:
Tested:
Confirm master unbroken with this pr.

```
dennis.won@mp-dennisw-ml2:~/harmony/harmony (pr_metrics_service_update) $ ./bin/wallet -p local transfer --from one103q7qe5t2505lypvltkqtddaef5tzfxwsse4z7 --to one1pdv9lrdwl0rg5vglh4xtyrv3wjk3wsqket7zxy --amount 40 --shardID 0
Unlock account succeeded! ''

Transaction Id for shard 0: 0xee147c7163255b01fb059f365d9b5efb4b4ed4b1b8306eec3a323ed0a0fe4880
```

cc: @flicker-harmony @LeoHChen @xChickens

xChickens:
brave_xHSW3NH3Sc
MobaXterm_i6ivTJUxAX

harmony-ek:
@flicker-harmony, next time you modify `scripts/node.sh` loop me in please.  I needed to fix some bugs.  This scripts is externally used by all of our node operators, so we must be extra careful.  Cc: @LeoHChen

PR #1323, opened 2019-08-09
[Foundation] Add 43 (repl 3) FN keys

xChickens:
- Add 43 new FN keys
- Replace NB, SD, MD1
- Switch out Kai's lost pubkey


internal/configs/sharding/mainnet.go
```diff
@@ -57,10 +64,20 @@ func (ms mainnetSchedule) IsLastBlock(blockNum uint64) bool {
...
+// ConsensusRatio ratio of new nodes vs consensus total nodes
+func (ms mainnetSchedule) ConsensusRatio() float64 {
+	return mainnetConsensusRatio
+}
+
+var mainnetReshardingEpoch = []*big.Int{big.NewInt(0), big.NewInt(mainnetV1Epoch), big.NewInt(mainnetV2Epoch), big.NewInt(mainnetV3Epoch)}
 var mainnetV0 = MustNewInstance(4, 150, 112, genesis.HarmonyAccounts, genesis.FoundationalNodeAccounts, mainnetReshardingEpoch)
 var mainnetV1 = MustNewInstance(4, 152, 112, genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV1, mainnetReshardingEpoch)
 var mainnetV2 = MustNewInstance(4, 200, 148, genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV2, mainnetReshardingEpoch)
+var mainnetV3 = MustNewInstance(4, 248, 138, genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV3, mainnetReshardingEpoch)
```
LeoHChen: Really.  248 per shard? It is wrong.  you added 40 more keys.  the number should be 210.  Also, why do you change the number of harmony nodes to 138?  it should be 148, no change.

internal/configs/sharding/mainnet.go
```diff
@@ -11,6 +11,10 @@ const (
 	blocksPerShard     = 16384  // 2^14
 	mainnetV1Epoch     = 1
 	mainnetV2Epoch     = 5
+	mainnetV3Epoch     = 8
+
+	mainnetVdfDifficulty  = 50000 // This takes about 100s to finish the vdf
```
LeoHChen: please remove the vdf and consensusRatio variable and functions here.

LeoHChen:
Also, I need a double check of all the keys.  Who can do that?  Please get a second eye on all the account keys.

PR #1326, opened 2019-08-10
add doc design of tx pool

mikedoan:
add doc design of tx pool


mikedoan:
jenkins, test this please.

PR #1348, opened 2019-08-13
Add cross link on beacon chain

rlan35:
tested locally with 2 shards.


api/proto/node/node.go
```diff
@@ -165,6 +165,17 @@ func ConstructBlocksSyncMessage(blocks []*types.Block) []byte {
 	return byteBuffer.Bytes()
 }
 
+// ConstructCrossLinkHeadersMessage constructs cross link header message to send blocks to beacon chain
```
chaosma: send blocks => send block headers

core/blockchain.go
```diff
@@ -1121,8 +1121,20 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 					header.Logger(utils.Logger()).Warn().Err(err).Msg("cannot store shard state")
 				}
 			}
+			if len(header.CrossLinks) > 0 {
+				crossLinks := &types.CrossLinks{}
+				err := rlp.DecodeBytes(header.CrossLinks, crossLinks)
+				if err != nil {
+					header.Logger(utils.Logger()).Warn().Err(err).Msg("[insertChain] cannot parse cross links")
```
chaosma: should we return here?

api/proto/node/node.go
rlan35: Done

core/blockchain.go

PR #1365, opened 2019-08-15
[Pangaea] Use explorer and new leaders for wallet.ini

xChickens:


LeoHChen:
Do we need to merge this one?  You can just put the wallet.ini into the .hmy/ directory where your wallet is running, then you should be fine then.

xChickens:
Other people won't have consistent balances themselves if we don't merge this

PR #1368, opened 2019-08-15
add incomingReceipts handling and verification

chaosma:


core/blockchain.go
```diff
@@ -407,7 +407,7 @@ func (bc *BlockChain) FastSyncCommitHead(hash common.Hash) error {
 
 // ShardID returns the shard Id of the blockchain.
 func (bc *BlockChain) ShardID() uint32 {
-	return uint32(bc.chainConfig.ChainID.Int64())
+	return bc.CurrentBlock().ShardID()
```
rlan35: This is hacky. Please add a TODO to have a better solution.

core/state_processor.go
```diff
@@ -95,9 +103,6 @@ func getTransactionType(header *types.Header, tx *types.Transaction) types.Trans
 	if tx.ShardID() != tx.ToShardID() && header.ShardID == tx.ShardID() {
 		return types.SubtractionOnly
 	}
-	if tx.ShardID() != tx.ToShardID() && header.ShardID == tx.ToShardID() {
```
rlan35: Why only this is deleted?

core/state_processor.go
```diff
@@ -151,7 +156,7 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
 
 	var cxReceipt *types.CXReceipt
 	if txType == types.SubtractionOnly {
-		cxReceipt = &types.CXReceipt{tx.Hash(), msg.Nonce(), msg.From(), msg.To(), tx.ShardID(), tx.ToShardID(), msg.Value()}
```
rlan35: I forget why we needed this and why we deleted it?

core/state_transition.go
```diff
@@ -134,6 +135,23 @@ func ApplyMessage(evm *vm.EVM, msg Message, gp *GasPool) ([]byte, uint64, bool,
 	return NewStateTransition(evm, msg, gp).TransitionDb()
 }
 
+// ApplyIncomingReceipt will add amount into ToAddress in the receipt
+func ApplyIncomingReceipt(db *state.DB, cxp *types.CXReceiptsProof) {
+	if cxp == nil {
+		return
+	}
+
```
rlan35: I thought here some verification is needed before we add the balance?

core/types/cx_receipt.go
```diff
@@ -72,3 +76,106 @@ type CXMerkleProof struct {
 	ShardIDs      []uint32      // order list, records destination shardID
 	CXShardHashes []common.Hash // ordered hash list, each hash corresponds to one destination shard's receipts root hash
 }
+
+// CXReceiptsProof carrys the cross shard receipts and merkle proof
+type CXReceiptsProof struct {
+	Receipts    CXReceipts
+	MerkleProof *CXMerkleProof
```
rlan35: Why CXMerkleProof is pointer while the CXReceipts is not?

node/worker/worker.go
```diff
@@ -112,14 +112,20 @@ func (w *Worker) CommitTransactions(txs types.Transactions, coinbase common.Addr
...
-	for _, receipts := range receiptsList {
-		// TODO: apply receipt
-		_ = receipts
+
+	if len(receiptsList) == 0 {
+		w.current.header.IncomingReceiptHash = types.EmptyRootHash
+	} else {
+		w.current.header.IncomingReceiptHash = types.DeriveSha(types.CXReceiptsProofs(receiptsList))
+	}
+
```
rlan35: You need to apply (add balance) with the receipts too. Take a look at the commitTransactions.

core/state_processor.go
chaosma: we don't have the additiononly type transaction; use receipts to add balance

core/blockchain.go
chaosma: I think this is the good way, since the shardID is dynamically changing after resharding. But let me what do youthink

core/types/cx_receipt.go
chaosma: CXReceipts is already an array:  []*CXReceipts, so we don't use another pointer here

node/worker/worker.go
chaosma: Good catch. I added. Still got the same error

core/state_processor.go
chaosma: the nonce of the receipt is not very useful, we have to sync between source and destination shard's nonce, if only put in one shard of the nonce is not enough. so we use key-value store to track unspent receipt

core/state_transition.go
chaosma: The verification is done before put into the receiptsList. I think it should be good to go

core/blockchain.go
rlan35: This is representing the blockchain. Blockchain itself can only have one ShardID

chaosma: Added todo list

core/blockchain.go
```diff
@@ -2105,3 +2105,44 @@ func (bc *BlockChain) CXMerkleProof(shardID uint32, block *types.Block) (*types.
...
+	newCheckpoint := lastCheckpoint
+
+	// the new checkpoint will not exceed currentNum+1
+	for num := lastCheckpoint; num <= currentNum+1; num++ {
+		hash, _ := rawdb.ReadCXReceiptsProofUnspent(bc.db, shardID, num)
+		if hash == rawdb.EmptyHash {
+			// TODO: check if there is IncompingReceiptsHash in crosslink header
+			// if the rootHash is non-empty, it means incomingReceipts are not delivered
+			// otherwise, it means there is no cross-shard transactions for this block
+			newCheckpoint = num
```
rlan35: Seems these "newCheckpoint = num" can be grouped into one call.

core/rawdb/accessors_chain.go
```diff
@@ -556,3 +563,46 @@ func DeleteCXReceipts(db DatabaseDeleter, shardID uint32, number uint64, hash co
...
+func ReadCXReceiptsProofUnspent(db DatabaseReader, shardID uint32, number uint64) (common.Hash, error) {
+	data, err := db.Get(cxReceiptUnspentKey(shardID, number))
+	if err != nil || len(data) == 0 {
+		return common.Hash{}, ctxerror.New("[ReadCXReceiptsProofUnspent] Cannot find the key", "shardID", shardID, "number", number).WithCause(err)
+	}
+	return common.BytesToHash(data), nil
+}
+
+// TryWriteCXReceiptsProofUnspent check whether a CXReceiptsProof is unspent, write to database if unspent
+func TryWriteCXReceiptsProofUnspent(dbr DatabaseReader, dbw DatabaseWriter, shardID uint32, number uint64, blockHash common.Hash) error {
```
rlan35: This is not used?

core/state_processor.go
```diff
@@ -74,17 +78,22 @@ func (p *StateProcessor) Process(block *types.Block, statedb *state.DB, cfg vm.C
 		}
 		receipts = append(receipts, receipt)
 		if cxReceipt != nil {
-			cxs = append(cxs, cxReceipt)
+			outcxs = append(outcxs, cxReceipt)
 		}
 		allLogs = append(allLogs, receipt.Logs...)
 	}
+
+	for _, cx := range block.IncomingReceipts() {
```
rlan35: Add a note: incomingReceipts should always be processed after transactions (to be consistent with the block proposal)

core/rawdb/accessors_chain.go
chaosma: It will be used to avoid double spent, this part is not written yet.

core/state_processor.go
chaosma: comments added

core/blockchain.go
chaosma: Since there is a TODO item, so I just make the separation of logic so that it is clearer to me. After finish TODO, we can optimize the code

core/state_processor.go
```diff
@@ -151,10 +157,32 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
...
+			utils.Logger().Warn().Msg("ApplyIncomingReceipts: Invalid incoming receipt!!")
+			continue
+		}
+		utils.Logger().Info().Msgf("ApplyIncomingReceipts: ADDING BALANCE %d", cx.Amount)
+
+		if !db.Exist(*cx.To) {
+			db.CreateAccount(*cx.To)
+		}
+		db.AddBalance(*cx.To, cx.Amount)
+		db.IntermediateRoot(config.IsEIP158(header.Number)).Bytes()
```
rlan35: Should we be consistent with ApplyTransaction:
        if config.IsByzantium(header.Number) {
		statedb.Finalise(true)
	} else {
		root = statedb.IntermediateRoot(config.IsEIP158(header.Number)).Bytes()
	}

core/types/cx_receipt.go
```diff
@@ -72,3 +76,106 @@ type CXMerkleProof struct {
...
+}
+
+// CXReceiptsProofs is a list of CXReceiptsProof
+type CXReceiptsProofs []*CXReceiptsProof
+
+// Len returns the length of s.
+func (cs CXReceiptsProofs) Len() int { return len(cs) }
+
+// Swap swaps the i'th and the j'th element in s.
+func (cs CXReceiptsProofs) Swap(i, j int) { cs[i], cs[j] = cs[j], cs[i] }
```
rlan35: What is this func for?

core/types/cx_receipt.go
```diff
@@ -72,3 +76,106 @@ type CXMerkleProof struct {
...
+
+	// (1) verify the CXReceipts trie root match
+	if sha != shardRoot {
+		return ctxerror.New("[IsValidCXReceiptsProof] Trie Root of ReadCXReceipts Not Match", "sourceShardID", sourceShardID, "sourceBlockNum", sourceBlockNum, "calculated", sha, "got", shardRoot)
+	}
+
+	// (2) verify the outgoingCXReceiptsHash match
+	outgoingHashFromSourceShard := crypto.Keccak256Hash(byteBuffer.Bytes())
+	if outgoingHashFromSourceShard != sourceOutgoingCXReceiptsHash {
+		return ctxerror.New("[ProcessReceiptMessage] IncomingReceiptRootHash from source shard not match", "sourceShardID", sourceShardID, "sourceBlockNum", sourceBlockNum, "calculated", outgoingHashFromSourceShard, "got", sourceOutgoingCXReceiptsHash)
```
rlan35: [IsValidCXReceiptsProof]

core/types/transaction_signing.go
```diff
@@ -42,15 +42,7 @@ type sigCache struct {
...
-	switch {
-	// TODO (chao) clean up different forks in ETH code
-	case config.IsEIP155(blockNumber):
-		signer = NewEIP155Signer(config.ChainID)
-	case config.IsHomestead(blockNumber):
-		signer = HomesteadSigner{}
-	default:
-		signer = FrontierSigner{}
-	}
+	signer = NewEIP155Signer(config.ChainID)
```
rlan35: This is the most current signing method for ethereum?

core/types/transaction_signing.go
```diff
@@ -232,11 +224,13 @@ func (fs FrontierSigner) Sender(tx *Transaction) (common.Address, error) {
 }
 
 func recoverPlain(sighash common.Hash, R, S, Vb *big.Int, homestead bool) (common.Address, error) {
+	V := byte(Vb.Uint64() - 27)
```
rlan35: Revert test code.

core/types/transaction_signing.go
chaosma: I think so. Also, the homestead signing method not work in our code.

core/types/transaction_signing.go
chaosma: Is this test code?? I thought it's the original code

core/types/cx_receipt.go
chaosma: changed

core/types/cx_receipt.go
chaosma: It's not used, can be used to sort, here we might not needed it.

core/state_processor.go
chaosma: Maybe, I am not quite sure about the byzantium block. But I think we should remove the isByzantium block (sadly we already passed that block number in ethereum). Let me know what do you think

core/types/transaction_signing.go
rlan35: I didn't mean to delete this line. I mean revert these changes (the prints and the moved line)

core/types/transaction_signing.go
rlan35: ok. EIP155 fork is at "EIP155Block:         big.NewInt(2675000)," which means the 1 transaction already in our mainnet is using default Frontier signer. Will these two compatible? If not, we need to add the fork logic.

core/state_processor.go
rlan35: we haven't "ByzantiumBlock:      big.NewInt(4370000)," it's 4.3 million blocks. Our single shard blocks haven't go beyond 1m

chaosma: yeah, wrong memory, we passed the EIP155 block. In this case, we could delete the Byzantium block checking here

core/types/transaction_signing.go
chaosma: Don't know if compatible with Frontier signer, I can do some test about it

core/types/transaction_signing.go
chaosma: deleted the ops ...

core/state_processor.go
chaosma: I checked the code, here it doesn't matter, both will use finalise(true). However, Byzantium fork contains several security improvement. I would suggest to make the byzantium block num smaller to benefit from that.

core/types/transaction_signing.go
chaosma: Local test not work for Frontier and Homestead signer

PR #1377, opened 2019-08-19
add one1 support to explorer API /address

lzl124631x:


api/service/explorer/service.go
```diff
@@ -103,7 +104,7 @@ func (s *Service) Run() *http.Server {
 	s.router.Path("/tx").HandlerFunc(s.GetExplorerTransaction)
 
 	// Set up router for address.
-	s.router.Path("/address").Queries("id", "{[0-9A-Fa-fx]*?}").HandlerFunc(s.GetExplorerAddress).Methods("GET")
+	s.router.Path("/address").Queries("id", fmt.Sprintf("{([0-9A-Fa-fx]*?)|(one1[%s]{38})}", bech32.Charset)).HandlerFunc(s.GetExplorerAddress).Methods("GET")
```
rlan35: our testnet address will be "tone1...."    better add wildcard before "one" too.

LeoHChen: should be specific, either tone1, or one1, IMHO, shouldn't be wildcard.

PR #1388, opened 2019-08-21
Fix node killing in local test

harmony-ek:
Do not target processes with `harmony`, `txgen`, `soldier`, `commander`, `profiler`, or `bootnode` somewhere in their full command line, but just kill processes whose binary name exactly matches one of these.

Also retire port-number-based node killer, which is unused.


yzhoujuly:
Tested in my local machine with "Harmony*" name. It works --- harmony processes run well.

PR #1389, opened 2019-08-21
avoid cross shard receipts double spent 

chaosma:


core/rawdb/accessors_chain.go
```diff
@@ -573,14 +573,12 @@ func ReadCXReceiptsProofUnspent(db DatabaseReader, shardID uint32, number uint64
...
 
-// TryWriteCXReceiptsProofUnspent check whether a CXReceiptsProof is unspent, write to database if unspent
-func TryWriteCXReceiptsProofUnspent(dbr DatabaseReader, dbw DatabaseWriter, shardID uint32, number uint64, blockHash common.Hash) error {
-	hash, _ := ReadCXReceiptsProofUnspent(dbr, shardID, number)
-	// only write to database for the first time given a specified block number
-	if hash == EmptyHash {
-		return dbw.Put(cxReceiptUnspentKey(shardID, number), blockHash.Bytes())
-	}
-	return ctxerror.New("[TryWriteCXReceiptsProofUnspent] blockHash already exist", "had", hash, "tryPut", blockHash)
+// WriteCXReceiptsProofUnspent check whether a CXReceiptsProof is unspent, write to database if unspent
```
rlan35: The code doesn’t check it’s unspent or not.

node/node_handler.go
```diff
@@ -630,6 +631,10 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block) {
 			Msg("BINGO !!! Reached Consensus")
 	}
 
+	// mark incomingReceipts as spent and clean up
+	node.Blockchain().MarkCXReceiptsSpent(newBlock)
```
rlan35: Should these be in the insert chain? So it’s more atomic operation?

core/blockchain.go
```diff
@@ -2132,17 +2132,57 @@ func (bc *BlockChain) NextCXReceiptsProofUnpentCheckpoint(currentNum uint64, sha
...
 	for num := lastCheckpoint; num < newCheckpoint; num++ {
 		rawdb.DeleteCXReceiptsProofUnspent(bc.db, shardID, num)
 	}
 }
+
+// MarkCXReceiptsSpent mark the incomingReceipts in a block as spent
+// avoid double spent
+func (bc *BlockChain) MarkCXReceiptsSpent(block *types.Block) {
+	for _, cxp := range block.IncomingReceipts() {
+		rawdb.WriteCXReceiptsProofUnspent(bc.db, cxp)
```
rlan35: This func should be named WriteCXReceiptsProofSpent, right?

chaosma: My original thought is mark cxreceiptsproof as spent

node/node_handler.go
chaosma: en. good point, let me take a look

core/rawdb/accessors_chain.go
chaosma: Yes, this is only write to database, the low level function. The checking is done in blockchain.go

chaosma: comments modified

node/node_handler.go
chaosma: move this into insertchain

PR #1394, opened 2019-08-21
support one address for rpc call

mikedoan:
support one address for rpc call

### DIFF
--- a/internal/hmyapi/blockchain.go
+++ b/internal/hmyapi/blockchain.go
@@ -15,6 +15,7 @@ import (
 	"github.com/harmony-one/harmony/core"
 	"github.com/harmony-one/harmony/core/types"
 	"github.com/harmony-one/harmony/core/vm"
+	internal_common "github.com/harmony-one/harmony/internal/common"
 	"github.com/harmony-one/harmony/internal/utils"
 )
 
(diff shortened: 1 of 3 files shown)


internal/hmyapi/blockchain.go
```diff
@@ -86,9 +87,10 @@ func (s *PublicBlockChainAPI) GetStorageAt(ctx context.Context, address common.A
 // GetBalance returns the amount of Nano for the given address in the state of the
 // given block number. The rpc.LatestBlockNumber and rpc.PendingBlockNumber meta
 // block numbers are also allowed.
-func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (*hexutil.Big, error) {
+func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address string, blockNr rpc.BlockNumber) (*hexutil.Big, error) {
 	// TODO: currently only get latest balance. Will add complete logic later.
-	return s.b.GetBalance(address)
+	adr := internal_common.ParseAddr(address)
```
rlan35: What about malformed adresses?

mikedoan:
Screen Shot 2019-08-21 at 2 41 02 PM
Screen Shot 2019-08-21 at 2 37 34 PM

internal/hmyapi/blockchain.go
mikedoan: i was thinking about that but i looked at deeper logic it can always return any address.

If the address input is specified as malformed it just returns 0x0000000000000000000000000000000000000000

mikedoan: fmt.Println(ParseAddr("minh3").Hex()) returns "0x0000000000000000000000000000000000000000"

rlan35: That address maybe used for burning tokens, so it may also have balances, please add a todo to better handle this

PR #1402, opened 2019-08-22
add crosslink verification

chaosma:


internal/configs/sharding/localnet.go
```diff
@@ -22,7 +22,7 @@ const (
 	localnetVdfDifficulty  = 5000 // This takes about 10s to finish the vdf
 	localnetConsensusRatio = float64(0.1)
 
-	localnetFirstCrossLinkBlock = 13
+	localnetFirstCrossLinkBlock = 3
```
denniswon: does this exist because cross shard operation (transaction) takes "at most" 3 blocks generation to finalize?

Is cross link created from the receiver side? @chaosma @rlan35

chaosma: It is used to deal with fork condition, because we don't have crosslink before. it will only take effect after some block

node/node_cross_shard.go
```diff
@@ -101,13 +101,43 @@ func (node *Node) verifyIncomingReceipts(block *types.Block) error {
...
+		block := beaconChain.GetBlockByNumber(blockNum)
+		if block == nil {
+			return ctxerror.New("[verifyCrosslinkWithReceipts] Cannot get beaconchain heaer", "blockNum", blockNum, "shardID", shardID)
+		}
+		hash = block.Hash()
+		outgoingReceiptHash = block.OutgoingReceiptHash()
+	} else {
+		crossLink, err := beaconChain.ReadCrossLink(shardID, blockNum, false)
+		if err != nil {
+			return ctxerror.New("[verifyCrosslinkWithReceipts] Cannot get crosslink", "blockNum", blockNum, "shardID", shardID).WithCause(err)
```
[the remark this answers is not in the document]
chaosma: good catch!

node/node_cross_shard.go
```diff
@@ -101,13 +101,43 @@ func (node *Node) verifyIncomingReceipts(block *types.Block) error {
...
+func (node *Node) compareCrosslinkWithReceipts(cxp *types.CXReceiptsProof) error {
+	var hash, outgoingReceiptHash common.Hash
+
+	shardID := cxp.MerkleProof.ShardID
+	blockNum := cxp.MerkleProof.BlockNum.Uint64()
+	beaconChain := node.Beaconchain()
+	if shardID == 0 {
+		block := beaconChain.GetBlockByNumber(blockNum)
+		if block == nil {
+			return ctxerror.New("[compareCrosslinkWithReceipts] Cannot get beaconchain heaer", "blockNum", blockNum, "shardID", shardID)
```
denniswon: nit: header

core/state/statedb.go
```diff
@@ -676,6 +676,6 @@ func (db *DB) Commit(deleteEmptyObjects bool) (root common.Hash, err error) {
 		}
 		return nil
 	})
-	log.Debug("Trie cache stats after commit", "misses", trie.CacheMisses(), "unloads", trie.CacheUnloads())
+	//log.Debug("Trie cache stats after commit", "misses", trie.CacheMisses(), "unloads", trie.CacheUnloads())
```
denniswon: ? should we just remove

node/node_syncing.go
```diff
@@ -360,6 +360,7 @@ func (node *Node) CalculateResponse(request *downloader_pb.DownloaderRequest, in
 				continue
 			}
 			encodedBlock, err := rlp.EncodeToBytes(block)
+
```
denniswon: ?

core/state/statedb.go
chaosma: Somebody might need it in future, don't know who add this

node/node_cross_shard.go
chaosma: changed

PR #1408, opened 2019-08-22
Cross shard rj

rlan35:


denniswon:
Seems like @chaosma merged changes here in https://github.com/harmony-one/harmony/pull/1402/files already. moved my comments there.

PR #1413, opened 2019-08-23
Merge cross shard tx feature into master

rlan35:


rlan35:
Jenkins, test this please.

api/client/client.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // Client represents a node (e.g. a wallet) which  sends transactions and receives responses from the harmony network
 type Client struct {
-	ShardID      uint32               // ShardID
+	ShardID      uint32               // ShardIDs
```
denniswon: just // ShardID

api/service/syncing/syncing.go
```diff
@@ -722,20 +733,24 @@ func (ss *StateSync) SyncLoop(bc *core.BlockChain, worker *worker.Worker, willJo
 	if !isBeacon {
 		ss.RegisterNodeInfo()
 	}
+	ticker := time.NewTicker(SyncLoopFrequency * time.Second)
```
denniswon: does this resolve a specific issue?

api/client/client.go
rlan35: Done.

api/service/networkinfo/service.go
```diff
@@ -132,6 +132,7 @@ func (s *Service) Init() error {
 	utils.Logger().Info().Str("Rendezvous", string(s.Rendezvous)).Msg("Announcing ourselves...")
 	s.discovery = libp2pdis.NewRoutingDiscovery(s.dht)
 	libp2pdis.Advertise(ctx, s.discovery, string(s.Rendezvous))
+	libp2pdis.Advertise(ctx, s.discovery, string(p2p.GroupIDBeaconClient))
```
rlan35: @harmony-ek Why do we advertise on BeaconClient group here?

harmony-ek: @rlan35, that is because beaconchain now sends new blocks to the network using `node.BroadcastNewBlock()`, which sends the block to the client group of the shard.  See #1391 for details.

PR #1423, opened 2019-08-24
[project] Move utils, bech32, common out from internal to pkg

fxfactorial:
As I am building out `go-sdk` which is in another repo, I need some functionalities contained in these packages. 

Note: the go toolchain does enforce `internal` so I moved these packages out from `internal` and into `pkg` signaling they are publicly consumable.


harmony-ek:

LeoHChen:
Jenkins, test this please.

fxfactorial:
From what I am understanding in the codebase, we are making assumption of existence of some test file/setup code under `internal/utils` and this is causing the test failures. Appreciate some guidance on what to move around back to under `internal/utils` (I can figure it out too but faster to ask methinks).

LeoHChen:
goimports FAILED!

go test passed actually. @fxfactorial

PR #1456, opened 2019-08-29
display error when balance rpc fails

rlan35:


cmd/client/wallet/main.go
```diff
@@ -852,8 +851,9 @@ func FetchBalance(address common.Address) []*AccountState {
 				}(rpcServerID)
 			}
 			wgShard.Wait()
-
-			result[shardID] = &AccountState{balance, nonce}
+			if balance.Cmp(big.NewInt(-1)) > 0 {
+				result[shardID] = &AccountState{balance, nonce}
+			}
```
mikedoan: what would we do with else condition?

PR #1459, opened 2019-08-29
add networkID for scheduling

mikedoan:
add networkID for scheduling

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -420,7 +420,7 @@ func main() {
 				err)
 			os.Exit(1)
 		}
-		core.ShardingSchedule = shardingconfig.NewFixedSchedule(devnetConfig, shardingconfig.DevNet)
+		core.ShardingSchedule = shardingconfig.NewFixedSchedule(devnetConfig)
 	}
 
 	initSetup()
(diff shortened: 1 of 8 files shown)


internal/configs/sharding/fixedschedule.go
```diff
@@ -86,8 +87,12 @@ func (s fixedSchedule) TxsThrottleConfig() *TxsThrottleConfig {
 	}
 }
 
+func (s fixedSchedule) GetNetworkID() NetworkID {
+	return s.networkID
```
LeoHChen: you don't need to do this, just return devnet here.

cmd/harmony/main.go
```diff
@@ -420,7 +420,7 @@ func main() {
 				err)
 			os.Exit(1)
 		}
-		core.ShardingSchedule = shardingconfig.NewFixedSchedule(devnetConfig)
+		core.ShardingSchedule = shardingconfig.NewFixedSchedule(devnetConfig, shardingconfig.DevNet)
```
LeoHChen: no need to do this, just return devnet in fixedschedule

internal/configs/sharding/instance.go
```diff
@@ -7,6 +7,18 @@ import (
 	"github.com/harmony-one/harmony/internal/genesis"
 )
 
+// NetworkID is the network type of the blockchain.
+type NetworkID byte
+
+//Consensus and other message categories
```
LeoHChen: please fix the comment.  Also, please add a String function to the NetworkID type, so that it will be easier for debugging.

PR #1462, opened 2019-08-30
fix syncLoop trap bug

chaosma:


api/service/syncing/syncing.go
```diff
@@ -734,14 +734,15 @@ func (ss *StateSync) SyncLoop(bc *core.BlockChain, worker *worker.Worker, willJo
...
+Loop:
 	for {
 		select {
 		case <-ticker.C:
 			otherHeight := ss.getMaxPeerHeight()
 			currentHeight := bc.CurrentBlock().NumberU64()
 			if currentHeight >= otherHeight {
-				utils.Logger().Info().Msgf("[SYNC] Node is now IN SYNC! (ShardID: %d)", bc.ShardID())
-				break
+				utils.Logger().Info().Msgf("[SYNC] Node is now IN SYNC! (ShardID: %d, otherHeight: %d, currentHeight: %d)", bc.ShardID(), otherHeight, currentHeight)
```
denniswon: wont this create infinite loop? how does this fix the issue?

chaosma: This will break the infinite loop, before it just cannot escape the loop. right?

PR #1463, opened 2019-08-30
Revert signer change and add shard info in tx hash for eip155

rlan35:
We shouldn't change the signer without adding correct forking logic.

### DIFF
--- a/core/types/transaction_signing.go
+++ b/core/types/transaction_signing.go
@@ -42,7 +42,14 @@ type sigCache struct {
 // MakeSigner returns a Signer based on the given chain config and block number.
 func MakeSigner(config *params.ChainConfig, blockNumber *big.Int) Signer {
 	var signer Signer
-	signer = NewEIP155Signer(config.ChainID)
+	switch {
+	case config.IsEIP155(blockNumber):
+		signer = NewEIP155Signer(config.ChainID)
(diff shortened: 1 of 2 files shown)


coolcottontail:
just FYI... once this is reverted. All transactions will be signed using old signer as blockNum is small.  The wallet won't be able to submit new transactions.

core/types/transaction_signing.go
```diff
@@ -42,7 +42,14 @@ type sigCache struct {
 // MakeSigner returns a Signer based on the given chain config and block number.
 func MakeSigner(config *params.ChainConfig, blockNumber *big.Int) Signer {
 	var signer Signer
-	signer = NewEIP155Signer(config.ChainID)
+	switch {
+	case config.IsEIP155(blockNumber):
+		signer = NewEIP155Signer(config.ChainID)
+	case config.IsHomestead(blockNumber):
```
LeoHChen: why do we need the homestead signer?

rlan35:
> just FYI... once this is reverted. All transactions will be signed using old signer as blockNum is small. The wallet won't be able to submit new transactions.

Good catch. also reverted wallet signing part.

rlan35:
> What about the wallet signer part? It uses nil as the chain config

also reverted.

PR #1466, opened 2019-08-30
[netinfo] limit the number of discovery

LeoHChen:


api/service/networkinfo/service.go
```diff
@@ -51,6 +52,8 @@ const (
 
 	// register to bootnode every ticker
 	dhtTicker = 6 * time.Hour
+
+	discoveryLimit = 32
```
denniswon: why 32?

denniswon: discussed offline. its random. default 100

PR #1485, opened 2019-09-01
fix ONE address issues for other rpcs

mikedoan:
fix ONE address issues for other rpcs

### DIFF
--- a/internal/hmyapi/blockchain.go
+++ b/internal/hmyapi/blockchain.go
@@ -64,7 +64,8 @@ func (s *PublicBlockChainAPI) GetBlockByHash(ctx context.Context, blockHash comm
 }
 
 // GetCode returns the code stored at the given address in the state for the given block number.
-func (s *PublicBlockChainAPI) GetCode(ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (hexutil.Bytes, error) {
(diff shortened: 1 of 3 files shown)


internal/hmyapi/blockchain.go
```diff
@@ -64,7 +64,8 @@ func (s *PublicBlockChainAPI) GetBlockByHash(ctx context.Context, blockHash comm
 }
 
 // GetCode returns the code stored at the given address in the state for the given block number.
-func (s *PublicBlockChainAPI) GetCode(ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (hexutil.Bytes, error) {
+func (s *PublicBlockChainAPI) GetCode(ctx context.Context, adr string, blockNr rpc.BlockNumber) (hexutil.Bytes, error) {
```
lzl124631x: ubernit: haven't seen `adr` as abbreviation for address, how about `addr`.

mikedoan: fixed

mikedoan:
```
// ParseAddr parses the given address, either as bech32 or as hex.
// The result can be 0x00..00 if the passing param is not a correct address.
func ParseAddr(s string) ethCommon.Address {
```
@lzl124631x ParseAddr never fails , in case the input string is invalid address it will parse to 0x000000 address which is never used

PR #1493, opened 2019-09-03
Add CrossLink, EIP155, and S3 fork

rlan35:
1. Refactored chain config to remove ethereum's forks and add our fork blocks for cross link, eip155 and S3 (which roll up all ethereum forks until constantinople)
2. Apply EIP155 forks to all relevant places.
3. Refactor node_newblock.go so it's more readable and less prone to mistakes (especially about the worker.UpdateCurrent part)

Tested locally with both intra-shard and cross-shard transactions.


rlan35:
jenkins, please test this.

rlan35:
Jenkins, test this please.

internal/keystore/keystore.go
```diff
@@ -42,7 +43,7 @@ func Unlock(account accounts.Account) {
 // SignTx signs transaction using account key
 func SignTx(account accounts.Account, tx *types.Transaction) (*types.Transaction, error) {
 	if hmyKeystore != nil {
-		return hmyKeystore.SignTx(account, tx, nil)
+		return hmyKeystore.SignTx(account, tx, big.NewInt(1)) // TODO: use chainID from config
```
chaosma: we need use chain config ID instead of 1 to distinguish mainnet from testnet and pangaea

internal/params/gas_table.go
```diff
@@ -20,11 +20,11 @@ type GasTable struct {
...
 
-// Variables containing gas prices for different ethereum phases.
+// Variables containing gas prices for different harmony phases.
 var (
-	// GasTableHomestead contain the gas prices for
-	// the homestead phase.
-	GasTableHomestead = GasTable{
+	// GasTableR3 contain the gas prices for
+	// the r3 phase.
+	GasTableR3 = GasTable{
```
chaosma: We already passed R3 phase, do we still need it?

node/worker/worker.go
```diff
@@ -85,6 +87,14 @@ func (w *Worker) throttleTxs(selected types.Transactions, recentTxsStats types.R
 
 // SelectTransactionsForNewBlock selects transactions for new block.
 func (w *Worker) SelectTransactionsForNewBlock(newBlockNum uint64, txs types.Transactions, recentTxsStats types.RecentTxsStats, txsThrottleConfig *shardingconfig.TxsThrottleConfig, coinbase common.Address) (types.Transactions, types.Transactions, types.Transactions) {
+	// Must update to the correct current state before processing potential txns
+	if err := w.UpdateCurrent(coinbase); err != nil {
```
chaosma: updatecurrent is called in here and commit transactions, can we remove one of them?

internal/keystore/keystore.go
rlan35: This needs wallet configuration. This part of code may not have reference to chainConfig.

internal/params/gas_table.go
rlan35: This is for backward compatibility

node/worker/worker.go
rlan35: We need both of them.

rlan35:
> To capture the offline conversation with @rlan35:
> 
> **tl;dr** – please use epoch-based upgrade instead of block-number-based upgrade.
> 
> Block-number-based upgrade risks desynchronized upgrade scenario, where one shard lagging behind the schedule and can't accept a new feature yet. For example, given:
> 
> * A cross-shard transaction block height _H__CX_ which introduces CX-related data fields (incoming/outgoing hash and cross links);
> * A shard _s__above_ whose block height is above (past) _H__CX_, and
> * A shard _s__below_ whose block height is below (before) _H__CX_,
> 
> Cross-shard transactions sent from _s__above_ to _s__below_ cannot be processed until _s__below_ reaches _H__CX_, because blocks in _s__below_ will not have CX fields such as `IncomingReceipts` yet.
> 
> One can do this upgrade with a two-threshold approach: Threshold 1 introduces the cross-shard transaction logic and data fields but nodes still reject CX submitted by clients (that is, the CX logic and data fields are not exercised) until threshold 2 comes around, and threshold 2 is introduced only when all shards have past threshold 1.
> 
> This two-threshold approach is, however, subject to another problem: The time gap between the fastest shard and the slowest shard each reaching threshold 1 will increase from now on, especially when shard 0 starts rejecting regular transactions—after its role changes into staking/clearinghouse-only mode—and admits less transactions and consequently less blocks.
> 
> Using epoch-based approach does not suffer from the latter problem, as an epoch will be time-synchronized across all shards by definition. (Right now it is not, and different shards enter the same epoch at different times, so we still need the two-epoch approach until we have proper resharding support that introduces synchronization of epoch across all shards.)

Updated to fork by epoch.

internal/params/config.go
```diff
@@ -170,136 +60,48 @@ type TrustedCheckpoint struct {
...
+	return isForked(c.EIP155Epoch, epoch)
 }
 
-// IsConstantinople returns whether num is either equal to the Constantinople fork block or greater.
-func (c *ChainConfig) IsConstantinople(num *big.Int) bool {
-	return isForked(c.ConstantinopleBlock, num)
+// IsCrossLink returns whether epoch is either equal to the CrossLink fork epoch or greater.
+func (c *ChainConfig) IsCrossLink(epoch *big.Int) bool {
+	return isForked(c.CrossLinkEpoch, epoch)
 }
```
harmony-ek:
```suggestion
}

// IsVrf returns whether blocks in the given epoch have VRF/VDF fields.
func (c *ChainConfig) IsVRF(epoch *big.Int) bool {
	// …
}
```

node/node_newblock.go
```diff
@@ -49,105 +47,90 @@ func (node *Node) WaitForConsensusReadyv2(readySignal chan struct{}, stopChan ch
...
-					selectedTxs := node.getTransactionsForNewBlock(coinbase)
-					if err := node.Worker.UpdateCurrent(coinbase); err != nil {
-						utils.GetLogger().Error("Failed updating worker's state", "Error", err)
-					}
-					utils.Logger().Info().
+					utils.Logger().Debug().
 						Uint64("blockNum", node.Blockchain().CurrentBlock().NumberU64()+1).
-						Int("selectedTxs", len(selectedTxs)).
-						Msg("PROPOSING NEW BLOCK ------------------------------------------------")
+						Msg("=========Proposing New Block==========")
```
harmony-ek: Can we keep the current message?  This and BINGO/HOORAY are landmark log lines that people may use often for liveness checks.  We should not change these messages without first communicating them out to the users.

node/staking_test.go
```diff
@@ -40,12 +40,16 @@ func TestUpdateStakingList(t *testing.T) {
 	node := New(host, consensus, testDBFactory, false)
 	node.BlockPeriod = 8 * time.Second
 
-	for i := 0; i < 5; i++ {
+	for i := 0; i < 1; i++ {
```
harmony-ek: If it's just once, why not unroll the block?

node/staking_test.go
```diff
@@ -40,12 +40,16 @@ func TestUpdateStakingList(t *testing.T) {
...
 		selectedTxs := node.getTransactionsForNewBlock(common.Address{})
 		node.Worker.CommitTransactions(selectedTxs, common.Address{})
-		block, _ := node.Worker.Commit([]byte{}, []byte{}, 0, common.Address{})
+		block, err := node.Worker.FinalizeNewBlock([]byte{}, []byte{}, 0, common.Address{}, nil, nil)
 
-		err := node.AddNewBlock(block)
+		if err != nil {
+			t.Errorf("Error when finalizing block: %v", err)
+		}
+		block.Header()
```
harmony-ek: Are you adding this to finalize the header and compute the hash?

node/worker/worker.go
```diff
@@ -258,16 +276,51 @@ func (w *Worker) IncomingReceipts() []*types.CXReceiptsProof {
...
 		copy(w.current.header.LastCommitSignature[:], sig[:])
 		w.current.header.LastCommitBitmap = append(signers[:0:0], signers...)
 	}
 	w.current.header.Coinbase = coinbase
 	w.current.header.ViewID = new(big.Int)
 	w.current.header.ViewID.SetUint64(viewID)
-	w.current.header.CrossLinks = crossLinks
+
+	// Cross Links
+	if crossLinks != nil && len(crossLinks) != 0 {
```
harmony-ek: OT but FWIW: Can we just define the `CrossLinks` header field not as `[]byte` but as `types.CrossLink` so that we do not have to do manual serialization/deserialization?  Or do we need to differentiate the “`CrossLinks` field absent” and “`CrossLinks` field present but is an empty slice” case?  (For `ShardState` that was the case from what I remember… we may need to revisit that as well.)

core/blockchain.go
```diff
@@ -241,6 +241,14 @@ func IsEpochBlock(block *types.Block) bool {
 	return ShardingSchedule.IsLastBlock(block.NumberU64() - 1)
 }
 
+// EpochFirstBlock returns the block number of the first block of an epoch.
+func EpochFirstBlock(epoch uint64) uint64 {
```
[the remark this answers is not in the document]

core/resharding.go
```diff
@@ -136,17 +135,18 @@ func Shuffle(list []types.NodeID) {
...
 
 // GetLastBlockNumberFromEpoch calculates the last block number for the given
 // epoch.  TODO ek – this is a temp hack.
 // TODO lc - use ShardingSchedule function
-func GetLastBlockNumberFromEpoch(epoch uint64) uint64 {
-	return (epoch+1)*ShardingSchedule.BlocksPerEpoch() - 1
-}
+//func GetLastBlockNumberFromEpoch(epoch uint64) uint64 {
+//	return (epoch+1)*ShardingSchedule.BlocksPerEpoch() - 1
+//}
```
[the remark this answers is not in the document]
rlan35: Done.

core/resharding.go
```diff
@@ -164,9 +164,9 @@ func GetShardingStateFromBlockChain(bc *BlockChain, epoch *big.Int) (*ShardingSt
 	}
 	shardState = shardState.DeepCopy()
 
-	blockNumber := GetBlockNumberFromEpoch(epoch.Uint64())
-	rndSeedBytes := bc.GetVdfByNumber(blockNumber)
-	rndSeed := binary.BigEndian.Uint64(rndSeedBytes[:])
+	//blockNumber := GetBlockNumberFromEpoch(epoch.Uint64())
+	//rndSeedBytes := bc.GetVdfByNumber(blockNumber)
+	rndSeed := uint64(0)
```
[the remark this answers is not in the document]
rlan35: because randomness is not launched yet. Basically these code is wrong (GetBlockNumberFromEpoch is wrong). We better disable it than letting these code slip through launch. Added todo here. This is part of the randomness and resharding work.

internal/keystore/keystore.go
```diff
@@ -42,7 +43,7 @@ func Unlock(account accounts.Account) {
 // SignTx signs transaction using account key
 func SignTx(account accounts.Account, tx *types.Transaction) (*types.Transaction, error) {
```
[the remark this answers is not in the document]
rlan35: Done.

internal/keystore/keystore.go
rlan35: Done.

internal/params/config.go
```diff
@@ -9,143 +9,33 @@ import (
 
 // Genesis hashes to enforce below configs on.
 var (
-	MainnetGenesisHash = common.HexToHash("0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3")
-	TestnetGenesisHash = common.HexToHash("0x41941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2d")
-	RinkebyGenesisHash = common.HexToHash("0x6341fd3daf94b748c72ced5a5b26028f2474f5f00d824504e4fa37a75767e177")
-	GoerliGenesisHash  = common.HexToHash("0xbf7e331f7f7c1dd2e05159666b3bf8bc7a8a3a9eb1d518969eab529dd9b88c1a")
+	MainnetGenesisHash = common.HexToHash("0x")
+	TestnetGenesisHash = common.HexToHash("0x")
```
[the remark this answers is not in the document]
rlan35: Added todo. Basically, I am removing the ethereum's trace here. because our genesis hash is not the one in the code. Added todo to populate our own genesis hash. (these hashes are not used anywhere either way for now)

internal/params/config.go
```diff
@@ -9,143 +9,33 @@ import (
...
-	//
-	// This configuration is intentionally not using keyed fields to force anyone
-	// adding flags to the config to also have to set these fields.
-	AllCliqueProtocolChanges = &ChainConfig{big.NewInt(1337), big.NewInt(0), nil, false, big.NewInt(0), common.Hash{}, big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), nil, nil, &CliqueConfig{Period: 0, Epoch: 30000}}
+	// AllProtocolChanges ...
+	AllProtocolChanges = &ChainConfig{big.NewInt(100), big.NewInt(0), big.NewInt(0), big.NewInt(0)}
 
 	// TestChainConfig ...
-	TestChainConfig = &ChainConfig{big.NewInt(1), big.NewInt(0), nil, false, big.NewInt(0), common.Hash{}, big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), big.NewInt(0), nil, new(EthashConfig), nil}
+	TestChainConfig = &ChainConfig{big.NewInt(99), big.NewInt(0), big.NewInt(0), big.NewInt(0)}
```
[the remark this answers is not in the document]
rlan35: Look at ethereum's original comment for this: 
// This configuration is intentionally not using keyed fields to force anyone	
// adding flags to the config to also have to set these fields.

I just added this comment back.

internal/params/config.go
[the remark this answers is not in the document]
harmony-ek: Then can we remove them?

internal/params/config.go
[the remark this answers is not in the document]
harmony-ek: Makes sense.

internal/params/config.go
rlan35: Let's leave this once vrf fork is ready

node/node_newblock.go
rlan35: ok, agreed. Reverted.

node/staking_test.go
rlan35: The staking code is basically outdated. We will need to revisit these code when implementing staking. This change just to make test pass.

node/staking_test.go
rlan35: yea, before, the finalization may fail, then the test will have nil pointer error.

harmony-ek: Makes sense, can you add a comment to explain that we are actually calling this accessor to rely upon its side effect?  The line looks very out of place.

node/worker/worker.go
rlan35: I think the serialization/deserailziation should be handled separately, rather than using the general serialization of the header. Just to have more flexibility and better control on the header data.

internal/params/config.go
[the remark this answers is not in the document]
rlan35: Done.

node/staking_test.go
rlan35: Done.

internal/params/config.go
```diff
@@ -170,136 +58,48 @@ type TrustedCheckpoint struct {
...
-	Period uint64 `json:"period"` // Number of seconds between blocks to enforce
-	Epoch  uint64 `json:"epoch"`  // Epoch length to reset votes and checkpoint
-}
+	CrossLinkEpoch *big.Int `json:"homesteadBlock,omitempty"`
 
-// String implements the stringer interface, returning the consensus engine details.
-func (c *CliqueConfig) String() string {
-	return "clique"
+	EIP155Epoch *big.Int `json:"eip155Block,omitempty"` // EIP155 hard fork epoch (include EIP158 too)
+	S3Epoch     *big.Int `json:"s3Block,omitempty"`     // S3 epoch is the first epoch containing S3 mainnet and all ethereum update up to Constantinople
```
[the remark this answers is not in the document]
rlan35: Done.

rlan35:
> General comment: Instead of comparing a block number against “first cross-link block number”, we should compare the epoch of the block against the “first cross-link epoch.”
> 
> (In general, getting the first/last block numbers of an epoch is not a function of a sharding schedule alone. The actual chain history determines when an epoch has ended on a shard, especially in the case of non-beacon shards.)
> 
> Also a minor nit below.

Added todo for this.

node/worker/worker.go
harmony-ek: I see.  Flexibility and control may be better achieved by implementing `rlp.Encoder` and `rlp.Decoder` interface IMO but this PR is not the right place to do that anyway.

rlan35:
Jenkins, test this please.

harmony-ek:
All 4 leaders stuck in block syncing.  Jenkins, test this please.

harmony-ek:
\*sigh\* Jenkins, test this please…

rlan35:
Jenkins, test this please

rlan35:
Jenkins, test this please

harmony-ek:
Jenkins, test this please.

rlan35:
Jenkins, test this please.

node/node_handler.go
```diff
@@ -791,6 +791,9 @@ func (node *Node) bootstrapConsensus() {
 					Msg("No peers, continue")
 				continue
 			}
+			utils.Logger().Info().
+				Int("numPeersNow", numPeersNow).
+				Msg("No peers, continue")
```
harmony-ek: 1. Print this only if `numPeersNow` has increased from the last tick, so as to prevent infinite loop of log messages if peer discovery fails.
2. The message is misleading.  How about moving this log line after the `MinPeers` check below, and saying "still waiting for more peers"?
3. It would help to log both `numPeersNow` and `minPeers` so as to make the target number explicit and aid comparison.

rlan35: Sorry. too rushed to add the change. Fixed.

PR #1498, opened 2019-09-03
Move types.Header into block package

harmony-ek:
This is a refactor-only change.  The new versioned block types will live in this package instead of the core package.

Unit tests still pass; local test reaches consensus.


harmony-ek:
Jenkins, test this please.

PR #1508, opened 2019-09-04
Move header fields into private struct; expose them using getters/setters

harmony-ek:
This is *mostly* a refactor.  It hides `block.Header` fields behind a private struct, and exposes the fields using field accessor methods (getters/setters).

Call sites using header field values have been modified to use getters (by simply appending the field selector with parentheses, compliant with Effective Go recommendation that [getters be named after the field they are exposing](https://golang.org/doc/effective_go.html#Getters), without a `Get` prefix).  Call sites *modifying* header field values directly have been retrofitted to use setters.  Some call sites that used to modify a header field by passing it by reference to another function (such as `GasUsed` calculation) now use a get-modify-set pattern.

The getters/setters protect the real header field value stored within the private struct by making a copy wherever needed.  This rule applies to `big.Int` pointers and `[]byte` slices.  This full protection also makes it possible to copy header by value without having to clone `big.Int` instances or `[]byte` slices.

Unit test still passes, and local test reaches consensus.

### DIFF
--- a/api/proto/node/node_test.go
+++ b/api/proto/node/node_test.go
@@ -60,14 +60,14 @@ func TestConstructBlocksSyncMessage(t *testing.T) {
 	statedb, _ := state.New(common.Hash{}, state.NewDatabase(db))
 
 	root := statedb.IntermediateRoot(false)
-	head := &block.Header{
-		Number:  new(big.Int).SetUint64(uint64(10000)),
-		Epoch:   big.NewInt(0),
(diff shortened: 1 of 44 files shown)


harmony-ek:
Jenkins failed; shards 0/1/3 failed to get past block syncing stage.  Shard 2 did achieve 25 consensuses.  Need to retry.  Jenkins, test this please.

PR #1523, opened 2019-09-06
[wallet] Add colored output in balance flag after doing transaction

fxfactorial:
Since checking balance is so commonly done after a transaction with wallet, I added a flag that will do a balance check after a transaction and show a colored highlight of the sender's address and shard and the receiver's address and shard. See picture which corresponds to 
Screen Shot 2019-09-05 at 7 30 46 PM

```
./bin/wallet -p local transfer --from one129r9pj3sk0re76f7zs3qz92rggmdgjhtwge62k --to one1d2rngmem4x2c6zxsjjz29dlah0jzkr0k2n88wc --amount 5.5 --shardID 1 --toShardID 0 --pass pass:'' --waitThenBal
```

### DIFF
--- a/cmd/client/wallet/main.go
+++ b/cmd/client/wallet/main.go
@@ -13,10 +13,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/harmony-one/harmony/internal/params"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
+	color "github.com/fatih/color"
 	ffi_bls "github.com/harmony-one/bls/ffi/go/bls"
 	"github.com/harmony-one/harmony/accounts"
 	"github.com/harmony-one/harmony/accounts/keystore"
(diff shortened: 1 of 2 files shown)


cmd/client/wallet/main.go
```diff
@@ -925,5 +945,13 @@ func submitTransaction(tx *types.Transaction, walletNode *node.Node, shardID uin
 	// FIXME (leo): how to we know the tx was successful sent to the network
 	// this is a hacky way to wait for sometime
 	time.Sleep(3 * time.Second)
```
rlan35: why wait for 3 seconds? the shown balance should have clear meaning whether it's before and after the transaction is applied. waiting for a few seconds will just make the result ambiguous. I think we just show the balance immediately, so it's always the balance before the transaction.

fxfactorial: I didn't add the 3 seconds wait logic but from your comment will add commit to remove it.

denniswon: This seems to have been added to wait for transactions to be committed in a block. Immediately calling get balance after submitting txn would get the balance before txn is actually processed.

LeoHChen:
please use your own fork to submit PR to the repo.

PR #1525, opened 2019-09-06
Wallet color

fxfactorial:
Cont of https://github.com/harmony-one/harmony/pull/1523

### DIFF
--- a/cmd/client/wallet/main.go
+++ b/cmd/client/wallet/main.go
@@ -13,10 +13,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/harmony-one/harmony/internal/params"
-
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


cmd/client/wallet/main.go
```diff
@@ -922,8 +942,13 @@ func submitTransaction(tx *types.Transaction, walletNode *node.Node, shardID uin
 		return err
 	}
 	fmt.Printf("Transaction Id for shard %d: %s\n", int(shardID), tx.Hash().Hex())
-	// FIXME (leo): how to we know the tx was successful sent to the network
-	// this is a hacky way to wait for sometime
-	time.Sleep(3 * time.Second)
```
LeoHChen: please add it back, otherwise, the send transaction may not always reach the network, due to libp2p's sending thread may not be done on time.

PR #1526, opened 2019-09-06
improve cross shard tx receipts verification design and logic

chaosma:
- add header and signature information when broadcasting cxreceipts
- avoid crosslink verification to speedup cross shard tx finalization by around 1 block time.

### DIFF
--- a/node/node_newblock.go
+++ b/node/node_newblock.go
@@ -195,7 +195,6 @@ func (node *Node) proposeLocalShardState(block *types.Block) {
 
 func (node *Node) proposeReceiptsProof() []*types.CXReceiptsProof {
 	validReceiptsList := []*types.CXReceiptsProof{}
-	pendingReceiptsList := []*types.CXReceiptsProof{}
 	node.pendingCXMutex.Lock()
 
 	sort.Slice(node.pendingCXReceipts, func(i, j int) bool {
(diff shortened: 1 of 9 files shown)


rlan35:
I think directly sending signature may not work in the case two shards change epoch at different time.

node/node_newblock.go
```diff
@@ -218,18 +217,13 @@ func (node *Node) proposeReceiptsProof() []*types.CXReceiptsProof {
...
-		if err := node.compareCrosslinkWithReceipts(cxp); err != nil {
-			utils.Logger().Debug().Err(err).Interface("cxp", cxp).Msg("[proposeReceiptsProof] CrossLink Verify Fail")
-			if err != ErrCrosslinkVerificationFail {
-				pendingReceiptsList = append(pendingReceiptsList, cxp)
-			}
-		} else {
-			utils.Logger().Debug().Interface("cxp", cxp).Msg("[proposeReceiptsProof] CXReceipts Added")
-			validReceiptsList = append(validReceiptsList, cxp)
-		}
+		utils.Logger().Debug().Interface("cxp", cxp).Msg("[proposeReceiptsProof] CXReceipts Added")
```
rlan35: Why don't we do "isvalidCX..." check in proposal? I think it's better to check here.

node/node_newblock.go
```diff
@@ -218,18 +217,13 @@ func (node *Node) proposeReceiptsProof() []*types.CXReceiptsProof {
...
+		validReceiptsList = append(validReceiptsList, cxp)
```
rlan35: We should have a upper limit on the number of receipts in one block.

rlan35: on the source: 2000 and here: 6000 limit.

node/node_newblock.go
chaosma: move it in proposal

node/node_newblock.go
chaosma: add upperlimit on incoming receipts as well as outgoing receipts

chaosma:
Jenkins, shut up! Test this, please!

PR #1536, opened 2019-09-07
Let 30% of the validator broadcast cx-receipts

rlan35:
desides, delay the cross-link function for later, and print consensus message on stdout.

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -731,7 +731,7 @@ func (consensus *Consensus) finalizeCommits() {
 	consensus.getLogger().Info().Int("NumCommits", len(consensus.commitSigs)).Msg("[Finalizing] Finalizing Block")
 
 	beforeCatchupNum := consensus.blockNum
-	beforeCatchupViewID := consensus.viewID
+	//beforeCatchupViewID := consensus.viewID
 
 	// Construct committed message
 	msgToSend, aggSig := consensus.constructCommittedMessage()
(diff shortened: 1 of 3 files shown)


rlan35:
Jenkins, test this please.

consensus/consensus_v2.go
```diff
@@ -792,11 +792,13 @@ func (consensus *Consensus) finalizeCommits() {
...
 	consensus.getLogger().Info().
-		Uint64("blockNum", beforeCatchupNum).
-		Uint64("ViewId", beforeCatchupViewID).
+		Uint64("blockNum", block.NumberU64()).
+		Uint64("ViewId", block.Header().ViewID().Uint64()).
 		Str("blockHash", block.Hash().String()).
 		Int("index", consensus.getIndexOfPubKey(consensus.PubKey)).
 		Msg("HOORAY!!!!!!! CONSENSUS REACHED!!!!!!!")
+	// Print to normal log too
+	utils.GetLogInstance().Info("HOORAY!!!!!!! CONSENSUS REACHED!!!!!!!", "BlockNum", block.NumberU64())
```
LeoHChen: why? we are phasing out the original logs.  will be no more original logs.

rlan35: We still have those “trie commit to database” stuff on the terminal, which is confusing people. It’s etheruem code so hard to remove. Remember a lots of people asking whether I joined consensus or not, right? . We might as well surface up consensus signal just for people to easily see.

rlan35: Just merged the code so other part get in master for release. Let me know if the log code has to be reverted.

PR #1420, opened 2019-08-24
fix new node state sync by correct block header verification

denniswon:
## Issue

https://github.com/harmony-one/harmony/issues/1313

#### Test Coverage Data

dennis.won@jongs-mbp:~/harmony/node (state_syncing) $ go test -cover
PASS
coverage: 15.4% of statements
ok  	github.com/harmony-one/harmony/node	4.416s

#### Test/Run Logs

Consensus done correctly and see the blockchain progressing.

PR #1539, opened 2019-09-09
correctly implement chain reader VerifyHeaders functions

denniswon:
Used in https://github.com/harmony-one/harmony/pull/1420/files


denniswon:
I did basic consensus, wallet balances/txn functionalities, etc. with this change. But I am not still 100% clear on how to "fully test" the safety of this change consensus, staking (my understanding is this is not in place yet), etc. that could be affected.

Please advise on what would be most confirming testing steps you guys would take when making changes to this core functionalities.

cc: @harmony-one/core

core/block_validator.go
```diff
@@ -113,70 +111,25 @@ func (v *BlockValidator) ValidateState(block, parent *types.Block, statedb *stat
 	return nil
 }
 
-// VerifyHeaderWithSignature verifies the header with corresponding commit sigs
-func VerifyHeaderWithSignature(header *block.Header, commitSig []byte, commitBitmap []byte) error {
```
chaosma: The logic of VerifyHeaderWithSignature is similar to VerifySeal. But the cross shard receipts verification requires signatures and bitmap from different source. If you remove it, it will break cross shard tx.

api/service/syncing/syncing.go
```diff
@@ -540,7 +540,7 @@ func (ss *StateSync) updateBlockAndStatus(block *types.Block, bc *core.BlockChai
 	// Verify block signatures
 	// TODO chao: only when block is verified against last commit sigs, we can update the block and status
 	if block.NumberU64() > 1 {
-		err := core.VerifyBlockLastCommitSigs(bc, block.Header())
```
chaosma: Previously, we use most common block hashes to determine which block to download, then we use that blocks to try to update the status. Now from my understanding, here we are improving the logic by verify the block first before put it into the queue.  If that is the case, VerifyHeader call needs cooperate with the retry download other blocks logic before updateBlockAndStatus.

PR #1549, opened 2019-09-10
prevent cross shard tx before V0 Epoch

chaosma:
In localnet config. The cross shard tx will be blocked at epoch=0. After epoch >= 1, the cross shard tx will be accepted. To test it ,just run the wallet command before and after epoch transition.


harmony-ek:
Jenkins, test this please.

PR #1550, opened 2019-09-10
Move CrossLinks to v2

harmony-ek:
## Issue

Introduce `CrossTxEpoch` before `CrossLinkEpoch`, then introduce cross-shard transaction fields in that epoch (`BodyV1`, `extblockV1`, and `HeaderV1`).

Introduce `CrossLinks` field only in `HeaderV2`.

Enable cross-shard transaction in `CrossTxEpoch`+1, in order to allow all 4 shards to roll into `CrossTxEpoch` and become ready to handle ingress receipts from other shards.

Also fix two bugs:

* EIP155 and S3 epochs were reversed in log output.
* `(*ChainConfig).IsS3` was using `CrossLinkEpoch` and not `S3Epoch`.

And for performance, avoid fetching entire block to use just its header, and instead, fetch only the header directly.  (This change was meant to go in the previous PR but got omitted.)

## Test

Local test passes.  Cross-shard transactions are rejected in epoch 0 (< `CrossTxEpoch`) and 1 (= `CrossTxEpoch`), and accepted from 2 (`CrossTxEpoch+1`) and on.


internal/params/config.go
```diff
@@ -19,20 +20,33 @@ var (
...
+		big.NewInt(0),   // EIP155Epoch
+		big.NewInt(0),   // S3Epoch
+	}
 
 	// TestChainConfig ...
 	// This configuration is intentionally not using keyed fields to force anyone
 	// adding flags to the config to also have to set these fields.
-	TestChainConfig = &ChainConfig{big.NewInt(99), big.NewInt(0), big.NewInt(0), big.NewInt(0)}
+	TestChainConfig = &ChainConfig{
+		big.NewInt(99), // ChainID
```
denniswon: declaring as constants would look better, like in mainnet/testnet/localnet.go, etc.

block/v2/header.go
```diff
@@ -0,0 +1,417 @@
...
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/rs/zerolog"
+
+	blockif "github.com/harmony-one/harmony/block/interface"
+	"github.com/harmony-one/harmony/crypto/hash"
+	"github.com/harmony-one/harmony/shard"
+)
+
+// Header is the V2 block header.
+type Header struct {
```
denniswon: can we use git cp (instead of cp then git add, which seems to have been used here) so that the change history log is maintained and we can see the changes made between v2/header.go vs. v1/header.go?

internal/params/config.go
harmony-ek: #1551

block/v2/header.go
harmony-ek: Actually:

```
$ git cp
git: 'cp' is not a git command. See 'git --help'.

The most similar command is
        fap
$ 
```

Unlike Subversion and other VCS, Git does not track renames/moves/copies as first-class metadata.  The `git mv` command is simply a convenience shortcut for a `mv`–`git rm`–`git add` sequence, and the “renames” shown in `git log`, `git show`, and `git status` are actually detected by heuristics, which can be turned off:

```
$ git mv CONTRIBUTING.md CONTRIBUTING.mdx
$ git status
On branch master
Your branch is up to date with 'harmony-one/master'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        renamed:    CONTRIBUTING.md -> CONTRIBUTING.mdx

$ git status --no-renames # turn off rename detection
On branch master
Your branch is up to date with 'harmony-one/master'.

Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

        deleted:    CONTRIBUTING.md
        new file:   CONTRIBUTING.mdx

$ 
```

PR #1554, opened 2019-09-10
[Pangaea] Change to functional DNS server

xChickens:
I messed up p.hmny.io and I don't want to use pga.hmny.io as there are old nodes connecting to it. Using n as it's close to p and is the second consonant in Pa**n**gaea.


charleszqwang:
test this please

Issue #1556, opened 2019-09-10
Mainnet V1 - Migration Checklist


john-harmony:
# Mainnet V1 - Release Checklist

Following are some tests to be completed for Mainnet V1

**Code Complete**
- [x]  [HTTPs support](https://github.com/harmony-one/harmony/blob/master/internal/configs/sharding/testnet.go#L143)
   - [x] hmy_getShardingStructure - Reflect new https DNS entries e.g.  https://api.s1.b.hmny.io/
   - [x] Websocket and RPC/API support for HTTPS
- [ ] Default Shard ID - @mikedoan | Harmony 
  - [ ] Return the default shard ID to SDK when calling hmy_getShardingStructure

- [x] [Explorer Issue](https://github.com/harmony-one/harmony/issues/1555) - not parsing blocks greater than 77

**Merge any outstanding Pull Requests**
 - [x] [Header Validation](https://github.com/harmony-one/harmony/pull/1539)

**Betanet Upgrade**
- [x] Decision on  whether to test S3 upgrade path @LeoHChen @harmony-ek @rlan35 
- [x] Upgrade Betanet to Master - @charleszqwang @AndyBoWu 
- [x]  Successful transaction testing - @chaosma @fxfactorial @johnwhitton 
- [ ]  MathWallet Validation - @johnwhitton 
- [ ] Explorer Validation - @flicker-harmony 
- [x]  RPC/API Testing - @mikedoan @john-harmony 
- [ ] Negative use case tests - (Transactions sent under adverse conditions) @chaosma @denniswon @harmony-ek 
  - [ ]  Beacon chain not synching
  - [ ] Conensus on 1 shard only
  - [ ] Test when one or more nodes are down for DNS entry
- [x] Transactions Throttling Tests @denniswon 
- [ ] Stress Testing @fxfactorial @denniswon @chaosma 
  - [ ] Send Number of txns to test txns per block and pooling (hmy_cli)
- [ ] Cross Shard
  - [ ] Test recovery process
  - [x] Test sending transactions from one shard to another 

- [ ] **Yes or No Decision on Mainnet V1 Launch**

- [ ] **Mainnet Upgrade and Validation**

john-harmony:
**From @denniswon**
 Transactions Throttling Tests
* Stress Testing
  * Send Number of txns to test txns per block and pooling (hmy_cli)
* Cross Shard
  * Test recovery process
  * Test sending transactions from one shard to another 

Conducted yesterday: 
[...]

PR #1557, opened 2019-09-10
Change endpoints to support for https and wss

john-harmony:
## Issue

Support https and wss as part of the mainnet v1 checklist https://github.com/harmony-one/harmony/issues/1556
This is needed for the SDK which is being used by math wallet.
As such math wallet cross shard transactions should be tested to verify this (from shard0 to shard 1)

#### Test Coverage Data

This needs to be tested against betanet, pangea and mainnet.
math wallet cross shard transactions should be tested to verify this (from shard0 to shard 1)

* Before
* After

## TODO
Merge and validate against betanet, pangaea and mainnet


LeoHChen:
Jenkins, test this please.

PR #1565, opened 2019-09-11
update epoch for cross shards txn

rlan35:


denniswon:
what's the rationale behind this change?

PR #1572, opened 2019-09-11
Add cross-shard tx backward compatibility

rlan35:


denniswon:
func (v *BlockValidator) ValidateCXReceiptsProof(cxp *types.CXReceiptsProof) error {

and 

func (cxp *CXReceiptsProof) GetToShardID() (uint32, error) {

only other places

But these don't seem to require the same change..

PR #1575, opened 2019-09-11
Fix explorer node stuck issue

rlan35:
And update local shard config so the keys will change in both epoch 1 and 2.

Tested locally.


AndyBoWu:
Related to https://github.com/harmony-one/harmony-ops/issues/73

PR #1597, opened 2019-09-14
Add RPC call for CXReceipt

chaosma:
- Add FromShardID and ToShardID fields in getTransactionByHash call
- Add write and retrieve for CXReceiptByHash
- Add RPC call getCXReceiptByHash

### DIFF
--- a/core/rawdb/accessors_indexes.go
+++ b/core/rawdb/accessors_indexes.go
@@ -123,3 +123,76 @@ func WriteBloomBits(db DatabaseWriter, bit uint, section uint64, head common.Has
 		utils.Logger().Error().Err(err).Msg("Failed to store bloom bits")
 	}
 }
+
+// ReadCxLookupEntry retrieves the positional metadata associated with a transaction
+// hash to allow retrieving the transaction or receipt by hash.
+// return nil if not found
+func ReadCxLookupEntry(db DatabaseReader, hash common.Hash) (common.Hash, uint64, uint64, uint64) {
+	data, _ := db.Get(cxLookupKey(hash))
+	if len(data) == 0 {
+		return common.Hash{}, 0, 0, 0
+	}
+	var entry TxLookupEntry
+	if err := rlp.DecodeBytes(data, &entry); err != nil {
+		utils.Logger().Error().Err(err).Str("hash", hash.Hex()).Msg("Invalid transaction lookup entry RLP")
+		return common.Hash{}, 0, 0, 0
+	}
+	return entry.BlockHash, entry.BlockIndex, entry.Index, entry.SubIndex
+}
+
+// WriteCxLookupEntries stores a positional metadata for every transaction from
+// a block, enabling hash based transaction and receipt lookups.
+func WriteCxLookupEntries(db DatabaseWriter, block *types.Block) {
+	for i, cxp := range block.IncomingReceipts() {
+		for j, cx := range cxp.Receipts {
+			entry := TxLookupEntry{
+				BlockHash:  block.Hash(),
+				BlockIndex: block.NumberU64(),
+				Index:      uint64(i),
+				SubIndex:   uint64(j),
+			}
+			data, err := rlp.EncodeToBytes(entry)
+			if err != nil {
+				utils.Logger().Error().Err(err).Msg("Failed to encode transaction lookup entry")
+			}
+			if err := db.Put(cxLookupKey(cx.TxHash), data); err != nil {
+				utils.Logger().Error().Err(err).Msg("Failed to store transaction lookup entry")
+			}
+		}
+	}
+}
+
+// DeleteCxLookupEntry removes all transaction data associated with a hash.
+func DeleteCxLookupEntry(db DatabaseDeleter, hash common.Hash) {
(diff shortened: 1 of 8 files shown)


core/rawdb/accessors_indexes.go
```diff
@@ -123,3 +123,76 @@ func WriteBloomBits(db DatabaseWriter, bit uint, section uint64, head common.Has
 		utils.Logger().Error().Err(err).Msg("Failed to store bloom bits")
 	}
 }
+
+// ReadCxLookupEntry retrieves the positional metadata associated with a transaction
+// hash to allow retrieving the transaction or receipt by hash.
```
rlan35: This only concern the cx receipt, not the original transaction, right? comment needs to be updated.

core/rawdb/accessors_indexes.go
```diff
@@ -123,3 +123,76 @@ func WriteBloomBits(db DatabaseWriter, bit uint, section uint64, head common.Has
...
+}
+
+// WriteCxLookupEntries stores a positional metadata for every transaction from
+// a block, enabling hash based transaction and receipt lookups.
+func WriteCxLookupEntries(db DatabaseWriter, block *types.Block) {
+	for i, cxp := range block.IncomingReceipts() {
+		for j, cx := range cxp.Receipts {
+			entry := TxLookupEntry{
+				BlockHash:  block.Hash(),
+				BlockIndex: block.NumberU64(),
```
rlan35: BlockNumber , not BlockIndex.

core/rawdb/schema.go
```diff
@@ -93,6 +94,7 @@ type TxLookupEntry struct {
 	BlockHash  common.Hash
 	BlockIndex uint64
 	Index      uint64
+	SubIndex   uint64 // used for lookup CXReceipt from CXReceiptsProof
```
rlan35: We shouldn't add additional field. The old Index alone is enough to pinpoint the position of the receipt in the 2d array.

internal/hmyapi/types.go
```diff
@@ -38,17 +70,19 @@ func newRPCTransaction(tx *types.Transaction, blockHash common.Hash, blockNumber
...
-		S:        (*hexutil.Big)(s),
+		From:        from,
+		Gas:         hexutil.Uint64(tx.Gas()),
+		GasPrice:    (*hexutil.Big)(tx.GasPrice()),
+		Hash:        tx.Hash(),
+		Input:       hexutil.Bytes(tx.Data()),
+		Nonce:       hexutil.Uint64(tx.Nonce()),
+		To:          tx.To(),
+		Value:       (*hexutil.Big)(tx.Value()),
+		FromShardID: tx.ShardID(),
```
rlan35: Need to check with John and Minh to see whether this break existing clients.

internal/hmyapi/types.go
```diff
@@ -22,11 +22,43 @@ type RPCTransaction struct {
 	To               *common.Address `json:"to"`
 	TransactionIndex hexutil.Uint    `json:"transactionIndex"`
 	Value            *hexutil.Big    `json:"value"`
+	FromShardID      uint32          `json:"fromShardID"`
```
rlan35: Keep it consistent with Transaction field "ShardID"

core/rawdb/accessors_indexes.go
chaosma: Here I used the TxLookupEntry defined by ethereum, but I added one more field here.

internal/hmyapi/types.go
chaosma: Would that be confusing to the user?

internal/hmyapi/types.go
chaosma: I run the curl request, it works fine

core/rawdb/schema.go
chaosma: we cannot use 2d array here. since each CXReceipts (array of CXReceipt) might have different length. so we need use 2d array to find location.

core/rawdb/accessors_indexes.go
chaosma: Will add more comment here.

core/rawdb/accessors_indexes.go
chaosma: TxLookupEntry is used in several places e.g. by getTransactionByHash

core/rawdb/schema.go
rlan35: What I mean is the index can be for the flattened receipts . Doesn’t matter what different length

internal/hmyapi/types.go
rlan35: We should be consistent . It will be more confusing if not consistent

core/rawdb/accessors_indexes.go
rlan35: Ok

internal/hmyapi/types.go
chaosma: OK

core/rawdb/schema.go
chaosma: Okay, let use one index to save some space.

PR #1611, opened 2019-09-17
[RPC] Expose node metadata

fxfactorial:
Add node meta data RPC so that we can build monitoring services. 

Tested with:

```
test/debug.sh
curl -X POST -H 'content-type:application/json' localhost:9500  \
     -d  '{"id":"1","jsonrpc":"2.0","method":"hmy_getNodeMetadata","params":[]}' | json_pp
{
   "id" : "1",
   "jsonrpc" : "2.0",
   "result" : {
      "network" : "",
      "version" : "Harmony (C) 2019. harmony, version - ( )",
      "blskey" : "65f55eb3052f9e9f632b2923be594ba77c55543f5c58ee1454b9cfd658d25e06373b0f7d42a19c84768139ea294f6204"
   }
}
```

Note:

1. network is empty, this is to be investigated as possible bug in existing code
2. version is empty, this is because of how local build of harmony binary proceeds. 
3. blskey is the public key used in consensus

### DIFF
--- a/internal/hmyapi/README.md
+++ b/internal/hmyapi/README.md
@@ -10,6 +10,7 @@
 * [x] hmy_protocolVersion - check protocol version
 * [ ] net_version - get network id
 * [ ] net_peerCount - peer count
+* [x] hmy_getNodeMetadata - get node's version, bls key
 
 ### BlockChain info related
 * [ ] hmy_gasPrice - return min-gas-price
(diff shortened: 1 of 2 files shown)


fxfactorial:
@chaosma please review

LeoHChen:
please fix the network is empty bug in another PR.

PR #1612, opened 2019-09-17
Add rpc latestleader information call

chaosma:


internal/hmyapi/types.go
```diff
@@ -41,6 +44,35 @@ type RPCCXReceipt struct {
...
+// HeaderInformation represents the latest consensus information
+type HeaderInformation struct {
+	BlockHash   common.Hash `json:"blockHash"`
+	BlockNumber uint64      `json:"blockNumber"`
+	ShardID     uint32      `json:"shardID"`
+	Leader      string      `json:"leader"`
+	ViewID      uint64      `json:"viewID"`
+	Epoch       uint64      `json:"epoch"`
+	Timestamp   string      `json:"timestamp"`
+	UnixTime    uint64      `json:"unixtime"`
```
rlan35: also return last sig and bitmap?

internal/hmyapi/blockchain.go
```diff
@@ -225,3 +225,9 @@ func doCall(ctx context.Context, b Backend, args CallArgs, blockNr rpc.BlockNumb
 	}
 	return res, gas, failed, err
 }
+
+// LatestHeader returns the latest header information
+func (s *PublicBlockChainAPI) LatestHeader(ctx context.Context) *HeaderInformation {
+	header, _ := s.b.HeaderByNumber(context.Background(), rpc.LatestBlockNumber) // latest header should always be available
```
rlan35: what is this rpc.LatestBlockNumber? why don't we just get the current block and return the header? as the func is "LatestHeader"

internal/hmyapi/types.go
chaosma: since it's the latest block, we don't have the corresponding sig and bitmap. if directly read from ReadLastCommits, it might not be synchronized in some cases

internal/hmyapi/blockchain.go
chaosma: since we have a rpc call to get latest block number:  "blockNumber", it used this function to get latest blocknumber, this is where I borrow the code.

internal/hmyapi/types.go
rlan35: but we should just return the sig as is, right? Just return the sig in this header itself.

chaosma: yes, that should be easier.

PR #1615, opened 2019-09-18
[node.sh] uses separate options to download harmony db

LeoHChen:

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 unset -v progname
 progname="${0##*/}"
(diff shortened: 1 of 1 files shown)


nyetwurk:
https is the worst possible way to download a large file, No way to do it incrementally. Rsync would make much much more sense.

nyetwurk:
also, recommend at least .gz so there is a bit of a wrapper around the tar file. Sums are a good backup, but bare tar files (especially large ones) are not suitable for passing around.

fxfactorial:
@nyetwurk `http` spec has multi-part upload, not sure if curl is smart enough to use that

nyetwurk:
@fxfactorial Either way, you're best off using rsync with untarred multiple files so it can incrementally download them. Also, if the snapshot gets updated, you can rsync from a newer snapshot and rsync will only download the changed files. That can't be done with http.

fxfactorial:
@nyetwurk I understand your PoV (and am well aware of the benefits of rsync) but assuming rsync exists on user machine is another assumption, plus assuming same behavior between all the possible versions/impl of rsync.

PR #1618, opened 2019-09-18
Fix balance and add cross-shard txs

flicker-harmony:
## Issue

Balance was incorrect, because it didn't correctly consider cross-shard txs. Added shard fields (FromShard, ToShard) to explorer transacrtion. Txs on frontend should be shown separately for each shard.

## Test

Postman, localnet testing.

#### Test Coverage Data

* Before
* After

### DIFF
--- a/api/service/explorer/service.go
+++ b/api/service/explorer/service.go
@@ -695,7 +695,6 @@ func (s *Service) GetExplorerAddress(w http.ResponseWriter, r *http.Request) {
 		balance, err := s.GetAccountBalance(address)
 		if err == nil {
 			balanceAddr = balance
-			data.Address.Balance = balance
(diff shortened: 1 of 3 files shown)


api/service/explorer/service.go
```diff
@@ -695,7 +695,6 @@ func (s *Service) GetExplorerAddress(w http.ResponseWriter, r *http.Request) {
 		balance, err := s.GetAccountBalance(address)
 		if err == nil {
 			balanceAddr = balance
-			data.Address.Balance = balance
```
rlan35: why remove this line?

rlan35:
Please return a copy of the big int balance from node.GetBalanceOfAddress

PR #1619, opened 2019-09-19
Cleanup state sync logic; fix explorer node sync wrong committee issue

rlan35:
Worker is not needed for state sync.
ShardID is not needed in worker.
Explorer node needs to update consensus info once sync'ed.

Tested locally on both normal and explorer node.

### DIFF
--- a/node/node_syncing.go
+++ b/node/node_syncing.go
@@ -190,7 +190,7 @@ func (node *Node) DoBeaconSyncing() {
 				continue
 			}
 		}
-		node.beaconSync.SyncLoop(node.Beaconchain(), node.BeaconWorker, false, true)
+		node.beaconSync.SyncLoop(node.Beaconchain(), node.BeaconWorker, true)
 		time.Sleep(BeaconSyncFrequency * time.Second)
 	}
 }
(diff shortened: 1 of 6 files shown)


node/node_syncing.go
```diff
@@ -230,7 +230,10 @@ SyncingLoop:
 			if willJoinConsensus {
 				node.Consensus.BlocksNotSynchronized()
 			}
-			node.stateSync.SyncLoop(bc, worker, willJoinConsensus, false)
+			node.stateSync.SyncLoop(bc, worker, false)
+			if node.NodeConfig.Role() == nodeconfig.ExplorerNode {
```
denniswon: NodeConfig.Role() == nodeconfig.ExplorerNode means willJoinConsensus == false means node. Currently, node being explorer is the only case willJoinConsensus == false.

It would be better if we use 'if (will_join_consensus}) { .sync state } else { .updateConsensusInformation }

denniswon: @chaosma @rlan35 could you correct me if there is misunderstanding on what I stated above? :)

PR #1622, opened 2019-09-19
hmyapi update

flicker-harmony:
## Issue

Add GetCommittee method similar to explorer.

## Test

Localnet, postman.
Screenshot 2019-09-19 at 16 10 07

#### Test Coverage Data

* Before
* After

## TODO

By the way, I guess we want to present balance in 10-digit format, not hex.

### DIFF
--- a/internal/hmyapi/blockchain.go
+++ b/internal/hmyapi/blockchain.go
@@ -138,11 +138,7 @@ func (s *PublicBlockChainAPI) GetStorageAt(ctx context.Context, addr string, key
 func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address string, blockNr rpc.BlockNumber) (*hexutil.Big, error) {
 	// TODO: currently only get latest balance. Will add complete logic later.
 	addr := internal_common.ParseAddr(address)
-	balance, err := s.b.GetBalance(addr)
-	if err != nil {
-		return nil, err
-	}
-	return balance, nil
+	return s.b.GetBalance(addr)
 }
 
 // BlockNumber returns the block number of the chain head.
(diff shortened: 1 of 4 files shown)


internal/hmyapi/blockchain.go
```diff
@@ -63,6 +63,36 @@ func (s *PublicBlockChainAPI) GetBlockByHash(ctx context.Context, blockHash comm
...
+func (s *PublicBlockChainAPI) GetCommittee(ctx context.Context, epoch int64) (map[string]interface{}, error) {
+	committee, err := s.b.GetCommittee(big.NewInt(epoch))
+	if err != nil {
+		return nil, err
+	}
+	validators := make([]map[string]interface{}, 0)
+	for _, validator := range committee.NodeList {
+		validatorBalance := new(hexutil.Big)
+		validatorBalance, err = s.b.GetBalance(validator.EcdsaAddress)
+		if err != nil {
```
rlan35: Committee should be a whole. We can't miss a single validator if there is error and return the rest. If any error for a single validator, we should just return error for the api.

internal/hmyapi/blockchain.go
```diff
@@ -108,7 +138,11 @@ func (s *PublicBlockChainAPI) GetStorageAt(ctx context.Context, addr string, key
 func (s *PublicBlockChainAPI) GetBalance(ctx context.Context, address string, blockNr rpc.BlockNumber) (*hexutil.Big, error) {
 	// TODO: currently only get latest balance. Will add complete logic later.
 	addr := internal_common.ParseAddr(address)
-	return s.b.GetBalance(addr)
+	balance, err := s.b.GetBalance(addr)
+	if err != nil {
+		return nil, err
```
rlan35: why this change?

flicker-harmony: Oh, I accidentally added that.

PR #1625, opened 2019-09-19
Unify json tags for IP

nyetwurk:
"ip" is already used for my local ip in other places

"IP" for the peer ip is just confusing, as are any json tags that have spaces.


harmony-ek:

PR #1626, opened 2019-09-19
Verify sig every 100 blocks during state sync

rlan35:
Also, revert some of the explorer address querying code to still support txn list query.

Tested locally.

### DIFF
--- a/api/service/syncing/syncing.go
+++ b/api/service/syncing/syncing.go
@@ -546,9 +546,18 @@ func (ss *StateSync) updateBlockAndStatus(block *types.Block, bc *core.BlockChai
 	// Verify block signatures
 	// TODO chao: only when block is verified against last commit sigs, we can update the block and status
 	if block.NumberU64() > 1 {
-		err := bc.Engine().VerifyHeader(bc, block.Header(), true)
+		verifySig := false
+		if block.NumberU64()%100 == 0 {
+			// Verify signature every 100 blocks
+			verifySig = true
+		}
+		err := bc.Engine().VerifyHeader(bc, block.Header(), verifySig)
 		if err != nil {
 			utils.Logger().Error().Err(err).Msgf("[SYNC] failed verifying signatures for new block %d", block.NumberU64())
+			utils.Logger().Debug().Interface("block", bc.CurrentBlock()).Msg("[SYNC] Rolling back last 99 blocks!")
+			for i := 0; i < 99; i++ {
+				bc.Rollback([]common.Hash{bc.CurrentBlock().Hash()})
+			}
 			return false
 		}
 	}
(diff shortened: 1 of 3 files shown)


api/service/syncing/syncing.go
```diff
@@ -546,9 +546,18 @@ func (ss *StateSync) updateBlockAndStatus(block *types.Block, bc *core.BlockChai
 	// Verify block signatures
 	// TODO chao: only when block is verified against last commit sigs, we can update the block and status
 	if block.NumberU64() > 1 {
-		err := bc.Engine().VerifyHeader(bc, block.Header(), true)
+		verifySig := false
```
denniswon: verifySig := block.NumberU64()%100 == 0

api/service/syncing/syncing.go
```diff
@@ -546,9 +546,18 @@ func (ss *StateSync) updateBlockAndStatus(block *types.Block, bc *core.BlockChai
...
+		if block.NumberU64()%100 == 0 {
+			// Verify signature every 100 blocks
+			verifySig = true
+		}
+		err := bc.Engine().VerifyHeader(bc, block.Header(), verifySig)
 		if err != nil {
 			utils.Logger().Error().Err(err).Msgf("[SYNC] failed verifying signatures for new block %d", block.NumberU64())
+			utils.Logger().Debug().Interface("block", bc.CurrentBlock()).Msg("[SYNC] Rolling back last 99 blocks!")
```
denniswon: using Interface it will print out logging error saying some fields are not json encodable (like chan byte[])
would be better to just use primitive type fields logging.

denniswon:
what about the last < 100 blocks? I guess with our current block height, 100 blocks segment size should be okay to just assume verification per 100 blocks is good enough?

Wondering what was tested other than consensus running to verify the correctness/security.

api/service/syncing/syncing.go
rlan35: Done

api/service/syncing/syncing.go
rlan35: I don't see other primitive field to use for Block object. Any suggestion?

rlan35:
> what about the last < 100 blocks? I guess with our current block height, 100 blocks segment size should be okay to just assume verification per 100 blocks is good enough?
> 
> Wondering what was tested other than consensus running to verify the correctness/security.

Tested state sync still works. This is a temporary workaround before the real fast state sync happens. The check every 100 blocks helps with validator sync'ing to the wrong chain. (note that the parentHash is still verified, this is very similar to the signature verification every epoch with jump link, as described in the whitepaper)

rlan35:
Jenkins, test this please.

PR #1627, opened 2019-09-20
Give Pangaea its own chain config

harmony-ek:
## Issue

Pangaea used to share chain config and feature upgrade schedule with Betanet (“Testnet”), but it should have its own schedule because it is to become a long-running network just like Betanet is.  Give Pangaea its own upgrade schedule.

## Test

UTs and local tests still pass.  Pangaea settings to be tested on the Pangaea network itself.


cmd/client/wallet/main.go
```diff
@@ -767,11 +767,12 @@ func processTransferCommand() {
...
-	chainConfig := params.MainnetChainConfig
-	if walletProfile.Profile != defaultProfile {
-		chainConfig = params.TestnetChainConfig
+	chainID, ok := new(big.Int).SetString(walletProfile.ChainID, 0)
+	if !ok {
+		fmt.Printf("invalid chain ID %#v in config", walletProfile.ChainID)
+		return
 	}
-	tx, err = ks.SignTx(account, tx, chainConfig.ChainID)
+	tx, err = ks.SignTx(account, tx, chainID)
```
rlan35: Where is the change to wallet profile file to add the chainid ?

cmd/client/wallet/main.go
```diff
@@ -767,11 +767,12 @@ func processTransferCommand() {
 
 	fmt.Printf("Unlock account succeeded! '%v'\n", senderPass)
 
-	chainConfig := params.MainnetChainConfig
-	if walletProfile.Profile != defaultProfile {
-		chainConfig = params.TestnetChainConfig
+	chainID, ok := new(big.Int).SetString(walletProfile.ChainID, 0)
```
rlan35: Can we not add a chain Id field in the profile and just infer it from the network name?

cmd/client/wallet/main.go
astralblue: If a chain ID is absent, the profile reader infers the chain ID from the profile name: 1 for `main` and `default`, 3 for `pangaea`, and 2 for everything else.

cmd/client/wallet/main.go
astralblue: No.  `.ini` section names are not supposed to be special.  People wanting to rename their sections should be free to do so, e.g. `pangaea` → `p`.  The ability to specify chain ID grants that freedom.

LeoHChen:
Don't you need to update the default wallet.ini and generated_wallet_ini.go file to add chainid field?

astralblue:
@LeoHChen the `ReadWalletProfile` function handles the no-`chain_id` case with a backward compatibility logic.  I will examine the embedding logic though.

harmony-ek:
@LeoHChen, I looked at the file embedding logic, and it seems that the file is just read and embedded verbatim, and parsing occurs at runtime.  So, since the default `wallet.ini` does not change, there is no reason to do anything special at the moment.

However, I found an issue with a name-backed backward compatibility logic: For example, if one changes `[main]` section in the default .ini to `[m]`, suddenly it will start using a different chain ID, surprising the user.  I'll add explicit `chain_id` to the default `wallet.ini` (and regenerate `generated_wallet_*.go`), and add a deprecation notice to the wallet code to tell the user to update their `wallet.ini`.

Issue #1630, opened 2019-09-21
Explorer search for transaction not working


rlan35:
this txn was submitted 0xcff4a3a6fd4eb34b9a7e48e3eca0c8c899de71a1ea21a126cc1048e65b332a72 to testnet and it's not searchable in testnet explorer: https://explorer.beta.harmony.one/#/

Reproduce it by sending another transaction and search for it in explorer.

john-harmony:
Here are two new transactions
- 0xfdfb4b514be1018a5a038d8f9c8e73d29a708c93ea2edf869893b92b91d227d7 (cross shard)
- 0xd73804146dbefb94edf5b54baaee8d9a57734b5c16ded05cbba3289435d40ea5 (same shard)

john-harmony:
Latest testnet transaction
https://explorer.beta.harmony.one/#/tx/0x1bb3cadb2031e5ce987b0c557abe134986b9834963dd0d54f75a9c7d966606ce

PR #1632, opened 2019-09-21
Negative balance fix for storage, ASC or DESC order for blocks and txs

flicker-harmony:
## Issue

This might fix https://github.com/harmony-one/harmony/issues/1630. Adds ASC or DESC order for blocks and txs on api requests (order param "ASC" or "DESC")

## Test

Tested locally with postman.

#### Test Coverage Data

* Before
* After

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -170,14 +170,9 @@ func (storage *Storage) UpdateAddressStorage(batch ethdb.Batch, addr string, exp
 
 	var address Address
 	if data, err := storage.db.Get([]byte(key)); err == nil {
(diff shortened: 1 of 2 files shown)


api/service/explorer/storage.go
```diff
@@ -172,11 +172,7 @@ func (storage *Storage) UpdateAddressStorage(batch ethdb.Batch, addr string, exp
 	if data, err := storage.db.Get([]byte(key)); err == nil {
 		err = rlp.DecodeBytes(data, &address)
 		if err == nil {
-			if explorerTransaction.Type == Received {
```
rlan35: why only add value now?

PR #1642, opened 2019-09-23
[transcationpool.go] fix bug on chain id and specify real chain id

mikedoan:
fix bug on chain id and specify real chain id


mikedoan:
Jenkins, test this please.

Issue #1633, opened 2019-09-21
Refactor and Consolidation of API's


john-harmony:
## Summary

Harmony currently has API's located in many areas. This ticket is to review and align Harmony's API's to best practices.

## Current Design
**API's are currently held in multiple locations including**
- [blockchain.go](https://github.com/harmony-one/harmony/blob/master/internal/hmyapi/blockchain.go)
- [api_backend.go](https://github.com/harmony-one/harmony/blob/master/hmy/api_backend.go)
- [explorer](https://github.com/harmony-one/harmony/blob/master/api/service/explorer/service.go)

**API's are referenced by** 
- Harmony Javascript SDK [here](https://github.com/harmony-one/sdk/blob/master/packages/harmony-network/src/rpcMethod/rpc.ts)
- Harmony CLI [here](https://github.com/harmony-one/go-sdk/blob/master/pkg/rpc/methods.go)

## Problems

- Code is fragmented
- Code is inconsistent
- Code is/was exposed on different ports

## Proposal

- Consolidate API's into functional/technical areas
- Ensure all API's accept one address format 
- Ensure all API's return address in one address format
- Ensure all API's are exposed on port 9500 (remove 5000)

Sample one address `one18t4yj4fuutj83uwqckkvxp9gfa0568uc48ggj7`

## Acceptance criteria

- [ ] Code has been consolidated
- [ ] API's have test included in [Postman](https://www.getpostman.com/collections/d3a30f1499c59dac4f84)
- [ ] API test suite is included in [Jenkins](https://github.com/harmony-one/jenkins)
- [ ] API's are [documented](https://docs.harmony.one/home/api-developers-guide/overview/methods)
- [ ] Testnet Migration and Validation complete
- [ ] Mainnet Migration and validation complete

Daniel-VDM:
As request by John, the following do not use the one1 address:

`hmy_getTransactionReceipt` has a 0x address in the 'from' and 'to' address of the result.
`hmy_getTransactionByHash` has a 0x address in the 'from' and 'to' address of the result.
`hmy_getTransactionByBlockNumberAndIndex` has a 0x address in the 'from' and 'to' address of the result.
`hmy_getTransactionByBlockHashAndIndex` has a 0x address in the 'from' and 'to' address of the result.
`hmy_getCSReceiptByHash` has a 0x address in the 'from' and 'to' address of the result.
`hmy_resendCx` has a 0x address in the 'from' and 'to' address of the result.

[...]

PR #1643, opened 2019-09-23
add hmy getTransactionsHistory

flicker-harmony:
## Issue

https://github.com/harmony-one/harmony/issues/1633

## Test

Screenshot 2019-09-24 at 00 07 29

#### Test Coverage Data

* Before
* After

## TODO

Make adjustments on explorer frontend. GetTransactionsHistory returns list of hashes, then query each tx by hash by GetTransactionByHash and its receipt by GetTransactionReceipt and show on explorer. Such design makes all methods atomic.

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -207,6 +207,12 @@ func (b *APIBackend) GetBalance(address common.Address) (*hexutil.Big, error) {
 	return (*hexutil.Big)(balance), err
 }
 
+// GetTransactionsHistory returns list of transactions hashes of address.
+func (b *APIBackend) GetTransactionsHistory(address string) ([]common.Hash, error) {
+	hashes, err := b.hmy.nodeAPI.GetTransactionsHistory(address)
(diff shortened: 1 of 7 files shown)


hmy/api_backend.go
```diff
@@ -207,6 +207,12 @@ func (b *APIBackend) GetBalance(address common.Address) (*hexutil.Big, error) {
 	return (*hexutil.Big)(balance), err
 }
 
+// GetTransactionsHistory returns list of transactions hashes of address.
+func (b *APIBackend) GetTransactionsHistory(address string) ([]common.Hash, error) {
```
rlan35: Why do we only return hashes? Should we add pagination? The number of txns can be huge.

flicker-harmony: I guess we can show 100 of txs by default, but shouldn't it be only on frontend side?

flicker-harmony:
Screenshot 2019-09-24 at 01 45 18

PR #1647, opened 2019-09-24
updated to use non-initializing vdf

coolcottontail:
Use the non-initializing vdf package from github.com/harmony.one/vdf@v3.0.0

A large array of prime numbers are pre-calculated instead of calculating, saving 3 seconds system initialization time. 

Somehow the tag v3.0.0 can't be fetched automatically so has to use this tag v0.0.0-20190924052046-3ac2a5bbd98a


coolcottontail:
jenkins test this please

PR #1649, opened 2019-09-24
Fnswap

benjaminbfong:
swapping fn keys


harmony-ek:

PR #1653, opened 2019-09-25
Node enable public RPC

LeoHChen:


LeoHChen:
Tested on my FN, the 9500 port is open after "-P" option enabled.

PR #1671, opened 2019-09-26
fix storage instance and 0x address case

flicker-harmony:
## Issue

Small fixes for tx transactions history api. Fixed 0x case (returned error if address wasn't in db), storage should be fetched with false flag.

## Test

Localnet, postman.

#### Test Coverage Data

* Before
* After


flicker-harmony:
Jenkins, test this please.

PR #1672, opened 2019-09-26
Add staking transaction struct and handler logic

rlan35:
Tested locally with cross-shard txn. It still works.


node/node.go
```diff
@@ -255,6 +260,17 @@ func (node *Node) reducePendingTransactions() {
 	}
 }
 
+func (node *Node) reducePendingStakingTransactions() {
+	txPoolLimit := core.ShardingSchedule.MaxTxPoolSizeLimit()
+	curLen := len(node.pendingStakingTransactions)
+
+	// If length of pendingStakingTransactions is greater than TxPoolLimit then by greedy take the TxPoolLimit recent transactions.
+	if curLen > txPoolLimit+txPoolLimit {
+		node.pendingStakingTransactions = append(types2.StakingTransactions(nil), node.pendingStakingTransactions[curLen-txPoolLimit:]...)
```
chaosma: we can first detect duplicated staking transactions, then remove the older transactions

chaosma: e.g.  remove the duplicated staking in add pending transactions function to save memory

node/worker/worker.go
```diff
@@ -96,13 +98,6 @@ func (w *Worker) throttleTxs(selected types.Transactions, recentTxsStats types.R
 
 // SelectTransactionsForNewBlock selects transactions for new block.
 func (w *Worker) SelectTransactionsForNewBlock(newBlockNum uint64, txs types.Transactions, recentTxsStats types.RecentTxsStats, txsThrottleConfig *shardingconfig.TxsThrottleConfig, coinbase common.Address) (types.Transactions, types.Transactions, types.Transactions) {
-	// Must update to the correct current state before processing potential txns
-	if err := w.UpdateCurrent(coinbase); err != nil {
```
chaosma: why removed

staking/types/staking_transaction.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"math/big"
+
+	"github.com/harmony-one/harmony/internal/common"
+)
+
+// StakingTransaction struct.
+type StakingTransaction struct {
+	AccountNonce uint64         `json:"nonce"      gencodec:"required"`
```
chaosma: renamed to Nonce ?

rlan35: Just following the transaction field. and This is the same as tx's nonce.

node/worker/worker.go
rlan35: Moved it out of this func. so it applied to both normal tx and staking tx.

node/node.go
rlan35: Done.

PR #1675, opened 2019-09-27
add delegation data structure and methods

chaosma:
- delegation data structure and methods
- validator data structure and methods
- commission data structure and methods


staking/types/delegation.go
```diff
@@ -0,0 +1,391 @@
...
+      Balance:                   %s
+`,
+			i, entry.CreationHeight, entry.CompletionTime, entry.InitialBalance, entry.SharesDst, entry.Balance,
+		)
+	}
+
+	return strings.TrimRight(out, "\n")
+}
+
+type redelegationRespAlias RedelegationResponse
```
fxfactorial: technically alias in go lang is of form: `type f = g`

staking/types/validator.go
```diff
@@ -5,6 +5,16 @@ import (
...
+)
+
+// Define validator staking related const
+const (
+	MaxNameLength            = 70
+	MaxIdentityLength        = 3000
+	MaxWebsiteLength         = 140
+	MaxSecurityContactLength = 140
+	MaxDetailsLength         = 280
 )
```
fxfactorial: How you picked these magic numbers

staking/types/delegation.go
chaosma: this is removed from latest commit

staking/types/validator.go
chaosma: I copied these numbers, we can always modify it later. This is just a start

PR #1680, opened 2019-09-27
Add edit validator;delegate;redelegate;undelegate msgs

rlan35:


internal/common/address.go
```diff
@@ -49,6 +50,12 @@ func IsBech32Address(s string) bool {
 	return true
 }
 
+// IsEmpty gets whether the address contains all 0 bytes
+func (a Address) IsEmpty() bool {
+	empty := Address{}
```
fxfactorial: pull out as plain var, no need to make each time.

internal/common/hash.go
```diff
@@ -34,6 +35,12 @@ func BytesToHash(b []byte) Hash {
 	return h
 }
 
+// IsEmpty gets whether the hash contains all 0 bytes
+func (h Hash) IsEmpty() bool {
+	empty := Hash{}
```
fxfactorial: pull out as plain var, no need to make each time.

staking/types/messages.go
```diff
@@ -1,10 +1,14 @@
...
 
 import (
+	"bytes"
+	"encoding/json"
 	"math/big"
 
-	"github.com/harmony-one/bls/ffi/go/bls"
-	"github.com/harmony-one/harmony/internal/common"
+	"github.com/harmony-one/harmony/shard"
+
```
fxfactorial: whitespace

staking/types/messages.go
```diff
@@ -18,16 +22,189 @@ type StakingMessage interface {
...
+	MinSelfDelegation *big.Int           `json:"min_self_delegation" yaml:"min_self_delegation"`
+	StakingAddress    common.Address     `json:"staking_address" yaml:"staking_address"`
+	ValidatingPubKey  shard.BlsPublicKey `json:"validating_pub_key" yaml:"validating_pub_key"`
+	Amount            *big.Int           `json:"amount" yaml:"amount"`
+}
+
+// NewMsgCreateValidator creates a new validator
+func NewMsgCreateValidator(
+	description Description, commission CommissionRates, minSelfDelegation *big.Int, stakingAddress common.Address, validatingPubKey shard.BlsPublicKey, amount *big.Int) MsgCreateValidator {
+
```
fxfactorial: same, whitespace

staking/types/messages.go
```diff
@@ -18,16 +22,189 @@ type StakingMessage interface {
...
+	return nil
+}
+
+// ValidateBasic quick validity check
+func (msg MsgCreateValidator) ValidateBasic() error {
+	// note that unmarshaling from bech32 ensures either empty or valid
+	if msg.StakingAddress.Big().Uint64() == 0 {
+		return errors.New("[CreateValidator] address is empty")
+	}
+	emptyKey := shard.BlsPublicKey{}
```
fxfactorial: pull out as plain var, no need to make each time

staking/types/messages.go
```diff
@@ -18,16 +22,189 @@ type StakingMessage interface {
...
+
+	return MsgDelegate{
+		DelegatorAddress: delegatorAddress,
+		ValidatorAddress: validatorAddress,
+		Amount:           amount,
+	}
+}
+
+// Type ...
+func (msg MsgDelegate) Type() string { return "delegate" }
```
fxfactorial: const string should be const, no need to make fresh string

staking/types/messages_test.go
```diff
@@ -0,0 +1,218 @@
...
+		Details:         "the best validator ever",
+	}, CommissionRates{
+		Rate:          commissionRate,
+		MaxRate:       maxRate,
+		MaxChangeRate: maxChangeRate,
+	}, minSelfDelegation, validatorAddress, blsPublickey, stakeAmount)
+
+	rlpBytes, err := rlp.EncodeToBytes(msgCreateValidator)
+	if err != nil {
+		t.Error("failed to rlp encode 'create validator' message")
```
fxfactorial: in general with these RLP decoding, wondering if it would be nice to include the RLP as Hex perhaps in the error message so later can do `rlpdump` when debugging when something went wrong, otherwise need to recompile code + print, aka waste time. Perhaps rlp itself could observe debug flag set in init from environment var

internal/common/address.go
rlan35: Done.

internal/common/hash.go
rlan35: Done.

staking/types/messages.go
rlan35: Done.

staking/types/messages.go
rlan35: Done.

staking/types/messages.go
rlan35: Done.

staking/types/messages.go
rlan35: Done.

staking/types/messages_test.go
rlan35: Will skip it for now. If bytes level there is problem, more debugging is needed either way.

staking/types/messages.go
```diff
@@ -18,16 +22,189 @@ type StakingMessage interface {
...
+		Commission:        commission,
+		MinSelfDelegation: minSelfDelegation,
+		StakingAddress:    stakingAddress,
+		ValidatingPubKey:  validatingPubKey,
+		Amount:            amount,
+	}
 }
 
 // Type ...
 func (msg MsgCreateValidator) Type() string { return "create_validator" }
```
chaosma: what are the other possible types?

PR #1687, opened 2019-09-29
Add transactions history api with full tx info

flicker-harmony:
## Issue

Richard requested tx history to have full tx info. I made a separate method to call tx history with full info using method that returns just hashes.

## Test

Localnet, postman.
Screenshot 2019-09-30 at 01 32 54

#### Test Coverage Data

* Before
* After

## TODO

Make adjustments on explorer back/front.

### DIFF
--- a/internal/hmyapi/transactionpool.go
+++ b/internal/hmyapi/transactionpool.go
@@ -20,6 +20,7 @@ type TxHistoryArgs struct {
 	Address string `json:"address"`
 	Offset  int    `json:"offset"`
 	Page    int    `json:"page"`
+	FullTx  bool   `json:"fullTx"`
 }
 
 // PublicTransactionPoolAPI exposes methods for the RPC interface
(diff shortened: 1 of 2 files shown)


internal/hmyapi/transactionpool.go
```diff
@@ -55,6 +55,20 @@ func (s *PublicTransactionPoolAPI) GetTransactionsHistory(ctx context.Context, a
...
 
+// GetTransactionsHistoryFull returns the list of transactions with full info that involve a particular address.
+func (s *PublicTransactionPoolAPI) GetTransactionsHistoryFull(ctx context.Context, args TxHistoryArgs) ([]*RPCTransaction, error) {
+	hashes, err := s.GetTransactionsHistory(ctx, args)
+	if err != nil {
+		return nil, err
+	}
+	result := []*RPCTransaction{}
+	for _, hash := range hashes {
+		tx := s.GetTransactionByHash(ctx, hash)
```
fxfactorial: How many hashes could this be? How much on average ? Why aren’t you getting them all at once instead of in sequence ? Does it have to be in sequence ?

internal/hmyapi/transactionpool.go
```diff
@@ -55,6 +55,20 @@ func (s *PublicTransactionPoolAPI) GetTransactionsHistory(ctx context.Context, a
 	return ReturnWithPagination(result, args), nil
 }
 
+// GetTransactionsHistoryFull returns the list of transactions with full info that involve a particular address.
+func (s *PublicTransactionPoolAPI) GetTransactionsHistoryFull(ctx context.Context, args TxHistoryArgs) ([]*RPCTransaction, error) {
```
LeoHChen: do we support pagination ?  This could be too huge of full history.

internal/hmyapi/transactionpool.go
flicker-harmony: Pagination makes it to 100 by default, can be more if offset is specified in TxHistoryArgs.

internal/hmyapi/transactionpool.go
flicker-harmony: Pagination already applied when tx history hashes are returned.

lzl124631x:
I think we should extend existing one instead of creating another one. These APIs will be used externally so we must have good interface design. https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_getblockbyhash has an example: 

> Boolean - If true it returns the full transaction objects, if false only the hashes of the transactions.

internal/hmyapi/transactionpool.go
```diff
@@ -34,25 +35,35 @@ func NewPublicTransactionPoolAPI(b Backend, nonceLock *AddrLocker) *PublicTransa
...
 		return nil, err
 	}
-	result, err := s.b.GetTransactionsHistory(oneAddress)
+	hashes, err := s.b.GetTransactionsHistory(oneAddress)
 	if err != nil {
 		return nil, err
 	}
-	return ReturnWithPagination(result, args), nil
+	result = ReturnWithPagination(hashes, args)
+	if !args.FullTx {
```
lzl124631x: looks great. Just want to confirm that, if this fullTx is not set (it should be optional parameter I think), it will go thru this `if`?

internal/hmyapi/transactionpool.go
```diff
@@ -34,25 +35,35 @@ func NewPublicTransactionPoolAPI(b Backend, nonceLock *AddrLocker) *PublicTransa
 }
 
 // GetTransactionsHistory returns the list of transactions hashes that involve a particular address.
-func (s *PublicTransactionPoolAPI) GetTransactionsHistory(ctx context.Context, args TxHistoryArgs) ([]common.Hash, error) {
+func (s *PublicTransactionPoolAPI) GetTransactionsHistory(ctx context.Context, args TxHistoryArgs) (map[string]interface{}, error) {
```
LeoHChen: Can we change the map to a one-member struct later on?

PR #1688, opened 2019-09-30
[p2p] add network type prefix to group id

LeoHChen:
this is solve the problem of validators in different network connected
with each others.

* mainet is still using the original harmony prefix to keep backward
compatibility

* pangaea uses "pangaea" as network prefix
* testnet uses "testnet" as network prefix

All nodes in Pangaea and Testnet need to restart to re-connect with each
other.  Mainnet nodes have no changes.

### DIFF
--- a/internal/configs/node/group.go
+++ b/internal/configs/node/group.go
@@ -1,12 +1,8 @@
-package p2p
+package nodeconfig
 
 import (
-	"context"
 	"fmt"
-	"io"
 	"strconv"
-
-	libp2p_peer "github.com/libp2p/go-libp2p-peer"
 )
 
 // GroupID is a multicast group ID.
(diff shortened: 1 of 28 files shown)


LeoHChen:
Local Network Test passed.

LeoHChen:
Jenkins, test this please.

internal/configs/node/group.go
```diff
@@ -25,32 +21,50 @@ func (id GroupID) String() string {
...
+	GroupIDUnknown           GroupID = "%s/B1acKh0lE"
 )
 
 // ShardID defines the ID of a shard
 type ShardID uint32
 
+func getNetworkPrefix(shardID ShardID) (netPre string) {
+	switch GetShardConfig(uint32(shardID)).GetNetworkType() {
+	case Mainnet:
+		netPre = "harmony"
```
rlan35: Let's do "harmony-mainnet", "harmony-testnet", "harmony-pangaea" ....  to avoid potential conflict with other projects.

rlan35: Just realize that it's related to backward compatibility, so should be "harmony", "harmony-testnet", "harmony-pangaea"

LeoHChen: sure will change it.

PR #1689, opened 2019-09-30
Staking dev cycle

fxfactorial:
Integrated Staking re RPC & added separate CLI `staking-iterate` so we can iterate faster on this feature. 

Run the binary with: `./bin/staking-standalone staking-iterate`


staking/types/transaction_signing.go
```diff
@@ -0,0 +1,269 @@
...
+//
+// The go-ethereum library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public License
+// along with the go-ethereum library. If not, see <http://www.gnu.org/licenses/>.
+
+package types
```
rlan35: Please add a NOTE here and in normal transaction_signing.go to keep these two files in sync, and add a TODO to refactor these copied code together.

staking/types/messages.go
```diff
@@ -22,13 +22,30 @@ const (
...
+	Kind string
 	// Signer returns the ECDSA address who must sign the outer transaction
-	Signer() common.Address
+	Signer common.Address
+}
+
+// NewMessage does sanity check on input
+func NewMessage(kind string, signer common.Address) (*Message, error) {
+	for _, k := range kinds {
+		if kind != k {
```
rlan35: should be kind == k?

staking/role/validator.go
```diff
@@ -0,0 +1,38 @@
...
+	Description       Ties           `json:"description" yaml:"description"`                 // description for the validator
+	Active            bool           `json:"active" yaml:"active"`                           // Is the validator active in the validating process or not
+	Stake             *big.Int       `json:"stake" yaml:"stake"`                             // The stake put by the validator itself
+	UnbondingHeight   *big.Int       `json:"unbonding_height" yaml:"unbonding_height"`       // if unbonding, height at which this validator has begun unbonding
+	Commission        Commission     `json:"commission" yaml:"commission"`                   // commission parameters
+	MinSelfDelegation *big.Int       `json:"min_self_delegation" yaml:"min_self_delegation"` // validator's self declared minimum self delegation
+}
+
+// Ties - some possible IRL connections
+type Ties struct {
```
rlan35: Why change to "Ties"?  What does it mean? We should use commonly used terms for understandability.

staking/message/messages_test.go
```diff
@@ -0,0 +1,207 @@
+package message
+
+// var (
```
rlan35: why are these commented?

staking/role/validator.go
fxfactorial: Ties means association, `I am tied to the land`, `What ties do you have to us?`

staking/role/commission.go
```diff
@@ -1,4 +1,4 @@
-package types
+package role
```
rlan35: what does role package mean? seems to me the files under "role" is not related to role (e.g. commission, decimal etc.)

staking/message/messages_test.go
fxfactorial: Because they assumed previous data structure but I want these tests implemented without losing idea of how to be structured, organized. (I prefer to move forward, while giving breadcrumbs for others)

staking/role/commission.go
fxfactorial: role means what part to play, that is, a validator is an entity, and nodes play the role of validator. Like I play the role of Banker, and part of my role is to have commissions

staking/transaction/transaction.go
```diff
@@ -0,0 +1,111 @@
...
+	// Signature values
+	V *big.Int `json:"v" gencodec:"required"`
+	R *big.Int `json:"r" gencodec:"required"`
+	S *big.Int `json:"s" gencodec:"required"`
+	// This is only used when marshaling to JSON.
+	Hash *common.Hash `json:"hash" rlp:"-"`
+}
+
+// Stake is a record captuing all staking operations
+type Stake struct {
```
rlan35: This should really just by StakingTransaction. Stake reads like it's containing information about a specific stake. Not quite easy to understand.

fxfactorial: I want to avoid stutter though: 
```
Avoid stutter. Since client code uses the package name as a prefix when referring to the package contents, the names for those contents need not repeat the package name. The HTTP server provided by the http package is called Server, not HTTPServer. Client code refers to this type as http.Server, so there is no ambiguity. 
```

https://blog.golang.org/package-names

caller ends up being then: `transaction.StakingTransaction`  But on this one I can see it going either way, can change

staking/transaction/transaction.go
```diff
@@ -0,0 +1,111 @@
...
+	data txdata
+	// caches
+	hash atomic.Value
+	size atomic.Value
+	from atomic.Value
+}
+
+type fulfill func() (message.Directive, interface{})
```
rlan35: This thing is confusing too. Better to just make Directive and payload as two params in NewStake()

fxfactorial: but as API consumer, (as hmy CLI is/will be), nicer to have one call to NewStake with different callbacks that I can dynamically make. That is, it decouples things for end user.

staking/transaction/transaction.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/harmony-one/harmony/crypto/hash"
+	"github.com/harmony-one/harmony/staking/message"
+)
+
+type txdata struct {
+	message.Directive
+	Stake        interface{}
```
rlan35: Change to To StakeMsg. "Stake" itself is really confusing.

staking/types/transaction.go
```diff
@@ -0,0 +1,110 @@
...
+	// Signature values
+	V *big.Int `json:"v" gencodec:"required"`
+	R *big.Int `json:"r" gencodec:"required"`
+	S *big.Int `json:"s" gencodec:"required"`
+	// This is only used when marshaling to JSON.
+	Hash *common.Hash `json:"hash" rlp:"-"`
+}
+
+// Stake is a record captuing all staking operations
+type Stake struct {
```
rlan35: Change this to "Transaction" or "StakingTransaction".

rlan35: After this. Good to merge.

rlan35:
LGTM. Approved.

cmd/staking/root.go
```diff
@@ -0,0 +1,173 @@
...
+	queryID       = 0
+	s             = &staker{}
+	localNetChain = big.NewInt(2)
+	dAddr         = common.ParseAddr(testAccount)
+)
+
+const (
+	// Harmony protocol assume beacon chain shard is only place to send
+	// staking, later need to consider logic when beacon chain shard rotates
+	stakingShard        = 0
```
chaosma: move it to upper level, so we can share this variable

chaosma:
Many codes from message.go and validator.go are removed. Are they deleted or refactor to some other places?

cmd/staking/root.go
fxfactorial: This is a temp CLI that later will remove once staking work done

fxfactorial:
@chaosma i felt it better to start with a small surface area, not everything was needed

PR #1690, opened 2019-10-01
offset, page -> pageSize, pageIndex

flicker-harmony:
## Issue

Change offset, page -> pageSize, pageIndex in hmy api transaction history.

#### Test Coverage Data

* Before
* After

### DIFF
--- a/internal/hmyapi/transactionpool.go
+++ b/internal/hmyapi/transactionpool.go
@@ -17,10 +17,10 @@ import (
 
 // TxHistoryArgs is struct to make GetTransactionsHistory request
 type TxHistoryArgs struct {
-	Address string `json:"address"`
-	Offset  int    `json:"offset"`
-	Page    int    `json:"page"`
-	FullTx  bool   `json:"fullTx"`
+	Address   string `json:"address"`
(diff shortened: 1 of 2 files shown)


internal/hmyapi/util.go
```diff
@@ -13,23 +13,23 @@ import (
 
 // defaultOffset is to have default pagination.
```
LeoHChen: Rename this too?

internal/hmyapi/transactionpool.go
```diff
@@ -17,10 +17,10 @@ import (
 
 // TxHistoryArgs is struct to make GetTransactionsHistory request
 type TxHistoryArgs struct {
-	Address string `json:"address"`
-	Offset  int    `json:"offset"`
-	Page    int    `json:"page"`
-	FullTx  bool   `json:"fullTx"`
+	Address   string `json:"address"`
+	PageIndex int    `json:"pageIndex"`
+	PageSize  int    `json:"pageSize"`
```
LeoHChen: Can you cross-ref the consumer side change too?

Also, does this require documentation update, if any?

PR #1700, opened 2019-10-04
Merklelize staking information  (WIP)

chaosma:
Discussed with RJ and Edgar.  We use Storage field (core/state/state_object.go) to store all the staking information including validator and delegation. To do this, we need first make the fields of Validator flatten (staking/types/validator.go). Then whenever we create a new validator, we use a specific key for each of the field in Validator structure to store the value. For delegation, it's even simpler, the key will be the delegator's address, value will be the amount of delegation. The Root field in the Account structure (state_object.go) will be the merkle root of this validator and its delegations' information.


fxfactorial:
subsumed in https://github.com/harmony-one/harmony/pull/1709

PR #1573, opened 2019-09-11
Optimize new node state sync by parallelizing insertion of blocks to chain

denniswon:
## Issue

https://github.com/harmony-one/harmony/issues/1313

optimization + error handling

Note that this PR is rebased off of https://github.com/harmony-one/harmony/pull/1420/ which "fixes" the state sync issue. this one optimizes the flow on top of the fix.

#### Test Coverage Data

dennis.won@jongs-mbp:~/harmony/node (optimize_sync) $ go test -cover
PASS
coverage: 15.5% of statements
ok  	github.com/harmony-one/harmony/node	4.228s

* Before
* After

#### Test/Run Logs

Consensus reached fine on localnet and see the block num increasing

PR #1701, opened 2019-10-04
remove quorum check and add verifyHaeder option to insertChains to blockchain

denniswon:
## Issue

1. false logic to check if the max consensus hash block header is signed by more than the quorum on shard is incorrect, and causes node to unable to reach consensus if the peers give them a max consensus hash that is invalid. Removed that, and instead, it will retry with the second max.

This should fix this issue of consensus failure. But this is still not optimal/robust enough. https://github.com/harmony-one/harmony/pull/1573 will further optimize the design.

#### Test Coverage Data

~/harmony/core $ go test -cover
newNodeList:  [ECDSA: one1qqqrvvpcxu6rwvfnxc6ngvecx5mrgvf3254yj4, BLS: ECDSA: one1qq6n2de3xuuryven8qcnqvfcxuurwd3s837n5r, BLS: ECDSA: one1qqcnjv3kxqcnydfcxc6nyd3kxg6rqvpe5srx9h, BLS: ECDSA: one1qqqrgvp5xy6nxwf5x5mngve4xsmnvdfhekh0gx, BLS: ECDSA: one1qqen2ve5xvengvekxuerzdpjxvmnyd33m0urmm, BLS:]
PASS
coverage: 24.3% of statements
ok  	github.com/harmony-one/harmony/core	6.581s


api/service/syncing/syncing.go
```diff
@@ -327,55 +326,39 @@ func (sc *SyncConfig) GetBlockHashesConsensusAndCleanUp() bool {
...
+					Int("respondSize", len(response.Payload)).
+					Msg("[SYNC] GetConsensusHashes: receive more blockHahses than request!")
+				peerConfig.blockHashes = response.Payload[:size+1]
+			} else {
+				peerConfig.blockHashes = response.Payload
+			}
+		}()
+		return
+	})
+	wg.Wait()
```
fxfactorial: Does this goroutine need to block, or can caller move on ?

api/service/syncing/syncing.go
```diff
@@ -745,28 +724,24 @@ func (ss *StateSync) SyncLoop(bc *core.BlockChain, worker *worker.Worker, isBeac
...
+			utils.Logger().Debug().Msgf("[SYNC] Node is Not in Sync (isBeacon: %t, ShardID: %d, otherHeight: %d, currentHeight: %d)", isBeacon, bc.ShardID(), otherHeight, currentHeight)
+		}
+		startHash := bc.CurrentBlock().Hash()
+		size := uint32(otherHeight - currentHeight)
+		if size > BatchSize {
+			size = BatchSize
 		}
+		ss.ProcessStateSync(startHash[:], size, bc, worker)
+		ss.purgeOldBlocksFromCache()
+		time.Sleep(SyncLoopFrequency * time.Second)
```
fxfactorial: Who is caller of this go routine, is it necessary to be explicitly blocking

chaosma:
discussed offline with Dennis, will fix the merge conflix

denniswon:
@chaosma per discussion with @LeoHChen, I will close old PRs and reopen as fresh new ones. Will link new ones here soon.

api/service/syncing/syncing.go
denniswon: state sync steps are sequential, so this needs to be blocked here.

api/service/syncing/syncing.go
denniswon: ditto: yes, it need to be blocking here. there should be a single state syncing happening as a batch-sequential sync loop

PR #1709, opened 2019-10-07
Further progress on staking

fxfactorial:
1. Incorporated code from @chaosma previously open PR
2. Removed dead staking related code from prior attempted implementation
3. Updated codebase to reflect removal of previous staking code usage
4. Small refactors along the way to pull out values, less usage of magic numbers, factored out many error values.


chaosma:
LGTM

core/resharding.go
```diff
@@ -171,42 +166,10 @@ func CalculateNewShardState(
 		return nil, ctxerror.New("cannot retrieve previous sharding state").
 			WithCause(err)
 	}
-	newNodeList := ss.UpdateShardingState(stakeInfo)
```
rlan35: Why do we remove these two line here? This func is not used but I don't see the need to remove these two line specifically now.

core/values/blockchain.go
```diff
@@ -0,0 +1,10 @@
+package values
+
+const (
+	// BeaconChainShardID is the ShardID of the BeaconChain
+	BeaconChainShardID = 0
+	// VotingPowerReduceBlockThreshold roughly corresponds to 3 hours
+	VotingPowerReduceBlockThreshold = 1350
+	// VotingPowerFullReduce roughly corresponds to 12 hours
+	VotingPowerFullReduce = 4 * VotingPowerReduceBlockThreshold
```
rlan35: no need for these threshold, we can just use the voting power == 0 to turn the validator to inactive. After four 1/4 voting power cut (or cumulatively 12 hours down time), the validator will be slashed.

core/types/block.go
```diff
@@ -181,12 +181,13 @@ func init() {
...
+// Block represents an entire block in the Harmony blockchain.
 type Block struct {
-	header           *block.Header
-	uncles           []*block.Header
-	transactions     Transactions
-	incomingReceipts CXReceiptsProofs
+	header              *block.Header
+	uncles              []*block.Header
+	transactions        Transactions
+	stakingTransactions staking.StakingTransactions
```
rlan35: This could potentially have compatibility issue. I checked the code and didn't find obvious issues for now. But we need to be careful.

core/resharding.go
fxfactorial: Sorry, optimistic dead code slashing

PR #1710, opened 2019-10-08
Limit concurrency

harmony-ek:
Do not spawn unbounded number of goroutines for parallel message processing.  Instead, use a worker pool of 32 goroutines sharing the same intake queue.  Log (`"rx overrun"`) and tail-drop messages if the queue is full.

## Tests

Local test pass; 100% coverage of the newly added core code (`./msgq`):

```sh
git checkout harmony-one/master
go test -count=1 -cover ./... > coverage-before.txt
git checkout limit_concurrency 
go test -count=1 -cover ./... > coverage-after.txt
sed -i.bak -E 's/  [0-9]+(\.[0-9]*)?s      /       /' coverage-*.txt
diff -U0 coverage-before.txt coverage-after.txt
```
```diff
--- coverage-before.txt 2019-10-08 16:13:42.000000000 -0700
+++ coverage-after.txt  2019-10-08 16:13:42.000000000 -0700
@@ -86 +86,2 @@
-ok     github.com/harmony-one/harmony/node     coverage: 15.4% of statements
+ok     github.com/harmony-one/harmony/msgq     coverage: 100.0% of statements
+ok     github.com/harmony-one/harmony/node     coverage: 14.3% of statements
```

## TODO

* [x] Add UTs

### DIFF
--- a/msgq/msgq.go
+++ b/msgq/msgq.go
@@ -47,5 +47,11 @@ func (q *Queue) HandleMessages(h MessageHandler) {
 	}
 }
 
+// Close closes the given queue.
+func (q *Queue) Close() error {
+	close(q.ch)
+	return nil
+}
+
 // ErrRxOverrun signals that a receive queue has been overrun.
 var ErrRxOverrun = errors.New("rx overrun")
--- a/msgq/msgq.go
+++ b/msgq/msgq.go
@@ -1,0 +1,51 @@
+// Package msgq implements a simple, finite-sized message queue.  It can be used
+// as a building block for a message processor pool.
+package msgq
+
+import (
+	"github.com/libp2p/go-libp2p-core/peer"
+	"github.com/pkg/errors"
+)
+
+type message struct {
+	content []byte
+	sender  peer.ID
+}
+
+// MessageHandler is a message handler.
+type MessageHandler interface {
+	HandleMessage(content []byte, sender peer.ID)
+}
+
+// Queue is a finite-sized message queue.
+type Queue struct {
+	ch chan message
+}
+
+// New returns a new message queue of the given size.
+func New(size int) *Queue {
+	return &Queue{ch: make(chan message, size)}
+}
+
+// AddMessage enqueues a received message for processing.  It returns without
+// blocking, and may return a queue overrun error.
+func (q *Queue) AddMessage(content []byte, sender peer.ID) error {
+	select {
+	case q.ch <- message{content, sender}:
+	default:
+		return ErrRxOverrun
+	}
+	return nil
+}
+
+// HandleMessages dequeues and dispatches incoming messages using the given
+// message handler, until the message queue is closed.  This function can be
+// spawned as a background goroutine, potentially multiple times for a pool.
+func (q *Queue) HandleMessages(h MessageHandler) {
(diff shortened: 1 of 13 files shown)


node/node.go
```diff
@@ -7,6 +7,8 @@ import (
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
+	libp2p_peer "github.com/libp2p/go-libp2p-core/peer"
+
```
fxfactorial: Is it desired that we leave one line clear between lib code outside `harmony-one`?

node/node.go
```diff
@@ -513,6 +527,11 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, chainDBFactory shardc
 		Interface("genesis block header", node.Blockchain().GetHeaderByNumber(0)).
 		Msg("Genesis block hash")
 
+	node.incomingMessages = make(chan incomingMessage, RxQueueSize)
+	for i := 0; i < RxWorkers; i++ {
```
fxfactorial: why explicit for loop, not range

fxfactorial:
Wondering if we should have go routine wrapper, like `withMaxConcurrency(max int, caller_provide_cb)`

node/node.go
harmony-ek: We do not have a team-level consensus yet; I personally prefer/use the stdlib-external-local grouping order, because that way the same-package imports are easier for human eyes to identify especially when dealing with import cycles and so on.

fxfactorial: I see, I prefer otherwise, no line breaks. Let's pick one way

node/node.go
harmony-ek: There is no channel/array/slice to range over; Go does not support plain integer range like in Python either.

harmony-ek:
> Wondering if we should have go routine wrapper, like `withMaxConcurrency(max int, caller_provide_cb)`

Thought about doing so, but didn't go for it because it would lose type safety: the intake channel's type must be made generic (`chan interface{}`) and the recipient would have to type-assert it from `interface{}` which opens a possibility of a panic when misused/abused (`msg := item.(incomingMessage)` where the received `item` is not an `incomingMessage`), and even with a type check (`if msg, ok := item.(incomingMessage); !ok { ... }`) it creates a spurious branch that hurts readability.

This is a cost-benefit trade off.  I'd say factor the logic out if we do this maybe the third time.

(O generics, where art thou…?)

msgq/msgq.go
```diff
@@ -0,0 +1,57 @@
...
+func (q *Queue) HandleMessages(h MessageHandler) {
+	for msg := range q.ch {
+		h.HandleMessage(msg.content, msg.sender)
+	}
+}
+
+// Close closes the given queue.
+func (q *Queue) Close() error {
+	close(q.ch)
+	return nil
```
fxfactorial: Why give back error at all if giving back nil

rlan35:
LGTM, [sidenote, it would be better if it support prioritization scheduling of the messages. Consensus should take higher priority]

msgq/msgq.go
harmony-ek: To be compliant with the stdlib `io.Closer` interface.

PR #1717, opened 2019-10-09
Factor out quorum logic from consensus, undo prior consolidation of values under core

fxfactorial:
This is a substantial refactor. 

1) Factor out quorum logic from consensus - So can drop in staking based quorum logic
(quorum.go shows what I think to be good example of composition of interfaces)
2) Remove naming of `pbft` to `fbft` and all such variants
3) Create comforts: spelling mistakes, long lines into multi-lines, prettier run on markdown, some removal of txgen as further entails full removal of txgen. 
4) Undo prior consolidation of values under `core/values`
5) Add panic safe ways for string versions of enums. 
6) Removal of dead code, dead comments. (bitrot)


cmd/client/txgen/main.go
```diff
@@ -98,7 +98,7 @@ func setUpTXGen() *node.Node {
 		fmt.Fprintf(os.Stderr, "Error :%v \n", err)
 		os.Exit(1)
 	}
-	consensusObj, err := consensus.New(myhost, uint32(shardID), p2p.Peer{}, nil)
+	consensusObj, err := consensus.NewOneVotePerValidator(myhost, uint32(shardID), p2p.Peer{}, nil)
```
chaosma: One validator can have different voting power, instead of each one has equal power, we can add a field votingPower in consensus object which is calculated from their stake

cmd/harmony/main.go
```diff
@@ -289,7 +289,7 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 	// Consensus object.
 	// TODO: consensus object shouldn't start here
 	// TODO(minhdoan): During refactoring, found out that the peers list is actually empty. Need to clean up the logic of consensus later.
-	currentConsensus, err := consensus.New(myHost, nodeConfig.ShardID, p2p.Peer{}, nodeConfig.ConsensusPriKey)
+	currentConsensus, err := consensus.NewOneVotePerValidator(myHost, nodeConfig.ShardID, p2p.Peer{}, nodeConfig.ConsensusPriKey)
```
chaosma: same comment as above

cmd/client/txgen/main.go
fxfactorial: yes, see the other constructor function which is for staking based voting.

consensus/consensus.go
```diff
@@ -254,6 +243,59 @@ func New(host p2p.Host, ShardID uint32, leader p2p.Peer, blsPriKey *bls.SecretKe
...
+	}
+	c.ConsensusMechanism = values.SuperMajorityVote
+	c.isQuorumSatisfied = func(p values.QuorumPhase) bool {
+		need := c.QuorumThreshold().Int64()
+		switch p {
+		case values.QuorumPrepare:
+			return int64(len(c.prepareSigs)) >= need
+		case values.QuorumCommit:
+			return int64(len(c.commitSigs)) >= need
+		default:
```
chaosma: Add view change quorum satisfy criterion, or still use previous method

cmd/client/txgen/main.go
fxfactorial: https://github.com/harmony-one/harmony/pull/1717/files#diff-3a5c8ea459d421751fb70343be214598R251

core/values/consensus.go
```diff
@@ -0,0 +1,56 @@
+package values
```
harmony-ek: 1. Please avoid putting new code into the `./core` package or its subpackages, which came from Ethereum and we apply a licensing (LGPL) exception there.
2. I don't think congregating constants into a package simply because they are constants is not a good idea.  Definitions should stay with the actual code to which they belong or at least which uses them the most.  I think the old location (`staking` and `consensus` packages) are more suitable.  (Plus, `types.StakingDirective` reads less natural than `staking.Directive`, for example.)

Could you consider moving them back to where they were?

consensus/consensus_viewchange_msg.go
```diff
@@ -29,8 +29,10 @@ func (consensus *Consensus) constructViewChangeMessage() []byte {
 	// next leader key already updated
 	vcMsg.LeaderPubkey = consensus.LeaderPubKey.Serialize()
 
-	preparedMsgs := consensus.PbftLog.GetMessagesByTypeSeqHash(msg_pb.MessageType_PREPARED, consensus.blockNum, consensus.blockHash)
-	preparedMsg := consensus.PbftLog.FindMessageByMaxViewID(preparedMsgs)
+	preparedMsgs := consensus.PBFTLog.GetMessagesByTypeSeqHash(
+		msg_pb.MessageType_PREPARED, consensus.blockNum, consensus.blockHash,
```
chaosma: Is this kind of format change automatically by editor plugin?

consensus/view_change.go
```diff
@@ -389,12 +348,12 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
 	utils.Logger().Debug().
 		Int("numSigs", len(consensus.viewIDSigs)).
-		Int("needed", consensus.Quorum()).
+		Int64("needed", consensus.QuorumThreshold().Int64()).
 		Msg("[onViewChange]")
 
 	// received enough view change messages, change state to normal consensus
-	if len(consensus.viewIDSigs) >= consensus.Quorum() {
-		consensus.mode.SetMode(Normal)
+	if int64(len(consensus.viewIDSigs)) >= consensus.QuorumThreshold().Int64() {
```
chaosma: Why change int to int64? It seems not necessary

internal/utils/bytes.go
```diff
@@ -89,7 +89,7 @@ func countOneBitsInByte(by byte) int {
 }
 
 // CountOneBits counts the number of 1 bit in byte array
-func CountOneBits(arr []byte) int {
+func CountOneBits(arr []byte) int64 {
```
chaosma: reason is to be system independent?

core/values/consensus.go
```diff
@@ -0,0 +1,56 @@
...
+	PBFTListening
+)
+
+var (
+	modes  = [...]string{"Normal", "ViewChanging", "Syncing", "Listening"}
+	phases = [...]string{"Announce", "Prepare", "Commit"}
+)
+
+func (s PBFTState) String() string {
+	return modes[s]
```
harmony-ek: Let's not panic on undefined/garbage values of PBFTState.  (`modes` is map as suggested above.)
```suggestion
	if name, ok := stateNames[s]; ok {
		return name
	}
	return fmt.Sprintf("PBFTState %v", byte(s))
```

core/values/consensus.go
```diff
@@ -0,0 +1,56 @@
...
+// Enum for node Mode
+const (
+	PBFTNormal PBFTState = iota
+	PBFTViewChanging
+	PBFTSyncing
+	PBFTListening
+)
+
+var (
+	modes  = [...]string{"Normal", "ViewChanging", "Syncing", "Listening"}
```
harmony-ek: Enum values start having undefined holes in their range, so it's usually more appropriate to use a map:
```suggestion
	stateNames  = map[PBFTState]string{
		PBFTNormal:       "Normal",
		PBFTViewChanging: "ViewChanging",
		PBFTSyncing:      "Syncing",
		PBFTListening:    "Listening",
	}
```

Ditto for `phases` below.

internal/utils/bytes.go
harmony-ek: If we are changing this, I'd suggest using uint64 (a bit count can never be negative, can it?).

fxfactorial: No, so that easier to deal with the result of `Quroum` count result from https://github.com/harmony-one/harmony/pull/1717/files#diff-3a5c8ea459d421751fb70343be214598R32

node/node.go
```diff
@@ -546,7 +546,7 @@ func (node *Node) CalculateInitShardState() (err error) {
 
 	// Get genesis epoch shard state from chain
 	blockNum := node.Blockchain().CurrentBlock().NumberU64()
-	node.Consensus.SetMode(consensus.Listening)
+	node.Consensus.SetMode(values.PBFTListening)
```
harmony-ek: Once again, `values.PBFTListening` is a duh-boilerplate-y name.  `consensus.PBFTListening` or `consensus.Listening` flows more natural.  (Or even better, if PBFT algorithm were factored into its own package, it could have been `pbft.Listening`… boom.)

consensus/view_change.go
fxfactorial: Because staking code uses Decimal in `core/numeric` and more uniform to pick larger int that can accommodate old & new code

consensus/consensus_viewchange_msg.go
fxfactorial: I changed it by hand because was wider than 100 col, code running off side of page.

consensus/consensus.go
fxfactorial: done

consensus/consensus.go
```diff
@@ -254,6 +243,59 @@ func New(host p2p.Host, ShardID uint32, leader p2p.Peer, blsPriKey *bls.SecretKe
...
+		}
+	}
+	c.QuorumThreshold = func() *big.Int {
+		return big.NewInt(int64(len(c.PublicKeys)*2/3 + 1))
+	}
+	c.rewardThreshold = func() *big.Int {
+		return big.NewInt(int64(len(c.PublicKeys) * 9 / 10))
+	}
+	return c, nil
+}
```
harmony-ek: Please avoid introducing multiple variant-specific factory methods, but consider factoring varied logic into separate implementation of a common interface then use DI idiom:

```go
type Quorum interface {
	Mechanism() values.ConsensusMechanism
	IsSatisfied(sigs, constituents int) bool
	QuorumThreshold(constituents int) int
	RewardThreshold(constituents int) int
}

type VoteBasedQuorum struct {}

type StakeBasedQuorum struct {}
```

With a few more steps you should also be able to eliminate `sigs` and `constituents` arguments from methods, by injecting accessors of the number of signatures/constituents in the current context into the quorum check implementations.

consensus/view_change.go
chaosma: based on the limitation of PBFT, the number of signatures will not too large. for core/numeric, the value is most used for float number, here the quorum is independent of the staking related value

fxfactorial: Understood; point was to have a uniform interface because Threshold in staking could be a big value, hence it uses big.Int, so not so much that its about number of signers, but more about threshold of staking amount.

chaosma: If using a uniform interface, then that's fine

node/node.go
chaosma: agreed

consensus/view_change.go
```diff
@@ -9,112 +9,66 @@ import (
...
-	ViewChanging
-	Syncing
-	Listening
-)
-
-// PbftMode contains mode and viewID of viewchanging
-type PbftMode struct {
-	mode   Mode
+// PBFTMode contains mode and viewID of viewchanging
+type PBFTMode struct {
```
rlan35: This should be PBFTState as it contains many things.

consensus/view_change.go
```diff
@@ -9,112 +9,66 @@ import (
...
-	Normal Mode = iota
-	ViewChanging
-	Syncing
-	Listening
-)
-
-// PbftMode contains mode and viewID of viewchanging
-type PbftMode struct {
-	mode   Mode
+// PBFTMode contains mode and viewID of viewchanging
```
rlan35: BTW, as you are refactoring the naming. Let's conform to use FBFT as it's our consensus's name.

core/values/consensus.go
```diff
@@ -0,0 +1,56 @@
...
+	PBFTPrepare
+	PBFTCommit
+)
+
+// PBFTState determines whether a node is in normal or viewchanging mode
+type PBFTState byte
+
+// Enum for node Mode
+const (
+	PBFTNormal PBFTState = iota
```
rlan35: The Mode should be for general Consensus, not specifically FBFT

PR #1721, opened 2019-10-12
Staking message verification and processing (part I)

chaosma:
- add header V3 and body V2 for staking transactions
- add rawdb database read/write methods


core/types/block.go
```diff
@@ -239,11 +244,21 @@ type extblockV1 struct {
 	IncomingReceipts CXReceiptsProofs
 }
 
+// includes staking transaction
+type extblockV2 struct {
+	Header           *block.Header
+	Txs              []*Transaction
+	Stks             []*staking.StakingTransaction
+	Uncles           []*block.Header
```
rlan35: can you remove Uncles. we don't need it.

internal/params/config.go
```diff
@@ -137,6 +146,12 @@ func (c *ChainConfig) IsCrossTx(epoch *big.Int) bool {
 	return isForked(crossTxEpoch, epoch)
 }
 
+// IsStaking determines whether it is staking epoch
+func (c *ChainConfig) IsStaking(epoch *big.Int) bool {
+	stkEpoch := new(big.Int).Add(c.StakingEpoch, common.Big1)
```
rlan35: no need to add 1. see the reasoning for IsCrossTx.

staking/types/messages.go
```diff
@@ -49,10 +50,10 @@ func (d Directive) String() string {
 type NewValidator struct {
 	Description       `json:"description" yaml:"description"`
 	CommissionRates   `json:"commission" yaml:"commission"`
-	MinSelfDelegation *big.Int           `json:"min_self_delegation" yaml:"min_self_delegation"`
-	StakingAddress    common.Address     `json:"staking_address" yaml:"staking_address"`
-	PubKey            shard.BlsPublicKey `json:"validating_pub_key" yaml:"validating_pub_key"`
-	Amount            *big.Int           `json:"amount" yaml:"amount"`
+	MinSelfDelegation *big.Int       `json:"min_self_delegation" yaml:"min_self_delegation"`
+	StakingAddress    common.Address `json:"staking_address" yaml:"staking_address"`
+	PubKey            bls.PublicKey  `json:"validating_pub_key" yaml:"validating_pub_key"`
```
rlan35: don't change this. we always use 48 bytes for pubkey, don't use the struct object.

node/worker/worker.go
```diff
@@ -145,63 +144,6 @@ func (w *Worker) SelectTransactionsForNewBlock(newBlockNum uint64, txs types.Tra
 	return selected, unselected, invalid
 }
 
-// SelectStakingTransactionsForNewBlock selects staking transactions for new block.
-func (w *Worker) SelectStakingTransactionsForNewBlock(
```
rlan35: use transaction flow for staking transaction too. So don't remove these.

core/types/block.go
chaosma: removed for extV3

internal/params/config.go
chaosma: removed

PR #1722, opened 2019-10-12
[testnet] shrink the size of the testnet

LeoHChen:


fxfactorial:
I think we should alternative each testnet release from a small sized testnet & big sized testnet. Could be that some problems only expose themselves at larger scale, poor man's random sampling.

PR #1726, opened 2019-10-13
Pr fn key swap

AndyBoWu:
swap 13 fn nodes' addresses and keys


harmony-ek:

internal/configs/sharding/mainnet.go
```diff
@@ -48,6 +49,9 @@ type mainnetSchedule struct{}
 
 func (mainnetSchedule) InstanceForEpoch(epoch *big.Int) Instance {
 	switch {
+	case epoch.Cmp(big.NewInt(mainnetV1_5Epoch)) >= 0:
+		//
```
LeoHChen: what's the estimated time of 48?  It has to happen after rolling update on Tuesday.  I suggest to use 49 or higher.

AndyBoWu: Will recalculate a proper epoch number. Also, might need to add two more nodes per Li's request.

PR #1738, opened 2019-10-16
update consensus and committee informatin for at end of every epoch

chaosma:
- update consensus and committee information at end of every epoch
- fix bls library source bug
- local test passed for new node join

### DIFF
--- a/node/node_syncing.go
+++ b/node/node_syncing.go
@@ -192,7 +192,7 @@ func (node *Node) DoBeaconSyncing() {
 				continue
 			}
 		}
-		node.beaconSync.SyncLoop(node.Beaconchain(), node.BeaconWorker, true)
+		node.beaconSync.SyncLoop(node.Beaconchain(), node.BeaconWorker, true, nil)
 		time.Sleep(BeaconSyncFrequency * time.Second)
 	}
 }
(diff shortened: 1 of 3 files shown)


scripts/setup_bls_build_flags.sh
```diff
@@ -35,10 +35,3 @@ case $OS in
       export DYLD_FALLBACK_LIBRARY_PATH=$LD_LIBRARY_PATH
       ;;
 esac
```
LeoHChen: why remove this?

chaosma: source scripts/setup_bls_build_flags.sh   will fail if not removing this

chaosma: I found this bug when I start a new node, because need to source bls library

LeoHChen: It can be fixed using "-v" I think.  Don't just remove it.

node/node_syncing.go
```diff
@@ -235,7 +235,9 @@ SyncingLoop:
 				node.Consensus.BlocksNotSynchronized()
 			}
 			node.stateSync.SyncLoop(bc, worker, false)
-			if node.NodeConfig.Role() == nodeconfig.ExplorerNode {
+			// update the consensus and committee information at the end of epoch including explorer node
```
rlan35: This is not enough right? This update happens only after SyncLoop finishes (which means it only happens after it's sync'ed, not during the sync)

chaosma:
Jenkins, test this please

node/node_syncing.go
chaosma: Right. I changed it into syncLoop, I think now it should be good

PR #1741, opened 2019-10-17
fn key swaps for epoch 54

AndyBoWu:
#### Test Coverage Data

* Before
* After


harmony-ek:

AndyBoWu:
This PR is good to go!!

There was a duplicated entry which was introduced in epoch 46. Fortunately, we have a script to do QC for this issue.

harmony-ek:
image

PR #1749, opened 2019-10-19
move syncing frequency as parameter in commandline flag

chaosma:


node/node.go
```diff
@@ -156,6 +156,10 @@ type Node struct {
 	peerRegistrationRecord map[string]*syncConfig // record registration time (unixtime) of peers begin in syncing
 	SyncingPeerProvider    SyncingPeerProvider
 
+	// syncing frequency parameters
+	syncFreq       int
+	beaconSyncFreq int
```
rlan35: add a default value to these too. Otherwise, it may cause infinite loop when it's 0.

Issue #1752, opened 2019-10-21
Peer Discovery returns the same set of peers every time - should return a new peers set every request


denniswon:
**Describe the bug**
Peer Discovery returns the same set of peers every time - should return a new peers set every request.
This is an issue with node syncing because when node syncing loop fails and retries from the peer discovery, it receives the same peers set and conducts retries with already-failed sync peer configs.

**To Reproduce**
1. Check out code with "sha1" commit
2. Build
3. run local test using "...." commands

**Expected behavior**
should return a new peers set every request for peers set

denniswon:
cc: @chaosma @LeoHChen

Issue #1313, opened 2019-08-09
verify blocks synced from state syncing to prevent blockchain corruption


LeoHChen:
The blocks sync'ed from state syncing have to be verified when adding them into blockchain.

PR #1753, opened 2019-10-21
remove incorrect quorum check but just use the max hash for new node syncing

denniswon:
## Issue

https://github.com/harmony-one/harmony/issues/1313

## TODO
https://github.com/harmony-one/harmony/issues/1752

### DIFF
--- a/cmd/client/wallet/main.go
+++ b/cmd/client/wallet/main.go
@@ -976,6 +976,8 @@ func submitTransaction(tx *types.Transaction, walletNode *node.Node, shardID uin
 		return err
 	}
 	fmt.Printf("Transaction Id for shard %d: %s\n", int(shardID), tx.Hash().Hex())
+	json, err := tx.MarshalJSON()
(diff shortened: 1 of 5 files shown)


api/service/syncing/syncing.go
```diff
@@ -635,10 +620,7 @@ func (ss *StateSync) generateNewState(bc *core.BlockChain, worker *worker.Worker
 // TODO: return error
 func (ss *StateSync) ProcessStateSync(startHash []byte, size uint32, bc *core.BlockChain, worker *worker.Worker) {
 	// Gets consensus hashes.
-	if !ss.GetConsensusHashes(startHash, size) {
-		utils.Logger().Debug().Msg("[SYNC] ProcessStateSync unable to reach consensus on ss.GetConsensusHashes")
-		return
-	}
+	ss.GetConsensusHashes(startHash, size)
```
LeoHChen: why log message is removed?

PR #1762, opened 2019-10-24
[WIP] Committee Assignment factored out, Block Rewarding factored out, Staked voting, EPoS logic

fxfactorial:
Caution: PR became larger than originally intended.

1. Factored out blockrewarder behind an interface
2. Factored out committee assignment behind an interface
3. golang naming conventions updates
4. Two leaking goroutines fixed
5. networkID not set in default node config, fixes lack of network ID in NodeMetaData RPC call
6. Removal of at least 1000 lines of code of dead code
7. Removal of core/resharding.*
8. Factor out logic from under `core` to own packages, namely `shard` & `committee`
9. Rename some types for clearer semantic of what they are
10. Rely on BeaconChain for supercommittee assignment
11. Mostly implement EPoS (but not enabled) 
12. Remove unused fields from some types


cmd/harmony/main.go
```diff
@@ -203,13 +208,13 @@ func passphraseForBls() {
 }
 
 func setupInitialAccount() (isLeader bool) {
-	genesisShardingConfig := core.ShardingSchedule.InstanceForEpoch(big.NewInt(core.GenesisEpoch))
+	genesisShardingConfig := shard.Schedule.InstanceForEpoch(big.NewInt(0))
```
rlan35: Why replace GenesisEpoch with 0?

node/node.go
```diff
@@ -554,7 +564,9 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, chainDBFactory shardc
 }
 
 // CalculateInitShardState initialize shard state from latest epoch and update committee pub keys for consensus and drand
-func (node *Node) CalculateInitShardState() (err error) {
+
+// LoadSuperCommitteeForLatestEpochOnChain ..
+func (node *Node) LoadSuperCommitteeForLatestEpochOnChain() (err error) {
```
rlan35: We may not sync'ed up yet. So saying "Current" is more appropriate than "Latest". Please change to LoadSuperCommitteeForCurrentEpochOnChain

consensus/consensus.go
```diff
@@ -24,8 +25,12 @@ const (
 
 // Consensus is the main struct with all states and data related to consensus process.
 type Consensus struct {
+	// Decider handles whether quorum achieved on the shard this consensus running on and decides on
+	// distribution of blockrewards - Both parameterized by quorum.Policy
 	Decider quorum.Decider
-
+	// CommitteeAssigner provides information about who the SuperCommittee members are, their public
+	// keys
+	CommitteeAssigner committee.MemberReader
```
rlan35: Why not CommitteeReader?

consensus/consensus_service.go
```diff
@@ -470,22 +470,21 @@ func (consensus *Consensus) getLeaderPubKeyFromCoinbase(header *block.Header) (*
...
 	consensus.getLogger().Info().Msg("[UpdateConsensusInformation] Updating.....")
 
 	if core.IsEpochLastBlockByHeader(header) {
 		// increase epoch by one if it's the last block
 		consensus.SetEpochNum(epoch.Uint64() + 1)
 		consensus.getLogger().Info().Uint64("headerNum", header.Number().Uint64()).
 			Msg("[UpdateConsensusInformation] Epoch updated for next epoch")
-		nextEpoch := new(big.Int).Add(epoch, common.Big1)
-		pubKeys = core.CalculatePublicKeys(nextEpoch, header.ShardID())
+		// nextEpoch := new(big.Int).Add(epoch, common.Big1)
```
rlan35: Why change the logic to not read for next epoch?

consensus/quorum/one-node-one-vote.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// Policy ..
+func (v *uniformVoteWeight) Policy() Policy {
+	return SuperMajorityVote
+}
+
+// IsQuorumAchieved ..
+func (v *uniformVoteWeight) IsQuorumAchieved(p Phase) bool {
+	return v.SignatoriesCount(p) >= v.QuorumThreshold()
```
rlan35: Shall we just use Signers instead of Signatories? for simplicity and consistency with other places.

consensus/quorum/one-node-one-vote.go
```diff
@@ -0,0 +1,64 @@
...
+
+	last := big.NewInt(0)
+	count := big.NewInt(int64(len(earners)))
+
+	for i, account := range earners {
+		cur := big.NewInt(0)
+		cur.Mul(Pie, big.NewInt(int64(i+1))).Div(cur, count)
+		diff := big.NewInt(0).Sub(cur, last)
+		hook(common.Address(account), diff)
+		payout = big.NewInt(0).Add(payout, diff)
```
rlan35: will payout be differnet from Pei?

consensus/quorum/one-node-one-vote.go
```diff
@@ -0,0 +1,64 @@
...
+) (payout *big.Int) {
+
+	last := big.NewInt(0)
+	count := big.NewInt(int64(len(earners)))
+
+	for i, account := range earners {
+		cur := big.NewInt(0)
+		cur.Mul(Pie, big.NewInt(int64(i+1))).Div(cur, count)
+		diff := big.NewInt(0).Sub(cur, last)
+		hook(common.Address(account), diff)
```
rlan35: Please add comment to describe what the hook() is.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/harmony-one/harmony/staking/effective"
+)
+
+var (
+	twoThirds = numeric.NewDec(2).QuoInt64(3)
+)
+
+type stakedVoter struct {
+	isActive, isHarmonyNode bool
+	effective               numeric.Dec
```
rlan35: I bet this is the amount of stake? Why it's Dec? It should be Int as it's the number of token, right? (Please add comment)

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+type stakedVoteWeight struct {
+	SignatureReader
+	// EPOS based staking
+	validatorStakes            map[*bls.PublicKey]stakedVoter
```
rlan35: Is mapping by *bls.PublicKey safe? How about mapping by the bytes of the key?

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+	totalEffectiveStakedAmount numeric.Dec
```
rlan35: same here. The token amount should be big.Int. (in the number of Nanos/Wei)

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+// Policy ..
+func (v *stakedVoteWeight) Policy() Policy {
+	return SuperMajorityStake
+}
+
+// We must maintain 2/3 quoroum, so whatever is 2/3 staked amount,
+// we divide that out & you
+// IsQuorumAchieved ..
+func (v *stakedVoteWeight) IsQuorumAchieved(p Phase) bool {
+	return true
```
rlan35: no TODO here?

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// QuorumThreshold ..
+func (v *stakedVoteWeight) QuorumThreshold() int64 {
+	return v.totalEffectiveStakedAmount.Mul(twoThirds).Int64()
+}
+
+// RewardThreshold ..
+func (v *stakedVoteWeight) IsRewardThresholdAchieved() bool {
+	return false
```
rlan35: no TODO？

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+// We must maintain 2/3 quoroum, so whatever is 2/3 staked amount,
+// we divide that out & you
+// IsQuorumAchieved ..
+func (v *stakedVoteWeight) IsQuorumAchieved(p Phase) bool {
+	return true
+}
+
+// QuorumThreshold ..
+func (v *stakedVoteWeight) QuorumThreshold() int64 {
+	return v.totalEffectiveStakedAmount.Mul(twoThirds).Int64()
```
rlan35: Is this returning the number of token needed ? Then it should be big.Int.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+var (
+	hSentinel          = big.NewInt(0)
+	hEffectiveSentinel = numeric.ZeroDec()
+)
+
+// Award ..
+func (v *stakedVoteWeight) Award(
+	Pie *big.Int, earners []common.Address, hook func(earner common.Address, due *big.Int),
+) *big.Int {
+	return nil
```
rlan35: Add TODO, please.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+
+	median := effective.Median(realStakes)
+	for _, voter := range v.validatorStakes {
+		if !voter.isHarmonyNode {
+			voter.effective = effective.Stake(median, voter.effective)
+			totalAmount.Add(voter.effective)
+		}
+	}
+
+	//
```
rlan35: add todo if it's not finished.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -0,0 +1,104 @@
...
+
+}
+
+func (v *stakedVoteWeight) ToggleActive(*bls.PublicKey) bool {
+
```
rlan35: ditto

core/resharding.go
```diff
@@ -1,259 +0,0 @@
...
-		if id < len(ss.shardState) {
-			ss.shardState[id].NodeList = append(ss.shardState[id].NodeList, nid)
-		} else {
-			utils.Logger().Error().Int("id", id).Int("shardState Count", len(ss.shardState)).Msg("assignNewNodes index out of range")
-		}
-	}
-}
-
-// cuckooResharding uses cuckoo rule to reshard X% of active committee(shards) into inactive committee(shards)
-func (ss *ShardingState) cuckooResharding(percent float64) {
```
rlan35: Why remove all the resharding code? Some will be needed when we do the actual resharding.

core/values/blockchain.go
```diff
@@ -1,8 +1,6 @@
 package values
 
 const (
-	// BeaconChainShardID is the ShardID of the BeaconChain
-	BeaconChainShardID = 0
```
rlan35: why remove?

consensus/consensus_service.go
fxfactorial: Because I didn't get around to asking question: What exactly is meaning of giving me committee of next epoch in staking? I don't know the next epoch's commitee until beacon chain gives it to me, and non shard-0 nodes presumably call this.

cmd/harmony/main.go
fxfactorial: Mistake; returned back.

node/node.go
fxfactorial: Done

consensus/consensus.go
fxfactorial: Better, assigner was older name I thought of using; done.

consensus/quorum/one-node-one-vote.go
fxfactorial: Done

node/node.go
```diff
@@ -503,9 +503,15 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, chainDBFactory shardc
 		node.recentTxsStats = make(types.RecentTxsStats)
 		node.TxPool = core.NewTxPool(core.DefaultTxPoolConfig, node.Blockchain().Config(), blockchain)
 		node.CxPool = core.NewCxPool(core.CxPoolSize)
+
+		// SetCommittee members
 		node.Worker = worker.New(node.Blockchain().Config(), blockchain, chain.Engine)
 
-		if node.Blockchain().ShardID() != values.BeaconChainShardID {
+		// Need to pick the right committee assignment - right
```
rlan35: Is this a TODO?

consensus/quorum/one-node-one-vote.go
fxfactorial: That's a good question; in the ideal case; they should always be the same. But possible that payout is less, because Pie is just the total size of the block-reward. Depending on how slashing implemented, it could be less.

node/node_genesis.go
```diff
@@ -34,28 +34,32 @@ const (
...
+
 type genesisInitializer struct {
 	node *Node
 }
 
 // InitChainDB sets up a new genesis block in the database for the given shard.
 func (gi *genesisInitializer) InitChainDB(db ethdb.Database, shardID uint32) error {
-	shardState := core.CalculateInitShardState()
-	if shardID != 0 {
+	shardState := committee.IncorporatingStaking.Read(big.NewInt(0))
```
rlan35: What does IncorporatingStaking mean? Any better names？

consensus/quorum/one-node-one-vote.go
fxfactorial: done

consensus/quorum/one-node-staked-vote.go
fxfactorial: Preference of using the Dec (wrapper around big.Int) is ease of use of it, good methods it provides.

consensus/quorum/one-node-staked-vote.go
fxfactorial: It should be safe, if concerned about pointer usage then invariant of type safety in code should be maintained because all the places that added a PublicKey used bls.PublicKey which would .Deserialize as the bytes, so I moved it up one level and let the called functions that operate with quorum handle the call to .Deserialize instead of prior to call it

node/node_resharding.go
```diff
@@ -1,300 +0,0 @@
...
-	err = syscall.Exec(execFile, args, os.Environ())
-	if err != nil {
-		utils.Logger().Error().
-			Err(err).
-			Msg("Failed to restart program after resharding")
-	}
-	panic("syscall.Exec() is not supposed to return")
-}
-
-func getRestartArguments(myShardID uint32) []string {
```
rlan35: The restart part will be needed when real resharding is added.

consensus/quorum/one-node-staked-vote.go
fxfactorial: Didn't because will be implementing soon, should I do TODOs even for things I will be doing soon?

node/node_syncing.go
```diff
@@ -160,15 +160,20 @@ func (p *LocalSyncingPeerProvider) SyncingPeers(shardID uint32) (peers []p2p.Pee
...
 func (node *Node) DoBeaconSyncing() {
-	go func(node *Node) {
+	go func() {
 		// TODO ek – infinite loop; add shutdown/cleanup logic
 		for {
 			select {
 			case beaconBlock := <-node.BeaconBlockChannel:
 				node.beaconSync.AddLastMileBlock(beaconBlock)
+				ss, sE := beaconBlock.Header().GetShardState()
+				fmt.Println("Beacon suite")
```
rlan35: Remove debugging print.

node/node_syncing.go
```diff
@@ -160,15 +160,20 @@ func (p *LocalSyncingPeerProvider) SyncingPeers(shardID uint32) (peers []p2p.Pee
...
+				if mems := ss.FindCommitteeByID(node.Blockchain().ShardID()); sE == nil && mems != nil {
+					node.Consensus.Decider.UpdateVotingPower(mems.NodeList)
```
rlan35: This should happen when the beacon block is actually accepted into the chain. Not here.

consensus/quorum/one-node-staked-vote.go
fxfactorial: in staked case yes, but quorum single vote case no. So I will make it both use the big.Int case

node/node_test.go
```diff
@@ -35,7 +35,7 @@ func TestNewNode(t *testing.T) {
 	}
 	decider := quorum.NewDecider(quorum.SuperMajorityVote)
 	consensus, err := consensus.New(
-		host, values.BeaconChainShardID, leader, blsKey, decider,
+		host, shard.BeaconChainID, leader, blsKey, decider,
```
rlan35: Actually this field should be "shard id", so if you want to refactor, better be "BeaconShardID". Because ChainID is a different thing than ShardID.

core/resharding.go
fxfactorial: Because it was confusing, not used and now new setup of resharding is ready. Can always go back in git history to pull this code. Having less code in place makes it easier to understand what is happening in big picture because don't need to worry if used being used or not. We can have a new package under shard, (like committee) that can provide own place; makes `core` smaller

core/values/blockchain.go
fxfactorial: Mistake, fixed

node/node.go
fxfactorial: leftover thinking outloud, removed

node/node_genesis.go
fxfactorial: Like, 'a committee incorporating the use of staking'. Maybe something like like `committee.WithStakingEnabled` or if assume that this value will always be the singular instance of `MemberReader`, then could say `committee.Reader`

node/node_resharding.go
fxfactorial: Removed because not used at moment & I see resharding being done in own package - but can add back

node/node_syncing.go
fxfactorial: I thought that could assume `node.beaconSync.AddLastMileBlock(beaconBlock)`  always succeeds, so I will do it in respective downstream of `AddLastMileBlock`

node/node_test.go
fxfactorial: right, confusing, done

PR #1764, opened 2019-10-24
update gomod

denniswon:
builds and consensus reached fine


denniswon:
not needed. sorry. it was my local setup problem

PR #1766, opened 2019-10-25
Update Pangaea/testnet epoch/feature schedule

harmony-ek:
Now it's 75 blocks/epoch, without epoch-0 exception.  At 8 seconds/block, it amounts to 10 minutes/epoch.

Resize testnet to 100 nodes/shard * 3 shards = 300 nodes.  Shard 0/1 are Harmony-operated for now; shard 2 is operated/managed by P-OPS.

Also disable crosslinks and staking transactions because they are not used yet, and enable cross-shard transaction at genesis.

We reuse existing testnet keys.

### DIFF
--- a/internal/configs/sharding/testnet.go
+++ b/internal/configs/sharding/testnet.go
@@ -15,9 +15,6 @@ var TestnetSchedule testnetSchedule
 type testnetSchedule struct{}
 
 const (
-	testnetV1Epoch = 1
-	testnetV2Epoch = 2
-
 	// 10 minutes per epoch (at 8s/block)
 	testnetBlocksPerEpoch = 75
 
@@ -38,10 +35,6 @@ const (
 
 func (testnetSchedule) InstanceForEpoch(epoch *big.Int) Instance {
 	switch {
-	case epoch.Cmp(big.NewInt(testnetV2Epoch)) >= 0:
-		return testnetV2
-	case epoch.Cmp(big.NewInt(testnetV1Epoch)) >= 0:
-		return testnetV1
 	default: // genesis
 		return testnetV0
 	}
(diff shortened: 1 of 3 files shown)


harmony-ek:
Part 1 of testnet reboot config update.  Key updates to follow.

LeoHChen:
Jenkins, test this please.

internal/params/config.go
```diff
@@ -34,9 +34,9 @@ var (
 	// TestnetChainConfig contains the chain parameters to run a node on the harmony test network.
 	TestnetChainConfig = &ChainConfig{
 		ChainID:        TestnetChainID,
-		CrossTxEpoch:   big.NewInt(1),
-		CrossLinkEpoch: big.NewInt(2),
-		StakingEpoch:   big.NewInt(3),
+		CrossTxEpoch:   big.NewInt(0),
+		CrossLinkEpoch: EpochTBD,
```
LeoHChen: why this is TBD? don't we need it to enable the crossshard?

internal/configs/sharding/testnet.go
```diff
@@ -48,40 +48,20 @@ func (testnetSchedule) InstanceForEpoch(epoch *big.Int) Instance {
...
 
 func (ts testnetSchedule) EpochLastBlock(epochNum uint64) uint64 {
-	blocks := ts.BlocksPerEpoch()
-	switch {
-	case epochNum == 0:
-		return testnetEpochBlock1 - 1
-	default:
-		return testnetEpochBlock1 - 1 + blocks*epochNum
-	}
+	return (ts.BlocksPerEpoch()+1)*epochNum - 1
```
LeoHChen: is it working for epoch 0?

harmony-ek: Good catch, will fix.

internal/params/config.go
harmony-ek: Cross-shard is enabled from epoch 0 (`CrossTxEpoch`); `CrossLinkEpoch` is for cross-links in beaconchain, which CX does not use yet.

PR #1755, opened 2019-10-22
Staking part2

chaosma:
- revert to previous design and fix conflicts
- add statedb operations
- add validator messages process

PR #1745, opened 2019-10-18
Validator api updates

flicker-harmony:
## Issue

Adds lots of hmy functionality, fixes some old stuff, stacking validator api: block signers, tx history implementation made cleaner, shard epoch and leader, number of blocks signed by validator for last 1500 blocks (GetSignedBlocks), validator methods like stake, info, address, delegators. Though last ones are not completely implemented because Chao and Edgar working on them.

## Test

Localnet, postman. Might be small fixes, I didn't complete testing.

#### Test Coverage Data

* Before
* After

## TODO

Need to add getTotalSupply, I need formula for total stacking supply of ONE in network.

PR #1767, opened 2019-10-25
Make Travis-CI see if binaries in ./cmd build fine

harmony-ek:
SSIA.  This is because the master branch needs to be deployable onto Pangaea/testnet at all times.

## Tests

Since the master branch is broken, the commit that introduces this change (a162c6f) fails Travis build.  Once the master branch builds again, I will merge from master then Travis should pass.

### DIFF
--- a/scripts/travis_checker.sh
+++ b/scripts/travis_checker.sh
@@ -79,6 +79,15 @@ else
 	ok=false
 fi
 
+echo "Building binaries in ./cmd/..."
+if go build ./cmd/...
+then
+	echo "go build succeeded."
+else
+	echo "go build FAILED!"
+	ok=false
+fi
+
 echo "Running go test..."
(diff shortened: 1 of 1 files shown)


harmony-ek:
Since #1755 broke the build, merge of this PR is blocked until the build is fixed.

harmony-ek:
@LeoHChen wrote:

> look good, but how did you test this one? Any test result.

Updated the initial comment.

harmony-ek:
I see the build is already checked directly from `.travis.yml`; rescinding this PR.

As for how the master branch broke: It's because #1745 (whose change uses the `ValidatorPubKeys` field) was not retested after #1755 (which retires/renames the field) was merged into master.  Apparently Travis-CI does not retest PRs upon master push.  Will check if Jenkins does this.

PR #1772, opened 2019-10-25
Disable throttling for txns

rlan35:
Transactions will be limited by gasLimit of the block. But all amount/account and num/account limit is not needed anymore.

### DIFF
--- a/node/worker/worker.go
+++ b/node/worker/worker.go
@@ -62,15 +62,15 @@ func (w *Worker) throttleTxs(selected types.Transactions, recentTxsStats types.R
 		sender = msg.From()
 	}
 
+	// do not throttle transactions if disabled
+	if !txsThrottleConfig.EnableTxnThrottling {
+		return sender, shardingconfig.TxSelect
+	}
+
 	// already selected max num txs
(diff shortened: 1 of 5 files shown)


node/worker/worker.go
```diff
@@ -62,17 +62,17 @@ func (w *Worker) throttleTxs(selected types.Transactions, recentTxsStats types.R
...
+		return sender, shardingconfig.TxSelect
+	}
+
 	// already selected max num txs
 	if len(selected) > txsThrottleConfig.MaxNumTxsPerBlockLimit {
 		utils.Logger().Info().Str("txId", tx.Hash().Hex()).Int("MaxNumTxsPerBlockLimit", txsThrottleConfig.MaxNumTxsPerBlockLimit).Msg("Throttling tx with max num txs per block limit")
 		return sender, shardingconfig.TxUnselect
 	}
 
-	// do not throttle transactions if disabled
```
denniswon: why did you remove this?

rlan35: I didn't remove. It's moved upper.

PR #1774, opened 2019-10-26
stop adding new txns when out of gas

rlan35:
make our target block gas limit to be around 100M gas. 10 times bigger than current ethereum's gas limit of 10M

Also stop processing txns with gasPrice == 0

### DIFF
--- a/node/worker/worker.go
+++ b/node/worker/worker.go
@@ -123,17 +123,21 @@ func (w *Worker) SelectTransactionsForNewBlock(newBlockNum uint64, txs types.Tra
 			invalid = append(invalid, tx)
 
 		case shardingconfig.TxSelect:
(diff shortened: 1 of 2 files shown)


node/worker/worker.go
```diff
@@ -419,8 +430,8 @@ func New(config *params.ChainConfig, chain *core.BlockChain, engine consensus_en
 		chain:   chain,
 		engine:  engine,
 	}
-	worker.gasFloor = 500000000000000000
-	worker.gasCeil = 1000000000000000000
+	worker.gasFloor = 80000000
+	worker.gasCeil = 120000000
```
fxfactorial: What was meaning of old numbers , new numbers ?

rlan35: Old number is just a meaninglessly large number.

PR #1782, opened 2019-10-28
Add non-zero exit to test/deploy.sh if Cx test failed

Daniel-VDM:
## Issue

Jenkins will not pick up a failure if script does not exit with a non-zero exit code.

### DIFF
--- a/test/deploy.sh
+++ b/test/deploy.sh
@@ -193,19 +193,23 @@ done < $config
 
 if [ "$DOTEST" == "true" ]; then
    echo "waiting for some block rewards"
-   sleep 60
+   sleep 110
    i=1
    echo "launching wallet cross shard transfer test"
    while [ $i -le $NUM_TEST ]; do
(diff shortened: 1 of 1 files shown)


test/deploy.sh
```diff
@@ -208,6 +208,10 @@ if [ "$DOTEST" == "true" ]; then
    sleep 20
    check_result
    [ -e $RESULT_FILE ] && cat $RESULT_FILE
+   if [ "$err" = true ] ; then
```
LeoHChen: can you use 
```"$err" == "true"```
 to be consistent with previous check.

Daniel-VDM: Addresed in latest

Daniel-VDM:
In my testing, waiting 20 seconds is *sometimes* not long enough for the Cx to go through. And infact this causes *some* Jenkins built tests to fail. Therefore, I always wait 25 seconds to check Cx on localnet.

Daniel-VDM:
In my testing, waiting 20 seconds is *sometimes* not long enough for the Cx to go through.

Daniel-VDM:
Jenkins, test this please.

Issue #1771, opened 2019-10-25
[Feature] latest/zerolog-*.log is automatically rotated but latest/validator-*.log is not


nyetwurk:
Currently zerolog is autorotated by harmony, but validator-*.log is not.

This means it needs to be externally rotated via something like logrotate, but the other does not.

This should be one or the other for both.

nyetwurk:
Having experimented with this, bad things happen when native zerolog rotation is disabled; logrotate CANNOT rotate out logs created by zerolog, it holds them open via what appears to be mmap(), so if logrotate does copy/truncate, the next time zerolog writes out an entry, the entire file gets recreated with leading zeros.

yelllowsin:
Closing this one as logs have been consolidated into zerolog-validator*.log file and being rotated accordingly. Feel free to reopen if any issue remains.

PR #1785, opened 2019-10-30
Replace ethereum log with zerolog to resolve logging rotation issue

flicker-harmony:
## Issue

https://github.com/harmony-one/harmony/issues/1771 we should either have no log rotation or log rotation for both zerolog and ethereum validator log, but the last one support for rotation is in question. Discussed with @nyetwurk and Leo that we should be consistent, better to allow for nodes to enable or disable rotation but at least to have disabled or enabled for both by default. Also when zerolog is rotated, files after first one gets uncompressed, which is bad. Easiest solution is to increase max log size (to 100 GB) for zerolog to disable rotation for now. 

## Test

Will need to run on testnet to see if log isn't rotated but it's obviously wouldn't as log size got increased by 1000 times.

#### Test Coverage Data

* Before
* After


flicker-harmony:
> > files after first one gets uncompressed
> 
> Actually it zerolog DOES compress the newly rotated log.
> The issue is a bit more confusing than that. By default, logrotate does this, but zerolog does _not_.
> 
> The reason logrotate does this is that immediately after rotation, the current log is 0. This means that the most recently rotated out log is the only one with data, so log rotate does NOT compress that one (so that scripts and utilities that scrape logs are always guaranteed to have uncompressed data to watch)
> 
> Zerolog does NOT behave like this, the newest rotated log is compressed.
> 
> I am a bit wary of disabling via making the max value high; is there a better way to disable actually disable it?

You said that validator log doesn't get rotated at all and zerolog does rotation, so I thought compression is referred to zerolog. Setting max value high is easier because it doesn't require cutting functionality from the code and doesn't create more confusion, simple solution.

LeoHChen:
Is this change applicable to both validator log and zerolog log?  I don't think we should set it to a very large number for zerolog as it is auto rotate.  I think one thing we should fix , is to redirect all validator logs to zerolog, and we have just one log file to maintain.  @flicker-harmony , can you finish that task to redirect all log messages in validator.log to zerolog?

nyetwurk:
IIRC validator logs still have BINGOs in them because of legacy monitoring that looks for bingos there and not in zero logs, but I could be wrong.

BTW here is a dump showing the difference between logrotate and zerolog rotation:
```
validator-34.94.115.155-9000.log
validator-34.94.115.155-9000.log.1
validator-34.94.115.155-9000.log.2.gz
validator-34.94.115.155-9000.log.3.gz
validator-34.94.115.155-9000.log.4.gz
validator-34.94.115.155-9000.log.5.gz
validator-34.94.115.155-9000.log.6.gz
validator-34.94.115.155-9000.log.7.gz
```
```
zerolog-validator-34.94.115.155-9000-2019-10-02T14-45-13.342.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-04T15-51-39.185.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-06T15-59-26.987.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-08T16-11-10.657.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-10T16-57-44.132.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-12T18-12-58.926.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-14T18-56-21.290.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-16T19-41-23.007.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-18T20-18-05.822.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-20T22-04-37.078.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-22T22-28-44.328.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-24T23-18-16.930.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-26T23-59-12.146.log.gz
zerolog-validator-34.94.115.155-9000-2019-10-29T01-50-32.814.log.gz
zerolog-validator-34.94.115.155-9000.log
```

Notice that for zerolog, all non-current logs are compressed, but for logrotate, current, and current.1 are uncompressed, all others are compressed.

nyetwurk:
This may not work. Experimentally, moving log out from underneath zero log via logrotate DOES NOT WORK, so if zerolog rotation is disabled, there is NO way to rotate it.

flicker-harmony:
I pushed changes that removes ethereum log to zerolog almost everywhere. We probably needed to that anyway. There are still places like connection log, p2p and some log handlers which are created using Ethereum log and replacing that is less trivial, @harmony-ek should take about that.

cmd/harmony/main.go
```diff
@@ -76,7 +75,7 @@ var (
 	ip               = flag.String("ip", "127.0.0.1", "ip of the node")
 	port             = flag.String("port", "9000", "port of the node.")
 	logFolder        = flag.String("log_folder", "latest", "the folder collecting the logs of this execution")
-	logMaxSize       = flag.Int("log_max_size", 100, "the max size in megabytes of the log file before it gets rotated")
+	logMaxSize       = flag.Int("log_max_size", 100000, "the max size in megabytes of the log file before it gets rotated")
```
nyetwurk: Please revert this for the time being, for the reasons I posted in the PR above.

nyetwurk: Or, alternately, make it an option.

flicker-harmony: It's an arg, so you can specify it.

nyetwurk: Oh sorry, I didn't see that.

nyetwurk: I'm still not sure people will want that, 1g is a pretty big log file, and if logrotate can't rotate it for you....

flicker-harmony: I guess 10gb might be better yeah

nyetwurk: I honestly think you may want to just leave it alone to avoid surprises, especially if validator log is going away, there is no more inconsistency, and I can get rid of my own logrotate games.

node/Find Results
```diff
@@ -0,0 +1,448 @@
+Searching 2456 files for "utils.GetLogInstance("
```
fxfactorial: looks like test code? probably remove from PR right

PR #1806, opened 2019-11-06
Use TxPool for worker's transaction proposal

rlan35:
Remove the logic of txn throttling and replace it with TxPool logic.

Also change Worker's commit logic from "first select then commit" to "only commit", the txPool will automatically update the pending txns when new block is confirmed.

Tested locally and made sure that txpool's pending transactions are looking good.


node/worker/worker.go
```diff
@@ -92,107 +96,97 @@ func (w *Worker) throttleTxs(selected types.Transactions, recentTxsStats types.R
...
+			continue
 		}
 
-		// log invalid or unselected txs
-		if flag == shardingconfig.TxUnselect || flag == shardingconfig.TxInvalid {
-			utils.Logger().Info().Str("txId", tx.Hash().Hex()).Str("txThrottleFlag", flag.String()).Msg("Transaction Throttle flag")
+		logs, err := w.commitTransaction(tx, coinbase)
+		sender, _ := common2.AddressToBech32(from)
+		switch err {
+		case core.ErrGasLimitReached:
```
denniswon: we are doing a double check for ErrGasLimitReached. we should remove the gas limit check above and have it failed here for modularity.

node/worker/worker.go
```diff
@@ -92,107 +96,97 @@ func (w *Worker) throttleTxs(selected types.Transactions, recentTxsStats types.R
...
-		_, err := w.commitStakingTransaction(tx, coinbase)
-		if err != nil {
-			w.current.state.RevertToSnapshot(snap)
-			invalid = append(invalid, tx)
-			utils.Logger().Error().Err(err).Str("stakingTxId", tx.Hash().Hex()).Msg("Commit staking transaction error")
-		} else {
-			selected = append(selected, tx)
-			utils.Logger().Info().Str("stakingTxId", tx.Hash().Hex()).Uint64("txGasLimit", tx.Gas()).Msg("StakingTransaction gas limit info")
+	// STAKING - only beaconchain process staking transaction
+	if w.chain.ShardID() == values.BeaconChainShardID {
```
denniswon: right now, this part is no-op right? we just log the coalescedLogs here?

test/chain/main.go
```diff
@@ -192,143 +199,6 @@ func playFaucetContract(chain *core.BlockChain) {
...
-	fmt.Println(state.GetBalance(allRandomUserAddress[2]))
-	fmt.Println("faucet contract balance (unchanged):")
-	fmt.Println(state.GetBalance(faucetContractAddress))
-	fmt.Printf("stake contract balance (should downup by %s)\n", withdraw)
-	fmt.Println(state.GetBalance(stakeContractAddress))
-	fmt.Println("stake address balance :")
-	fmt.Println(state.GetBalance(StakingAddress))
-}
-
-func playStakingContract(chain *core.BlockChain) {
```
denniswon: why remove this test?

denniswon:
@rlan35 so throttling completely removed now?
If yes, then we should remove all relevant code lines in the repo for future confusion prevention.

core/blockchain.go
```diff
@@ -1155,7 +1155,6 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
 // After insertion is done, all accumulated events will be fired.
 func (bc *BlockChain) InsertChain(chain types.Blocks, verifyHeaders bool) (int, error) {
 	n, events, logs, err := bc.insertChain(chain, verifyHeaders)
-	bc.PostChainEvents(events, logs)
```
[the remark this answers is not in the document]
rlan35: Yes, technically the Insertion is not done until all the following operations are successful.

core/tx_pool.go
```diff
@@ -301,12 +301,11 @@ func (pool *TxPool) loop() {
 		case ev := <-pool.chainHeadCh:
 			if ev.Block != nil {
 				pool.mu.Lock()
-				//if pool.chainconfig.IsHomestead(ev.Block.Number()) {
-				//	pool.homestead = true
-				//}
-				//pool.reset(head.Header(), ev.Block.Header())
-				//head = ev.Block
-
+				if pool.chainconfig.IsS3(ev.Block.Epoch()) {
```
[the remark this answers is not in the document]
rlan35: s3 fork is our rollup fork for all ethereum forks up to Constantinople, so it includes Homestead.

node/node.go
```diff
@@ -277,18 +277,14 @@ func (node *Node) tryBroadcast(tx *types.Transaction) {
...
-	for _, tx := range newTxs {
-		if _, ok := node.pendingTransactions[tx.Hash()]; !ok {
-			node.pendingTransactions[tx.Hash()] = tx
-		}
-		if len(node.pendingTransactions) > txPoolLimit {
-			break
-		}
-	}
+
+	node.TxPool.AddRemotes(newTxs)
```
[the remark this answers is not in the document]
rlan35: local is for txns submitted from local client, basically the client runner's own txn. Remote is the txn from the network. local txns has special logic for persistence and so on. For now, we are not supporting local txns.

node/node_handler_test.go
```diff
@@ -65,8 +69,9 @@ func TestVerifyNewBlock(t *testing.T) {
 	}
 	node := New(host, consensus, testDBFactory, false)
 
-	selectedTxs, stks := node.getTransactionsForNewBlock(common.Address{})
-	node.Worker.CommitTransactions(selectedTxs, stks, common.Address{})
+	txs := make(map[common.Address]types.Transactions)
+	stks := types2.StakingTransactions{}
+	node.Worker.CommitTransactions(txs, stks, common.Address{})
```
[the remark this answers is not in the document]
rlan35: This is the existing logic, it's just for testing out whether the worker flow works or not. I didn't change the logic.

node/node_newblock.go
```diff
@@ -80,10 +82,21 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 	// Update worker's current header and state data in preparation to propose/process new transactions
 	coinbase := node.Consensus.SelfAddress
 
-	// Prepare transactions including staking transactions
-	selectedTxs, selectedStakingTxs := node.getTransactionsForNewBlock(coinbase)
+	// Prepare transactions including staking transactions\
```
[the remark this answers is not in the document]
rlan35: my typo. removed.

node/worker/worker.go
rlan35: it's not the same and this logic is following what ethereum is doing. The first check is block level check to say that there is no way any more txn can be added. This current check is for a specific txn. If the current txn is big and it may fail the check, but it doesn't mean the next txn won't succeed.

node/worker/worker.go
rlan35: yes.

test/chain/main.go
rlan35: These are old test for  the previous staking logic implemented via smart contract. It's deprecated.

rlan35:
> @rlan35 so throttling completely removed now?
> If yes, then we should remove all relevant code lines in the repo for future confusion prevention.

Done.

rlan35:
> Please add test results and description for this PR!

Done.

core/blockchain.go
```diff
@@ -1197,6 +1196,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks, verifyHeaders bool) (int,
 		}
 	}
 
+	bc.PostChainEvents(events, logs)
```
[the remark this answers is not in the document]

mikedoan:
I would recommend to deploy this pr to testnet as it's pretty critical

internal/configs/sharding/fixedschedule.go
```diff
@@ -55,43 +52,6 @@ func (s fixedSchedule) RandomnessStartingEpoch() uint64 {
...
-func (s fixedSchedule) MaxNumRecentTxsPerAccountLimit() uint64 {
-	return mainnetMaxNumRecentTxsPerAccountLimit
-}
-
-func (s fixedSchedule) MaxTxPoolSizeLimit() int {
-	return mainnetMaxTxPoolSizeLimit
-}
-
-func (s fixedSchedule) MaxNumTxsPerBlockLimit() int {
-	return mainnetMaxNumTxsPerBlockLimit
```
denniswon: wouldnt we still need some throttling of this kind? This sets max tx num to 8000 per block and this logic existed before I added additional throttling

node/node_handler_test.go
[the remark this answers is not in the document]
denniswon: oh this is a test class. hmm in the future, we should actually calling selecting txns from the txpool.

internal/configs/sharding/fixedschedule.go
denniswon: Oh i see. Below we now hardcoded that max number of txns per block. It was 8000 before and now 1000 I guess.

PR #1815, opened 2019-11-07
Reduce wait time for bootnode connection to 30s

rlan35:
Previously the retry time is 10 mins and it will block the whole client process for 10 mins if a single bootnode fails to connect. Reduce the retry time to 30s, so it won't appear that the node is stuck for a long time.

### DIFF
--- a/api/service/networkinfo/service.go
+++ b/api/service/networkinfo/service.go
@@ -41,8 +41,8 @@ type Service struct {
 
 // ConnectionRetry set the number of retry of connection to bootnode in case the initial connection is failed
 var (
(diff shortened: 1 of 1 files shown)


gaia:
This might seem obvious, but it isn't being addressed here? Will is proceed even if the unreachable bootnode is still down after 30 seconds?

rlan35:
> This might seem obvious, but it isn't being addressed here: will it proceed even if the unreachable bootnode is still down after 30 seconds?

Yes, it will be unblocked once the retry finishes: https://github.com/harmony-one/harmony/blob/master/api/service/networkinfo/service.go#L147

PR #1818, opened 2019-11-10
[committee] Factor out committee membership, provide entry for alternative committee membership

fxfactorial:
This is PR #1 of breaking this PR https://github.com/harmony-one/harmony/pull/1762 into smaller PRs.

1. Provide more fine grained control over committee membership
2. Maintain current committee membership as to not break mainnet, until stakingEpoch past
3. Remove dead code, often times was inconsistent with itself. 
4. With single spot for committee shard assignment, easy to now add new committee membership business logic via epoch rather than scattered.


node/node.go
```diff
@@ -545,44 +546,41 @@ func New(host p2p.Host, consensusObj *consensus.Consensus, chainDBFactory shardc
...
-
 	go node.bootstrapConsensus()
-
 	return &node
 }
 
-// CalculateInitShardState initialize shard state from latest epoch and update committee pub keys for consensus and drand
-func (node *Node) CalculateInitShardState() (err error) {
+// InitConsensusWithMembers initialize shard state from latest epoch and update committee pub keys for consensus and drand
+func (node *Node) InitConsensusWithMembers() (err error) {
```
rlan35: InitConsensusWithValidators

node/node_newblock.go
```diff
@@ -144,13 +145,15 @@ func (node *Node) proposeShardState(block *types.Block) error {
...
-	if !core.IsEpochLastBlock(block) {
+	if !shard.Schedule.IsLastBlock(block.Number().Uint64()) {
 		// We haven't reached the end of this epoch; don't propose yet.
 		return nil
 	}
-	nextEpoch := new(big.Int).Add(block.Header().Epoch(), common.Big1)
-	// TODO: add logic for EPoS
-	shardState, err := core.CalculateNewShardState(node.Blockchain(), nextEpoch)
+	// TODO Use ReadFromComputation
+	prevEpoch := new(big.Int).Sub(block.Header().Epoch(), common.Big1)
```
rlan35: Why use prevEpoch, instead of the original nextEpoch?

node/node_resharding.go
```diff
@@ -51,14 +51,17 @@ func (node *Node) validateNewShardState(block *types.Block) error {
...
-	if block.ShardID() == 0 {
+	if block.ShardID() == shard.BeaconChainShardID {
 		// Beacon validators independently recalculate the master state and
 		// compare it against the proposed copy.
-		nextEpoch := new(big.Int).Add(block.Header().Epoch(), common.Big1)
 		// TODO ek – this may be called from regular shards,
 		//  for vetting beacon chain blocks received during block syncing.
 		//  DRand may or or may not get in the way.  Test this out.
-		expected, err := core.CalculateNewShardState(node.Blockchain(), nextEpoch)
+		expected, err := committee.WithStakingEnabled.ReadFromChain(
```
rlan35: The validation should be using a ground truth data, rather than reading what’s in the shard chain. Previously the truth is the calculation. Now the truth should be the beacon chain’s shard state. Here it should be reading from beacon chain.

shard/committee/assignment.go
```diff
@@ -0,0 +1,261 @@
...
+				&shard.StakedMember{big.NewInt(0)},
+			},
+		)
+	}
+
+	utils.Logger().Info().Ints("distribution of Stakers in Shards", fillCount)
+	return superComm, nil
+}
+
+func (def partialStakingEnabled) ReadPublicKeysFromChain(
```
rlan35: ReadPublicKeysFromDB is better.

shard/committee/assignment.go
```diff
@@ -0,0 +1,261 @@
...
+		committerKeys = append(committerKeys, committerKey)
+	}
+	return committerKeys, nil
+
+	return nil, nil
+}
+
+// ReadPublicKeysFromChain produces publicKeys of entire supercommittee per epoch, optionally providing a
+// shard specific subcommittee
+func (def partialStakingEnabled) ReadPublicKeysFromComputation(
```
rlan35: ComputePublicKeys. Read means directly fetch data without doing anything. Read...Computation is confusing. No need to follow the same pattern with the above method.

node/node_newblock.go
fxfactorial: The original logic confused me:

Original:
1. Take the block's header.epoch, then add one. 
https://github.com/harmony-one/harmony/blob/master/node/node_newblock.go#L151
2. Call CalculateNewShardState and decrement the epoch by 1
https://github.com/harmony-one/harmony/blob/master/core/resharding.go#L159

Calculate doesn't actually compute anything, just reads off the chain in non-epoch == 0 case. 

So actually it always ends up reading from same epoch, No?

PR #1819, opened 2019-11-11
[quorum] Staked quorum member

fxfactorial:
This is PR #2 of breaking this PR #1762 into smaller PRs, assumes PR#1 merged into master, then will rebase this one to reflect that.

1. Factored out quorum's single implm and provide implement for both plain one vote vs staked vote
2. Provide way for dependency injection to Decider from consensus

### DIFF
--- a/shard/committee/assignment.go
+++ b/shard/committee/assignment.go
@@ -21,12 +21,12 @@ import (
 // a shardID parameter
 const StateID = -1
 
-// MembershipList  ..
-type MembershipList interface {
-	ReadFromComputation(
+// ValidatorList ..
+type ValidatorList interface {
+	Compute(
 		epoch *big.Int, config params.ChainConfig, reader StakingCandidatesReader,
 	) (shard.State, error)
-	ReadFromChain(epoch *big.Int, reader ChainReader) (shard.State, error)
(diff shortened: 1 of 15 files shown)


shard/committee/assignment.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/harmony-one/harmony/shard"
+	staking "github.com/harmony-one/harmony/staking/types"
+)
+
+// StateID means reading off whole network when using calls that accept
+// a shardID parameter
+const StateID = -1
+
+// MembershipList  ..
+type MembershipList interface {
```
rlan35: ValidatorList?

shard/committee/assignment.go
```diff
@@ -0,0 +1,261 @@
...
+	ReadFromComputation(
```
rlan35: Compute

shard/committee/assignment.go
```diff
@@ -0,0 +1,261 @@
...
+		epoch *big.Int, config params.ChainConfig, reader StakingCandidatesReader,
+	) (shard.State, error)
+	ReadFromChain(epoch *big.Int, reader ChainReader) (shard.State, error)
```
rlan35: ReadFromDB

PR #1823, opened 2019-11-11
[beacon] Only change epoch when beaconchain changed to higher epoch

fxfactorial:
Only compute new committee when last block of shard0 (beaconchain)

### DIFF
--- a/node/worker/worker.go
+++ b/node/worker/worker.go
@@ -281,15 +281,35 @@ func (w *Worker) IncomingReceipts() []*types.CXReceiptsProof {
 	return w.current.incxs
 }
 
-// ProposeShardStateWithoutBeaconSync proposes the next shard state for next epoch.
-func (w *Worker) ProposeShardStateWithoutBeaconSync() shard.State {
-	if !shard.Schedule.IsLastBlock(w.current.header.Number().Uint64()) {
-		return nil
-	}
-	shardState, _ := committee.WithStakingEnabled.Compute(
-		new(big.Int).Add(w.current.header.Epoch(), common.Big1), *w.config, nil,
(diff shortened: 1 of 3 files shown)


node/worker/worker.go
```diff
@@ -281,15 +281,35 @@ func (w *Worker) IncomingReceipts() []*types.CXReceiptsProof {
...
+				new(big.Int).Add(w.current.header.Epoch(), common.Big1),
+				*w.config,
+				beacon,
+			)
+		}
+	default:
+		switch beacon.CurrentHeader().Epoch().Cmp(w.current.header.Epoch()) {
+		case 1:
+			nextCommittee, oops = committee.WithStakingEnabled.ReadFromDB(
+				w.current.header.Epoch(), beacon,
```
rlan35: This should be beacon.CurrentHeader().Epoch().  Not w.current.header.Epoch()

node/worker/worker.go
```diff
@@ -281,15 +281,35 @@ func (w *Worker) IncomingReceipts() []*types.CXReceiptsProof {
...
+	case shard.BeaconChainShardID:
+		if shard.Schedule.IsLastBlock(w.current.header.Number().Uint64()) {
+			nextCommittee, oops = committee.WithStakingEnabled.Compute(
+				new(big.Int).Add(w.current.header.Epoch(), common.Big1),
+				*w.config,
+				beacon,
+			)
+		}
+	default:
+		switch beacon.CurrentHeader().Epoch().Cmp(w.current.header.Epoch()) {
```
rlan35: This logic may work in normal case. But we need to add compatibility logic during the enabling of staking. Please add a TODO to take care of migration.

PR #1826, opened 2019-11-13
Implementation of EPoS 

fxfactorial:
EPoS implementation of shard state (super committee) assignment. See https://medium.com/harmony-one/introducing-harmonys-effective-proof-of-stake-epos-2d39b4b8d58 for explanation. Includes extension to original design of using multiple BLS keys for same one-address staker. 

Commit 37770e3 useful for testing locally - TODO need to incorporate that & JSON dumps as test cases later.


shard/committee/assignment.go
```diff
@@ -131,47 +140,156 @@ func with400Stakers(
...
+
+	unsortedStakes := make([]int, len(stakers))
+	eposStakes := make([]*big.Int, len(stakers))
+
+	for i, j := range stakers {
+		unsortedStakes[i] = int(j.Stake.Int64())
+		eposStakes[i] = j.Stake
+	}
+
+	s3 := effective.Apply(eposStakes)
```
rlan35: The Epos calculation should apply on the individual slot level. Not on the validator level.

shard/committee/assignment.go
```diff
@@ -131,47 +140,156 @@ func with400Stakers(
...
-		// scratchPad.DeserializeHexStr()
-		pubKey := shard.BlsPublicKey{}
-		pubKey.FromLibBLSPublicKey(scratchPad)
-		superComm[spot].NodeList = append(
-			superComm[spot].NodeList,
-			shard.NodeID{
-				top[i].Address,
+	shardBig := big.NewInt(int64(s.NumShards()))
+
+	for i := 0; i < len(s.FnAccounts()); i++ {
```
rlan35: This should be hardcoded 320. Not len(s.FnAccounts()). FnAccounts will be deprecated after staking.

shard/committee/assignment.go
```diff
@@ -131,47 +140,156 @@ func with400Stakers(
...
+		bucket := int(new(big.Int).Mod(stakers[i].Address.Big(), shardBig).Int64())
+		org := stakers[i].Stake
+		epos := big.NewInt(s3[i].TruncateInt64())
+		fmt.Println("stakes", org, epos)
+		superComm[bucket].NodeList[fillCount[bucket]] = shard.NodeID{
+			stakers[i].Address,
+			stakers[i].SlotPubKeys[0],
```
rlan35: stakers[i].SlotPubKeys[0]    the 0 here should really be an "i" to iterate over all keys and add them each as a separate node in the supperComm.

shard/committee/assignment.go
```diff
@@ -131,47 +140,156 @@ func with400Stakers(
...
+			epos,
+		}
+		fillCount[bucket]++
+	}
+
+	hAccounts := s.HmyAccounts()
+	offset := 0
+
+	for i := range fillCount {
```
rlan35: We should first fill in our node, 680 of them. And the rest is filled with open staking keys.

shard/committee/assignment.go
fxfactorial: I thought that is taken into account by: https://github.com/harmony-one/harmony/pull/1826/files#diff-5306cae6810bfd2a339378a4b76c1406R145

shard/committee/assignment.go
rlan35: Otherwise， our nodes will be put into different shards depending on the open staking accounts's count.

fxfactorial: excellent point, order matters

shard/committee/assignment.go
fxfactorial: true - will remedy

shard/committee/assignment.go
fxfactorial: actually just realized, my scheme doesn't work because it assume linearity, but epos is non-linear, so will change. (I initially thought of this way as an optimization on memory allocation)

internal/params/config.go
```diff
@@ -36,9 +36,10 @@ var (
 		ChainID:        TestnetChainID,
 		CrossTxEpoch:   big.NewInt(0),
 		CrossLinkEpoch: big.NewInt(0),
-		StakingEpoch:   EpochTBD,
-		EIP155Epoch:    big.NewInt(0),
-		S3Epoch:        big.NewInt(0),
+		// MinEpoch needed is at least 1, crashes on 0
```
rlan35: remove this comment?

numeric/decimal.go
```diff
@@ -482,11 +483,6 @@ func (d Dec) RoundInt64() int64 {
 	return chopped.Int64()
 }
 
-// RoundInt round the decimal using bankers rounding
```
rlan35: I have change in here. please merge.

rlan35: I need these funcs.

staking/effective/calculate.go
```diff
@@ -0,0 +1,80 @@
...
+		spread := numeric.NewDecFromBigInt(shortHand[staker].Stake).QuoInt64(slotsCount)
+		var i int64
+		for ; i < slotsCount; i++ {
+			eposedSlots = append(eposedSlots, SlotPurchase{staker, spread})
+		}
+	}
+	if len(eposedSlots) < len(shortHand) {
+		// WARN Should never happen
+	}
+	median := median(eposedSlots)
```
rlan35: The median should only be counting among the top 320 slots.

shard/committee/assignment.go
```diff
@@ -116,62 +125,148 @@ func preStakingEnabledCommittee(s shardingconfig.Instance) shard.State {
...
+		// WARN unlikely to happen in production but will happen as we are developing
+	}
+
+	for i := 0; i < stakedSlotsCount; i++ {
+		bucket := int(new(big.Int).Mod(staked[i].Address.Big(), shardBig).Int64())
+		slot := staked[i]
+		pubKey := essentials[slot.Address].SpreadAmong[slotUsage[slot.Address]-1]
+		slotUsage[slot.Address]--
+		// Keep going round till find an open spot
+		for j := bucket; ; j = (j + 1) % shardCount {
```
rlan35: We should just put the slot into a fixed shard based on the slot bls key. We don't need to enforce the maxNodePerShard in this release.

shard/committee/assignment.go
```diff
@@ -116,62 +125,148 @@ func preStakingEnabledCommittee(s shardingconfig.Instance) shard.State {
...
+	subCommitteeIdentities := make([]*bls.PublicKey, len(subCommittee.NodeList))
+	spot = 0
+	for i := range subCommittee.NodeList {
+		identity := &bls.PublicKey{}
+		subCommittee.NodeList[i].BlsPublicKey.ToLibBLSPublicKey(identity)
+		subCommitteeIdentities[spot] = identity
+		spot++
+	}
+
+	return allIdentities, subCommitteeIdentities
```
rlan35: Why do we need allIdentities as a list? Why just return all identities grouped by shard id and remove the subCommitteeIdentities?

fxfactorial: because caller has to then find their shardID rather than ignore the data they don't need

rlan35: Remove the flattened version and group by shards.

staking/effective/calculate.go
```diff
@@ -0,0 +1,101 @@
...
+			eposedSlots = append(eposedSlots, SlotPurchase{staker, spread})
+		}
+	}
+	if len(eposedSlots) < len(shortHand) {
+		// WARN Should never happen
+	}
+
+	sort.SliceStable(
+		eposedSlots,
+		func(i, j int) bool { return eposedSlots[i].Dec.GTE(eposedSlots[j].Dec) },
```
rlan35: GT should be good, right? It's a "less" func, so if E also qualifies as less, it will do some unnecessary swaps.

staking/effective/calculate.go
```diff
@@ -0,0 +1,101 @@
...
+
+func stake(median, actual numeric.Dec) numeric.Dec {
+	left := numeric.MinDec(onePlusC.Mul(median), actual)
+	right := oneMinusC.Mul(median)
+	return numeric.MaxDec(left, right)
+}
+
+// SlotPurchase ..
+type SlotPurchase struct {
+	common.Address `json:"slot-owner"`
```
rlan35: Should contain the bls pub key here.

shard/committee/assignment.go
```diff
@@ -116,62 +124,131 @@ func preStakingEnabledCommittee(s shardingconfig.Instance) shard.State {
...
+
+	shardBig := big.NewInt(int64(shardCount))
+
+	if l := len(staked); l < stakedSlotsCount {
+		// WARN unlikely to happen in production but will happen as we are developing
+		stakedSlotsCount = l
+	}
+
+	for i := 0; i < stakedSlotsCount; i++ {
+		bucket := int(new(big.Int).Mod(staked[i].Address.Big(), shardBig).Int64())
```
rlan35: Should mod on the bls pub key, not the validator's address.

shard/committee/assignment.go
```diff
@@ -116,62 +124,131 @@ func preStakingEnabledCommittee(s shardingconfig.Instance) shard.State {
...
-func with400Stakers(
-	s shardingconfig.Instance, stakerReader StakingCandidatesReader,
+func eposStakedCommittee(
+	s shardingconfig.Instance, stakerReader DataProvider, stakedSlotsCount int,
 ) (shard.State, error) {
 	// TODO Nervous about this because overtime the list will become quite large
 	candidates := stakerReader.ValidatorCandidates()
-	stakers := make([]*staking.Validator, len(candidates))
+	essentials := map[common.Address]effective.SlotOrder{}
+	slotUsage := map[common.Address]int{}
```
rlan35: probably don't need slotUsage if we return a ordered slots after median and only use the top 320.

staking/effective/calculate.go
```diff
@@ -0,0 +1,101 @@
...
+
+	pull := 320
+	if l := len(eposedSlots); l < pull {
+		pull = l
+	}
+	picks := eposedSlots[:pull]
+	median := median(picks)
+
+	for i := range picks {
+		picks[i].Dec = stake(median, picks[i].Dec)
```
rlan35: rename stake to effectiveStake(

shard/committee/assignment.go
```diff
@@ -116,62 +124,131 @@ func preStakingEnabledCommittee(s shardingconfig.Instance) shard.State {
...
+		superComm[spot].NodeList = append(superComm[spot].NodeList, shard.NodeID{
+			common2.ParseAddr(hAccounts[i].Address),
+			pubKey,
+			nil,
+		})
+	}
+
+	staked := effective.Apply(essentials)
+
+	sort.SliceStable(
```
rlan35: It should already be ordered (even after converting to effective stake)

staking/effective/calculate.go
```diff
@@ -0,0 +1,105 @@
...
+	if len(eposedSlots) < len(shortHand) {
+		// WARN Should never happen
+	}
+
+	sort.SliceStable(
+		eposedSlots,
+		func(i, j int) bool { return eposedSlots[i].Dec.GT(eposedSlots[j].Dec) },
+	)
+
+	pull := 320
```
rlan35: please make this a parameter, just pass the one from the caller.

PR #1830, opened 2019-11-14
Add delegate/undelegate and block reward core state transition

rlan35:


core/state_transition.go
```diff
@@ -366,9 +374,77 @@ func (st *StateTransition) applyEditValidatorTx(ev *staking.EditValidator, block
...
+	for i := range wrapper.Delegations {
+		delegation := wrapper.Delegations[i]
+		if bytes.Equal(delegation.DelegatorAddress.Bytes(), delegate.DelegatorAddress.Bytes()) {
+			delegatorExist = true
+			if CanTransfer(stateDB, delegate.DelegatorAddress, delegate.Amount) {
+				delegation.Amount.Add(delegation.Amount, delegate.Amount)
+				err := stateDB.UpdateStakingInfo(wrapper.Validator.Address, wrapper)
+				if err == nil {
+					// TODO: use undelegated token in locking period
+					stateDB.SubBalance(delegate.DelegatorAddress, delegate.Amount)
```
fxfactorial: is it ever possible to have a negative balance?

rlan35: Good question. It's possible. And the protocol should eliminate those transactions in the very beginning.

core/state/statedb.go
```diff
@@ -726,3 +728,43 @@ func (db *DB) IsValidator(addr common.Address) bool {
 	}
 	return so.IsValidator(db.db)
 }
+
+// AddReward distributes the reward to all the delegators based on stake percentage.
+func (db *DB) AddReward(validator common.Address, reward *big.Int) error {
```
fxfactorial: so if this fails, do I early return as well at call site?

core/state_transition.go
```diff
@@ -366,9 +374,93 @@ func (st *StateTransition) applyEditValidatorTx(ev *staking.EditValidator, block
...
+		delegation := wrapper.Delegations[i]
+		if bytes.Equal(delegation.DelegatorAddress.Bytes(), delegate.DelegatorAddress.Bytes()) {
+			delegatorExist = true
+			totalInUndelegation := delegation.TotalInUndelegation()
+			// If the sum of normal balance and the total amount of tokens in undelegation is greater than the amount to delegate
+			if big.NewInt(0).Add(totalInUndelegation, stateDB.GetBalance(delegate.DelegatorAddress)).Cmp(delegate.Amount) >= 0 {
+				// Firstly use the tokens in undelegation to delegate (redelegate)
+				undelegateAmount := big.NewInt(0).Set(delegate.Amount)
+				// Use the latest undelegated token first as it has the longest remaining locking time.
+				i := len(delegation.Entries) - 1
```
fxfactorial: How we know this invariant (latest has longest remaining locking time) is always true? Can sorting list in some way make us keep the invariant?

core/state/statedb.go
rlan35: Yes, but the caller should really check the existence of the validator too.

core/state_transition.go
rlan35: Because we are using this list always as a FIFO queue. It's always ordered by time.

rlan35: But added a TODO to sort the list just in case.

PR #1836, opened 2019-11-17
Fix active validator DB api; Correct validator snapshots; Finish commission rate check

rlan35:
Also added basic staking wireframes for block body.

and more sanity check on commission rate of validators

and rpcs for validator addresses and validator info.

### DIFF
--- a/core/types/bodyv2.go
+++ b/core/types/bodyv2.go
@@ -51,6 +51,15 @@ func (b *BodyV2) TransactionAt(index int) *Transaction {
 	return b.f.Transactions[index].Copy()
 }
 
+// StakingTransactionAt returns the staking transaction at the given index in this block.
+// It returns nil if index is out of bounds.
+func (b *BodyV2) StakingTransactionAt(index int) *staking.StakingTransaction {
+	if index < 0 || index >= len(b.f.StakingTransactions) {
+		return nil
+	}
+	return b.f.StakingTransactions[index].Copy()
+}
+
 // CXReceiptAt returns the CXReceipt at given index in this block
 // It returns nil if index is out of bounds
 func (b *BodyV2) CXReceiptAt(index int) *CXReceipt {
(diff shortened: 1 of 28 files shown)


core/types/bodyv2.go
```diff
@@ -51,6 +51,15 @@ func (b *BodyV2) TransactionAt(index int) *Transaction {
 	return b.f.Transactions[index].Copy()
 }
 
+// StakingTransactionAt returns the staking transaction at the given index in this block.
+// It returns nil if index is out of bounds.
+func (b *BodyV2) StakingTransactionAt(index int) *staking.StakingTransaction {
+	if index < 0 || index >= len(b.f.StakingTransactions) {
```
fxfactorial: `.f` is never nil?

staking/types/delegation.go
```diff
@@ -93,3 +93,20 @@ func (d *Delegation) DeleteEntry(epoch *big.Int) {
 		d.Entries = entries
 	}
 }
+
+// RemoveUnlockedUndelegations removes all fully unlocked undelegations and returns the total sum
+func (d *Delegation) RemoveUnlockedUndelegations(curEpoch *big.Int) *big.Int {
+	totalWithdraw := big.NewInt(0)
+	count := 0
+	for j := range d.Entries {
+		if curEpoch.Cmp(d.Entries[j].Epoch) > 14 { // need to wait at least 14 epochs to withdraw;
```
fxfactorial: maybe the 14 should be a named const? like in staking/effective

fxfactorial: oh wait, the Cmp gives back 1, 0, -1 right, so not the > 14

rlan35: Goot catch. Done.

core/types/bodyv2.go
rlan35: It's a struct, so it can't be nil

PR #1852, opened 2019-11-21
Fix dependency possible mismatch with test/deploy tests

Daniel-VDM:
* Remove dependency of harmony-ops repo by cloning it into /tmp on each run. Time-wise this should be fine as the localnet is still booting in the bg. 
* Checks CLI version on each run and asks to make the latest if out of date.

### DIFF
--- a/test/deploy.sh
+++ b/test/deploy.sh
@@ -60,22 +60,32 @@ function cleanup_and_result() {
 }
 
 function debug_staking() {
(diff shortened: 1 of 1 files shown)


Daniel-VDM:
Jenkins, test this please

Daniel-VDM:
Jenkins, test this please

LeoHChen:
why don't we make the hmy-cli part of the build process?  I don't like it being part of the test.  It has to be separated.  You may check the version or so, but pls don't mix them up.

PR #1855, opened 2019-11-22
[staking] Implement Quorum in staked case while taking into account hmy slots

fxfactorial:
Thank you Chao for the help

### DIFF
--- a/consensus/quorum/one-node-staked-vote.go
+++ b/consensus/quorum/one-node-staked-vote.go
@@ -10,18 +10,22 @@ import (
 	"github.com/harmony-one/harmony/numeric"
 	"github.com/harmony-one/harmony/shard"
 	"github.com/harmony-one/harmony/staking/slash"
+	"github.com/pkg/errors"
 )
 
 var (
-	one   = numeric.OneDec()
-	two   = numeric.NewDec(2)
-	three = numeric.NewDec(3)
-
-	oneThird        = one.Quo(three)
-	twoThirds       = two.Quo(three)
 	ninetyPercent   = numeric.MustNewDecFromStr("0.90")
-	quorumThreshold = twoThirds.Mul(numeric.NewDec(100)).RoundInt()
+	quorumThreshold = numeric.MustNewDecFromStr("0.67")
+	stakersShare    = numeric.MustNewDecFromStr("0.33")
+	totalShare      = numeric.MustNewDecFromStr("1.00")
 )
+
+// TallyResult is the result of when we calculate voting power,
+// recall that it happens to us at epoch change
+type TallyResult struct {
+	ourPercent   numeric.Dec
+	theirPercent numeric.Dec
+}
 
 type stakedVoter struct {
 	isActive, isHarmonyNode bool
@@ -35,9 +39,11 @@ type stakedVoteWeight struct {
 	DependencyInjectionWriter
 	DependencyInjectionReader
 	slash.ThresholdDecider
-	validatorStakes map[shard.BlsPublicKey]stakedVoter
-	stakedTotal     numeric.Dec
-	hmySlotCount    int64
+	voters                map[shard.BlsPublicKey]stakedVoter
+	ourVotingPowerTotal   numeric.Dec
+	theirVotingPowerTotal numeric.Dec
+	stakedTotal           numeric.Dec
+	hmySlotCount          int64
 }
 
 // Policy ..
@@ -68,7 +74,7 @@ func (v *stakedVoteWeight) computeCurrentTotalPower(p Phase) numeric.Dec {
 		if v.ReadSignature(p, members[i]) != nil {
 			w.FromLibBLSPublicKey(members[i])
 			currentTotalPower = currentTotalPower.Add(
-				v.validatorStakes[w].effectivePercent,
+				v.voters[w].effectivePercent,
 			)
 		}
 	}
@@ -78,7 +84,7 @@ func (v *stakedVoteWeight) computeCurrentTotalPower(p Phase) numeric.Dec {
 
 // QuorumThreshold ..
 func (v *stakedVoteWeight) QuorumThreshold() *big.Int {
-	return quorumThreshold
+	return quorumThreshold.RoundInt()
 }
 
 // RewardThreshold ..
@@ -95,8 +101,8 @@ func (v *stakedVoteWeight) Award(
 	count := big.NewInt(int64(len(earners)))
 	// proportional := map[common.Address]numeric.Dec{}
 
-	for _, details := range v.validatorStakes {
-		if details.isHarmonyNode == false {
+	for _, voter := range v.voters {
+		if voter.isHarmonyNode == false {
 			// proportional[details.earningAccount] = details.effective.QuoTruncate(
 			// 	v.stakedTotal,
 			// )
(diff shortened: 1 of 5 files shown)


consensus/quorum/one-node-staked-vote.go
```diff
@@ -39,31 +45,52 @@ func (v *stakedVoteWeight) Policy() Policy {
...
 	for i := range members {
-		w.FromLibBLSPublicKey(members[i])
-		// isHMY := v.validatorStakes[w].isHarmonyNode
-		if v.ReadSignature(p, members[i]) == nil {
-			// TODO TODO finish this logic
+		if v.ReadSignature(p, members[i]) != nil {
+			w.FromLibBLSPublicKey(members[i])
+			if v.validatorStakes[w].isHarmonyNode {
+				currentTotalPower = currentTotalPower.Add(
+					two.Mul(v.stakedTotal).Add(one).Quo(ourCount),
```
rlan35: Use hard coded voting power percentage for harmony (68%) and staker (32%)

fxfactorial: yep - mostly done with the refactor that way already - will finish in early morning before you're in office 👍

rlan35: Cool. Tomorrow, I may not be in office as my kid is sick again with fever and cough. I will see tomorrow morning his condition.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -13,14 +13,23 @@ import (
...
-	hSentinel = numeric.ZeroDec()
+	one   = numeric.OneDec()
+	two   = numeric.NewDec(2)
+	three = numeric.NewDec(3)
+
+	twoThirds         = two.Quo(three)
+	ninetyPercent     = numeric.MustNewDecFromStr("0.90")
+	sixtyEightPercent = numeric.MustNewDecFromStr("0.68")
+	thirtyTwoPercent  = numeric.MustNewDecFromStr("0.32")
+	quorumThreshold   = sixtyEightPercent.Mul(numeric.NewDec(100)).RoundInt()
```
rlan35: quorum threshold should be 1/3 in Dec. basically 0.6666666666666666666..  And the condition of quorum is > 1/3

consensus/quorum/one-node-staked-vote.go
```diff
@@ -39,31 +49,44 @@ func (v *stakedVoteWeight) Policy() Policy {
...
+		if v.ReadSignature(p, members[i]) != nil {
+			w.FromLibBLSPublicKey(members[i])
+			currentTotalPower = currentTotalPower.Add(
+				*v.validatorStakes[w].effective,
+			)
 		}
 	}
 
-	return true
+	return currentTotalPower
```
rlan35: We should make this a persisted variable, otherwise, we have to compute it everytime a new messge is received.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -103,26 +126,49 @@ func (v *stakedVoteWeight) Award(
...
+			effective:      nil,
 		}
+
+		// Real Staker
+		if staked[i].StakeWithDelegationApplied != nil {
+			member.isHarmonyNode = false
+			stakerP := staked[i].StakeWithDelegationApplied.
+				Quo(v.stakedTotal).
+				Mul(thirtyTwoPercent)
+			member.effective = &stakerP
```
rlan35: name effective to effectivePercent

consensus/quorum/one-node-staked-vote.go
fxfactorial: good point, the math is slow for cpu

consensus/quorum/one-node-staked-vote.go
```diff
@@ -10,26 +10,40 @@ import (
...
 	"github.com/harmony-one/harmony/shard"
 	"github.com/harmony-one/harmony/staking/slash"
+	"github.com/pkg/errors"
 )
 
 var (
-	twoThirds = numeric.NewDec(2).QuoInt64(3)
-	hSentinel = numeric.ZeroDec()
+	ninetyPercent   = numeric.MustNewDecFromStr("0.90")
+	quorumThreshold = numeric.MustNewDecFromStr("0.67")
```
rlan35: quorumThreshold is 2/3, not 0.67.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -10,26 +10,40 @@ import (
...
+	stakersShare    = numeric.MustNewDecFromStr("0.33")
```
rlan35: Staker's shard is 32%, and harmony node's shard is 68%.

consensus/quorum/one-node-staked-vote.go
```diff
@@ -39,31 +53,43 @@ func (v *stakedVoteWeight) Policy() Policy {
...
 	}
 
-	return true
+	return currentTotalPower
 }
 
 // QuorumThreshold ..
 func (v *stakedVoteWeight) QuorumThreshold() *big.Int {
-	return v.total.Mul(twoThirds).Ceil().RoundInt()
+	return quorumThreshold.RoundInt()
```
rlan35: I believe this will round your 0.67 to 1.   Man, we shouldn't commit this code.

fxfactorial: fixing quick now

consensus/quorum/one-node-staked-vote.go
```diff
@@ -101,38 +127,89 @@ func (v *stakedVoteWeight) Award(
...
+		// Real Staker
+		if staked[i].StakeWithDelegationApplied != nil {
+			member.isHarmonyNode = false
+			member.effectivePercent = staked[i].StakeWithDelegationApplied.
+				Quo(v.stakedTotal).
+				Mul(stakersShare)
+			theirPercentage = theirPercentage.Add(member.effectivePercent)
+		} else { // Our node
+			member.effectivePercent = numeric.NewDecFromBigInt(
+				v.QuorumThreshold(),
```
rlan35: This should be 68%, not quorumThreshold. We will have more voting power than the minimum quorum requirement.

fxfactorial: yes, implemented in new pr

consensus/quorum/one-node-staked-vote.go
```diff
@@ -101,38 +127,89 @@ func (v *stakedVoteWeight) Award(
...
 	utils.Logger().Info().
+		Str("our-percentage", ourPercentage.String()).
+		Str("their-percentage", theirPercentage.String()).
 		Uint32("on-shard", s).
-		Str("Staked", v.total.String()).
+		Str("Raw-Staked", v.stakedTotal.String()).
 		Msg("Total staked")
+
+	switch {
+	case totalStakedPercent.Equal(totalShare) == false:
```
rlan35: With the current code logic, these two cases are checking the same thing. totalStakedPercent = ours + theirs in the above logic. Due to precision problem, the final sum may not be 100%, we need to hard code it to make it 100%, such as giving the residual to the last voter to sum up to 100%

fxfactorial: the decimal wrapper was designed for this usage though, to take into account the precision problems.

PR #1856, opened 2019-11-22
accept and verify multiple viewID of view change messages

chaosma:


consensus/view_change.go
```diff
@@ -202,10 +201,13 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
 	defer consensus.vcLock.Unlock()
 
+	// update the dictionary key if the viewID is first time received
+	consensus.addViewIDKeyIfNotExist(recvMsg.ViewID)
+
 	// TODO: remove NIL type message
 	// add self m1 or m2 type message signature and bitmap
-	_, ok1 := consensus.nilSigs[consensus.PubKey.SerializeToHexStr()]
-	_, ok2 := consensus.bhpSigs[consensus.PubKey.SerializeToHexStr()]
+	_, ok1 := consensus.nilSigs[recvMsg.ViewID][consensus.PubKey.SerializeToHexStr()]
```
rlan35: Do you need to check both the outer map returns ok and the inner map returns ok? What if there is no map associated with recvMsg.ViewID, any nil pointer issue?

rlan35: I see you have the addViewIDKeyIfNotExist, can you make sure it's called before any potential nil pointer happens?

consensus/view_change.go
```diff
@@ -341,16 +340,17 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 	utils.Logger().Debug().
 		Str("validatorPubKey", senderKey.SerializeToHexStr()).
 		Msg("[onViewChange] Add M3 (ViewID) type message")
```
LeoHChen: do you need to add viewID into the log messages?  that can help us triage the issues later.

consensus/view_change.go
```diff
@@ -318,13 +318,12 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 		utils.Logger().Debug().
 			Str("validatorPubKey", senderKey.SerializeToHexStr()).
 			Msg("[onViewChange] Add M1 (prepared) type message")
```
LeoHChen: log message to include viewID as well.

consensus/view_change.go
chaosma: yes, I added it.

consensus/view_change.go
chaosma: yeap

chaosma:
> Please test the case of a view change happen and then epoch change happen (basically replaying the case of the shard down)

Local test

consensus/view_change.go
chaosma: It is called before any access to the map

PR #1866, opened 2019-11-23
Fix potential lastmileblocks crash; Remove validator.Stake; refactor Compute;

rlan35:
Tested locally. 

Can see voting power changes with new validator created and delegation added.


shard/committee/assignment.go
```diff
@@ -126,11 +126,15 @@ func eposStakedCommittee(
 	for i := range candidates {
 		// TODO Should be using .ValidatorStakingWithDelegation, not implemented yet
 		validator, err := stakerReader.ReadValidatorData(candidates[i])
+		validatorStake := big.NewInt(0)
+		for _, delegation := range validator.Delegations {
```
fxfactorial: If you are doing the sum, then I suppose need to remove the TODO right above, to use `.ValidatorStakingWithDelegation`

fxfactorial: actually now I think, .ValidatorStakingWithDelegation should be removed, perhaps remove the bottom 3 methods in core/blockchain.go

PR #1869, opened 2019-11-24
[staking][reward] Give out block reward for cross links

fxfactorial:
1. Give out block reward to cross links on beacon chain, processing of which done concurrently for persisted in correct sequence. 
2. Give out block reward to own beacon chain
3. Factor out votepower as separate package, need to have computation of voting power also at reward stage.


internal/chain/reward.go
```diff
@@ -1,39 +1,140 @@
...
+					commitBitmap := cxLink.Header().LastCommitBitmap()
+
+					if err := mask.SetMask(commitBitmap); err != nil {
+						// return ctxerror.New("cannot set group sig mask bits").WithCause(err)
+					}
+
+					totalAmount := numeric.ZeroDec()
+
+					for j := range stakers {
+						totalAmount = totalAmount.Add(*stakers[j].StakeWithDelegationApplied)
```
rlan35: effective stake.

internal/chain/reward.go
```diff
@@ -1,39 +1,140 @@
...
+
+					for j := range stakers {
+						switch signed, err := mask.IndexEnabled(j); true {
+						case err != nil:
+							// return ctxerror.New(
+							// 	"cannot check for committer bit", "committerIndex", j,
+							// ).WithCause(err)
+						case signed:
+							go func(signersDue numeric.Dec, addr common.Address) {
+								payable <- slotPayable{addr, signersDue}
```
rlan35: number first, then address.

internal/chain/reward.go
```diff
@@ -1,157 +1,312 @@
...
 	parentShardState, err := bc.ReadShardState(parentHeader.Epoch())
 	if err != nil {
-		return ctxerror.New("cannot read shard state",
-			"epoch", parentHeader.Epoch(),
+		return nil, nil, nil, ctxerror.New(
+			"cannot read shard state", "epoch", parentHeader.Epoch(),
 		).WithCause(err)
 	}
-	parentCommittee := parentShardState.FindCommitteeByID(parentHeader.ShardID())
+	parentCommittee := parentShardState.FindCommitteeByID(shardID)
```
rlan35: shardID can be read from header directly.

internal/chain/reward.go
```diff
@@ -1,157 +1,312 @@
...
+// included uncles. The coinbase of each uncle block is also rewarded.
+func AccumulateRewards(
+	bc engine.ChainReader, state *state.DB, header *block.Header,
+	rewarder reward.Distributor, slasher slash.Slasher,
+	beaconChain engine.ChainReader,
+) error {
+
+	blockNum := header.Number().Uint64()
+	if blockNum == 0 {
+		// Epoch block has no parent to reward.
```
rlan35: you mean genesis block?

internal/chain/reward.go
```diff
@@ -1,157 +1,312 @@
...
-			// Slash if missing block was long enough
-			if slasher.ShouldSlash(missing[member].BlsPublicKey) {
-				// TODO Logic
+		votingPower := votepower.Compute(members)
+
+		for beaconMember := range payable {
+			voter := votingPower.Voters[payable[beaconMember].BlsPublicKey]
+			if !voter.IsHarmonyNode {
+				due := BlockRewardStakedCase.Mul(voter.EffectivePercent)
+				state.AddBalance(voter.EarningAccount, due.RoundInt())
```
rlan35: Add todo to give out the leftover to the last voter

internal/chain/reward.go
fxfactorial: this is original comment from code, will update to `genesis block `

internal/chain/reward.go
```diff
@@ -1,157 +1,312 @@
...
+							due := BlockRewardStakedCase.Mul(voter.EffectivePercent)
+							to := voter.EarningAccount
+							go func(signersDue numeric.Dec, addr common.Address, j int) {
+								payable <- slotPayable{
+									effective: signersDue,
+									payee:     addr,
+									nonce:     i * (i + j),
+									oops:      nil,
+								}
+							}(due, to, member)
```
rlan35: just give the nonce directly here. the formula i * (i + j) doesn't work. The sizes of payableSigners varies. No single formula does the work. Maybe just make all this single threaded?

internal/chain/reward.go
```diff
@@ -1,157 +1,312 @@
...
-	signers := make([]string, len(payable))
-
 	for i := range payable {
-		signers[i] = payable[i].string
 		state.AddBalance(payable[i].Address, payable[i].Int)
 	}
 
 	header.Logger(utils.Logger()).Debug().
-		Int("NumAccounts", len(accounts)).
+		Int("NumAccounts", len(payable)).
```
rlan35: It's better to print out for each shard separately. With numTotalVoters/numSigners/TotalRewards for each. Aggregated info is not helpful here. The totalAmount can be used to increase the TotalSupply record later.

rlan35: It see. this is the old logic. It's a bit confusing here.

internal/chain/reward.go
fxfactorial: yes know, but did this for the ` func ballotResultBeaconchain(` so can call with `return ballotResult(bc, header, shard.BeaconChainShardID)`

consensus/votepower/roster.go
```diff
@@ -0,0 +1,90 @@
...
+
+		// Real Staker
+		if staked[i].TotalStake != nil {
+			member.IsHarmonyNode = false
+			member.EffectivePercent = staked[i].TotalStake.
+				Quo(roster.RawStakedTotal).
+				Mul(StakersShare)
+			theirPercentage = theirPercentage.Add(member.EffectivePercent)
+		} else { // Our node
+			member.EffectivePercent = HarmonysShare.Quo(ourCount)
```
rlan35: Add a TODO to make sure it sums to 100% of the voting powers.

internal/chain/reward.go
```diff
@@ -2,133 +2,302 @@ package chain
...
+		}
+
+		votingPower := votepower.Compute(members)
+
+		for beaconMember := range payable {
+			// TODO Give out whatever leftover to the last voter/handle
+			// what to do about share of those that didn't sign
+			voter := votingPower.Voters[payable[beaconMember].BlsPublicKey]
+			if !voter.IsHarmonyNode {
+				due := BlockRewardStakedCase.Mul(voter.EffectivePercent)
```
rlan35: EffectivePercent is only the percent within 32%, right? Then we are only giving out at most 32% of the total block rewards.

internal/chain/reward.go
```diff
@@ -2,133 +2,302 @@ package chain
...
+
+					if err != nil {
+						slotError(err, payable)
+						return
+					}
+
+					for member := range payableSigners {
+						voter := votingPower.Voters[payableSigners[member].BlsPublicKey]
+						if !voter.IsHarmonyNode {
+							due := BlockRewardStakedCase.Mul(voter.EffectivePercent)
```
rlan35: same here.

PR #1870, opened 2019-11-24
Synchronize shard chains with beacon chain after stakingEpoch. Add PreStakingEpoch

rlan35:
Also correct the shard computation logic to read directly from DB in the correct place. Only beacon chain do Compute() at the proposal phase.


consensus/consensus_service.go
```diff
@@ -453,12 +453,69 @@ func (consensus *Consensus) getLeaderPubKeyFromCoinbase(header *block.Header) (*
...
+	}
+
 	pubKeys := []*bls.PublicKey{}
 	hasError := false
 	header := consensus.ChainReader.CurrentHeader()
 	epoch := header.Epoch()
-	curPubKeys := committee.WithStakingEnabled.ComputePublicKeys(
+
+	// TODO: change GetCommitteePublicKeys to read from DB
+	curShardState, err := committee.WithStakingEnabled.Compute(
```
fxfactorial: if read from DB, do have: `.ReadFromDB`

rlan35: Already did.

node/node_newblock.go
```diff
@@ -88,14 +89,16 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 
 	// TODO: integrate staking transaction into tx pool
 	pendingStakingTransactions := types2.StakingTransactions{}
-	node.pendingStakingTxMutex.Lock()
-	for _, tx := range node.pendingStakingTransactions {
-		pendingStakingTransactions = append(pendingStakingTransactions, tx)
+	// Only process staking transactions after pre-staking epoch happened.
+	if node.Blockchain().Config().IsPreStaking(node.Worker.GetNewEpoch()) {
```
fxfactorial: this will make testing annoying for rest of team - need to let them know only happens after epoch 3 on debug.sh

node/worker/worker.go
```diff
@@ -288,39 +288,65 @@ func (w *Worker) SuperCommitteeForNextEpoch(
...
-			switch beacon.CurrentHeader().Epoch().Cmp(w.current.header.Epoch()) {
+			switch beaconEpoch.Cmp(w.current.header.Epoch()) {
 			case 1:
-				nextCommittee, oops = committee.WithStakingEnabled.ReadFromDB(
-					beacon.CurrentHeader().Epoch(), beacon,
+				// If beacon chain is bigger than shard chain in epoch, it means I should catch up with beacon chain now
+				nextCommittee, err = committee.WithStakingEnabled.ReadFromDB(
+					beaconEpoch, beacon,
 				)
+				blockEpoch := big.NewInt(0).Set(beaconEpoch).Sub(beaconEpoch, big.NewInt(1)) // Set this block's epoch to be beaconEpoch - 1, so the next block will have beaconEpoch
```
fxfactorial: What is wrong about setting current epoch to beaconEpoch?

node/node_newblock.go
rlan35: In test/local net, PreStakingEpoch = big.NewInt(0). There is no issue for testing.

node/worker/worker.go
rlan35: It's already explicit. beaconEpoch - 1 so the next block will have beaconEpoch itself. If we set current block to be beaconEpoch, then next block will have beaconEpoch + 1, which is wrong. The logic is whenever there is a block with Shard State in it, the next block is the new epoch and that epoch is preBlock.Epoch + 1.

node/node_newblock.go
fxfactorial: I meant 3 is slow for testing today, if staking was at epoch 1 , would iterate faster , no?

node/node_newblock.go
```diff
@@ -78,7 +78,16 @@ func (node *Node) WaitForConsensusReadyV2(readySignal chan struct{}, stopChan ch
 func (node *Node) proposeNewBlock() (*types.Block, error) {
 	// Update worker's current header and state data in preparation to propose/process new transactions
 	coinbase := node.Consensus.SelfAddress
-	node.Worker.UpdateCurrent(coinbase)
+
+	// After staking, all coinbase will be the address of bls pub key
```
fxfactorial: would be nice if was a longer comment explaining why specifically after staking, the connection between address(bls_pub_key) (I thought address were always ecdsa based) and coin base

node/worker/worker.go
fxfactorial: Thank you, I mean put that directly into a longer /* */, because valuable information for programmer

rlan35:
jenkins, test this please.

chaosma:
@rlan35  Have you tested the cross-shard transaction?

PR #1874, opened 2019-11-24
[WIP] [reward] Adjust block reward by % of current staked

fxfactorial:
Adjust block reward in staked case when deviates from the 35% target rate.

### DIFF
--- a/internal/chain/reward.go
+++ b/internal/chain/reward.go
@@ -32,7 +32,8 @@ var (
 	BlockRewardStakedCase = numeric.NewDecFromBigInt(new(big.Int).Mul(
 		big.NewInt(18), big.NewInt(denominations.One),
 	))
-
+	totalTokens                  = numeric.NewDec(12600000000)
+	targetStakedPercentage       = numeric.MustNewDecFromStr("0.35")
 	errPayoutNotEqualBlockReward = errors.New("total payout not equal to blockreward")
 )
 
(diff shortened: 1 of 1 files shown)


internal/chain/reward.go
```diff
@@ -144,6 +168,19 @@ func AccumulateRewards(
 	if bc.Config().IsStaking(header.Epoch()) &&
 		bc.CurrentHeader().ShardID() == shard.BeaconChainShardID {
 
+		defaultReward := BlockRewardStakedCase
+
+		if shard.Schedule.IsLastBlock(blockNum) {
```
rlan35: Should use ShardState size > 0 for the condition of last block. The schedule is only for pre-staking.

internal/chain/reward.go
```diff
@@ -144,6 +168,19 @@ func AccumulateRewards(
...
+			// TODO Use cached result in off-chain db instead of full computation
+			percentageStaked, err := whatPercentStakedNow(beaconChain)
+			if err != nil {
+				return err
+			}
+			howMuchOff := targetStakedPercentage.Sub(*percentageStaked)
+			adjustBy := howMuchOff.MulTruncate(numeric.NewDec(100)).Mul(dynamicAdjust)
+			defaultReward = defaultReward.Add(adjustBy)
```
rlan35: We need to make sure the final rewards are not negative.

PR #1877, opened 2019-11-25
Simplified CrossLink structure;

chaosma:


core/types/crosslink.go
```diff
@@ -4,51 +4,62 @@ import (
...
-
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/harmony-one/harmony/block"
 )
 
 // CrossLink is only used on beacon chain to store the hash links from other shards
+// signature and bitmap correspond to |blockNumber|parentHash| byte array
+// Captial to enable rlp encoding
 type CrossLink struct {
-	ChainHeader *block.Header
```
fxfactorial: where is the written proof/explanation why possible to have same level of verification without full Header? Needs to be written out, explained

node/node_cross_shard.go
```diff
@@ -120,124 +119,75 @@ func (node *Node) VerifyBlockCrossLinks(block *types.Block) error {
...
-			node.Blockchain().WriteCrossLinks(types.CrossLinks{crossLink}, true)
+				Msgf("[ProcessingCrossLink] committing for shardID %d, blockNum %d", cl.ShardID(), cl.Number().Uint64())
 		}
-
-		// Queue up the cross links that's in the future
-		node.pendingClMutex.Lock()
-		node.pendingCrossLinks = append(node.pendingCrossLinks, headersToQuque...)
-		node.pendingClMutex.Unlock()
+		node.pendingCLMutex.Lock()
+		node.pendingCrossLinks = append(node.pendingCrossLinks, candidates...)
```
rlan35: We still need to put cross link into the DB before it's proposed in blockchain. If a node is down, the in-memory data is all gone.

core/rawdb/accessors_chain.go
```diff
@@ -555,25 +555,18 @@ func ReadCXReceipts(db DatabaseReader, shardID uint32, number uint64, hash commo
...
 
 // WriteCXReceipts stores all the transaction receipts given destination shardID, blockNumber and blockHash
-func WriteCXReceipts(db DatabaseWriter, shardID uint32, number uint64, hash common.Hash, receipts types.CXReceipts, temp bool) error {
+func WriteCXReceipts(db DatabaseWriter, shardID uint32, number uint64, hash common.Hash, receipts types.CXReceipts) error {
 	bytes, err := rlp.EncodeToBytes(receipts)
 	if err != nil {
 		utils.Logger().Error().Msg("[WriteCXReceipts] Failed to encode cross shard tx receipts")
 	}
 	// Store the receipt slice
-	if err := db.Put(cxReceiptKey(shardID, number, hash, temp), bytes); err != nil {
```
rlan35: Why is the temp flag removed? If have to remove, any compatibility issue?

core/blockchain.go
```diff
@@ -2093,31 +2092,28 @@ func (bc *BlockChain) WriteEpochVdfBlockNum(epoch *big.Int, blockNum *big.Int) e
 }
 
 // WriteCrossLinks saves the hashes of crosslinks by shardID and blockNum combination key
-// temp=true is to write the just received cross link that's not committed into blockchain with consensus
-func (bc *BlockChain) WriteCrossLinks(cls []types.CrossLink, temp bool) error {
```
rlan35: I think we still need the temp flag to store crossLinks that's pending.

core/rawdb/accessors_chain.go
```diff
@@ -555,25 +555,18 @@ func ReadCXReceipts(db DatabaseReader, shardID uint32, number uint64, hash commo
...
+	if err := db.Put(cxReceiptKey(shardID, number, hash), bytes); err != nil {
 		utils.Logger().Error().Msg("[WriteCXReceipts] Failed to store cxreceipts")
 	}
 	return err
 }
 
-// DeleteCXReceipts removes all receipt data associated with a block hash.
-func DeleteCXReceipts(db DatabaseDeleter, shardID uint32, number uint64, hash common.Hash, temp bool) {
```
rlan35: DeleteCXReceipts is never used. Then do we ever remove receipts that's not needed anymore?

core/types/crosslink.go
```diff
@@ -4,51 +4,62 @@ import (
...
+	ParentHashF  common.Hash
```
rlan35: What is "F" suffix for?

node/node.go
```diff
@@ -108,12 +107,13 @@ type Node struct {
 	ConfirmedBlockChannel chan *types.Block    // The channel to send confirmed blocks
 	BeaconBlockChannel    chan *types.Block    // The channel to send beacon blocks for non-beaconchain nodes
 	DRand                 *drand.DRand         // The instance for distributed randomness protocol
-	pendingCrossLinks     []*block.Header
-	pendingClMutex        sync.Mutex
 
 	pendingCXReceipts map[string]*types.CXReceiptsProof // All the receipts received but not yet processed for Consensus
 	pendingCXMutex    sync.Mutex
 
+	pendingCrossLinks []types.CrossLink
```
rlan35: Let's not use in-memory cache anymore. Our normal txns are already using TxPool, which I believe keeps txns in local db for persistence.

node/node_cross_shard.go
```diff
@@ -120,124 +119,75 @@ func (node *Node) VerifyBlockCrossLinks(block *types.Block) error {
...
-
-		if crossLink.BlockNum().Cmp(firstCrossLinkBlock) > 0 { // TODO: verify genesis block
-			err = node.VerifyCrosslinkHeader(lastLink.Header(), crossLink.Header())
-			if err != nil {
-				return ctxerror.New("cannot ValidateNewBlock",
+	for _, crossLink := range *crossLinks {
+		cl, err := node.Blockchain().ReadCrossLink(crossLink.ShardID(), crossLink.BlockNum())
+		if err == nil && cl != nil {
+			if !bytes.Equal(cl.Serialize(), crossLink.Serialize()) {
+				return ctxerror.New("[CrossLinkVerification] Double signed crossLink",
```
rlan35: When there is a double sign, add a TODO here. Don't return error, then double-sign will fail the system.

node/node_cross_shard.go
```diff
@@ -120,124 +119,75 @@ func (node *Node) VerifyBlockCrossLinks(block *types.Block) error {
...
 			return
 		}
-
-		// Try to reprocess all the pending cross links
-		node.pendingClMutex.Lock()
-		crossLinkHeadersToProcess := node.pendingCrossLinks
-		node.pendingCrossLinks = []*block.Header{}
-		node.pendingClMutex.Unlock()
-
 		firstCrossLinkBlock := core.EpochFirstBlock(node.Blockchain().Config().CrossLinkEpoch)
```
rlan35: If crosslink has Epoch in it, now we can just compare directly the epoch, no need to get the first block anymore as we don't care about the order now (the first block's crosslink doesn't have to be processed first anymore). Also after staking, the logic of EpochFirstBlock will be invalid.

node/node_cross_shard.go
```diff
@@ -120,124 +119,75 @@ func (node *Node) VerifyBlockCrossLinks(block *types.Block) error {
...
-			if len(headersToQuque) > crossLinkBatchSize {
-				break
+		for i, cl := range crosslinks {
+			if cl.Number() == nil || cl.Number().Cmp(firstCrossLinkBlock) < 0 {
+				utils.Logger().Debug().
+					Msgf("[ProcessingCrossLink] Crosslink %d skipped: %v", i, cl)
+				continue
 			}
-			exist, err := node.Blockchain().ReadCrossLink(header.ShardID(), header.Number().Uint64(), false)
+			exist, err := node.Blockchain().ReadCrossLink(cl.ShardID(), cl.Number().Uint64())
```
rlan35: Add todo to still preserve the cross-link if it's a double-sign.

core/types/crosslink.go
```diff
@@ -4,51 +4,62 @@ import (
...
+	BlockNumberF *big.Int
+	SignatureF   [96]byte //aggregated signature
+	BitmapF      []byte   //corresponding bitmap mask for agg signature
+	ShardIDF     uint32   //need first verify signature on |blockNumber|blockHash| is correct
+	EpochF       *big.Int //need first verify signature on |blockNumber|blockHash| is correct
```
rlan35: This should be the parentEpoch as you used it to retrieve committee.

node/node_cross_shard.go
```diff
@@ -278,22 +228,18 @@ func (node *Node) verifyIncomingReceipts(block *types.Block) error {
...
-	parentHash := header.ParentHash()
-	if prevHeader.Hash() != parentHash {
-		return ctxerror.New("[CrossLink] Invalid cross link header - parent hash mismatch", "shardID", header.ShardID(), "blockNum", header.Number())
-	}
 
 	// Verify signature of the new cross link header
 	// TODO: check whether to recalculate shard state
-	shardState, err := node.Blockchain().ReadShardState(prevHeader.Epoch())
-	committee := shardState.FindCommitteeByID(prevHeader.ShardID())
+	shardState, err := node.Blockchain().ReadShardState(cl.Epoch())
```
rlan35: the epoch here is wrong.

core/types/crosslink.go
chaosma: We don't have same level of verification. Cannot verify the block header content, but trust the majority of the committee is not corrupted. I will write some comments

core/rawdb/accessors_chain.go
chaosma: If we support the CXReceipt at any given time, then we should not delete the CXReceipt. Otherwise, we can clean the old CXReceipt when we know it's been delivered. We don't have the support to check and archive old CXReceipt yet.

core/types/crosslink.go
chaosma: Field, just used to distinguish between the method from the field.

node/node.go
chaosma: will add the tmp flag back to local db persistency

node/node_cross_shard.go
chaosma: use local db instead of in-memory

core/rawdb/accessors_chain.go
chaosma: it's no longer needed and has no compatibility issue

node/node_cross_shard.go
chaosma: fixed

core/types/crosslink.go
chaosma: fixed

core/blockchain.go
```diff
@@ -1191,22 +1195,23 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
...
 		}
 		for _, crossLink := range *crossLinks {
-			if err := bc.WriteCrossLinks(types.CrossLinks{crossLink}, false); err == nil {
-				utils.Logger().Info().Uint64("blockNum", crossLink.BlockNum().Uint64()).Uint32("shardID", crossLink.ShardID()).Msg("[InsertChain] Cross Link Added to Beaconchain")
+			if err := bc.WriteCrossLinks(types.CrossLinks{crossLink}); err == nil {
+				utils.Logger().Info().Uint64("blockNum", crossLink.BlockNum()).Uint32("shardID", crossLink.ShardID()).Msg("[insertChain/crosslinks] Cross Link Added to Beaconchain")
 			}
-			bc.DeleteCrossLinks(types.CrossLinks{crossLink}, true)
-			bc.WriteShardLastCrossLink(crossLink.ShardID(), crossLink)
+			//bc.WriteShardLastCrossLink(crossLink.ShardID(), crossLink)
```
rlan35: remove commented line?

core/blockchain.go
```diff
@@ -2131,6 +2133,40 @@ func (bc *BlockChain) WriteShardLastCrossLink(shardID uint32, cl types.CrossLink
 	return rawdb.WriteShardLastCrossLink(bc.db, cl.ShardID(), cl.Serialize())
 }
 
+// UpdateShardLastCrossLink saves the last crosslink of a shard
+func (bc *BlockChain) UpdateShardLastCrossLink(shardID uint32, cl types.CrossLink) error {
```
rlan35: please describe the logic in this func. Even by reading the code I don't fully get what's the meaning of this func.

core/blockchain.go
```diff
@@ -2142,6 +2178,61 @@ func (bc *BlockChain) ReadShardLastCrossLink(shardID uint32) (*types.CrossLink,
 	return crossLink, err
 }
 
+// ReadLastPendingCrossLinks retrieves pending crosslinks
+func (bc *BlockChain) ReadLastPendingCrossLinks() ([]types.CrossLink, error) {
```
rlan35: ReadPendingCrossLinks, no need for "Last"

core/types/crosslink.go
```diff
@@ -4,51 +4,76 @@ import (
...
 )
 
 // CrossLink is only used on beacon chain to store the hash links from other shards
+// signature and bitmap correspond to |blockNumber|parentHash| byte array
+// Captial to enable rlp encoding
+// Here we replace header to signatures only, the basic assumption is the committee will not be
+// corrupted during one epoch, which is the same as consensus assumption
 type CrossLink struct {
-	ChainHeader *block.Header
+	ParentHashF  common.Hash
```
rlan35: Let's remove all the "Parent" prefix. Everything here is about the same block. So no need to say parent or not.

core/types/crosslink.go
```diff
@@ -4,51 +4,76 @@ import (
...
 // NewCrossLink returns a new cross link object
 func NewCrossLink(header *block.Header) CrossLink {
-	return CrossLink{header}
+	if header.Number().Uint64() == 0 {
+		return CrossLink{}
+	}
+	epoch := big.NewInt(0)
+	parentBlockNum := big.NewInt(0)
+	// check if previous header is the last block of an epoch
+	if shard.Schedule.IsLastBlock(header.Number().Uint64() - 1) {
```
rlan35: This doesn't work  after staking. Instead, should really fetch the parent Block and read its Epoch. As there can be gaps between epochs now due to staking epoch synchronization.

node/node_cross_shard.go
rlan35: It's still returning error? If there is really a cross-link with double sign, it's a proof for slashing. not an invalid block.

node/node_cross_shard.go
```diff
@@ -120,124 +120,77 @@ func (node *Node) VerifyBlockCrossLinks(block *types.Block) error {
...
-// ProcessHeaderMessage verify and process Node/Header message into crosslink when it's valid
-func (node *Node) ProcessHeaderMessage(msgPayload []byte) {
-	if node.NodeConfig.ShardID == 0 {
-
-		var headers []*block.Header
-		err := rlp.DecodeBytes(msgPayload, &headers)
+// ProcessCrossLinkMessage verify and process Node/CrossLink message into crosslink when it's valid
+func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
+	// TODO: non-leader in beaconchain doesn't need process crosslink message, but still need to monitor leader's behavior
+	if node.NodeConfig.ShardID == 0 && node.Consensus.IsLeader() {
```
rlan35: Everyone needs to process cross-links locally and remove them once it's confirmed in blocks. Otherwise, if the leader is down or leader switched, the new leader won't have the pending crosslinks to propose, then those links will be lost.

node/node_newblock.go
```diff
@@ -112,9 +112,14 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 	// Prepare cross links
 	var crossLinks types.CrossLinks
 	if node.NodeConfig.ShardID == 0 {
-		crossLinksToPropose, localErr := node.ProposeCrossLinkDataForBeaconchain()
-		if localErr == nil {
-			crossLinks = crossLinksToPropose
+		crossLinks, err = node.Blockchain().ReadLastPendingCrossLinks()
```
rlan35: Still need mutex on read/write of the pending cross links. Otherwise, there can be race conditions. The newly added pending cross links may be deleted if they are added in between Read and Delete.

node/node_cross_shard.go
chaosma: I move the delete pending crosslinks to insertchain code.

core/types/crosslink.go
chaosma: In this case, I will remove epoch field to make the code cleaner

core/blockchain.go
```diff
@@ -1212,22 +1216,27 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
...
-				utils.Logger().Info().Uint64("blockNum", crossLink.BlockNum().Uint64()).Uint32("shardID", crossLink.ShardID()).Msg("[InsertChain] Cross Link Added to Beaconchain")
+			if err := bc.WriteCrossLinks(types.CrossLinks{crossLink}); err == nil {
+				utils.Logger().Info().Uint64("blockNum", crossLink.BlockNum()).Uint32("shardID", crossLink.ShardID()).Msg("[insertChain/crosslinks] Cross Link Added to Beaconchain")
 			}
-			bc.DeleteCrossLinks(types.CrossLinks{crossLink}, true)
-			bc.WriteShardLastCrossLink(crossLink.ShardID(), crossLink)
+			bc.UpdateShardLastCrossLink(crossLink.ShardID(), crossLink)
+		}
+		//clean local database cache after crosslink inserted into blockchain
+		err = bc.DeletePendingCrossLinks()
```
rlan35: why delete here? there could be pending cross links that's not proposed in the block.

core/blockchain.go
rlan35: Then it shouldn't be called last right? It should be called "LastContinuousCrossLink"

core/rawdb/schema.go
```diff
@@ -58,18 +58,18 @@ var (
 	shardStatePrefix = []byte("ss") // shardStatePrefix + num (uint64 big endian) + hash -> shardState
 	lastCommitsKey   = []byte("LastCommits")
 
+	lastPendingCrosslinkKey = []byte("lastPendingCL") // prefix for shard last pending crosslink
```
rlan35: Still, why call last Pending? pending is enough.

internal/chain/reward.go
```diff
@@ -239,9 +239,10 @@ func AccumulateRewards(
 
 				cxLink := crossLinks[i]
 
-				shardState, err := bc.ReadShardState(cxLink.ChainHeader.Epoch())
-				if !bc.Config().IsStaking(cxLink.Header().Epoch()) {
-					shardState, err = committee.WithStakingEnabled.Compute(cxLink.ChainHeader.Epoch(), bc)
+				epoch := bc.GetHeaderByNumber(cxLink.BlockNum()).Epoch()
```
rlan35: This should be directly using the cxLink.Epoch().

internal/chain/reward.go
```diff
@@ -252,7 +253,7 @@ func AccumulateRewards(
 
 				subComm := shardState.FindCommitteeByID(cxLink.ShardID())
 				// _ are the missing signers, later for slashing
-				payableSigners, _, err := blockSigners(cxLink.Header(), subComm)
+				payableSigners, _, err := blockSigners(cxLink.Bitmap(), subComm)
 
 				if err != nil {
```
rlan35: please remove the two HACK places. and directly return err.

core/blockchain.go
chaosma: checking and deleting only existing cross links

core/rawdb/schema.go
chaosma: remove "last"

core/blockchain.go
chaosma: good name! changed

core/blockchain.go
```diff
@@ -2124,6 +2128,43 @@ func (bc *BlockChain) WriteShardLastCrossLink(shardID uint32, cl types.CrossLink
...
+		return nil
+	}
+	if err != nil {
+		return err
+	}
+	newCheckpoint := uint64(0)
+	for i := cl0.BlockNum() + 1; i < cl.BlockNum(); i++ {
+		tmp, err := bc.ReadShardLastCrossLink(shardID)
+		if err != nil || tmp == nil {
+			if i > cl0.BlockNum() {
```
rlan35: isn't i always greater than cl0.BlockNum()

core/blockchain.go
```diff
@@ -2124,6 +2128,43 @@ func (bc *BlockChain) WriteShardLastCrossLink(shardID uint32, cl types.CrossLink
...
+				newCheckpoint = i - 1
```
rlan35: Should you break here? Here you found a new checkPoint right?

core/blockchain.go
```diff
@@ -2135,6 +2176,73 @@ func (bc *BlockChain) ReadShardLastCrossLink(shardID uint32) (*types.CrossLink,
...
+
+// DeleteCommittedFromPendingCrossLinks delete pending crosslinks that already committed
+func (bc *BlockChain) DeleteCommittedFromPendingCrossLinks() (int, error) {
+	cls, err := bc.ReadPendingCrossLinks()
+	if err != nil || len(cls) == 0 {
+		return 0, err
+	}
+	pendingCLs := []types.CrossLink{}
+	for _, cl := range cls {
+		cl0, err := bc.ReadCrossLink(cl.ShardID(), cl.BlockNum())
```
rlan35: Let feed the crosslink from the block directly here instead of reading it from DB. Reading db is slow and it will further increase the risk of race condition (as we didn't lock on this method).

core/blockchain.go
chaosma: right, removed

core/blockchain.go
chaosma: yes, that's right

core/blockchain.go
```diff
@@ -77,6 +77,7 @@ const (
 	validatorStatsCacheLimit           = 1024
 	validatorListCacheLimit            = 10
 	validatorListByDelegatorCacheLimit = 1024
+	pendingCrossLinksCacheLimit        = 2
```
rlan35: should make this larger.

chaosma: actually 1 is enough, since there is one key and corresponding to a list.

PR #1878, opened 2019-11-25
Enable staking for pangaea; Fix cross-shard verification during staking epoch transition.

rlan35:


rlan35:
Jenkins, test this please.

PR #1886, opened 2019-11-26
[shard] Wrapper on decoding ShardState

fxfactorial:
Backward compat by hooking place where ShardState is read. 

Includes unit test of old shard state structure being decoded into new shard state structure.

### DIFF
--- a/shard/shard_state.go
+++ b/shard/shard_state.go
@@ -46,7 +46,10 @@ type Committee struct {
 	Slots   SlotList `json:"subcommittee"`
 }
 
-/* Legacy */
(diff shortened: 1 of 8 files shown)


shard/shard_state.go
```diff
@@ -45,6 +46,81 @@ type Committee struct {
 	Slots   SlotList `json:"subcommittee"`
 }
 
+/* Legacy */
+
+// StateLegacy ..
```
rlan35: explain in comment, that Legacy means before staking is enabled.

PR #1890, opened 2019-11-26
[devnet] reduce the number of nodes per shard to reach consensus in b…

LeoHChen:
…ootstrap


rlan35:
FYI, you may also wanna change these two number later if after staking the consensus is not reached: https://github.com/harmony-one/harmony/blob/master/consensus/votepower/roster.go#L11 68% voting power basically requires all our nodes to sign (if there is external validators yet)

PR #1892, opened 2019-11-26
[quorum] Add interface method to check if I am in committee, implemen…

fxfactorial:
Don't deal with any message coming in if I am not in commitee


consensus/consensus_v2.go
```diff
@@ -27,6 +27,13 @@ import (
 
 // handleMessageUpdate will update the consensus state according to received message
 func (consensus *Consensus) handleMessageUpdate(payload []byte) {
+	if !consensus.Decider.AmIMemberOfCommitee() {
```
rlan35: Move this condition under, so only "consensus.onCommitted(msg)" is executed when I am not in committee. This so the node can still catch up with blocks.

PR #1893, opened 2019-11-26
Fix staking txn replication issue

rlan35:
Tested locally.

### DIFF
--- a/api/proto/node/node.go
+++ b/api/proto/node/node.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"log"
 
+	types2 "github.com/harmony-one/harmony/staking/types"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
 	peer "github.com/libp2p/go-libp2p-peer"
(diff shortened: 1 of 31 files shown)


api/proto/node/node.go
```diff
@@ -137,6 +139,21 @@ func ConstructTransactionListMessageAccount(transactions types.Transactions) []b
 	return byteBuffer.Bytes()
 }
 
+// ConstructStakingTransactionListMessageAccount constructs serialized staking transactions in account model
+func ConstructStakingTransactionListMessageAccount(transactions types2.StakingTransactions) []byte {
+	byteBuffer := bytes.NewBuffer([]byte{byte(proto.Node)})
```
fxfactorial: maybe `[]byte{byte(proto.Node), byte(Staking), byte(Send)})`?

rlan35: Ack. Let's do it later.

node/node.go
```diff
@@ -256,6 +257,21 @@ func (node *Node) tryBroadcast(tx *types.Transaction) {
 	}
 }
 
+func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
+	msg := proto_node.ConstructStakingTransactionListMessageAccount(staking.StakingTransactions{stakingTx})
+
+	shardGroupID := nodeconfig.NewGroupIDByShardID(nodeconfig.ShardID(0)) // broadcast to beacon chain
+	utils.Logger().Info().Str("shardGroupID", string(shardGroupID)).Msg("tryBroadcastStaking")
+
+	for attempt := 0; attempt < NumTryBroadCast; attempt++ {
```
LeoHChen: will it causing problem to resend the same staking tx multiple times in retry?  will it create multiple validators on the same account?

LeoHChen:
also, please create a new branch on top of the master and squash your commits.  The current PR will cause messy commit history.

node/node.go
rlan35: No, the p2p network will screen that out if it's the same message. Also transactions can't be repeated because of nonce.

PR #1901, opened 2019-11-27
Fix staking txns reader

rlan35:


core/types/block.go
```diff
@@ -511,14 +511,16 @@ func (b *Block) WithSeal(header *block.Header) *Block {
 }
 
 // WithBody returns a new block with the given transaction and uncle contents.
-func (b *Block) WithBody(transactions []*Transaction, uncles []*block.Header, incomingReceipts CXReceiptsProofs) *Block {
+func (b *Block) WithBody(transactions []*Transaction, stakingTxns []*staking.StakingTransaction, uncles []*block.Header, incomingReceipts CXReceiptsProofs) *Block {
 	block := &Block{
 		header:           CopyHeader(b.header),
 		transactions:     make([]*Transaction, len(transactions)),
+		stakingTransactions: make([]*staking.StakingTransaction, len(stakingTxns)),
```
harmony-ek: This function is supposed to initialize all block fields by design but failed to do so, and we didn't detect that when we added `stakingTransactions` because member initializations were tagged; can you remove tags so that similar future mistakes result in compile error right here?  (Thanks Ethereum for the hint.)

harmony-ek: #1902

PR #1904, opened 2019-11-27
fix staking msg conversion crash

chaosma:


staking/types/transaction.go
```diff
@@ -39,7 +39,11 @@ func (d *txdata) CopyFrom(d2 *txdata) {
 	restored, _ := RLPDecodeStakeMsg(
```
rlan35: Should we also directly change d2.StakeMsg == nil? and avoid RLPDecodeStakeMsg?

chaosma: but in this case, we don't know if RLPDecodeStakeMsg will success or not, only if it returns nil

PR #1908, opened 2019-11-27
move stats account in the correct place; Fix edit desc bug

rlan35:


staking/types/validator.go
```diff
@@ -250,12 +250,14 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
 	if validator.Address != edit.ValidatorAddress {
 		return errAddressNotMatch
 	}
-	desc, err := UpdateDescription(edit.Description)
-	if err != nil {
-		return err
-	}
+	if edit.Description != nil {
+		desc, err := UpdateDescription(edit.Description)
```
gupadhyaya: we might still update the individual fields of the Description. In that case the check has to happen inside UpdateDescription?

rlan35: We don't have that feature yet. To update, you have to put every fields.

PR #1921, opened 2019-11-30
Staking devnet fixes

rlan35:
1. Fix block reward TotalSupply denomination problem
2. Change unlock period to 7 epochs
3. Remove node.AddNewBlock as it's basically insertChain.
4. Make more node in beacon chain broadcast new blocks
5. Add initial validator snapshot when newly created to avoid nil pointer issue.
6. Fix Decider public key provider nil pointer bug.
7. Add emergency rollback logic

Tested locally.


api/service/syncing/syncing.go
```diff
@@ -580,7 +596,7 @@ func (ss *StateSync) generateNewState(bc *core.BlockChain, worker *worker.Worker
 		if block == nil {
 			break
 		}
-		err = ss.UpdateBlockAndStatus(block, bc, worker)
+		err = ss.UpdateBlockAndStatus(block, bc, worker, false)
```
chaosma: if default is false, what if a node sent a lot of invalid block to cause new node roll back?

core/blockchain.go
```diff
@@ -68,15 +68,15 @@ const (
...
+	shardCacheLimit                    = 4
 	commitsCacheLimit                  = 10
 	epochCacheLimit                    = 10
 	randomnessCacheLimit               = 10
 	validatorCacheLimit                = 1024
 	validatorStatsCacheLimit           = 1024
 	validatorListCacheLimit            = 10
 	validatorListByDelegatorCacheLimit = 1024
-	pendingCrossLinksCacheLimit        = 2
+	pendingCrossLinksCacheLimit        = 10
```
chaosma: the shardCache and crossLinkCache only contains one key, increase cache size will not take any effect

core/blockchain.go
```diff
@@ -2546,6 +2558,7 @@ func (bc *BlockChain) ReadValidatorSnapshot(addr common.Address) (*staking.Valid
 }
 
 // WriteValidatorSnapshots writes the snapshot of provided list of validators
+// Note: this should only be called within the blockchain insertBlock process.
```
chaosma: In this case, we could make it low case method to avoid accidentally called outside.

api/service/syncing/syncing.go
rlan35: false is the same logic as before. This change is to make beacon sync work.

The state sync'ing will need more careful work in general.

core/blockchain.go
rlan35: Good point. Done for other cases.

core/blockchain.go
rlan35: ok. sounds good.

api/service/syncing/syncing.go
chaosma: I see

core/blockchain.go
rlan35: ShardCache is for shard state. I made it 10.

PR #1926, opened 2019-12-01
add validator voting power calculation based on bitmap mask

chaosma:


consensus/quorum/one-node-one-vote.go
```diff
@@ -43,6 +63,13 @@ func (v *uniformVoteWeight) IsRewardThresholdAchieved() bool {
 	return v.SignersCount(Commit) >= (v.ParticipantsCount() * 9 / 10)
 }
 
+// ComputeTotalPowerByMask computes the total power indicated by bitmap mask
+func (v *uniformVoteWeight) computeTotalPowerByMask(mask *bls_cosi.Mask) *numeric.Dec {
```
rlan35: Use int here. Don't use Dec for uniform vote

consensus/quorum/one-node-one-vote.go
```diff
@@ -33,6 +34,25 @@ func (v *uniformVoteWeight) IsQuorumAchieved(p Phase) bool {
 	return r
 }
 
+// IsQuorumAchivedByMask ..
+func (v *uniformVoteWeight) IsQuorumAchievedByMask(mask *bls_cosi.Mask) bool {
+	threshold := v.QuorumThreshold()
+	currentTotalPower := v.computeTotalPowerByMask(mask)
```
rlan35: Just use: r := v.SignersCount(p) >= v.TwoThirdsSignersCount()   Int comparison, don't use dec, it's less error-prone.

consensus/quorum/one-node-one-vote.go
chaosma: removed this function

consensus/quorum/one-node-one-vote.go
chaosma: yeah, use countsonebit

PR #1927, opened 2019-12-01
Add epoch number in shard state for use of beacon epoch sync

rlan35:
Add epoch number in the new shard state. It's necessary because with beacon epoch sync, we have to mark explicitly which epoch the new shard state is for.

Also fix cross link block proposal conditions.

Tested locally, will deploy to devnet.

### DIFF
--- a/node/node_cross_shard.go
+++ b/node/node_cross_shard.go
@@ -167,16 +167,10 @@ func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
 
 			if err = node.VerifyCrossLink(cl); err != nil {
 				utils.Logger().Debug().
-					Msgf("[ProcessingCrossLink] Crosslink blockNum %d epochNum %d shard %d skipped: %v", cl.BlockNum(), cl.Epoch().Uint64(), cl.ShardID(), cl)
+					Msgf("[ProcessingCrossLink] Failed to verify new cross link for blockNum %d epochNum %d shard %d skipped: %v", cl.BlockNum(), cl.Epoch().Uint64(), cl.ShardID(), cl)
 				continue
 			}
 
-			if err != nil {
-				utils.Logger().Error().
-					Err(err).
-					Msgf("[ProcessingCrossLink] Failed to verify new cross link for shardID %d, blockNum %d", cl.ShardID(), cl.Number())
-				continue
-			}
 			candidates = append(candidates, cl)
 			utils.Logger().Debug().
 				Msgf("[ProcessingCrossLink] committing for shardID %d, blockNum %d", cl.ShardID(), cl.Number().Uint64())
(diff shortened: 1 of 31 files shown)


staking/types/validator.go
```diff
@@ -297,6 +284,19 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
 			validator.SlotPubKeys = append(validator.SlotPubKeys[:index], validator.SlotPubKeys[index+1:]...)
 		}
 	}
+
```
rlan35: Changed the order to put add after remove. To avoid the problem when add/remove the same existing key.

node/node_cross_shard.go
```diff
@@ -170,7 +165,12 @@ func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
...
 			}
 
-			err = node.VerifyCrossLink(cl)
+			if err = node.VerifyCrossLink(cl); err != nil {
+				utils.Logger().Debug().
+					Msgf("[ProcessingCrossLink] Crosslink blockNum %d epochNum %d shard %d skipped: %v", cl.BlockNum(), cl.Epoch().Uint64(), cl.ShardID(), cl)
+				continue
+			}
+
 			if err != nil {
```
chaosma: duplicate with above err != nil

node/worker/worker.go
```diff
@@ -262,10 +262,15 @@ func (w *Worker) GetNewEpoch() *big.Int {
...
 
-	// TODO: Don't depend on sharding state for epoch change.
-	if len(parent.Header().ShardState()) > 0 && parent.NumberU64() != 0 {
-		// ... except if parent has a resharding assignment it increases by 1.
-		epoch = epoch.Add(epoch, common.Big1)
+	shardState, err := parent.Header().GetShardState()
+	if err == nil && shardState.Epoch != nil {
+		// For shard state of staking epochs, the shard state will have an epoch and it will decide the next epoch for following blocks
+		epoch = new(big.Int).Set(shardState.Epoch)
+	} else {
```
chaosma: when does the "else" case happen, should we require shardState.Epoch always not empty?

node/worker/worker.go
```diff
@@ -315,15 +320,11 @@ func (w *Worker) SuperCommitteeForNextEpoch(
 					beaconEpoch, beacon,
```
chaosma: Is it possible that different validator has different speed, their database are not synced. Then their proposed epoch will be different

node/worker/worker.go
rlan35: For pre-staking case, the state.Epoch is nil. After staking, it's not nil.

node/worker/worker.go
rlan35: resolved offline.

node/worker/worker.go
chaosma: In this case, it's better to use IsStakeEpoch instead of checking it's nil, which just depending on the proposer

rlan35: Already added the check in local. Won't bother triggering another jenkins. Will add in following PR.

PR #1929, opened 2019-12-01
[reward] Add reward Schedule table from spreadsheet

fxfactorial:
Add Initial table from spreadsheet, need to change keys as unix time stamps to be used by block in reward.go


rlan35:
Looks good. I assume the timestamps and percentage are programmically converted to the code, right?

PR #1930, opened 2019-12-01
[votepower] Allow negative diff, explicit check sum to one

fxfactorial:


consensus/votepower/roster.go
```diff
@@ -91,10 +94,18 @@ func Compute(staked shard.SlotList) *Roster {
 			Msg("sum of voting power of hmy & staked slots not equal to 1, gave diff to staked slot")
 	}
 
+	if !ourPercentage.Add(theirPercentage).Equal(numeric.OneDec()) {
+		utils.Logger().Error().
+			Str("theirs", theirPercentage.String()).
+			Str("ours", ourPercentage.String()).
+			Msg("Voting power not equal in percentage")
```
rlan35: -> "Total voting power not equal 100 percent"

PR #1935, opened 2019-12-02
[votingpower] Extend vote power snapshot, use existing roster data structure

fxfactorial:
Save vote power snapshot after each block

### DIFF
--- a/consensus/quorum/one-node-staked-vote.go
+++ b/consensus/quorum/one-node-staked-vote.go
@@ -230,7 +230,7 @@ func (v *stakedVoteWeight) JSON() string {
 		IsHarmony   bool   `json:"is-harmony-slot"`
 		Identity    string `json:"bls-public-key"`
 		VotingPower string `json:"voting-power-%"`
-		RawStake    string `json:"raw-stake,omitempty"`
+		EPoSedStake string `json:"effective-stake,omitempty"`
 	}
 
 	type t struct {
@@ -254,7 +254,7 @@ func (v *stakedVoteWeight) JSON() string {
 			"",
 		}
 		if !voter.IsHarmonyNode {
-			member.RawStake = voter.RawStake.String()
+			member.EPoSedStake = voter.EPoSedStake.String()
 		}
 		parts[i] = member
 		i++
--- a/consensus/quorum/one-node-staked-vote.go
+++ b/consensus/quorum/one-node-staked-vote.go
@@ -227,10 +227,10 @@ func (v *stakedVoteWeight) JSON() string {
 	voterCount := len(v.roster.Voters)
 
 	type u struct {
-		IsHarmony   bool   `json:"is-harmony-slot"`
-		Identity    string `json:"bls-public-key"`
-		VotingPower string `json:"voting-power-%"`
-		EPoSedStake string `json:"effective-stake,omitempty"`
+		IsHarmony      bool   `json:"is-harmony-slot"`
+		Identity       string `json:"bls-public-key"`
+		VotingPower    string `json:"voting-power-%"`
+		EffectiveStake string `json:"effective-stake,omitempty"`
 	}
 
 	type t struct {
@@ -254,7 +254,7 @@ func (v *stakedVoteWeight) JSON() string {
 			"",
 		}
 		if !voter.IsHarmonyNode {
-			member.EPoSedStake = voter.EPoSedStake.String()
+			member.EffectiveStake = voter.EffectiveStake.String()
 		}
 		parts[i] = member
 		i++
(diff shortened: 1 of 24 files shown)


core/blockchain.go
```diff
@@ -1344,6 +1349,12 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
 		bc.insert(block)
 	}
 	bc.futureBlocks.Remove(block.Hash())
+
+	if header.ShardID() == shard.BeaconChainShardID {
+		r, _ := committee.EPOSStakedCommittee(shard.Schedule.InstanceForEpoch(header.Epoch()), bc, 320)
```
rlan35: It shouldn't be 320 hardcoded. It should be the same logic as in Compute.

core/blockchain.go
```diff
@@ -1344,6 +1349,12 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
 		bc.insert(block)
 	}
 	bc.futureBlocks.Remove(block.Hash())
+
+	if header.ShardID() == shard.BeaconChainShardID {
```
rlan35: This should only be executed during pre-staking time, bascially preStakingEpoch <= Epoch < StakingEpoch

core/blockchain.go
```diff
@@ -2628,50 +2636,72 @@ func (bc *BlockChain) UpdateValidatorVotingPower(state *shard.State) error {
...
+				)
+				stat.validator.BLSKeyPerShard = append(
+					stat.validator.BLSKeyPerShard, staking.KeyPerShard{
+						shardID, roster.Voters[slot.BlsPublicKey].Identity.Hex(),
+					},
+				)
+				stat.validator.AvgVotingPower = stat.validator.AvgVotingPower.Quo(
+					numeric.NewDec(int64(len(rosters))),
+				)
+				stats = append(stats, stat)
```
rlan35: This logic doesn't seems right. The stats should be by validator. but here you have a stats for each bls key. Also the voting power should be total voting power of the validator in each shard divided by 4. But since you have separated voting powers for each key, probably better just remove the avgVotingPower and let frontend compute it.

fxfactorial: I thought about that too, but wasn't sure what the right answer would be (about stats by validator vs stats for each bls key). So would it be then a group by on the earning one address, right?

consensus/quorum/one-node-staked-vote.go
```diff
@@ -230,7 +230,7 @@ func (v *stakedVoteWeight) JSON() string {
 		IsHarmony   bool   `json:"is-harmony-slot"`
 		Identity    string `json:"bls-public-key"`
 		VotingPower string `json:"voting-power-%"`
-		RawStake    string `json:"raw-stake,omitempty"`
+		EPoSedStake string `json:"effective-stake,omitempty"`
```
rlan35: What exactly is rawStatke? If it's the effective stake, just call it EffectiveStake here.

fxfactorial: right, its the stake post epos processing, the rawstake naming was misleading

consensus/votepower/roster.go
```diff
@@ -18,11 +22,12 @@ var (
...
-	EarningAccount          common.Address
-	Identity                shard.BlsPublicKey
-	EffectivePercent        numeric.Dec
-	RawStake                numeric.Dec
+	IsActive         bool               `json:"is-active"`
+	IsHarmonyNode    bool               `json:"is-harmony"`
+	EarningAccount   common.Address     `json:"earning-account"`
+	Identity         shard.BlsPublicKey `json:"bls-public-key"`
+	EffectivePercent numeric.Dec        `json:"voting"`
+	EPoSedStake      numeric.Dec        `json:"eposed-stake"`
```
rlan35: EffectiveStake

core/blockchain.go
```diff
@@ -2628,50 +2636,39 @@ func (bc *BlockChain) UpdateValidatorVotingPower(state *shard.State) error {
...
+	for key, value := range networkWide {
+		statsFromDB, err := rawdb.ReadValidatorStats(bc.db, key)
+		if statsFromDB == nil {
+			statsFromDB = &staking.ValidatorStats{
+				big.NewInt(0), big.NewInt(0), big.NewInt(0),
+				numeric.NewDec(0), numeric.NewDec(0),
+				[]staking.VotePerShard{}, []staking.KeysPerShard{},
+			}
+		}
+		statsFromDB.AvgVotingPower = value.AverageVotingPower
```
rlan35: This can be removed. You can counting the avg voting power per slot, which is not interesting. It should be the avg of powers in four different shards, which can be computed in the VotingPowerPerShard. So let's just remove this field.

PR #1936, opened 2019-12-03
Fixes crosslinks proposal and verification bugs

rlan35:
Also make the peer discovery initially have more tries.

Tested on devnet.


api/service/discovery/service.go
```diff
@@ -86,6 +85,8 @@ func (s *Service) contactP2pPeers() {
 	msgBuf := host.ConstructP2pMessage(byte(0), pingMsg.ConstructPingMessage())
 	s.sentPingMessage(s.config.ShardGroupID, msgBuf)
 
+	pingInterval := 5
+	initialFlatRetries := 20 // no expotential backoff for 20 times.
```
chaosma: Is this related to the bug fix or just some optimization?

rlan35: optimization

PR #1937, opened 2019-12-03
remove ShardStateHash from header v3

chaosma:


node/worker/worker.go
```diff
@@ -390,7 +390,10 @@ func (w *Worker) FinalizeNewBlock(sig []byte, signers []byte, viewID uint64, coi
 
 	// Shard State
 	if shardState != nil && len(shardState.Shards) != 0 {
-		w.current.header.SetShardStateHash(shardState.Hash())
+		//we store shardstatehash in header only before prestaking epoch (header v0,v1,v2)
+		if !w.config.IsPreStaking(shardState.Epoch) {
```
rlan35: We shouldn't count in the shardState.epoch, it should be the block's epoch. they are not the same thing.

chaosma: shardState.Epoch is the next epoch?

PR #1940, opened 2019-12-03
Staking devnet fixes

rlan35:


consensus/consensus_v2.go
```diff
@@ -788,7 +788,6 @@ func (consensus *Consensus) finalizeCommits() {
 		return
 	}
 	consensus.FBFTLog.AddMessage(pbftMsg)
-	consensus.ChainReader.WriteLastCommits(pbftMsg.Payload)
```
rlan35: Move this after try catchup, because try catchup can fail.

PR #1943, opened 2019-12-03
[SHELVED][COME-BACK-TO] Unavailability: Kick out inactive validators 

fxfactorial:
Kick out inactive validators once met threshold of epoch worth of blocks passing. Resets to 0 when signs a single block. 

EDIT: Shelved for now - do not close PR; we will come back to this because nuance in blocknumber for epoch for shard-chain and incorporation of that logic requires more nuance in thought re crosslink.


core/blockchain.go
```diff
@@ -2617,9 +2619,32 @@ func (bc *BlockChain) UpdateValidatorUptime(slots shard.SlotList, mask *bls.Mask
...
+				}
+
+				faithful := []common.Address{}
+				mia := addr.Bytes()
+				for i := range active {
+					if bytes.Compare(mia, active[i].Bytes()) != 0 {
+						faithful = append(faithful, active[i])
+					}
+				}
+				if err := bc.WriteActiveValidatorList(faithful); err != nil {
```
rlan35: This is not the way. ActiveValidatorList is the list of validators that's selected in the committee. By removing it from the list doesn't rule them out from next epoch committee selection. Instead, you should flip the IsActive flag of the validator information in the account state. And when you compute committee in next epoch. don't select the ones with IsActive==false.

rlan35: Since IsActive is part of the account state so it affect state root. So flipping the IsActive field should happen before the state root is calculated. Somewhere in the Finalize() method should work. Similar to the delegation withdrawal payout code.

fxfactorial: 1) I thought it was probably not the right way but was irresistibly easy, will amend. 
2) Also thought that issue would happen but wasn't sure right place, will amend

shard/committee/assignment.go
```diff
@@ -117,7 +117,12 @@ func eposStakedCommittee(
 	// TODO benchmark difference if went with data structure that sorts on insert
 	for i := range candidates {
 		validator, err := stakerReader.ReadValidatorInformation(candidates[i])
```
rlan35: Just realized this part may have potential problem. This ReadValidatorInformation is reading from the latest snapshot of the validator. Which means the eposStakedCommittee and the Compute() func will only work if it's called for the head of the chain. It won't work if it's called to compute for a history epoch.

rlan35: But I just checked the codebase, seems we never run Compute to calculate an epoch that's in the history, so it's good for now.

internal/chain/engine.go
```diff
@@ -254,9 +254,28 @@ func (e *engineImpl) Finalize(
...
+
+				count := len(stats.Delinquency.Missing)
+				didMiss := 0
+
+				for _, mia := range stats.Delinquency.Missing {
+					if mia.Count >= slash.MissedThresholdForInactive {
+						didMiss++
+					}
+				}
+				if didMiss >= ((count * (2 / 3)) + 1) {
```
rlan35: This is only counting 66% of nodes that' not signing for the whole epoch. I was saying counting 66% of all the blocks that the validator should sign but missed across all keys.

fxfactorial:
Closed - will implement clean slate from latest master.

PR #1950, opened 2019-12-03
add shardState verification

chaosma:
- refactor shardState proposal into blockchain which can be shared between leader and validators
- add validator verification on shardState

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -49,6 +49,7 @@ import (
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/numeric"
 	"github.com/harmony-one/harmony/shard"
+	"github.com/harmony-one/harmony/shard/committee"
 	staking "github.com/harmony-one/harmony/staking/types"
 	lru "github.com/hashicorp/golang-lru"
 )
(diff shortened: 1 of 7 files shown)


internal/chain/engine.go
```diff
@@ -145,6 +146,34 @@ func ReadPublicKeysFromLastBlock(bc engine.ChainReader, header *block.Header) ([
 	return GetPublicKeys(bc, parentHeader, false)
 }
 
+// VerifyShardState implements Engine, checking the shardstate is valid at epoch transition
+func (e *engineImpl) VerifyShardState(bc engine.ChainReader, beacon engine.ChainReader, header *block.Header) error {
+	headerShardState, _ := header.GetShardState()
```
rlan35: Just read the raw bytes directly. This first decode to shard.State. And you later encode it again... no need.

core/blockchain.go
```diff
@@ -2964,3 +2965,74 @@ func (bc *BlockChain) GetECDSAFromCoinbase(header *block.Header) (common.Address
...
+	case shard.BeaconChainShardID:
+		if shard.Schedule.IsLastBlock(header.Number().Uint64()) {
+			nextCommittee, err = committee.WithStakingEnabled.Compute(
+				new(big.Int).Add(header.Epoch(), common.Big1),
+				beacon,
+			)
+		}
+	default:
+		// TODO: needs to make sure beacon chain sync works.
+		beaconEpoch := beacon.CurrentHeader().Epoch()
```
rlan35: When this is called as verification instead of proposing. It shouldn't read from the CurrentHeader of beacon because it will change based on different timing. It should use the shardState.Epoch as beaconEpoch. In the case of before-staking, the shardState.epoch is nil, and in that case, the beaconEpoch is not needed for verification either.

internal/chain/engine.go
chaosma: This is when there are issue to debug. will change back

core/blockchain.go
```diff
@@ -2987,7 +2991,17 @@ func (bc *BlockChain) SuperCommitteeForNextEpoch(
...
 	default:
 		// TODO: needs to make sure beacon chain sync works.
-		beaconEpoch := beacon.CurrentHeader().Epoch()
+		if isVerify {
+			//verify
+			shardState, err := header.GetShardState()
+			if err != nil {
+				return &shard.State{}, err
+			}
+			beaconEpoch = shardState.Epoch
```
rlan35: shardState.Epoch is nil when it's not staking epoch.

node/node_handler.go
```diff
@@ -264,6 +264,16 @@ func (node *Node) VerifyNewBlock(newBlock *types.Block) error {
 			"my shard ID", node.Blockchain().ShardID(),
 			"new block's shard ID", newBlock.ShardID())
 	}
```
rlan35: This VerifyNewBlock is only called for consensus block verification, but not for block sync verification. We should consolidate it into a single func called by all places that requires block verification. Please add a todo.

chaosma: I first want to put it into verifyHeader, but verifyShardState need beaconchain as parameter, so I put it outside. will add a todo

PR #1951, opened 2019-12-04
Use one1 address for staking rpcs.

rlan35:
tested locally.

### DIFF
--- a/internal/hmyapi/types.go
+++ b/internal/hmyapi/types.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/harmony-one/go-sdk/pkg/address"
-
 	"github.com/harmony-one/harmony/shard"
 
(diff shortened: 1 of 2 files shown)


internal/hmyapi/types.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/harmony-one/go-sdk/pkg/address"
```
fxfactorial: wrong import - this will give go-sdk cyclic imports problem

rlan35: yes, corrected.

rlan35:
> Travis failed. Better to use internal/common as internal_common, and methods from there rather than address2 from go-sdk/pkg/address. As less as possible outside of repo libs usage.

good point. corrected.

rlan35:
jenkins, test this please.

PR #1981, opened 2019-12-07
correct the beneficiary of coinbase address

chaosma:


node/node_newblock.go
```diff
@@ -84,6 +84,7 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 
 	// Update worker's current header and state data in preparation to propose/process new transactions
 	coinbase := node.Consensus.SelfAddress
+	beneficiary := node.Consensus.SelfAddress
```
rlan35: Just realize this won't work after staking, after staking, the SelfAddress is the bls address. We need to derive the ecdsa address from shard state. p.bc.GetECDSAFromCoinbase

rlan35:
Already fixed in another PR.

PR #1982, opened 2019-12-07
reduce crosslink batch and disable validator block broadcasting

rlan35:


node/node_handler.go
```diff
@@ -360,7 +360,7 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block, commitSigAndBit
 		// 15% of the validator also need to do broadcasting
 		rand.Seed(time.Now().UTC().UnixNano())
 		rnd := rand.Intn(100)
-		if rnd < 1 {
+		if rnd < 0 {
```
denniswon: so this is where we turn off for non-leaders

node/node_newblock.go
```diff
@@ -127,7 +127,7 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 
 	// Prepare cross links
 	var crossLinksToPropose types.CrossLinks
-	if node.NodeConfig.ShardID == 0 {
+	if node.NodeConfig.ShardID == 0 && node.Blockchain().Config().IsStaking(node.Worker.GetCurrentHeader().Epoch()) {
```
denniswon: and this is where we turn off for the leader "until the staking epoch"?
Is the intention to disable crosslink during the pre-staking period?
What is the reason for turning off the crosslink?

cc: @harmony-ek

node/node_handler.go
rlan35: yes

node/node_newblock.go
rlan35: Cross link potentially is causing network trouble.

PR #1986, opened 2019-12-07
[rpc][staking] Quick way to add an staking transaction error sink for…

fxfactorial:
Quick way to add a way for downstream consumers to see what is going wrong with latest staking transactions. 

```
loc2='http://localhost:9500'

 curl -d '{
    "jsonrpc": "2.0",
     "method": "hmy_getCurrentTransactionErrorSink",
     "params": [],
     "id": 1
}' -H "Content-Type: application/json" -X POST ${loc2}
```

### DIFF
--- a/internal/hmyapi/harmony.go
+++ b/internal/hmyapi/harmony.go
@@ -6,8 +6,8 @@ import (
 
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/harmony-one/harmony/api/proto"
-	"github.com/harmony-one/harmony/core"
 	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
+	staking "github.com/harmony-one/harmony/staking/types"
 )
 
 // PublicHarmonyAPI provides an API to access Harmony related information.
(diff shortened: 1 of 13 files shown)


internal/hmyapi/harmony.go
```diff
@@ -65,3 +66,21 @@ func (s *PublicHarmonyAPI) GetNodeMetadata() NodeMetadata {
...
+func (s *PublicHarmonyAPI) GetCurrentTransactionErrorSink() []core.StakingTransactionError {
+	result := []core.StakingTransactionError{}
+	count := 0
+	core.StakingTransactionErrorSink.Range(func(key, value interface{}) bool {
+		count++
+		result = append(result, value.(core.StakingTransactionError))
+		return true
+	})
+	const maxMapSize = 1024
+	if count == maxMapSize {
```
rlan35: This logic is confusing to me. Something is wrong.

fxfactorial: If I reached the max size that I wanted to keep in memory, then delete it all

fxfactorial: alternative i thought of using is a circular queue, or something circular data structure

node/rpc.go
```diff
@@ -49,10 +49,17 @@ func (node *Node) IsCurrentlyLeader() bool {
 	return node.Consensus.IsLeader()
 }
 
+// ErroredStakingTransactionSink is the inmemory failed staking transactions this node has
+func (node *Node) ErroredStakingTransactionSink() []staking.RPCTransactionError {
+	return node.errorSink.failedTxns
```
rlan35: This will also need lock too. Reading when other writing.

PR #1988, opened 2019-12-07
fix crosslink bugs

rlan35:


core/blockchain.go
```diff
@@ -2337,11 +2336,11 @@ func (bc *BlockChain) DeleteCommittedFromPendingCrossLinks(crossLinks []types.Cr
 	}
 
 	pendingCLs := []types.CrossLink{}
-Loop:
+
 	for _, cl := range cls {
 		if _, ok := m[cl.ShardID()]; ok {
 			if _, ok1 := m[cl.ShardID()][cl.BlockNum()]; ok1 {
-				continue Loop
```
chaosma: with or without Loop both are correct: https://play.golang.org/p/kD5ZDZepoaT

rlan35: ok, please approve.

PR #1995, opened 2019-12-08
add flag to revert only beacon chain; also fix a epoch calc bug

rlan35:


internal/configs/sharding/testnet.go
```diff
@@ -39,7 +39,7 @@ func (testnetSchedule) BlocksPerEpoch() uint64 {
 }
 
 func (ts testnetSchedule) CalcEpochNumber(blockNum uint64) *big.Int {
-	epoch := blockNum % ts.BlocksPerEpoch()
+	epoch := blockNum / ts.BlocksPerEpoch()
```
fxfactorial: what about the other networks, is this mistake there too?

rlan35: i checked it’s only a one off mistake and had no real impact before because this method is not really used

PR #1996, opened 2019-12-09
fix beneficiary issue

rlan35:


cmd/harmony/main.go
```diff
@@ -295,7 +294,9 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 		return currentConsensus.PubKey, nil
 	})
 
-	currentConsensus.SelfAddress = common.ParseAddr(initialAccount.Address)
+	if initialAccount.Address != "" { // staking validator doesn't have to specify ECDSA address
+		currentConsensus.SelfAddress = common.ParseAddr(initialAccount.Address)
```
chaosma: Is initialAccount.Address always bls address?

node/node_newblock.go
```diff
@@ -83,14 +84,27 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
+		err         error
+	)
 
 	// After staking, all coinbase will be the address of bls pub key
 	if node.Blockchain().Config().IsStaking(node.Worker.GetCurrentHeader().Epoch()) {
 		addr := common.Address{}
-		blsPubKeyBytes := node.Consensus.PubKey.GetAddress()
-		addr.SetBytes(blsPubKeyBytes[:])
-		coinbase = addr
+		if bytes.Compare(coinbase[:], addr[:]) == 0 { // empty SelfAddress means it's a staking validator
```
chaosma: what if coinbase not equal bls address (unless selfaddress always equal bls address in this case)?

cmd/harmony/main.go
rlan35: Nop. I changed it to always be the ECDSA address or empty "" if it's staking validator.

node/node_newblock.go
rlan35: Here it's checking whether SelfAddress is empty. If it's empty, it means it's staking validator, and the ECDSA address needs to be looked up.

PR #1998, opened 2019-12-09
[build] add back library files to md5sum

LeoHChen:


LeoHChen:
missing library files in md5sum will cause new node not able to download them and won't be able to start the node software if the software is not statically linked.

PR #2006, opened 2019-12-11
[config] Add Local development config

fxfactorial:


internal/configs/node/config.go
```diff
@@ -297,6 +297,6 @@ func (t NetworkType) ChainConfig() params.ChainConfig {
 	case Pangaea:
 		return *params.PangaeaChainConfig
 	default:
-		return *params.TestnetChainConfig
```
rlan35: This is not enough. You also need to change the deployment script which I think it putting testnet or devnet as network type.

PR #2013, opened 2019-12-11
Update the BLS pub key and signature aggregation to prevent rogue key attack

coolcottontail:
## Update the BLS pub key and signature aggregation to prevent rogue key attack

This PR implements  the proposal from  https://crypto.stanford.edu/~dabo/pubs/papers/BLSmultisig.html

NOTE:
It is not a GENERIC BLS aggregation algorithm.  It is ONLY USEFUL if the message been signed is the same.  It works for Harmony's use case.

To make the transition process as easy as possible, I updated the constructor for creating a new Mask.

func NewMask(publics []*bls.PublicKey, myKey *bls.PublicKey, optionalUsePubKeyHash...bool) (*Mask, error) 

The be aggregation algorithm can ONLY be turned on when the Mask constructor has an new optional parameter optionalUsePubKeyHash == true. 

To use it, just replace the old one

1) mask creation:
from:
prepareBitmap, _ := bls_cosi.NewMask(members, nil)
to:
prepareBitmap, _ := bls_cosi.NewMask(members, nil, true)

2) signature aggregation

m2Sig := bls_cosi.AggregateSig(...)

with the new one:

m2Sig := bls_cosi.AggregateSigWithPublicKey(...)

### DIFF
--- a/crypto/bls/bls_test.go
+++ b/crypto/bls/bls_test.go
@@ -245,14 +245,13 @@ func TestSetMask(test *testing.T) {
 	}
 }
 
-
 func TestNewImplementationWithPublicKeyHash1(test *testing.T) {
 	secs := make([]*bls.SecretKey, 10)
 	pubs := make([]*bls.PublicKey, 10)
 	sigs := make([]*bls.Sign, 10)
 
 	m := "test test"
-	for i:= 0; i < 10; i++ {
+	for i := 0; i < 10; i++ {
 		secs[i] = RandPrivateKey()
 		pubs[i] = secs[i].GetPublicKey()
 		sigs[i] = secs[i].Sign(m)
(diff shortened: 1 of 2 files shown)


rlan35:
Does this invalidate the need for key ownership proof? Also, any performance impact and what's the migration plan?

crypto/bls/bls_test.go
```diff
@@ -244,3 +244,84 @@ func TestSetMask(test *testing.T) {
...
+	aggSig := sigs[0]
+
+	for i := 1; i < 10; i++ {
+		aggPub.Add(TransformPublicKey(pubs[i]))
+		if i != 5 {
+			aggSig.Add(TransformSignature(sigs[i], pubs[i].Serialize()))
+		}
+	}
+
+	aggPub.Sub(TransformPublicKey(pubs[5]))
```
chaosma: what is this test case used for?

rlan35:
Please do complexity analysis of the new multi-sig algorithm compared to the existing one.

fxfactorial:
@coolcottontail what else needed for this PR?

PR #2021, opened 2019-12-12
add debug flag to avoid too much log info in crosslink verification

chaosma:


chaosma:
Jenkins test this please

Issue #2009, opened 2019-12-11
Able to send tx while wallet account has been staked


sophoah:
**Describe the bug**
In devnet, I was able to send ONE token while the fund was supposed to be blocked through staking.

**To Reproduce**
**1) Make sure we have fund in the wallet:**
root@vmi317161:~ /hmy-devnet# ./hmy balances one15knq45tf5psjze6q65dpm4lumqhm8jny997mph
[
  {
    "shard": 0,
    "amount": 7488.0000
  },
  {
    "shard": 1,
    "amount": 0.0000
  }
]
root@vmi317161:~ /hmy-devnet

**2) Create a new validator and stake the entire fund**
root@vmi317161:~ /hmy-devnet# ./hmy staking create-validator --validator-addr one15knq45tf5psjze6q65dpm4lumqhm8jny997mph --name Soph --identity Soph --website Soph@soph.one --security-contact Soph --details "Soph the validator" --rate 0.1 --max-rate 0.9 --max-change-rate 0.05 --min-self-delegation 2 --max-total-delegation 99999999999 --bls-pubkeys 0xb3727f8a497174dba2d4859a131a79f4065fa8d18b5e51e5624bc7dcc1e39ae25b0eb5a0b9a4b25a49c9ee3d3ebf7804 --amount 7487 --chain-id devnet
Enter the bls passphrase:
{"transaction-receipt":"0x156605f7a0b05bad73195171e4c05dfe1e6a3fcd6e89c9f00132bd9fa7d2c9c3"}

**3) checking wallet again** 
root@vmi317161:~ /hmy-devnet# ./hmy balances one15knq45tf5psjze6q65dpm4lumqhm8jny997mph
[
  {
    "shard": 0,
    "amount": 7487.999960252
  },
  {
    "shard": 1,
    "amount": 0.0000
  }
]

Gas fees only deducted from the wallet, shouldn't the stake amount be blocked by sending it somewhere ? or I was hoping it would be known to be blocked by the protocol

**4) try to send token**
root@vmi317161:~ /hmy-devnet# ./hmy transfer --from one15knq45tf5psjze6q65dpm4lumqhm8jny997mph --to one1v9dq279psll5lyxmtkuv4s4xzwn7z2x3dxk445 --from-shard 0 --to-shard 0 --amount 50 --chain-id devnet
{"transaction-receipt":"0x3655af121d1e687b910c2f5973370363955c2ed601bfd3d1a866cda8917e1a0e"}
[...]

PR #2029, opened 2019-12-13
Fix create validator balance reduce bug; Add LastEpochInCommittee

rlan35:
## Issue

Fixing the issue that createValidator doesn't reduce balance from the validator address: https://github.com/harmony-one/harmony/issues/2009

Add LastEpochInCommittee so delegation for validators that's never in committee or have been out of committee for 7 epochs can be immediately unlocked.

## Test

locally tested.


fxfactorial:
@rlan35 why intend a merge to staking_pangaea, not master?

PR #2035, opened 2019-12-14
[rpc] Expose median raw stake for explorer nodes only, remove dead st…

fxfactorial:


hmy/api_backend.go
```diff
@@ -307,6 +308,41 @@ func (b *APIBackend) GetValidatorInformation(addr common.Address) *staking.Valid
...
+			stake.Add(stake, validator.Delegations[i].Amount)
+		}
+		stakes = append(stakes, stake)
+	}
+	sort.SliceStable(
+		stakes,
+		func(i, j int) bool { return stakes[i].Cmp(stakes[j]) == -1 },
+	)
+	const isEven = 0
+	switch l := len(stakes); l % 2 {
```
rlan35: Should be on only the top 320. Not everyone who have staked.

hmy/api_backend.go
```diff
@@ -307,6 +308,41 @@ func (b *APIBackend) GetValidatorInformation(addr common.Address) *staking.Valid
 	return &val.Validator
 }
 
+var (
+	two = big.NewInt(2)
+)
+
+// GetMedianRawStakeSnapshot  ..
+func (b *APIBackend) GetMedianRawStakeSnapshot() *big.Int {
+	candidates := b.hmy.BlockChain().ValidatorCandidates()
```
rlan35: This only works on beacon chain. So better just expose this rpc for beacon chain explorer.

PR #2037, opened 2019-12-15
fix mint new tokens when transfer money to validator addr

chaosma:
- checking to.address() is staking validator addr or not
- avoid create cxReceipts in case of evm execution fail

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -57,6 +57,7 @@ func NewEVMContext(msg Message, header *block.Header, chain ChainContext, author
 	return vm.Context{
 		CanTransfer: CanTransfer,
 		Transfer:    Transfer,
+		IsValidator: IsValidator,
 		GetHash:     GetHashFn(header, chain),
 		Origin:      msg.From(),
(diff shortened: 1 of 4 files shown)


fxfactorial:
I asked @Daniel-VDM to explicitly add this case as part of our automated staking tests.

sophoah:
@fxfactorial @Daniel-VDM could you guys maybe shared with the P-OPS how you guys are developping those test and see if we can help, there are other scenarios like the create-validator with self-delegation that result today with a non deduction of the fund that was stake (ie issue 2008/2009 that was worked by @rlan35 )

we could then also develop more automated test that we could use in the future.

fxfactorial:
@sophoah Check here https://github.com/harmony-one/harmony-ops/blob/master/test-automation/api-tests/test.py#L166 and @Daniel-VDM has an open branch of further work.

LeoHChen:
do we need to cherry-pick this change to s3 branch?  @rlan35

PR #2041, opened 2019-12-21
[node] option to specify the pub port

LeoHChen:


scripts/node.sh
```diff
@@ -183,8 +191,10 @@ do
    p) blspass="${OPTARG}";;
    d) download_only=true;;
    D) do_not_download=true;;
+   M) multi_key=true;;
```
rlan35: Disable multi key before it’s actually deployed?

LeoHChen: default is off. only enable it with "-M" option, still testing.

PR #2045, opened 2020-01-07
Extend failure sink

fxfactorial:
Change implementation of in memory buffer to ring buffer, expand failed txns to include plain txns as well.


rlan35:
What's the test result?

rlan35:
@gupadhyaya please also review.

fxfactorial:
@rlan35 Output should be same as it was with the prior usage of staking txn failures: 

`./hmy failures plain` or `./hmy failures staking`

PR #2046, opened 2020-01-07
[node][rpc] Backport logic of failure sink for plain txn

fxfactorial:
Later we should revert this commit before we merge in master into s3 to avoid needless merge conflicts.


node/worker/worker.go
```diff
@@ -239,9 +243,17 @@ func (w *Worker) CommitTransactions(
...
 	for _, tx := range txs {
 		snap := w.current.state.Snapshot()
 		_, err := w.commitTransaction(tx, coinbase)
+		if err != nil {
+			erroredTxns = append(erroredTxns, types.RPCTransactionError{
+				tx.Hash().Hex(), time.Now().Unix(), err.Error(),
+			})
+		}
+
 		if err != nil {
```
gupadhyaya: The two checks should be combined and also `txnErrorSink(erroredTxns)` should happen inside the branch. Otherwise, the second if condition will skip it and return right?

core/types/transaction.go
```diff
@@ -33,6 +33,13 @@ import (
 	common2 "github.com/harmony-one/harmony/internal/common"
 )
 
+// RPCTransactionError ..
+type RPCTransactionError struct {
+	TxHashID             string `json:"tx-hash-id"`
```
LeoHChen: are we consistent with all the json string?  I saw some complaint about inconsistent json field name in the API.

node/worker/worker.go
fxfactorial: good catch - this file changed heavily from s3 to master so this mistake

node/node_newblock.go
```diff
@@ -82,8 +82,16 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 
 	// Prepare transactions including staking transactions
 	selectedTxs, selectedStakingTxs := node.getTransactionsForNewBlock(coinbase)
-
-	if err := node.Worker.CommitTransactions(selectedTxs, selectedStakingTxs, coinbase); err != nil {
+	if err := node.Worker.CommitTransactions(
+		selectedTxs, selectedStakingTxs, coinbase,
+		func(payload []types.RPCTransactionError) {
+			node.errorSink.Lock()
```
LeoHChen: do you need to unlock?

core/types/transaction.go
fxfactorial: JSON APIs should be kebab-case, but not all places do that, can clean up the snake_case JSON APIs later

node/node_newblock.go
fxfactorial: ah, good catch, copy paste from master commit was cut off , fixed

PR #2047, opened 2020-01-07
update shard configs for testnet launching

AndyBoWu:
### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**

### DIFF
--- a/internal/configs/sharding/pangaea.go
+++ b/internal/configs/sharding/pangaea.go
@@ -12,9 +12,9 @@ import (
 
 const (
 	// PangaeaHTTPPattern is the http pattern for pangaea.
-	PangaeaHTTPPattern = "https://api.s%d.t2.hmny.io"
+	PangaeaHTTPPattern = "https://api.s%d.pga.hmny.io"
 	// PangaeaWSPattern is the websocket pattern for pangaea.
-	PangaeaWSPattern = "wss://ws.s%d.t2.hmny.io"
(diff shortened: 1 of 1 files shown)


internal/configs/sharding/pangaea.go
```diff
@@ -56,7 +56,7 @@ func (ps pangaeaSchedule) ConsensusRatio() float64 {
 var pangaeaReshardingEpoch = []*big.Int{common.Big0}
 
 var pangaeaV0 = MustNewInstance(
-	2, 50, 40, genesis.PangaeaAccounts, genesis.FoundationalPangaeaAccounts, pangaeaReshardingEpoch)
+	3, 250, 250, genesis.PangaeaAccounts, genesis.FoundationalPangaeaAccounts, pangaeaReshardingEpoch)
```
LeoHChen: 1. PangaeaAccounts has not enough keys, please use TNHarmony accounts. 
2. do we really need 250 own nodes?  Can we just use 220 so that we can leave some room for potential FN in testnet?

AndyBoWu: 1. did some verification.

genesis.PangaeaAccounts has 80 pre-created accounts, which is not enough for the testnet creation.

genesis.tn_harmony has 800 accounts. updated this change.

2. 250 * 2 / 3 = 167 account to reach consensus. So, it should be fine as long as the number of nodes is more than that. 220 is a good number too.

Issue #2033, opened 2019-12-14
[Feature] RPC API upgrades


flicker-harmony:
# Problem/limitation At Hand

1. Right now api.s0.p.hmny.io (s1, s2, s3, etc and for other networks) don't support hmy_getTransactionsHistory, but explorer nodes like e0.t.hmny.io and for others shards/networks
support that and support all other hmy rpc methods except https://github.com/harmony-one/harmony/blob/master/internal/hmyapi/private_account.go which involves transaction signing with accounts password. But private account method shouldn't be allowed to use for public APIs anyway - you can't send your pass to an arbitrary node or even our explorer nodes.

2. Some external partners like TrustWallet want us to switch from 0x format in return values for all RPC structs to decimals if possible. 

**Which component?**

- [ ] rpc
- [ ] block-explorer
- [ ] staking-explorer
- [ ] go-sdk
- [ ] test automation
- [ ] harmony-sdk
- [ ] deployment

# Proposed Solution

1. Solution for that is to just make explorer nodes as public endpoints but make load balancing and protection against attacks and wrap direct endpoints with api.s0.p.hmny.io endpoint. 
  a) Update docs, API endpoints for apps and notify partners
  b) @AndyBoWu @john-harmony can help and setup api endpoints to block explorer nodes securely
  c) @Daniel-VDM or P-OPS can test if all APIs after that (except private_account.go APIs) are available and return correct results
  d) Update wiki and guides and notify partners (biz or any engineer can do that).

2. It will require adjusting RPC APIs and other apps for that like staking explorer, block explorer, go-sdk, harmony-sdk, harmony docs and of course make testing for that and for and notifying partners for their apps.
   a) It should involve me on RPC
   b) @potvik for block/staking explorer frontends
   c) @mikedoan and me for block explorer, staking explorer
   d) @gupadhyaya for harmony-sdk
   e) @fxfactorial or @Daniel-VDM for go-sdk
   f) Test that to return correct results for all changed methods @Daniel-VDM and P-OPS
   g) update wiki/guides when its done, notify partners (anyone, biz or P-OPS or @Daniel-VDM)

We can also contact and ask P-OPS team for proper testing. 

# Timeline

We will starting working on it next month after vacation. We can make adjustments and discussions from now. Delivery shouldn't take long from one week to two weeks, but we should consider pessimistic scenarios if bugs/issues are found or engineers who own components aren't available to make upgrades because of others tasks

[...]

PR #2048, opened 2020-01-07
RPC formatting upgrade

flicker-harmony:
## Issue

First part of rpc formatting upgrade in https://github.com/harmony-one/harmony/issues/2033. 

## Test
 
Tested locally. But needs to be tested by others, on testnet, by p-ops team possibly with @Daniel-VDM help.
Screenshot 2020-01-07 at 23 45 24
Screenshot 2020-01-07 at 23 44 56

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    NO

    NO

    Will need to adjust rpc format in external apps like explorers, sdk, etc.

   NO

## TODO

1. Testing: on testnet, by p-ops team, by partners, with @Daniel-VDM  help. Test all rpc methods one by one.
2. SDK, explorers backends adjustments, notify partners and node operators to adjust accordingly, update wiki (just replace hex to decimal everywhere and examples


LeoHChen:
As requested by TrustWallet for backward compatibility and based on my understanding, we may have to use apiv2 to support the new format of decimal values.  Please check with Victor from trustwallet team.  I am afraid this change will break their wallet.  Also, how about our own explorer?

flicker-harmony:
@LeoHChen it will break lots of apps and I noted it in issue and this PR, there is no other way except making it as v2 apis.

LeoHChen:
then we need to make it an api v2.

flicker-harmony:
We will make separate folders v1 v2 and apis with prefixes to support old apis.

internal/hmyapi/blockchain.go
```diff
@@ -300,19 +299,19 @@ func (s *PublicBlockChainAPI) GetLeader(ctx context.Context) string {
 }
 
 // GetValidatorSelfDelegation returns validator stake.
-func (s *PublicBlockChainAPI) GetValidatorSelfDelegation(ctx context.Context, address string) hexutil.Uint64 {
-	return hexutil.Uint64(s.b.GetValidatorSelfDelegation(internal_common.ParseAddr(address)).Uint64())
+func (s *PublicBlockChainAPI) GetValidatorSelfDelegation(ctx context.Context, address string) uint64 {
```
rlan35: For all token related value, we should use big.Int. uint64 is not enough for represent token more than 18 ONE. which is 18000000000000000000

neeboo:
Thats breaking changes.

flicker-harmony:
@neeboo it supports new and old one now. Developers can adjust whenever they see fit and we can make smooth updates for sdk, explorers, apps. Of course at some point we will stop updating v1 except issues.

PR #2051, opened 2020-01-08
[security] Audit fixes

fxfactorial:
Addressed report issues.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -310,7 +310,7 @@ func (node *Node) AddPendingStakingTransaction(
 	if node.Consensus.IsLeader() &&
 		node.NodeConfig.ShardID == shard.BeaconChainShardID {
 		const txPoolLimit = 1000
-		if s := len(node.pendingStakingTransactions); s > txPoolLimit {
+		if s := len(node.pendingStakingTransactions); s >= txPoolLimit {
 			utils.Logger().Info().
 				Int("tx-pool-size", s).
 				Int("tx-pool-limit", txPoolLimit).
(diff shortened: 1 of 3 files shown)


node/node.go
```diff
@@ -337,20 +342,55 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
 	blockNum := receipts.Header.Number().Uint64()
 	shardID := receipts.Header.ShardID()
+
+	// Sanity checks
+
+	// cross-shard receipt should not be coming from our shard
+	if s := node.Consensus.ShardID; s == shardID {
+		utils.Logger().Info().
+			Uint32("my-shard", s).
+			Msg("Already Got Same Receipt message")
```
rlan35: This message seems wrong.

node/node.go
```diff
@@ -337,20 +345,56 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
+		utils.Logger().Info().
+			Int("totalPendingReceipts", len(node.pendingCXReceipts)).
+			Msg("CXReceiptsProof contains empty field")
 		return
 	}
 
 	blockNum := receipts.Header.Number().Uint64()
 	shardID := receipts.Header.ShardID()
+
+	// Sanity checks
```
rlan35: I think what we need is this:

		if err := node.Blockchain().Validator().ValidateCXReceiptsProof(cxp); err != nil {
			if strings.Contains(err.Error(), rawdb.MsgNoShardStateFromDB) {
				pendingReceiptsList = append(pendingReceiptsList, cxp)
			} else {
				utils.Logger().Error().Err(err).Msg("[proposeReceiptsProof] Invalid CXReceiptsProof")
			}
			continue
		}

PR #2056, opened 2020-01-09
Enable libp2p logging option

LeoHChen:


cmd/harmony/main.go
```diff
@@ -504,6 +509,19 @@ func main() {
...
 		memprofiling.GetMemProfiling().Start()
 	}
+
+	if *logP2P {
+		f, err := os.OpenFile(path.Join(*logFolder, "libp2p.log"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Failed to open libp2p.log. %v\n", err)
+		} else {
+			defer f.Close()
+			backend1 := gologging.NewLogBackend(f, "", 0)
```
denniswon: why is there 1 at the end at var name?

go.mod
```diff
@@ -62,7 +62,6 @@ require (
 	golang.org/x/lint v0.0.0-20190409202823-959b441ac422
 	golang.org/x/sys v0.0.0-20190801041406-cbf593c0f2f3 // indirect
 	golang.org/x/tools v0.0.0-20190924052046-3ac2a5bbd98a
-	google.golang.org/appengine v1.4.0 // indirect
```
denniswon: ?

LeoHChen: just one misc commit to remove the unused lib.

cmd/harmony/main.go
LeoHChen: you can set multiple backend though for gologging.

PR #2070, opened 2020-01-11
 merge Txpool into s3

rlan35:
Cherry pick txpool logic from master to s3

Tested locally:
====== WALLET BALANCES ======
    Balance in Shard 0:  183.38074238, nonce: 10
    Balance in Shard 1:  184.952170952, nonce: 10
====== RESULTS ======
PASS

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1136,6 +1136,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 	n, events, logs, err := bc.insertChain(chain)
 	bc.PostChainEvents(events, logs)
 	if err == nil {
+		// TODO: incorporate these into insertChain
 		for idx, block := range chain {
 			header := block.Header()
 			header.Logger(utils.Logger()).Info().
@@ -1176,6 +1177,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 		}
 	}
 
+	// This should be done after everything about adding a block is done.
 	bc.PostChainEvents(events, logs)
 	return n, err
 }
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1136,7 +1136,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 	n, events, logs, err := bc.insertChain(chain)
 	bc.PostChainEvents(events, logs)
 	if err == nil {
-		// TODO: incorporate these into insertChain
+		// TODO: incorporate these into insertChain (already done in master)
 		for idx, block := range chain {
 			header := block.Header()
 			header.Logger(utils.Logger()).Info().
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1176,6 +1176,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 		}
 	}
 
+	bc.PostChainEvents(events, logs)
 	return n, err
 }
 
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -292,7 +292,7 @@ func (pool *TxPool) loop() {
 	defer journal.Stop()
 
 	// Track the previous head headers for transaction reorgs
-	//head := pool.chain.CurrentBlock()
+	head := pool.chain.CurrentBlock()
 
 	// Keep waiting for and reacting to the various events
 	for {
@@ -301,12 +301,25 @@ func (pool *TxPool) loop() {
 		case ev := <-pool.chainHeadCh:
 			if ev.Block != nil {
 				pool.mu.Lock()
-				//if pool.chainconfig.IsHomestead(ev.Block.Number()) {
-				//	pool.homestead = true
-				//}
-				//pool.reset(head.Header(), ev.Block.Header())
-				//head = ev.Block
+				if pool.chainconfig.IsS3(ev.Block.Epoch()) {
+					pool.homestead = true
+				}
+				pool.reset(head.Header(), ev.Block.Header())
+				head = ev.Block
 
+				// DEBUG-
+				pending, queued := pool.stats()
+				stales := pool.priced.stales
+
+				if pending != prevPending || queued != prevQueued || stales != prevStales {
+					utils.Logger().Debug().
+						Int("executable", pending).
+						Int("queued", queued).
+						Int("stales", stales).
+						Msg("Transaction pool status report")
+					prevPending, prevQueued, prevStales = pending, queued, stales
+				}
+				// -
 				pool.mu.Unlock()
 			}
 		// Be unsubscribed due to system stopped
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -306,20 +306,6 @@ func (pool *TxPool) loop() {
 				}
 				pool.reset(head.Header(), ev.Block.Header())
 				head = ev.Block
-
-				// DEBUG-
-				pending, queued := pool.stats()
-				stales := pool.priced.stales
-
-				if pending != prevPending || queued != prevQueued || stales != prevStales {
-					utils.Logger().Debug().
-						Int("executable", pending).
-						Int("queued", queued).
-						Int("stales", stales).
-						Msg("Transaction pool status report")
-					prevPending, prevQueued, prevStales = pending, queued, stales
-				}
-				// -
 				pool.mu.Unlock()
 			}
 		// Be unsubscribed due to system stopped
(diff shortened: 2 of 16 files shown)


node/worker/worker.go
```diff
@@ -420,8 +384,8 @@ func New(config *params.ChainConfig, chain *core.BlockChain, engine consensus_en
 		chain:   chain,
 		engine:  engine,
 	}
-	worker.gasFloor = 500000000000000000
-	worker.gasCeil = 1000000000000000000
+	worker.gasFloor = 80000000
```
LeoHChen: any comments on why change this number?

node/worker/worker.go
```diff
@@ -420,8 +384,8 @@ func New(config *params.ChainConfig, chain *core.BlockChain, engine consensus_en
...
+	worker.gasCeil = 120000000
```
LeoHChen: any reason on using this number?

core/blockchain.go
```diff
@@ -1136,6 +1136,7 @@ func (bc *BlockChain) InsertChain(chain types.Blocks) (int, error) {
 	n, events, logs, err := bc.insertChain(chain)
 	bc.PostChainEvents(events, logs)
 	if err == nil {
+		// TODO: incorporate these into insertChain
```
LeoHChen: any impact? can you open a ticket on this?

core/tx_pool.go
```diff
@@ -301,12 +301,11 @@ func (pool *TxPool) loop() {
 		case ev := <-pool.chainHeadCh:
 			if ev.Block != nil {
 				pool.mu.Lock()
-				//if pool.chainconfig.IsHomestead(ev.Block.Number()) {
-				//	pool.homestead = true
-				//}
-				//pool.reset(head.Header(), ev.Block.Header())
-				//head = ev.Block
-
+				if pool.chainconfig.IsS3(ev.Block.Epoch()) {
```
LeoHChen: which signer shall we use?

go.mod
```diff
@@ -62,6 +62,7 @@ require (
 	golang.org/x/lint v0.0.0-20190409202823-959b441ac422
 	golang.org/x/sys v0.0.0-20190801041406-cbf593c0f2f3 // indirect
 	golang.org/x/tools v0.0.0-20190924052046-3ac2a5bbd98a
+	google.golang.org/appengine v1.4.0 // indirect
```
LeoHChen: this is not needed.

node/node.go
```diff
@@ -286,23 +284,19 @@ func (node *Node) tryBroadcast(tx *types.Transaction) {
...
-	utils.Logger().Info().Int("length of newTxs", len(newTxs)).Int("totalPending", len(node.pendingTransactions)).Msg("Got more transactions")
+
+	pendingCount, queueCount := node.TxPool.Stats()
+	utils.Logger().Info().Int("length of newTxs", len(newTxs)).Int("totalPending", pendingCount).Int("totalQueued", queueCount).Msg("Got more transactions")
 }
 
 // Add new staking transactions to the pending staking transaction list.
 func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTransactions) {
-	txPoolLimit := core.ShardingSchedule.MaxTxPoolSizeLimit()
+	txPoolLimit := 1000 // TODO: incorporate staking txn into TxPool
```
LeoHChen: use a constant please.

node/node.go
```diff
@@ -313,7 +307,7 @@ func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTra
 		}
 	}
 	node.pendingStakingTxMutex.Unlock()
-	utils.Logger().Info().Int("length of newStakingTxs", len(newStakingTxs)).Int("totalPending", len(node.pendingTransactions)).Msg("Got more staking transactions")
+	utils.Logger().Info().Int("length of newStakingTxs", len(newStakingTxs)).Int("totalPending", len(node.pendingStakingTransactions)).Msg("Got more staking transactions")
```
LeoHChen: do we have staking transaction?

node/node_handler_test.go
```diff
@@ -34,11 +36,10 @@ func TestAddNewBlock(t *testing.T) {
 	nodeconfig.SetNetworkType(nodeconfig.Devnet)
 	node := New(host, consensus, testDBFactory, false)
 
-	selectedTxs, selectedStakingTxs := node.getTransactionsForNewBlock(common.Address{})
-	node.Worker.CommitTransactions(
-		selectedTxs, selectedStakingTxs, common.Address{},
-		func([]types.RPCTransactionError) {},
-	)
+	txs := make(map[common.Address]types.Transactions)
+	stks := types2.StakingTransactions{}
```
LeoHChen: we don't have staking tx in s3, right?

node/node_handler_test.go
```diff
@@ -69,12 +70,10 @@ func TestVerifyNewBlock(t *testing.T) {
...
 	node := New(host, consensus, testDBFactory, false)
 
-	selectedTxs, selectedStakingTxs := node.getTransactionsForNewBlock(common.Address{})
-	node.Worker.CommitTransactions(
-		selectedTxs, selectedStakingTxs, common.Address{},
-		func([]types.RPCTransactionError) {},
-	)
-
+	txs := make(map[common.Address]types.Transactions)
+	stks := types2.StakingTransactions{}
```
LeoHChen: we don't have staking tx in s3, right?

node/node_newblock.go
```diff
@@ -81,18 +83,28 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
-	if err := node.Worker.CommitTransactions(
-		selectedTxs, selectedStakingTxs, coinbase,
+	pending, err := node.TxPool.Pending()
+	if err != nil {
+		utils.Logger().Err(err).Msg("Failed to fetch pending transactions")
+		return nil, err
+	}
+
+	// TODO: integrate staking transaction into tx pool
+	pendingStakingTransactions := types2.StakingTransactions{}
```
LeoHChen: do we need staking tx here?

node/worker/worker.go
```diff
@@ -50,99 +53,98 @@ type Worker struct {
...
-
-	return sender, shardingconfig.TxSelect
-}
-
-// SelectTransactionsForNewBlock selects transactions for new block.
-func (w *Worker) SelectTransactionsForNewBlock(newBlockNum uint64, txs types.Transactions, recentTxsStats types.RecentTxsStats, txsThrottleConfig *shardingconfig.TxsThrottleConfig, coinbase common.Address) (types.Transactions, types.Transactions, types.Transactions) {
+// CommitTransactions commits transactions for new block.
+func (w *Worker) CommitTransactions(
+	pendingNormal map[common.Address]types.Transactions,
+	pendingStaking staking.StakingTransactions, coinbase common.Address,
```
LeoHChen: pendingStaking is not needed, right?

LeoHChen:
please also provide a test log on this change.

node/worker/worker.go
rlan35: Previous setting doesn't make sense when we first launch mainnet. It was too high. The 80m to 120m limit is 10x that of Ethereum. So we have 10x more capacity than ethereum. And it translate to roughly max 5700 normal txns in a block

core/blockchain.go
rlan35: the TODO is already done in master. But the code at cherrypick doesn't

core/tx_pool.go
rlan35: S3 is the condition we used to include homestead when we did the upgrade before.

go.mod
rlan35: Reverted.

node/node.go
rlan35: Staking is not supported in s3. But somehow the initial staking txns skeleton code is in s3. It's no-op so shouldn't be a problem.

LeoHChen:
Thanks for all the clarification.  Please add some of the key clarification into the code and I am okay to approve it. Also, the gas fee changes have to be added to master branch as well.  Or maybe it is already changed in master branch. If not, we need to open an issue to keep track of it. Thanks.

rlan35:
> Thanks for all the clarification. Please add some of the key clarification into the code and I am okay to approve it. Also, the gas fee changes have to be added to master branch as well. Or maybe it is already changed in master branch. If not, we need to open an issue to keep track of it. Thanks.

All the change in this PR are cherry-picked from master, so they are already done in master. Added comment. Also removed the staking txns arg in CommitTransactions method to be safe.

rlan35:
> LGTM in general. I assume the txpool functions were tested in master/s3 branch.

Yea, basic txpool logic was tested in master when it was implemented. It's good we can test it more extensively for token swap now.

PR #2076, opened 2020-01-13
Add debug info for overrun issue & fix OOB bugs from the audit

rlan35:
## Issue

Fix all the OOB issue reported from security audit: https://github.com/harmony-one/security-audit/issues

### DIFF
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -25,6 +25,8 @@ import (
 	libp2p_peer "github.com/libp2p/go-libp2p-core/peer"
 )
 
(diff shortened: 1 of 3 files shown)


node/node_handler.go
```diff
@@ -38,6 +38,11 @@ func (node *Node) receiveGroupMessage(
 				Msg("cannot receive from group")
 			continue
 		}
+		if len(msg) < 5 {
```
fxfactorial: Maybe const name the magic number 5

rlan35: Done.

PR #2077, opened 2020-01-14
[availability] Implement inactive toggle for validators that miss threshold of signing required; (66%) of epoch

fxfactorial:
Implementation of unavailability setting to inactive if signing threshold measure not met.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1212,7 +1212,9 @@ func (bc *BlockChain) WriteBlockWithState(
 			mask, _ := bls.NewMask(members, nil)
 			mask.SetMask(block.Header().LastCommitBitmap())
 
-			if err = bc.UpdateValidatorUptime(committee.Slots, mask); err != nil {
+			if err = bc.UpdateValidatorUptime(
+				committee.Slots, mask, header.Epoch(),
+			); err != nil {
 				utils.Logger().Err(err).Msg("[Uptime] Failed updating validator uptime")
 			}
 		} else {
@@ -1267,7 +1269,9 @@ func (bc *BlockChain) WriteBlockWithState(
 					mask, _ := bls.NewMask(members, nil)
 					mask.SetMask(crossLink.Bitmap())
 
-					if err = bc.UpdateValidatorUptime(committee.Slots, mask); err != nil {
+					if err = bc.UpdateValidatorUptime(
+						committee.Slots, mask, header.Epoch(),
+					); err != nil {
 						utils.Logger().Err(err).Msg("[Uptime] Failed updating validator uptime")
 					}
 				} else {
@@ -2593,7 +2597,9 @@ func (bc *BlockChain) ReadValidatorStats(addr common.Address) (*staking.Validato
 }
 
 // UpdateValidatorUptime writes the stats for the committee and bumps up availability counts
-func (bc *BlockChain) UpdateValidatorUptime(slots shard.SlotList, mask *bls.Mask) error {
+func (bc *BlockChain) UpdateValidatorUptime(
+	slots shard.SlotList, mask *bls.Mask, epoch *big.Int,
+) error {
 	blsToAddress := make(map[shard.BlsPublicKey]common.Address)
 	for _, slot := range slots {
 		blsToAddress[slot.BlsPublicKey] = slot.EcdsaAddress
@@ -2621,6 +2627,7 @@ func (bc *BlockChain) UpdateValidatorUptime(slots shard.SlotList, mask *bls.Mask
 				return err
 			}
 
+			wrapper.Snapshot.Epoch = epoch
 			wrapper.Snapshot.NumBlocksToSign.Add(wrapper.Snapshot.NumBlocksToSign, one)
 
 			enabled, err := mask.IndexEnabled(i)
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -51,6 +51,7 @@ import (
 	"github.com/harmony-one/harmony/numeric"
 	"github.com/harmony-one/harmony/shard"
 	"github.com/harmony-one/harmony/shard/committee"
+	"github.com/harmony-one/harmony/staking/availability"
 	staking "github.com/harmony-one/harmony/staking/types"
 	lru "github.com/hashicorp/golang-lru"
 )
@@ -1241,6 +1242,8 @@ func (bc *BlockChain) WriteBlockWithState(
 			}
 
 			members := []*bls2.PublicKey{}
+			addrs := []common.Address{}
+
 			for _, slot := range committee.Slots {
 				pubKey := &bls2.PublicKey{}
 				err := pubKey.Deserialize(slot.BlsPublicKey[:])
@@ -1248,6 +1251,7 @@ func (bc *BlockChain) WriteBlockWithState(
 					utils.Logger().Err(err).Msg("[Uptime] Failed to deserialize bls public key")
 				}
 				members = append(members, pubKey)
+				addrs = append(addrs, slot.EcdsaAddress)
 			}
 
 			mask, _ := bls.NewMask(members, nil)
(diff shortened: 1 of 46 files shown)


staking/types/messages.go
```diff
@@ -74,6 +74,7 @@ type EditValidator struct {
 	SlotKeyToRemove    *shard.BlsPublicKey `json:"slot_key_to_remove" yaml:"slot_key_to_remove" rlp:"nil"`
 	SlotKeyToAdd       *shard.BlsPublicKey `json:"slot_key_to_add" yaml:"slot_key_to_add" rlp:"nil"`
 	SlotKeyToAddSig    shard.BlsSignature  `json:"slot_key_to_add_sig" yaml:"slot_key_to_add_sig" rlp:"nil"`
+	IsActive           *bool               `json:"is_active" rlp:"nil"`
```
rlan35: let's just use Active instead. Same as Validator object.

staking/availability/measure.go
```diff
@@ -0,0 +1,43 @@
...
+func SetInactiveUnavailableValidators(
+	addrs []common.Address, batch ethdb.Batch, bc consensus_engine.ChainReader,
+) error {
+	for i := range addrs {
+		stats, err := bc.ReadValidatorStats(addrs[i])
+		if err != nil {
+			return err
+		}
+		if r := new(big.Int).Div(
+			stats.NumBlocksSigned, stats.NumBlocksToSign,
```
rlan35: this condition should be counting only the stats for last epoch. not for the whole history. So you will need stats by epoch.

core/blockchain.go
```diff
@@ -1240,13 +1241,18 @@ func (bc *BlockChain) WriteBlockWithState(
...
+
 			for _, slot := range committee.Slots {
 				pubKey := &bls2.PublicKey{}
 				err := pubKey.Deserialize(slot.BlsPublicKey[:])
 				if err != nil {
 					utils.Logger().Err(err).Msg("[Uptime] Failed to deserialize bls public key")
 				}
 				members = append(members, pubKey)
+				if isNewEpoch {
+					addrs = append(addrs, slot.EcdsaAddress)
```
rlan35: You can directly call ReadActiveValidatorList to get all the validators of all shards at this epoch. And I think you can do the SetInactiveUnavailableValidators once for all validators right? Also don't need to embed this logic in crosslinks, just do it separately.

staking/types/validator.go
```diff
@@ -74,6 +74,16 @@ type ValidatorStats struct {
 	BLSKeyPerShard []KeysPerShard
 }
 
+// ValidatorSnapshot ..
+type ValidatorSnapshot struct {
+	Validator *ValidatorWrapper
+	Epoch     *big.Int
+	// The number of blocks the validator should've signed when in active mode (selected in committee)
+	NumBlocksToSign *big.Int `rlp:"nil"`
```
rlan35: Let's just put NumBlocksToSign NumBlocksSigned into validatorWrapper, which is in state. It will be needed for fast state sync.

rlan35: It's better to have a permanent record of a validator's signing records.

core/blockchain.go
```diff
@@ -2567,14 +2586,21 @@ func (bc *BlockChain) writeValidatorSnapshots(addrs []common.Address) error {
...
 	// Batch write the current data as snapshot
 	batch := bc.db.NewBatch()
 	for i := range validators {
-		err := rawdb.WriteValidatorSnapshot(batch, validators[i])
+		stats, err := rawdb.ReadValidatorStats(bc.db, validators[i].Address)
 		if err != nil {
 			return err
 		}
+		validators[i].Snapshot.Epoch = epoch
+		validators[i].Snapshot.NumBlocksSigned = stats.NumBlocksSigned
```
rlan35: Remove the NumBlocksSigned and ToSign in ValidatorStats and only keep those in Wrapper as source of truth. And update those number in Wrapper every block. Then here the snapshot process will be simply copy and store the wrapper as is.

core/blockchain.go
```diff
@@ -2614,19 +2638,26 @@ func (bc *BlockChain) UpdateValidatorUptime(slots shard.SlotList, mask *bls.Mask
...
+			wrapper.Snapshot.NumBlocksToSign.Add(wrapper.Snapshot.NumBlocksToSign, one)
 
 			enabled, err := mask.IndexEnabled(i)
 			if err != nil {
 				return err
 			}
 
 			if enabled {
-				stats.NumBlocksSigned.Add(stats.NumBlocksSigned, big.NewInt(1))
+				wrapper.Snapshot.NumBlocksSigned.Add(wrapper.Snapshot.NumBlocksSigned, one)
```
rlan35: As we talked about last time. The update of wrapper needs to be within the block proposal because it's change the state. Here it's just changing the copy of wrapper. And here it's not the right place to change state.

fxfactorial: Gotcha, I thought about this spot would still count because the increment was happening here and state isn’t written yet

rlan35: Everything in WriteBlockWithState happens after the state is modified.

rlan35:
The new code looks good.

core/blockchain.go
```diff
@@ -2820,8 +2703,12 @@ func (bc *BlockChain) UpdateStakingMetaData(tx *staking.StakingTransaction, root
 		if err != nil {
 			return err
 		}
-		err = rawdb.WriteValidatorSnapshot(bc.db, validator)
-		if err != nil {
+		zero := big.NewInt(0)
+		validator.Snapshot.Epoch = epoch
```
rlan35: don't need to modify this. When the wrapper is created. you already populated these fields with 0. Should just copy and record the wrapper as is.

internal/chain/engine.go
```diff
@@ -324,6 +329,44 @@ func (e *engineImpl) Finalize(
 			}
 		}
 	}
+
+	if inStakingEra {
+		if isBeaconChain {
```
rlan35: should also add isNewEpoch, because we should only do the tallying at the end of epoch. And with this, we should put the tallying before the shard state proposal, so kicked out nodes won't be part of shard anymore.

internal/chain/engine.go
```diff
@@ -324,6 +329,44 @@ func (e *engineImpl) Finalize(
...
+					}
+				}
+			}
+
+			if err := availability.IncrementValidatorSigningCounts(
+				chain, header, header.ShardID(), state, processed,
+			); err != nil {
+				return nil, nil, err
+			}
+			if err := availability.SetInactiveUnavailableValidators(
```
rlan35: And with this, we should put the tallying before the shard state proposal, so kicked out nodes won't be part of shard anymore. So this part should be separate from IncrementValidatorSigningCounts.

rlan35:
generally looks good.

core/blockchain.go
fxfactorial: gotcha, so removed the two counters but i think i should still keep `validator.Snapshot.Epoch = epoch`

rlan35: why add an epoch? Then it's not really a snapshot anymore.

core/state_transition.go
```diff
@@ -389,22 +389,28 @@ func (st *StateTransition) applyCreateValidatorTx(createValidator *staking.Creat
...
 
-	delegations := []staking.Delegation{
+	zero := big.NewInt(0)
+	wrapper := staking.ValidatorWrapper{}
+	wrapper.Validator = *v
+	wrapper.Delegations = []staking.Delegation{
 		staking.NewDelegation(v.Address, createValidator.Amount),
 	}
-	wrapper := staking.ValidatorWrapper{*v, delegations}
+	wrapper.Snapshot.Epoch = zero
```
rlan35: if you like to add epoch, better add here. But I don't see any use of it yet.

fxfactorial: hmm, i thought i remember hearing you say that we might want to hold onto it for state sync? and i was using it for the temp sanity check that later would remove once know things stable, correct.

PR #2078, opened 2020-01-14
[project] Update Dockerfile to go 1.13.6

fxfactorial:
Bump docker


mattlockyer:
To anyone encountering this error with previous versions of Go:

```
build command-line-arguments: cannot load crypto/ed25519: cannot find module providing package crypto/ed25519
```

The solution is to up your Go version to go 1.13.6.

PR #2098, opened 2020-01-15
[rpc] Backport fix to rpc on named return value

fxfactorial:


fxfactorial:
@LeoHChen original error is that on empty, the named return value was compiled as a null instead of [] when it hit the RPC layer, 

example before and after:

```
 ./hmy failures plain 
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": null
}(Wed Jan-1 2:41:23pm)-(CPU 3.8%:0:Net 1108)-(edgar:~/go/src/github.com/harmony-one/go-sdk)-(20M:12)
> ./hmy failures plain 
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": []
}
```

PR #2100, opened 2020-01-16
fix cross-shard receipt merkle root bug

rlan35:
Tested locally with 3 shards.


node/node_cross_shard.go
```diff
@@ -47,7 +47,7 @@ func (node *Node) BroadcastCXReceipts(newBlock *types.Block, lastCommits []byte)
 		if i == int(myShardID) {
 			continue
 		}
-		go node.BroadcastCXReceiptsWithShardID(newBlock, commitSig, commitBitmap, uint32(i))
+		node.BroadcastCXReceiptsWithShardID(newBlock, commitSig, commitBitmap, uint32(i))
```
fxfactorial: why remove go

node/node_handler.go
```diff
@@ -362,7 +362,7 @@ func (node *Node) PostConsensusProcessing(newBlock *types.Block, commitSigAndBit
 		// 15% of the validator also need to do broadcasting
 		rand.Seed(time.Now().UTC().UnixNano())
 		rnd := rand.Intn(100)
-		if rnd < 15 {
+		if rnd < 1 {
```
fxfactorial: leftover debug?

LeoHChen:
Is this change compatible with previous code?  Will we have problem replay the blockchain because of this change?  Do we need to use an epoch to enable this change for cx?

core/types/cx_receipt.go
```diff
@@ -82,6 +86,36 @@ func (cs CXReceipts) MaxToShardID() uint32 {
...
+	}
+	return crypto.Keccak256Hash(byteBuffer.Bytes())
+}
+
+// GetToShardReceipts filters the cross shard receipts with given destination shardID
+func (cs CXReceipts) GetToShardReceipts(shardID uint32) CXReceipts {
+	cxs := CXReceipts{}
+	for i := range cs {
+		cx := cs[i]
+		if cx.ToShardID == shardID {
```
LeoHChen: To be more protective, better check if cx is nil ?

core/types/cx_receipt.go
```diff
@@ -82,6 +86,36 @@ func (cs CXReceipts) MaxToShardID() uint32 {
...
+	byteBuffer := bytes.NewBuffer([]byte{})
+	for i := 0; i <= int(cs.MaxToShardID()); i++ {
+		shardReceipts := cs.GetToShardReceipts(uint32(i))
+		if len(shardReceipts) != 0 {
+			sKey := make([]byte, 4)
+			binary.BigEndian.PutUint32(sKey, uint32(i))
+			byteBuffer.Write(sKey)
+			hash := DeriveSha(shardReceipts)
+			byteBuffer.Write(hash[:])
+		}
```
LeoHChen: do you need to print a log here if len(shardReceipt) == 0?

core/types/cx_receipt.go
```diff
@@ -82,6 +86,36 @@ func (cs CXReceipts) MaxToShardID() uint32 {
...
+	}
+	return crypto.Keccak256Hash(byteBuffer.Bytes())
```
LeoHChen: will it cause issue if byteBuffer.Bytes() is empty?

core/blockchain.go
```diff
@@ -1081,8 +1081,8 @@ func (bc *BlockChain) WriteBlockWithState(block *types.Block, receipts []*types.
 			if i == int(block.ShardID()) {
 				continue
 			}
-			shardReceipts := GetToShardReceipts(cxReceipts, uint32(i))
-			err := rawdb.WriteCXReceipts(batch, uint32(i), block.NumberU64(), block.Hash(), shardReceipts, false)
+			shardReceipts := types.CXReceipts(cxReceipts).GetToShardReceipts(uint32(i))
+			err := rawdb.WriteCXReceipts(batch, uint32(i), block.NumberU64(), block.Hash(), shardReceipts)
 			if err != nil {
 				utils.Logger().Debug().Err(err).Interface("shardReceipts", shardReceipts).Int("toShardID", i).Msg("WriteCXReceipts cannot write into database")
```
LeoHChen: this should be Logger().Error(), instead of Logger.Debug().

node/node_cross_shard.go
```diff
@@ -56,21 +56,24 @@ func (node *Node) BroadcastCXReceiptsWithShardID(block *types.Block, commitSig [
 	myShardID := node.Consensus.ShardID
 	utils.Logger().Info().Uint32("toShardID", toShardID).Uint32("myShardID", myShardID).Uint64("blockNum", block.NumberU64()).Msg("[BroadcastCXReceiptsWithShardID]")
 
-	cxReceipts, err := node.Blockchain().ReadCXReceipts(toShardID, block.NumberU64(), block.Hash(), false)
+	cxReceipts, err := node.Blockchain().ReadCXReceipts(toShardID, block.NumberU64(), block.Hash())
 	if err != nil || len(cxReceipts) == 0 {
-		utils.Logger().Info().Err(err).Uint32("ToShardID", toShardID).Int("numCXReceipts", len(cxReceipts)).Msg("[BroadcastCXReceiptsWithShardID] No ReadCXReceipts found")
+		utils.Logger().Info().Err(err).Uint32("ToShardID", toShardID).Int("numCXReceipts", len(cxReceipts)).Msg("[CXMerkleProof] No receipts found for the destination shard")
```
LeoHChen: use Logger().Error() to replace Logger.Info()

core/blockchain.go
```diff
@@ -2098,49 +2098,26 @@ func (bc *BlockChain) GetVMConfig() *vm.Config {
...
-	}
 
 	epoch := block.Header().Epoch()
 	shardingConfig := ShardingSchedule.InstanceForEpoch(epoch)
 	shardNum := int(shardingConfig.NumShards())
 
 	for i := 0; i < shardNum; i++ {
-		receipts, err := bc.ReadCXReceipts(uint32(i), block.NumberU64(), block.Hash(), false)
+		receipts, err := bc.ReadCXReceipts(uint32(i), block.NumberU64(), block.Hash())
 		if err != nil || len(receipts) == 0 {
```
LeoHChen: Do we need to log an error here?

node/node_handler.go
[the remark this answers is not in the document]
rlan35: 1% of validator which is 2-3 nodes. Leader alway broadcast anyway. The issue is the messages are broadcasted duplicately which may clog the network

core/types/cx_receipt.go
rlan35: no need. it's expected.

core/types/cx_receipt.go
rlan35: no

node/node_cross_shard.go
rlan35: because there is another go routine inside BroadcastCXReceiptsWithShardID for the p2p sending.

node/node_handler.go
rlan35: intentional. 15% is too much duplicated msg. (leader is always broadcasting anyway)

rlan35:
> how did you test it? can you post some test log?

Added

PR #2117, opened 2020-01-18
Accept both old and new merkle root logic.

rlan35:
Also filed an issue for later forking logic. https://github.com/harmony-one/harmony/issues/2118

====== WALLET BALANCES ======
    Balance in Shard 0:  184.428361428, nonce: 10
    Balance in Shard 1:  186.809313809, nonce: 10
====== RESULTS ======
PASS


core/block_validator.go
```diff
@@ -215,6 +218,9 @@ func (v *BlockValidator) ValidateCXReceiptsProof(cxp *types.CXReceiptsProof) err
 
 	// (2) verify the outgoingCXReceiptsHash match
 	outgoingHashFromSourceShard := crypto.Keccak256Hash(byteBuffer.Bytes())
+	if byteBuffer.Len() == 0 {
+		outgoingHashFromSourceShard = types.EmptyRootHash
+	}
```
rlan35: This should be an no-op as if there is no receipts, then the dest shard shouldn't receive it. But just to make sure the logic are the same with the source side. add it here too.

PR #2130, opened 2020-01-22
[project] Remove unused, deprecated docker

fxfactorial:
Not needed anymore, use root of project' `Dockerfile` 

`docker run -it  harmonyone/main:stable /bin/bash`


fxfactorial:
Talked with @LeoHChen offline, seems like we want to keep this for now.

PR #2152, opened 2020-01-24
[slash][consensus] Notice double sign & broadcast, factor out tech debt of consensus 

fxfactorial:
This PR implements capturing a double sign event in consensus and broadcasting it over p2p to beaconchain. Beaconchain correspondingly receives the protobuf message and adds to its block proposal state. Verify & Applying to state not implemented but stubbed out for next work, computed signature and bitmap also not implemented, for next PR. This PR also includes the off-chain caching of pending slashing candidates. 

PR also includes substantial refactoring of consensus, addressing many leftover TODOs and removing unnecessary marshaling/unmarshaling, other small optmizations. Intent of BFT is clearer in code as well.


consensus/consensus_v2.go
```diff
@@ -762,8 +765,43 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 		Int64("numReceivedSoFar", consensus.Decider.SignersCount(quorum.Commit)).
 		Logger()
 	logger.Info().Msg("[OnCommit] Received new commit message")
+
+	// proceed only when the message is not received before
+	if signed := consensus.Decider.ReadBallot(
```
rlan35: Moving the check lower here will incur lots of unnecessary checks above.

consensus/consensus_v2.go
```diff
@@ -762,8 +765,43 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
...
+		// need to verify the block hash, and if block hash is different, then
+		// that is a clear case of double signing
+		if h1, h2 :=
+			signed.BlockHash,
+			recvMsg.BlockHash; signed.BlockHeight == recvMsg.BlockNum &&
+			bytes.Compare(h1.Bytes(), h2.Bytes()) != 0 {
+			go func() {
+				// TODO(Edgar) Compute the aggregate vote, bitmap
+				consensus.SlashChan <- slash.Record{
+					BlockHash:   h2,
```
rlan35: Better add both block hashes and the signatures. So this record is verifiable by itself. Otherwise, it has to come with the crossLink to prove the double signing.

consensus/quorum/quorum.go
```diff
@@ -24,8 +25,8 @@ const (
 
 var phaseNames = map[Phase]string{
 	Prepare:    "Prepare",
-	Commit:     "Prepare",
-	ViewChange: "Commit",
+	Commit:     "Commit",
+	ViewChange: "viewChange",
```
rlan35: Ops. good fix.

consensus/votepower/roster.go
```diff
@@ -24,10 +25,22 @@ var (
 
 // Ballot is a vote cast by a validator
 type Ballot struct {
-	SignerPubKey      shard.BlsPublicKey `json:"bls-public-key"`
-	BlockLeader       shard.BlsPublicKey `json:"leader-when-signed"`
-	BlockHeightHeight uint64             `json:"block-height"`
-	Signature         *bls.Sign          `json:"signature"`
+	SignerPubKey shard.BlsPublicKey `json:"bls-public-key"`
+	BlockLeader  shard.BlsPublicKey `json:"leader-when-signed"`
```
rlan35: on a second thought, actually using viewID will be an easier way to distinguish signing blocks from different leaders.

node/node_handler.go
```diff
@@ -59,6 +59,29 @@ func (node *Node) receiveGroupMessage(
 	}
 }
 
+func (node *Node) processSkippedMsgType(cat proto_node.BlockMessageType, content []byte) {
```
rlan35: This func's naming and description can be better.

node/worker/worker.go
```diff
@@ -328,6 +331,21 @@ func (w *Worker) FinalizeNewBlock(
 	w.current.header.SetCoinbase(coinbase)
 	w.current.header.SetViewID(new(big.Int).SetUint64(viewID))
 
+	// Slashes
```
rlan35: We may need to slash later (not immediately) so as to wait for all the double sign on the same block to appear. Probably at the end of epoch is better.

consensus/consensus_v2.go
fxfactorial: ya, I thought about that but there is information that i need and i didn't want to spend time on more thinking of how to organize both.

consensus/consensus_v2.go
fxfactorial: gotca, will add the double signed block hash as well

consensus/votepower/roster.go
fxfactorial: but will viewID from various shards be meaningful to beaconchain?

staking/slash/double-sign.go
```diff
@@ -2,23 +2,23 @@ package slash
...
-	DoubleSignature votepower.BallotResults
-	ShardID         uint32         `json:"shard-id"`
-	Epoch           *big.Int       `json:"epoch"`
-	Beneficiary     common.Address // the reporter who will get rewarded
+	Offender shard.BlsPublicKey
+	Signed   struct {
+		Header *block.Header
+	} `json:"signed"`
+	DoubleSigned struct {
+		Header *block.Header
```
rlan35: still need the signatures.

staking/slash/report.go
```diff
@@ -0,0 +1,28 @@
...
+// DoubleSignWebHooks ..
+type DoubleSignWebHooks struct {
+	WebHooks struct {
+		OnNoticeDoubleSign     string `yaml:"notice-double-sign"`
+		OnThisNodeDoubleSigned string `yaml:"this-node-double-signed"`
+	} `yaml:"web-hooks"`
+}
+
+// NewDoubleSignWebHooksFromPath ..
+func NewDoubleSignWebHooksFromPath(yamlPath string) (*DoubleSignWebHooks, error) {
```
rlan35: Is this webhook feature tested?

rlan35:
Regarding the refactoring. Generally looks good. Please test to make sure the consensus and staking logic still works. Let’s submit this PR first. It’s too big now.

staking/slash/report.go
fxfactorial: Next PR consumes it

PR #2162, opened 2020-01-28
[rpc] Stub out estimate gas cost

fxfactorial:
@mattlockyer lets maybe work together to have this working for you 

EDIT: thank you @flicker-harmony for finishing it, only one thing about the unnecessarily duplicated function, (since wasn't a method)


internal/hmyapi/apiv2/blockchain.go
```diff
@@ -597,6 +598,69 @@ func (s *PublicBlockChainAPI) GetDelegationByDelegatorAndValidator(ctx context.C
 	return nil, nil
 }
 
+// doEstimateGas ..
+func doEstimateGas(ctx context.Context, b Backend, args CallArgs, gasCap *big.Int) (hexutil.Uint64, error) {
```
fxfactorial: since this wasn't a method, it could have been defined just once

PR #2172, opened 2020-02-01
Add a blacklist

Daniel-VDM:
## Issue

In the (unlikely) event of a successful attack/hack, implementing and rolling out a solution will take some time. A blacklist offers an easy and fast way to contain the situation while we work out a solution. 

The blacklist that we will be using will be maintained in the main repo's root for everyone to see (it will start off empty). Moreover, any transaction that was dropped because of a blacklisted account will be reported as such in our newly added `hmy_getCurrentTransactionErrorSink` RPC; which means it will be displayed in our explorer. 

#### Implementation highlights:
* The blacklist is read from a newline delimited file (by default it's `./blacklist.txt`).
* Each transaction added to the tx-pool has its `to` and `from` address checked against this blacklist. If there is a hit, the transaction is considered invalid and is dropped from the tx-pool. 
* Easy to rip out / revert in the future.

**Sample blacklist file:**
```
one1spshr72utf6rwxseaz339j09ed8p6f8ke370zj
one1zksj3evekayy90xt4psrz8h6j2v3hla4qwz4ur
one129r9pj3sk0re76f7zs3qz92rggmdgjhtwge62k
one12fuf7x9rgtdgqg7vgq0962c556m3p7afsxgvll
one1658znfwf40epvy7e46cqrmzyy54h4n0qa73nep
one17ys3gsd5gpuhx3juqe73sgtpzpvdusktuq7kfc

```

### Unit Test Coverage

Before:

```
PASS
coverage: 20.4% of statements
ok  	github.com/harmony-one/harmony/core	6.711s
```

```
PASS
coverage: 14.0% of statements
ok  	github.com/harmony-one/harmony/node	2.160s
```

After:

```
PASS
coverage: 20.5% of statements
ok  	github.com/harmony-one/harmony/core	6.813s
```

```
PASS
coverage: 14.2% of statements
ok  	github.com/harmony-one/harmony/node	2.128s
```

## Operational Checklist

    **NO**

    **NO**

    **YES**

    **They can specify which blacklist they want to enforce, by default it will be `./blacklist.txt`. Moreover, they should know that the file is a newline delimited file (reference example above).**

    **NO**


mindstyle85:
think this would be a good safety net for starters yes, some FN members are already using their own blacklist for some nodes (those that go down a lot etc.). 

Question: guessing this file will be filled by everyone by himself? If so how can we ensure the attackers wont use it too?

sophoah:
This is a good one, we would just need to be mindful on how, who decide for adding/removing addresses to that list stored on the harmony main repo. Of course, we won't be able to prevent node operator to could create their own blacklist.

So my concern is the distribution and update of all validator. It could take 1-2 days to update the network. and we want to be able to react fast if there is a hack or something.

What about having all the nodes to download that file let's say every x time from our main github repo or an s3 bucket. Node would be updated automatically without having the need for the node to restart, like it is right now via the bash script node.sh.

For now harmony still have 2/3+1 of nodes so this isn't yet a problem, we could deploy fast, but it would be an issue once we are going full external.

CryptosWolfOfAllStreets:
From what I have read regarding blacklists in general this would have to be a file filled out and constantly updated by the team themselves for it to be effective because not everyone running a node is accessing their node daily so if not upheld by a central authority it would not seem to work because new transactions are checked against the blacklist but everyone's would be different etc so I take it that its centrally operated.

Therefore is it such a good idea to leave it openly readable to non node runners too? Would this not just give potential attackers an idea of which wallets not to use / create new ones not on the list. 

I'm guessing for the really talented hackers this would not be a major issue to overcome but it is heading in the right direction of security and heads away from decentralisation - a possible way around this could be if the FN channels were locked and only actual runners could join the channels and then those channels could discuss additions to the list between themselves and update the list themselves without it being central to the team - this would need its operations working out but I think it could be done.

There is a lot to think about with this so I will probably comment again soon :)

BrunoMarshall:
Great stuff! The only question/concern from my side is that the transaction speed will possibly decrease if the blacklist get big over time, because each transaction added to the tx-pool need to check the address against this blacklist. What happen is this list get tremendous big?

rlan35:
This change is only temporarily necessary because we still maintain the leaders. After we go fully decentralized, if any bad thing happened, there should be a hard fork to introduce the blocking of addresses in the protocol and this change will be deprecated.

mindstyle85:
> This change is only temporarily necessary because we still maintain the leaders. After we go fully decentralized, if any bad thing happened, there should be a hard fork to introduce the blocking of addresses in the protocol and this change will be deprecated.

so this blacklist is something you guys will maintain and we as node runners will just download it with node.sh?

cmd/harmony/main.go
```diff
@@ -322,9 +326,15 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 		currentConsensus.DisableViewChangeForTestingOnly()
 	}
 
+	blacklist, err := setupBlacklist()
+	if err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "Blacklist error: %s\n", err.Error())
```
rlan35: If there is no blacklist file, we should just continue normally. we shouldn't block the node from running by blacklist.

Daniel-VDM:
> Question: guessing this file will be filled by everyone by himself? If so how can we ensure the attackers wont use it too?

While everyone could define their own blacklist, they can use ours as a reference. Regarding attackers, they could use this, but it would require > 1/3 nodes to agree on banning any given account. 

> What about having all the nodes to download that file let's say every x time from our main github repo or an s3 bucket. Node would be updated automatically without having the need for the node to restart, like it is right now via the bash script node.sh.

I like this idea. We could pull the latest blacklist every time the harmony binary is run. So every time `node.sh` automatically updates, we could also refresh the blacklist. One thing though, the node's binary would need to be restarted every time the list is updated. I don't like the idea of the binary itself look for a list every x second as a blacklist is not meant to be constantly updated.

> Therefore is it such a good idea to leave it openly readable to non node runners too? Would this not just give potential attackers an idea of which wallets not to use / create new ones not on the list.

We want to leave it open for maximum transparency. While attackers can create new wallets, they cannot get the funds of the backlisted wallets, so it is okay for anyone to know who is on the blacklist. 

> A possible way around this could be if the FN channels were locked and only actual runners could join the channels and then those channels could discuss additions to the list between themselves and update the list themselves without it being central to the team.

As we approach open staking, the group of node runners will be forever changing. Also, this is a consortium of people deciding who is banned, which is not as decentralized or as transparent as we would like. 

> What happen is this list get tremendous big?

The blacklist is stored in memory as a hash set/map. This has an O(1) lookup time for each address, so it will have *very* little effect on the performance of the node. Moreover, I don't expect the list to be super large, since in that case, a hard fork may be in order. 

Thanks for the questions, everyone!

core/tx_pool.go
```diff
@@ -860,7 +892,11 @@ func (pool *TxPool) AddLocals(txs []*types.Transaction) []error {
 // If the senders are not among the locally tracked ones, full pricing constraints
 // will apply.
 func (pool *TxPool) AddRemotes(txs []*types.Transaction) []error {
-	return pool.addTxs(txs, false)
+	errs := []error{}
+	for _, err := range pool.addTxs(txs, false) {
```
rlan35: why doing this?

core/tx_pool_test.go
```diff
@@ -261,6 +261,52 @@ func TestInvalidTransactions(t *testing.T) {
 	}
 }
 
+func TestBlacklistedTransactions(t *testing.T) {
```
rlan35: what's the format of the blacklist file?

rlan35: better have a test example file for it.

core/tx_pool.go
Daniel-VDM: goimports test ran on Travis require that I use the built-in `errors` package for the unit tests and does not allow me to use `github.com/pkg/errors`

core/tx_pool_test.go
Daniel-VDM: It's a newline delimited file of addresses. I'll add comments in the setup function of main

cmd/harmony/main.go
Daniel-VDM: Yes, that's a good idea, will be updated on latest

JB273:
Hey Daniel, got a couple of questions too, didn't have the time this afternoon (CET) to put them here, but like to do so now (some of these seem to be answered already and are just for archive needs):
It's good to have a brainstorm about it together though to find out what exactly is the use case for this, don't want to bring extra features into this that will possible could break this... or create instability in the network!

I'm curious who decides on this blacklist?
Is it a file local to your own node?
If all operators create their own blacklist, will the network become instable?
Example:
I put an ip in the blacklist-file, but you don't on your node, will the tx pass from your node?
Who's tracking changes on the blacklist and decides on giving free the ip's on the list (whitelist your node again)?

Eventually you will not have majority anymore (also this sounds somehow not so decentralized :grin:).
That said I understand the need/wish for something like this, only need to make it secure, workable, scaleable and futureproof...

Security is a very valid discussion, we are hammering on security for node runners!
It's the noderunner responsibilty to have a save system.
We need to somehow make them proof it is secure! That's your first line of defense.

cmd/harmony/main.go
```diff
@@ -127,6 +129,8 @@ var (
 	doRevertBefore = flag.Int("do_revert_before", 0, "If the current block is less than do_revert_before, revert all blocks until (including) revert_to block")
 	revertTo       = flag.Int("revert_to", 0, "The revert will rollback all blocks until and including block number revert_to")
 	revertBeacon   = flag.Bool("revert_beacon", false, "Whether to revert beacon chain or the chain this node is assigned to")
+	// Blacklist of addresses
+	blacklistPath = flag.String("blacklist", "./blacklist.txt", "Path to blacklist file.")
```
denniswon: indentation for =

fxfactorial: Default supposed to be empty string

cmd/harmony/main.go
```diff
@@ -416,6 +425,27 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 	return currentNode
 }
 
+func setupBlacklist() (*map[ethCommon.Address]bool, error) {
+	if _, err := os.Stat(*blacklistPath); os.IsNotExist(err) {
+		return nil, errors.New(fmt.Sprintf("blacklist file not found at `%s`", *blacklistPath))
+	}
```
fxfactorial: No point to check file and then read. Just read and if no file then error will happen

core/tx_pool.go
```diff
@@ -83,6 +84,12 @@ var (
 	// ErrKnownTransaction is returned if a transaction that is already in the pool
 	// attempting to be added to the pool.
 	ErrKnownTransaction = errors.New("known transaction")
+
+	// ErrBlacklistFrom is returned if a transaction's from/source address is blacklisted
+	ErrBlacklistFrom = errors.New("`from` address of transaction is blacklisted")
```
fxfactorial: ... transaction in blacklist

core/tx_pool.go
```diff
@@ -162,6 +171,8 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	GlobalQueue:  1024,
 
 	Lifetime: 3 * time.Hour,
+
+	Blacklist: &map[common.Address]bool{},
```
fxfactorial: Use struct{} not bool

cmd/harmony/main.go
```diff
@@ -416,6 +425,27 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
...
+	dat, err := ioutil.ReadFile(*blacklistPath)
+	if err != nil {
+		return nil, err
+	}
+	addrMap := make(map[ethCommon.Address]bool)
+	for _, line := range strings.Split(string(dat), "\n") {
+		if len(line) != 0 { // blacklist file may have trailing empty string line
+			addr, err := common.Bech32ToAddress(line)
+			if err != nil {
+				return nil, err
```
denniswon: you can also just continue and skip the line with error

core/tx_pool.go
```diff
@@ -83,6 +84,12 @@ var (
 	// ErrKnownTransaction is returned if a transaction that is already in the pool
 	// attempting to be added to the pool.
 	ErrKnownTransaction = errors.New("known transaction")
+
+	// ErrBlacklistFrom is returned if a transaction's from/source address is blacklisted
+	ErrBlacklistFrom = errors.New("`from` address of transaction is blacklisted")
+
+	// ErrBlacklistTo is returned if a transaction's to/destination address is blacklisted
+	ErrBlacklistTo = errors.New("`to` address of transaction is blacklisted")
```
denniswon: what is the risk that this 'to' blacklist feature trying to mitigate? Was not clear why we need this in addition to 'from' blacklist

core/tx_pool.go
```diff
@@ -901,14 +934,15 @@ func (pool *TxPool) addTxsLocked(txs []*types.Transaction, local bool) []error {
...
+		if err == nil && !replace {
 			from, _ := types.Sender(pool.signer, tx) // already validated
 			dirty[from] = struct{}{}
 		}
-		if errs[i] != nil && errors.Cause(errs[i]) != ErrKnownTransaction {
-			erroredTxns = append(erroredTxns, *types.NewRPCTransactionError(tx.Hash(), errs[i]))
+		if err != nil && err != ErrKnownTransaction {
+			erroredTxns = append(erroredTxns, *types.NewRPCTransactionError(tx.Hash(), err))
 		}
+		errs[i] = errors.Cause(err)
```
denniswon: why were these lines here changed? Not sure what the change adds

core/tx_pool_test.go
```diff
@@ -261,6 +261,52 @@ func TestInvalidTransactions(t *testing.T) {
...
+	bannedFromAcc, _ := deriveSender(badTx)
+	bannedToAcc := *badTx.To()
+	goodFromAcc, _ := deriveSender(goodTx)
+
+	// Fund from accounts
+	pool.currentState.AddBalance(bannedFromAcc, big.NewInt(50100))
+	pool.currentState.AddBalance(goodFromAcc, big.NewInt(50100))
+
+	(*DefaultTxPoolConfig.Blacklist)[bannedFromAcc] = false
+	(*DefaultTxPoolConfig.Blacklist)[bannedToAcc] = true
```
denniswon: it should be commented somewhere about the format of the blacklist file (that false => from blacklist, true => to blacklist)

cmd/harmony/main.go
```diff
@@ -322,9 +326,14 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
...
 
+	blacklist, err := setupBlacklist()
+	if err != nil {
+		utils.Logger().Warn().Msgf("Blacklist error: %s", err.Error())
+	}
+
 	// Current node.
 	chainDBFactory := &shardchain.LDBFactory{RootDir: nodeConfig.DBDir}
-	currentNode := node.New(myHost, currentConsensus, chainDBFactory, *isArchival)
+	currentNode := node.New(myHost, currentConsensus, chainDBFactory, blacklist, *isArchival)
```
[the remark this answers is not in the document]
Daniel-VDM: Yes, it needs to be restarted. I think this is fine since the blacklist is not be updated frequently.

cmd/harmony/main.go
Daniel-VDM: I prefer all or nothing when loading addresses.

core/tx_pool.go
Daniel-VDM: One should not be able to send a transaction to a blacklisted account. Stopping someone from burning tokens.

core/tx_pool.go
Daniel-VDM: The function should return error causes and not wrapped errors with messages. Look at `validateTx`.

core/tx_pool_test.go
Daniel-VDM: yes, will add to usage msg

core/tx_pool.go
Daniel-VDM: will change, didn't realize empty structs use no mem.

core/tx_pool.go
Daniel-VDM: will change

cmd/harmony/main.go
Daniel-VDM: Ah right, will change

cmd/harmony/main.go
Daniel-VDM: Will change

cmd/harmony/main.go
Daniel-VDM: I prefer all or nothing

cmd/harmony/main.go
```diff
@@ -416,6 +425,25 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 	return currentNode
 }
 
+func setupBlacklist() (*map[ethCommon.Address]struct{}, error) {
+	utils.Logger().Debug().Msgf("Using blacklist file at `%s`", *blacklistPath)
+	dat, err := ioutil.ReadFile(*blacklistPath)
```
rlan35: can we have a default path for the blacklist? So user don't have to manually set it all the time?

LeoHChen:
> > What about having all the nodes to download that file let's say every x time from our main github repo or an s3 bucket. Node would be updated automatically without having the need for the node to restart, like it is right now via the bash script node.sh.
> 
> I like this idea. We could pull the latest blacklist every time the harmony binary is run. So every time `node.sh` automatically updates, we could also refresh the blacklist. One thing though, the node's binary would need to be restarted every time the list is updated. I don't like the idea of the binary itself look for a list every x second as a blacklist is not meant to be constantly updated.

I don't think we should automate the blacklist update process to start with.  It has to be a mindful decision for node runners to update the blacklist file in order to really blacklist the account approved by the community.  The blacklist update is maintained by the community.

Of course, if later on, the community vote to have auto-update blacklist, we can always implement it later. But not at the initial stage.

scripts/node.sh
```diff
@@ -123,6 +123,7 @@ options:
    -y             run in legacy, foundational-node mode (default)
    -M             support multi-key mode (default: off)
    -A             enable archival node mode (default: off)
+   -B blacklist   specify file containing blacklisted accounts as a newline delimited file (default: ./blacklist.txt)
```
LeoHChen: default in .hmy directory would be better.

cmd/harmony/main.go
Daniel-VDM: Okay will do

cmd/client/txgen/main.go
```diff
@@ -98,7 +98,7 @@ func setUpTXGen() *node.Node {
 	decider := quorum.NewDecider(quorum.SuperMajorityVote)
```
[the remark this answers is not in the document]
Daniel-VDM: https://github.com/harmony-one/harmony/issues/2183

cmd/client/wallet_stress_test/main.go
```diff
@@ -189,7 +189,7 @@ func createWalletNode() *node.Node {
 		utils.FatalErrMsg(err, "cannot initialize network")
```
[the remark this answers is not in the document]
Daniel-VDM: https://github.com/harmony-one/harmony/issues/2183

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

Daniel-VDM:
jenkins test this please

cmd/harmony/main.go
```diff
@@ -322,9 +326,14 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 		currentConsensus.DisableViewChangeForTestingOnly()
 	}
 
+	blacklist, err := setupBlacklist()
+	if err != nil {
+		utils.Logger().Warn().Msgf("Blacklist error: %s", err.Error())
```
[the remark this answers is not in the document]
Daniel-VDM: will change

scripts/node.sh
Daniel-VDM: will move

cmd/harmony/main.go
```diff
@@ -416,6 +425,25 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
...
 }
 
+func setupBlacklist() (*map[ethCommon.Address]struct{}, error) {
+	utils.Logger().Debug().Msgf("Using blacklist file at `%s`", *blacklistPath)
+	dat, err := ioutil.ReadFile(*blacklistPath)
+	if err != nil {
+		return nil, err
+	}
+	addrMap := make(map[ethCommon.Address]struct{})
+	for _, line := range strings.Split(string(dat), "\n") {
```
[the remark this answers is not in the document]
Daniel-VDM: comments supported in the latest, and added docs

PR #2173, opened 2020-02-03
Update description fields individually; Cleanup staking txn fields

rlan35:


staking/types/validator.go
```diff
@@ -19,8 +19,8 @@ import (
 
 // Define validator staking related const
 const (
-	MaxNameLength            = 70
-	MaxIdentityLength        = 3000
+	MaxNameLength            = 140
```
gupadhyaya: max lengths are changed now? We need to update cli too.

rlan35: Yea, please. take a look at the test plan.

PR #2179, opened 2020-02-04
Report duplicate bls keys on the same validator as error

rlan35:


staking/types/validator.go
```diff
@@ -259,6 +262,14 @@ func (w *ValidatorWrapper) SanityCheck() error {
 		)
 	}
 
+	allKeys := map[shard.BlsPublicKey]struct{}{}
```
gupadhyaya: we can never add a duplicate key to `validator.SlotPubKeys`, so why check for duplicates?

gupadhyaya: I see this logic is processing the input. Then it is fine I guess.

PR #2180, opened 2020-02-04
Add one-off block verification rule for beacon chain bad block

rlan35:
Tested on mainnet.


LeoHChen:
please cherry-pick to s3 as well.

PR #2181, opened 2020-02-05
multi bls key support

gupadhyaya:
- [testnet] use a new testnet for multi-key test

- adding support for multiple initial accounts for multi-key scenarios

- minor fixes

- changing viewchange duration to exponential

[multi-key] use 2 shards for testnet of multi-key (#2)

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/hex"
 	"flag"
 	"fmt"
 	"io/ioutil"
@@ -9,6 +8,7 @@ import (
 	"math/rand"
 	"os"
 	"path"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
@@ -116,8 +116,10 @@ var (
 	enableMemProfiling = flag.Bool("enableMemProfiling", false, "Enable memsize logging.")
 	enableGC           = flag.Bool("enableGC", true, "Enable calling garbage collector manually .")
 	blsKeyFile         = flag.String("blskey_file", "", "The encrypted file of bls serialized private key by passphrase.")
-	blsPass            = flag.String("blspass", "", "The file containing passphrase to decrypt the encrypted bls file.")
-	blsPassphrase      string
+	blsFolder          = flag.String("blsfolder", ".hmy/blskeys", "The folder that stores the bls keys; same blspass is used to decrypt all bls keys; all bls keys mapped to same shard")
+
+	blsPass       = flag.String("blspass", "", "The file containing passphrase to decrypt the encrypted bls file.")
+	blsPassphrase string
 
 	// Sharding configuration parameters for devnet
 	devnetNumShards   = flag.Uint("dn_num_shards", 2, "number of shards for -network_type=devnet (default: 2)")
@@ -131,7 +133,7 @@ var (
 
 	keystoreDir = flag.String("keystore", hmykey.DefaultKeyStoreDir, "The default keystore directory")
 
-	initialAccount = &genesis.DeployAccount{}
+	initialAccounts = make([]*genesis.DeployAccount, 0)
 
 	// logging verbosity
 	verbosity = flag.Int("verbosity", 5, "Logging verbosity: 0=silent, 1=error, 2=warn, 3=info, 4=debug, 5=detail (default: 5)")
@@ -202,8 +204,12 @@ func passphraseForBls() {
 		return
 	}
 
-	if *blsKeyFile == "" || *blsPass == "" {
-		fmt.Println("Internal nodes need to have pass to decrypt blskey")
+	if *blsKeyFile == "" && *blsFolder == "" {
+		fmt.Println("blskey_file or blsfolder option must be provided")
+		os.Exit(101)
+	}
+	if *blsPass == "" {
+		fmt.Println("Internal nodes need to have blspass to decrypt blskey")
 		os.Exit(101)
 	}
 	passphrase, err := utils.GetPassphraseFromSource(*blsPass)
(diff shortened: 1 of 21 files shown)


gupadhyaya:
@rlan35 @fxfactorial @LeoHChen when you get chance, please review this PR, so that we can make progress. Fixed the bug that was spoiling shard chain explorer. The multi-bls network is running fine. I will continue to test the cross-shard tx.

cmd/harmony/main.go
```diff
@@ -116,8 +115,10 @@ var (
 	enableMemProfiling = flag.Bool("enableMemProfiling", false, "Enable memsize logging.")
 	enableGC           = flag.Bool("enableGC", true, "Enable calling garbage collector manually .")
 	blsKeyFile         = flag.String("blskey_file", "", "The encrypted file of bls serialized private key by passphrase.")
-	blsPass            = flag.String("blspass", "", "The file containing passphrase to decrypt the encrypted bls file.")
-	blsPassphrase      string
+	blsFolder          = flag.String("blsfolder", "", "The encrypted file of bls serialized private key by passphrase.")
```
rlan35: description not updated.

cmd/harmony/main.go
```diff
@@ -214,59 +219,103 @@ func passphraseForBls() {
 	blsPassphrase = passphrase
 }
 
-func setupInitialAccount() (isLeader bool) {
+func findAccountsByPubKeys(config shardingconfig.Instance, pubKeys []*bls.PublicKey) {
+	for _, key := range pubKeys {
+		_, account := config.FindAccount(key.SerializeToHexStr())
+		if account != nil {
+			initialAccounts = append(initialAccounts, account)
+		}
```
rlan35: let's printout if a key is not found for informational purpose.

consensus/consensus_leader_msg.go
```diff
@@ -42,7 +42,7 @@ func (consensus *Consensus) constructPreparedMessage() ([]byte, *bls.Sign) {
 	}
 
 	consensusMsg := message.GetConsensus()
-	consensus.populateMessageFields(consensusMsg)
+	consensus.populateMessageFields(consensusMsg, consensus.LeaderPubKey)
```
rlan35: Why put as a separate field. Isn't populateMessageFields already have access to consensus.LeaderPubKey?

consensus/consensus_v2.go
```diff
@@ -290,16 +297,19 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
-		utils.Logger().Info().
-			Str("blockHash", hex.EncodeToString(consensus.blockHash[:])).
-			Msg("[OnAnnounce] Sent Prepare Message!!")
+	groupID := []nodeconfig.GroupID{nodeconfig.NewGroupIDByShardID(nodeconfig.ShardID(consensus.ShardID))}
+	for i, key := range consensus.PubKey.PublicKey {
+		// Construct and send prepare message
+		msgToSend := consensus.constructPrepareMessage(key, consensus.priKey.PrivateKey[i])
+		// TODO: this will not return immediatey, may block
+
+		if err := consensus.msgSender.SendWithoutRetry(groupID, host.ConstructP2pMessage(byte(17), msgToSend)); err != nil {
```
rlan35: So we are not grouping the messages together in a single message?

consensus/consensus_v2.go
```diff
@@ -103,7 +103,12 @@ func (consensus *Consensus) announce(block *types.Block) {
 
 	consensus.block = encodedBlock
 	consensus.blockHeader = encodedBlockHeader
-	msgToSend := consensus.constructAnnounceMessage()
+	key, err := consensus.GetConsensusLeaderPrivateKey()
```
rlan35: This call is a bit expensive. Why can't we store the key in a field and directly read it? It's doing the looping everytime.

cmd/harmony/main.go
```diff
@@ -311,7 +360,11 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 	currentConsensus, err := consensus.New(
 		myHost, nodeConfig.ShardID, p2p.Peer{}, nodeConfig.ConsensusPriKey, decider,
 	)
-	currentConsensus.SelfAddress = common.ParseAddr(initialAccount.Address)
+
+	currentConsensus.SelfAddress = make(map[string]ethCommon.Address)
+	for _, initialAccount := range initialAccounts {
```
fxfactorial: leave def of selfaddress alone, instead add new field of  a map[shard.BlsPublicKey]struct{}

fxfactorial:
So I have a suggestion for a different approach because the current approach will mean reimplenting again on master because otherwise the merge conflicts will be insane. 

Instead consider this. 

You make some kind of interface, like so: 

```go
type BLSSigner interface {
    SignForKeyIndex(index int, payload []byte) *bls.Sign
}
```

or something like that, and you attach a concrete type implementation of this interface to the consensus record itself, like how Decider is done. Now there are no function signatures to change, just use this field when need to sign. Adding this kind of thing will be much easier to master and will make it easy later for if we want to change the bls key management impl.

cmd/harmony/main.go
LeoHChen: also, please add a message saying that all blskey has to be in the same shard, also share the same pass.

cmd/harmony/main.go
LeoHChen: ditto

cmd/harmony/main.go
```diff
@@ -214,59 +219,103 @@ func passphraseForBls() {
...
+	for _, account := range initialAccounts {
+		fmt.Printf("My Genesis Account: %v\n", *account)
+	}
 }
 
-func setupConsensusKey(nodeConfig *nodeconfig.ConfigType) *bls.PublicKey {
-	consensusPriKey, err := blsgen.LoadBlsKeyWithPassPhrase(*blsKeyFile, blsPassphrase)
+func readMultiBlsKeys(consensusMultiBlsPriKey *nodeconfig.MultiBlsPrivateKey, consensusMultiBlsPubKey *nodeconfig.MultiBlsPublicKey) error {
+	multiBlsKeyDir := blsFolder
+	blsKeyFiles, err := ioutil.ReadDir(*multiBlsKeyDir)
```
LeoHChen: please do a better job to read only .key files.

consensus/view_change.go
```diff
@@ -121,20 +121,22 @@ func (consensus *Consensus) startViewChange(viewID uint64) {
 	consensus.current.SetViewID(viewID)
 	consensus.LeaderPubKey = consensus.GetNextLeaderKey()
 
-	diff := viewID - consensus.viewID
-	duration := time.Duration(int64(diff) * int64(viewChangeDuration))
+	diff := int64(viewID - consensus.viewID)
+	duration := time.Duration(diff * diff * int64(viewChangeDuration))
```
LeoHChen: this change has to create a separate PR.

internal/configs/sharding/testnet.go
```diff
@@ -16,9 +16,9 @@ const (
 	testnetVdfDifficulty = 10000 // This takes about 20s to finish the vdf
 
 	// TestNetHTTPPattern is the http pattern for testnet.
-	TestNetHTTPPattern = "https://api.s%d.b.hmny.io"
+	TestNetHTTPPattern = "https://api.s%d.b2.hmny.io"
```
LeoHChen: this change b2 shouldn't be here.

cmd/harmony/main.go
gupadhyaya: `blsFolder          = flag.String("blsfolder", ".hmy/blskeys", "The folder that stores the bls keys; same blspass is used to decrypt all bls keys; all bls keys mapped to same shard")`

cmd/harmony/main.go
gupadhyaya: `fmt.Printf("Bls key not found: %s\n", keyStr)`

cmd/harmony/main.go
gupadhyaya: `
if filepath.Ext(blsKeyFile.Name()) != ".key" {
			continue
		}
`

consensus/consensus_leader_msg.go
gupadhyaya: Also used by non-leader during `constructPrepareMessage` which needs non-leader pub key to be populated.

consensus/consensus_v2.go
gupadhyaya: Should be possible, but requires more changes to signing and verifying. I was planning to consider this for the open staking multi key support.

consensus/consensus_v2.go
gupadhyaya: Agreed. I will add this optimization to the open staking multi-key support.

cmd/harmony/main.go
gupadhyaya: why do we need two separate fields? The only use of the SelfAddress is to correctly assign the address of the block proposer while proposing the block. 
`coinbase := node.Consensus.SelfAddress[node.Consensus.LeaderPubKey.SerializeToHexStr()]` 
SelfAddress is not used anywhere else.

gupadhyaya:
> So I have a suggestion for a different approach because the current approach will mean reimplenting again on master because otherwise the merge conflicts will be insane.
> 
> Instead consider this.
> 
> You make some kind of interface, like so:
> 
> ```go
> type BLSSigner interface {
>     SignForKeyIndex(index int, payload []byte) *bls.Sign
> }
> ```
> 
> or something like that, and you attach a concrete type implementation of this interface to the consensus record itself, like how Decider is done, the dependency injection way. Now there are no function signatures to change, just use this field when need to sign. Adding this kind of thing will be much easier to master and will make it easy later for if we want to change the bls key management impl.

Great suggestion. But, I want to consider this for the master only. Anyways, we will have to discard these changes when we move to master (as master has deviated quite a bit from current s3). Do you agree?

fxfactorial:
No, that's the point. By writing behind an interface and your concrete type, will be less code to throw away since you won't touch function signatures by by way suggested and the work will have less chance of merge conflict with master if you try a cherry pick with work already done.

cmd/harmony/main.go
fxfactorial: To lessen the pain of merge conflicts to master

PR #2185, opened 2020-02-05
[state][staking][availability] Handle availability only in finalize

fxfactorial:


core/state/dump.go
```diff
@@ -55,31 +57,41 @@ func (db *DB) RawDump() Dump {
 		if err := rlp.DecodeBytes(it.Value, &data); err != nil {
 			panic(err)
 		}
-
 		obj := newObject(nil, common.BytesToAddress(addr), data)
+		var wrapper staking.ValidatorWrapper
+		wrap := ""
+		if err := rlp.DecodeBytes(obj.Code(db.db), &wrapper); err != nil {
```
rlan35: what is this code?

staking/availability/measure.go
```diff
@@ -16,8 +16,12 @@ import (
...
 var (
 	measure                    = new(big.Int).Div(big.NewInt(2), big.NewInt(3))
-	errValidatorEpochDeviation = errors.New("validator snapshot epoch not exactly one epoch behind")
-	errNegativeSign            = errors.New("impossible period of signing")
+	errValidatorEpochDeviation = errors.New(
+		"validator snapshot epoch not exactly one epoch behind",
+	)
+	errNegativeSign = errors.New("impossible period of signing")
+	// ErrDivByZero ..
+	ErrDivByZero = errors.New("toSign of availability cannot be 0, mistake in protocol")
```
rlan35: "mistake in protocol" => "This shouldn't happen"; this is no mistake in protocol :)

core/state/dump.go
fxfactorial: makes debugging easy, dumping the validator wrapper if it exists in the code field

PR #2192, opened 2020-02-06
Optimization for median raw stake and staking new methods

flicker-harmony:
## Issue

Optimization for median raw stake - recalculates only with new block and new methods: getStake, getTotalStaking and getTotalSupply. For last one I still need formula for total ONE supply, it was requested.

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**


hmy/api_backend.go
```diff
@@ -27,7 +28,12 @@ import (
 
 // APIBackend An implementation of internal/hmyapi/Backend. Full client.
 type APIBackend struct {
-	hmy *Harmony
+	hmy              *Harmony
+	MedianStakeCache struct {
+		Mut            sync.Mutex
```
fxfactorial: No need it give it a name at all, just embed the Mutex alone, sync.Mutex, then we can do MedianStakeCache.Lock() directly.

hmy/api_backend.go
```diff
@@ -321,11 +327,20 @@ var (
 	two = big.NewInt(2)
 )
 
-// GetMedianRawStakeSnapshot  ..
+// GetMedianRawStakeSnapshot ..
 func (b *APIBackend) GetMedianRawStakeSnapshot() *big.Int {
+	b.MedianStakeCache.Mut.Lock()
```
fxfactorial: No need to later have the unlock scattered all over, just immediately after grabbing the lock, do defer b.MedianStakeCache.Unlock()

hmy/api_backend.go
```diff
@@ -321,11 +327,19 @@ var (
...
 
-// GetMedianRawStakeSnapshot  ..
+// GetMedianRawStakeSnapshot ..
 func (b *APIBackend) GetMedianRawStakeSnapshot() *big.Int {
+	b.MedianStakeCache.Lock()
+	defer b.MedianStakeCache.Unlock()
+	if b.MedianStakeCache.BlockHeight != -1 && b.MedianStakeCache.BlockHeight > int64(rpc.LatestBlockNumber)-20 {
+		return b.MedianStakeCache.MedianRawStake
+	}
+	b.MedianStakeCache.Unlock()
```
fxfactorial: no need to lock, unlock anymore since this initial two calls.

Issue #2126, opened 2020-01-21
Add rpc to return APR (based on circulatingSupply and stakeRatio)


rlan35:
APR = (num_blocks_per_year * block_rewards) / total_staked_tokens

block_rewards = max[   18 - (stake_ratio - 0.35) * 100 * 0.4    ,     0   ]

stake_ratio = total_staked_tokens / circulating_supply

circulating_supply can be read from chain DB.
total_staked_tokens can be calculated by going through all validators (better to keep a record on chain DB too to avoid recalculation)

PR #2204, opened 2020-02-07
[staking][economics] Compute APR for validators

fxfactorial:
Closes https://github.com/harmony-one/harmony/issues/2126

compute is pure, no off-chain db caching of values (that is already a TODO in internal/chain/reward.go) We can do that later/someone else 
Using the `hmy` cli 
```
./hmy blockchain utility-metrics
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "accumulated-rewards": {
      "total-block-reward-accum": 0,
      "validator-rewards": []
    },
    "active-validators-apr": [],
    "current-percent-network-deviation": "0.350000000000000000",
    "current-percent-token-staked": "0.000000000000000000",
    "reward-bonus": "14000000000000000000.000000000000000000"
  }
}
```


staking/economics/reward.go
```diff
@@ -12,6 +12,10 @@ import (
 	"github.com/harmony-one/harmony/numeric"
 )
 
+const (
+	numBlocksPerYear = 300_000_000
```
rlan35: We shouldn't hard code the block numbers per year. it's a dynamic number. Depending how you calculate the average block time (e.g. averaging over last 1000 blocks or 10000 blocks). Let's avg over last epoch's block time, instead of hard coded like this.

fxfactorial: ah, yes, my bad. I left it in while developing, will do avg

staking/economics/reward.go
```diff
@@ -41,49 +49,73 @@ func adjust(amount numeric.Dec) numeric.Dec {
...
 	}
-	percentage := stakedNow.Quo(totalTokens.Mul(
+
+	circulatingSupply := totalTokens.Mul(
 		reward.PercentageForTimeStamp(timestamp),
-	).Add(dole))
+	).Add(dole)
+
+	for i := range rates {
+		rates[i].StakeRatio = numeric.NewDecFromBigInt(
```
rlan35: This is wrong. The APR in the formula I wrote is for the general APR for staking assuming all stake are fully utilized. It's not for individual validator's yield. Individual validator's yield should be dependent on their uptime (basically their actual block reward earned). Basically, individual validator's yield <= general APR

consensus/votepower/roster.go
```diff
@@ -41,6 +42,30 @@ func (b BallotResults) EncodePair() (string, string) {
...
 }
 
+// ShardReward ..
+type ShardReward struct {
+	ShardID      uint32
+	EarnedReward *big.Int
+}
+
+// VoterReward ..
+type VoterReward struct {
```
rlan35: Let's be consistent, "ValidatorReward" is better.  Voter can mean many things.

consensus/votepower/roster.go
```diff
@@ -41,6 +42,30 @@ func (b BallotResults) EncodePair() (string, string) {
...
+	Validator   common.Address
+	BlockNumber uint64
+	ByShards    []ShardReward
+}
+
+// RewardAccumulation ..
+type RewardAccumulation struct {
+	NetworkTotalPayout *big.Int      `json:"total-block-reward-accum"`
+	ValidatorReward    []VoterReward `json:"validator-rewards"`
```
rlan35: "ValidatorRewards"

staking/economics/reward.go
```diff
@@ -0,0 +1,284 @@
...
+	return p.blockNumber
+}
+
+// Read ..
+func (p *Produced) Read() *votepower.RewardAccumulation {
+	return &p.accum
+}
+
+// ReadRewarded ..
+func (p *Produced) ReadRewarded() []votepower.VoterReward {
```
rlan35: Bad naming, ReadValidatorRewards()

core/blockchain.go
```diff
@@ -2808,41 +2804,82 @@ func (bc *BlockChain) UpdateStakingMetaData(tx *staking.StakingTransaction, root
...
+			}
+		} else {
+			newWinners = append(newWinners, rewardsSoFar[i])
+		}
 	}
-	return bc.WriteBlockRewardAccumulator(new(big.Int).Add(current, diff), number)
+
+	return bc.WriteBlockRewardAccumulator(economics.NewProduced(
+		rewarded.ReadBlockNumber(),
+		append(rewardsSoFar, newWinners...),
```
rlan35: This should be an accumulator for each validator, right? Not just recording rewards for each block. it will blow up the disk.

core/blockchain.go
```diff
@@ -2808,41 +2804,82 @@ func (bc *BlockChain) UpdateStakingMetaData(tx *staking.StakingTransaction, root
...
+		return bytes.Compare(
+			rewardsSoFar[i].Validator.Bytes(),
+			rewardsSoFar[j].Validator.Bytes(),
+		) == -1
+	})
+
+	newlyRewarded, existingTotal := rewarded.ReadRewarded(), rewarded.ReadTotalPayout()
+	existingTotal.Add(existingTotal, rewarded.ReadTotalPayout())
+
+	for i := range newlyRewarded {
```
rlan35: I know you have a tweet on this. But seems it's not the best way, why don't we just key by the validator address+epoch and value with the accumulated rewards? Then there is no need to search. Just O(1) lookup. This search will get very heavy when number of validators get big. and it's happening for every block.

staking/economics/reward.go
```diff
@@ -0,0 +1,284 @@
...
+	potentialAdjust        = oneHundred.Mul(dynamicAdjust)
+	zero                   = numeric.ZeroDec()
+	// ErrPayoutNotEqualBlockReward ..
+	ErrPayoutNotEqualBlockReward = errors.New("total payout not equal to blockreward")
+	errNotTwoEpochsPastStaking   = errors.New("two epochs ago was not >= staking epoch")
+	oneYear                      = big.NewInt(int64(nanoSecondsInYear))
+)
+
+// ExpectedValueBlocksPerYear ..
+func ExpectedValueBlocksPerYear(
```
rlan35: ExpectedNumBlocksPerYear

staking/economics/reward.go
```diff
@@ -0,0 +1,284 @@
...
+				return bytes.Compare(
+					rewardsSoFar[i].Validator.Bytes(),
+					rewardsSoFar[j].Validator.Bytes(),
+				) == -1
+			})
+
+			for i := range rates {
+				lookingFor := rates[i].Validator.Bytes()
+				if k :=
+					sort.Search(soFarCount, func(j int) bool {
```
rlan35: This part is too complicated too. Basically validator's APR is estimated_reward_per_year/total_stake; estimated_reward_per_year=avg_reward_per_block * blocks_per_year; avg_reward_per_block = total_reward_per_epoch / blocks_per_epoch;

fxfactorial:
merge conflicts now are too difficult to fix, will reattempt in fresh PR

PR #2226, opened 2020-02-08
[consensus] Remove unnecessary usage of bandwidth in network messages…

fxfactorial:
This should have a noticeable impact on our aws bill for network costs for bandwidth usage

### DIFF
--- a/consensus/construct.go
+++ b/consensus/construct.go
@@ -57,7 +57,6 @@ func (consensus *Consensus) construct(
 		}
 	case msg_pb.MessageType_COMMIT:
 		if s := consensus.priKey.SignHash(payloadForSignOverride); s != nil {
-			consensusMsg.Block = consensus.block
 			consensusMsg.Payload = s.Serialize()
 		}
 	case msg_pb.MessageType_COMMITTED:
@@ -69,7 +68,7 @@ func (consensus *Consensus) construct(
 		buffer.Write(consensus.commitBitmap.Bitmap)
 		consensusMsg.Payload = buffer.Bytes()
 	case msg_pb.MessageType_ANNOUNCE:
-		consensusMsg.Payload = consensus.blockHeader
+		consensusMsg.Payload = consensus.blockHash[:]
 	}
 
 	marshaledMessage, err := consensus.signAndMarshalConsensusMessage(message)
--- a/consensus/validator.go
+++ b/consensus/validator.go
@@ -24,7 +24,10 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
 		return
 	}
 
-	consensus.onAnnounceSanityChecks(recvMsg)
+	// NOTE let it handle its own logs
+	if !consensus.onAnnounceSanityChecks(recvMsg) {
+		return
+	}
 
 	consensus.getLogger().Debug().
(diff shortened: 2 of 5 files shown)


consensus/validator.go
```diff
@@ -172,17 +96,17 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
 		Uint64("MsgViewID", recvMsg.ViewID).
 		Msg("[OnPrepared] Received prepared message")
 
-	if recvMsg.BlockNum < consensus.blockNum {
+	if recvMsg.BlockNum != consensus.blockNum {
```
rlan35: We should still store the blocks with Num > curBlock, in the case of missed messages or messages out of order. But for old blocks for sure we can ignore.

consensus/validator.go
```diff
@@ -26,80 +24,7 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
-	header := new(block.Header)
-	if err := rlp.DecodeBytes(encodedHeader, header); err != nil {
-		consensus.getLogger().Warn().
-			Err(err).
-			Uint64("MsgBlockNum", recvMsg.BlockNum).
-			Msg("[OnAnnounce] Unparseable block header data")
-		return
-	}
-
-	if recvMsg.BlockNum < consensus.blockNum ||
```
rlan35: why is block num verification also discarded? Then what's left in the announce?

LeoHChen:
aws bandwidth is impacted by our pubsub gossiping, the size of the message is not a huge factor IMHO.

consensus/validator.go
fxfactorial: In unpushed code, I just didn't push yet. Have some more on the way

fxfactorial:
@LeoHChen We'll find out for sure, soon enough :+1:

LeoHChen:
> @LeoHChen We'll find out for sure, soon enough 👍

this has to be tested out before we merge it. we can actually find out the amount of data being transmitted using a smaller test net and do control group testing.  I am not comfortable of merging the code and test it on mainnet.

consensus/construct.go
```diff
@@ -57,7 +57,6 @@ func (consensus *Consensus) construct(
 		}
 	case msg_pb.MessageType_COMMIT:
 		if s := consensus.priKey.SignHash(payloadForSignOverride); s != nil {
-			consensusMsg.Block = consensus.block
```
LeoHChen: so we don't need to include the block in commit message? why there is no explanation in the commit message of about saving?

consensus/construct.go
```diff
@@ -69,7 +68,7 @@ func (consensus *Consensus) construct(
 		buffer.Write(consensus.commitBitmap.Bitmap)
 		consensusMsg.Payload = buffer.Bytes()
 	case msg_pb.MessageType_ANNOUNCE:
-		consensusMsg.Payload = consensus.blockHeader
+		consensusMsg.Payload = consensus.blockHash[:]
```
LeoHChen: why change this one?  please elaborate in the commit message.

consensus/construct.go
LeoHChen: was the block being used before? or it is totally just discarded before.

PR #2160, opened 2020-01-28
Report failed (non-staking) transactions

Daniel-VDM:
## Issue

Similar to #2157 #2151, this PR will report failed (non-staking) transactions on all nodes instead of just on the leader. 

### Unit Test Coverage

Before:

```
PASS
coverage: 13.6% of statements
ok  	github.com/harmony-one/harmony/node	2.391s
```

```
PASS
coverage: 19.5% of statements
ok  	github.com/harmony-one/harmony/core	6.853s
```

After:

```
PASS
coverage: 14.0% of statements
ok  	github.com/harmony-one/harmony/node	2.067s
```

```
PASS
coverage: 20.4% of statements
ok  	github.com/harmony-one/harmony/core	6.656s
```

## Operational Checklist

1. **Does this PR introduce backward-incompatible changes to the on-disk data structure and/or the over-the-wire protocol?**. (If no, skip to question 8.)

    **NO**

8. **Does this PR introduce backward-incompatible changes *NOT* related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)

    **NO**

11. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**

    **NO**

PR #2236, opened 2020-02-10
Abstract transactions in tx pool and add staking transaction to pool with error report

Daniel-VDM:
## Issue

The transaction pool only accepts plain transactions and needs to be generalized to handle staking transactions. Moreover, only the leader can report failed staking transaction errors ( similar to #2160 ). Lastly, staking messages are only verified when they are applied, this should be separated so staking msg verification can be done at an earlier stage like the tx pool.

### Highlights
* Abstracts transactions in the tx pool by using a `PoolTransaction` interface. This also required some rework of the `tx_list.go` and `tx_journal.go`. This is most of the PR. 
* Implement methods for `StakingTransactions` to satisfy the tx pool interface. 
* Refactor error reporting in tx pool to handle multiple error sinks.
* Add staking transaction checks in the tx pool and report any errors to the error sink.
* Update consumers and providers of tx pool to interface staking transactions.
* Fix RPC `GetTransactionByHash` to never look in tx pool.
* Add pending staking transaction to hmy API. 
* Remove case when old transactions in tx pool report to error sink even though it is finalized. 
* Refactor staking message verification and staking msg application in `state_transition.go`.
* Make all nodes broadcast transactions if they received a message via RPC.

### Unit Test Coverage

Before:

```
PASS
coverage: 20.3% of statements
ok  	github.com/harmony-one/harmony/core	6.815s
```

```
PASS
coverage: 14.3% of statements
ok  	github.com/harmony-one/harmony/node	1.949s
```

After:

```
PASS
coverage: 20.5% of statements
ok  	github.com/harmony-one/harmony/core	8.504s
```

```
PASS
coverage: 14.7% of statements
ok  	github.com/harmony-one/harmony/node	3.389s
```

### Test/Run Logs

* [Transaction Generation](https://gist.github.com/Daniel-VDM/490c711c982209fd3d654b0ea90bda35) (~500 txs in 60 sec on localnet, all succeeded)
* [Basic Staking Tx test](https://gist.github.com/Daniel-VDM/6a275e3c06147438065282a1ed00f7a6) (create validator, edit validator, delegate, undelegate). This test was done with `GenesisGasLimit uint64 = 10712388` since currently create validator costs more than block gas limit on localnet.

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/core/staking_verifier.go
+++ b/core/staking_verifier.go
@@ -1,0 +1,266 @@
+package core
+
+import (
+	"bytes"
+	"math/big"
+
+	"github.com/pkg/errors"
+
+	"github.com/harmony-one/harmony/core/vm"
+	common2 "github.com/harmony-one/harmony/internal/common"
+	staking "github.com/harmony-one/harmony/staking/types"
+)
+
+var (
+	errStateDBIsMissing    = errors.New("no stateDB was provided")
+	errChainContextMissing = errors.New("no chain context was provided")
+	errEpochMissing        = errors.New("no epoch was provided")
+	errBlockNumMissing     = errors.New("no block number was provided")
+)
+
+// TODO: add unit tests to check staking msg verification
+
+// VerifyAndCreateValidatorFromMsg verifies the create validator message using
+// the stateDB, epoch, & blocknumber and returns the validatorWrapper created
+// in the process.
+//
+// Note that this function never updates the stateDB, it only reads from stateDB.
+func VerifyAndCreateValidatorFromMsg(
+	stateDB vm.StateDB, epoch *big.Int, blockNum *big.Int, msg *staking.CreateValidator,
+) (*staking.ValidatorWrapper, error) {
+	if stateDB == nil {
+		return nil, errStateDBIsMissing
+	}
+	if epoch == nil {
+		return nil, errEpochMissing
+	}
+	if blockNum == nil {
+		return nil, errBlockNumMissing
+	}
+	if msg.Amount.Sign() == -1 {
+		return nil, errNegativeAmount
+	}
+	if stateDB.IsValidator(msg.ValidatorAddress) {
+		return nil, errors.Wrapf(errValidatorExist, common2.MustAddressToBech32(msg.ValidatorAddress))
+	}
+	if !CanTransfer(stateDB, msg.ValidatorAddress, msg.Amount) {
+		return nil, errInsufficientBalanceForStake
+	}
+	v, err := staking.CreateValidatorFromNewMsg(msg, blockNum)
+	if err != nil {
+		return nil, err
+	}
+	wrapper := &staking.ValidatorWrapper{}
+	wrapper.Validator = *v
+	wrapper.Delegations = []staking.Delegation{
+		staking.NewDelegation(v.Address, msg.Amount),
+	}
+	wrapper.Snapshot.Epoch = epoch
+	wrapper.Snapshot.NumBlocksSigned = big.NewInt(0)
+	wrapper.Snapshot.NumBlocksToSign = big.NewInt(0)
+	if err := wrapper.SanityCheck(); err != nil {
+		return nil, err
+	}
+	return wrapper, nil
+}
+
+// VerifyAndEditValidatorFromMsg verifies the edit validator message using
+// the stateDB, chainContext and returns the edited validatorWrapper.
+//
+// Note that this function never updates the stateDB, it only reads from stateDB.
+func VerifyAndEditValidatorFromMsg(
+	stateDB vm.StateDB, chainContext ChainContext, blockNum *big.Int, msg *staking.EditValidator,
+) (*staking.ValidatorWrapper, error) {
+	if stateDB == nil {
+		return nil, errStateDBIsMissing
+	}
+	if chainContext == nil {
+		return nil, errChainContextMissing
+	}
+	if blockNum == nil {
+		return nil, errBlockNumMissing
+	}
+	if !stateDB.IsValidator(msg.ValidatorAddress) {
+		return nil, errValidatorNotExist
+	}
+	wrapper := stateDB.GetStakingInfo(msg.ValidatorAddress)
+	if wrapper == nil {
+		return nil, errValidatorNotExist
+	}
+	if err := staking.UpdateValidatorFromEditMsg(&wrapper.Validator, msg); err != nil {
+		return nil, err
+	}
+	newRate := wrapper.Validator.Rate
+	if newRate.GT(wrapper.Validator.MaxRate) {
+		return nil, errCommissionRateChangeTooHigh
+	}
+
+	// TODO: make sure we are reading from the correct snapshot
+	snapshotValidator, err := chainContext.ReadValidatorSnapshot(wrapper.Address)
+	if err != nil {
+		return nil, err
+	}
+	rateAtBeginningOfEpoch := snapshotValidator.Validator.Rate
+
+	if rateAtBeginningOfEpoch.IsNil() || (!newRate.IsNil() && !rateAtBeginningOfEpoch.Equal(newRate)) {
+		wrapper.Validator.UpdateHeight = blockNum
+	}
+
+	if newRate.Sub(rateAtBeginningOfEpoch).Abs().GT(wrapper.Validator.MaxChangeRate) {
+		return nil, errCommissionRateChangeTooFast
+	}
+
+	if err := wrapper.SanityCheck(); err != nil {
+		return nil, err
+	}
+	return wrapper, nil
+}
+
+// VerifyAndDelegateFromMsg verifies the delegate message using the stateDB
+// and returns the balance to be deducted by the delegator as well as the
+// validatorWrapper with the delegation applied to it.
+//
+// Note that this function never updates the stateDB, it only reads from stateDB.
+func VerifyAndDelegateFromMsg(
+	stateDB vm.StateDB, msg *staking.Delegate,
+) (*staking.ValidatorWrapper, *big.Int, error) {
+	if stateDB == nil {
+		return nil, nil, errStateDBIsMissing
+	}
+	if msg.Amount.Sign() == -1 {
+		return nil, nil, errNegativeAmount
+	}
+	if !stateDB.IsValidator(msg.ValidatorAddress) {
+		return nil, nil, errValidatorNotExist
+	}
+	wrapper := stateDB.GetStakingInfo(msg.ValidatorAddress)
+	if wrapper == nil {
+		return nil, nil, errValidatorNotExist
+	}
+	// Check for redelegation
+	for i := range wrapper.Delegations {
+		delegation := &wrapper.Delegations[i]
+		if bytes.Equal(delegation.DelegatorAddress.Bytes(), msg.DelegatorAddress.Bytes()) {
+			totalInUndelegation := delegation.TotalInUndelegation()
+			balance := stateDB.GetBalance(msg.DelegatorAddress)
+			// If the sum of normal balance and the total amount of tokens in undelegation is greater than the amount to delegate
+			if big.NewInt(0).Add(totalInUndelegation, balance).Cmp(msg.Amount) >= 0 {
+				// Check if it can use tokens in undelegation to delegate (redelegate)
+				delegateBalance := big.NewInt(0).Set(msg.Amount)
+				// Use the latest undelegated token first as it has the longest remaining locking time.
+				i := len(delegation.Undelegations) - 1
+				for ; i >= 0; i-- {
+					if delegation.Undelegations[i].Amount.Cmp(delegateBalance) <= 0 {
+						delegateBalance.Sub(delegateBalance, delegation.Undelegations[i].Amount)
+					} else {
+						delegation.Undelegations[i].Amount.Sub(delegation.Undelegations[i].Amount, delegateBalance)
+						delegateBalance = big.NewInt(0)
+						break
+					}
+				}
+				delegation.Undelegations = delegation.Undelegations[:i+1]
+				delegation.Amount.Add(delegation.Amount, msg.Amount)
+				if err := wrapper.SanityCheck(); err != nil {
+					return nil, nil, err
+				}
+				// Return remaining balance to be deducted for delegation
+				if delegateBalance.Cmp(big.NewInt(0)) < 0 {
+					return nil, nil, errInsufficientBalanceForStake // shouldn't really happen
+				}
+				return wrapper, delegateBalance, nil
+			}
+			return nil, nil, errors.Wrapf(
+				errInsufficientBalanceForStake,
+				"total-delegated %s own-current-balance %s amount-to-delegate %s",
+				totalInUndelegation.String(),
+				balance.String(),
+				msg.Amount.String(),
+			)
+		}
+	}
+	// If no redelegation, create new delegation
+	if !CanTransfer(stateDB, msg.DelegatorAddress, msg.Amount) {
+		return nil, nil, errInsufficientBalanceForStake
+	}
+	wrapper.Delegations = append(wrapper.Delegations, staking.NewDelegation(msg.DelegatorAddress, msg.Amount))
+	if err := wrapper.SanityCheck(); err != nil {
+		return nil, nil, err
+	}
+	return wrapper, msg.Amount, nil
+}
+
+// VerifyAndUndelegateFromMsg verifies the undelegate validator message
+// using the stateDB & chainContext and returns the edited validatorWrapper
+// with the undelegation applied to it.
+//
+// Note that this function never updates the stateDB, it only reads from stateDB.
+func VerifyAndUndelegateFromMsg(
+	stateDB vm.StateDB, epoch *big.Int, msg *staking.Undelegate,
+) (*staking.ValidatorWrapper, error) {
+	if stateDB == nil {
+		return nil, errStateDBIsMissing
+	}
+	if epoch == nil {
+		return nil, errEpochMissing
+	}
+	if msg.Amount.Sign() == -1 {
+		return nil, errNegativeAmount
+	}
+	if !stateDB.IsValidator(msg.ValidatorAddress) {
+		return nil, errValidatorNotExist
+	}
+	wrapper := stateDB.GetStakingInfo(msg.ValidatorAddress)
+	if wrapper == nil {
+		return nil, errValidatorNotExist
+	}
+	for i := range wrapper.Delegations {
+		delegation := &wrapper.Delegations[i]
+		if bytes.Equal(delegation.DelegatorAddress.Bytes(), msg.DelegatorAddress.Bytes()) {
+			if err := delegation.Undelegate(epoch, msg.Amount); err != nil {
+				return nil, err
+			}
+			if err := wrapper.SanityCheck(); err != nil {
+				return nil, err
+			}
+			return wrapper, nil
+		}
+	}
+	return nil, errNoDelegationToUndelegate
+}
+
+// VerifyAndCollectRewardsFromDelegation verifies and collects rewards
+// from the given delegation slice using the stateDB. It returns all of the
+// edited validatorWrappers and the sum total of the rewards.
+//
+// Note that this function never updates the stateDB, it only reads from stateDB.
+func VerifyAndCollectRewardsFromDelegation(
+	stateDB vm.StateDB, delegations []staking.DelegationIndex,
+) ([]*staking.ValidatorWrapper, *big.Int, error) {
+	if stateDB == nil {
+		return nil, nil, errStateDBIsMissing
+	}
+	updatedValidatorWrappers := []*staking.ValidatorWrapper{}
+	totalRewards := big.NewInt(0)
+	for i := range delegations {
+		delegation := &delegations[i]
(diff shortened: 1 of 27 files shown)


cmd/harmony/main.go
```diff
@@ -425,7 +425,7 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 	return currentNode
 }
 
-func setupBlacklist() (*map[ethCommon.Address]struct{}, error) {
+func setupBlacklist() (map[ethCommon.Address]struct{}, error) {
```
LeoHChen: why you have a blacklist change in this PR? Please use a separate one for blacklist fix.

core/tx_journal.go
```diff
@@ -76,33 +100,57 @@ func (journal *txJournal) load(add func([]*types.Transaction) []error) error {
...
+		var tx types.PoolTransaction
+		switch txType, err := stream.Uint(); txType {
+		case plainTxID:
+			tx = new(types.Transaction)
+		case stakingTxID:
+			tx = new(staking.StakingTransaction)
+		default:
+			if err != nil {
+				if err == io.EOF {
+					err = nil
```
LeoHChen: why ignore the EOF error? please comment.

core/tx_journal.go
```diff
@@ -76,33 +100,57 @@ func (journal *txJournal) load(add func([]*types.Transaction) []error) error {
...
+					loadBatch(batch)
+				}
+				return err
+			}
+		}
+
 		if err = stream.Decode(tx); err != nil {
-			if err != io.EOF {
-				failure = err
+			if err == io.EOF {
```
LeoHChen: why ignore EOF?

core/tx_list.go
```diff
@@ -380,11 +380,11 @@ func (h priceHeap) Less(i, j int) bool {
 		return false
 	}
 	// If the prices match, stabilize via nonces (high nonce is worse)
-	return h[i].Nonce() > h[j].Nonce()
+	return h[i].Nonce() > h[i].Nonce()
```
LeoHChen: why this change? is it related?

core/types/tx_pool.go
```diff
@@ -0,0 +1,102 @@
...
+	staking "github.com/harmony-one/harmony/staking/types"
+)
+
+var (
+	// ErrUnknownPoolTxType is returned when attempting to assert a PoolTransaction to its concrete type
+	ErrUnknownPoolTxType = errors.New("unknown transaction type in tx-pool")
+)
+
+// PoolTransaction is the general transaction interface used by the tx pool
+type PoolTransaction interface {
```
LeoHChen: do you need to document each field in this interface?

sophoah: Actually comments are good for external parties that do not understanding entirely the code.

core/tx_list.go
Daniel-VDM: Ah was a typo, thx for catching

core/tx_journal.go
Daniel-VDM: ignore EOF to exit reading from journal final without error, will document

core/tx_journal.go
Daniel-VDM: mm yes, we should not ignore EOF here, will change

cmd/harmony/main.go
Daniel-VDM: okay, will remove blacklist change and add in followup pr

core/types/tx_pool.go
Daniel-VDM: I believe we force all exposed components to have documentation (via golint in travis check?)

core/tx_pool.go
```diff
@@ -675,14 +718,282 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
+	}
+}
+
+func (pool *TxPool) validateCreateValidatorMsg(stkMsg *staking.CreateValidator) error {
+	if pool.currentState.IsValidator(stkMsg.ValidatorAddress) {
+		return errValidatorExist
+	}
+
+	// create temp validator to do sanity check
+	validatorWrapper, err := staking.CreateValidatorFromNewMsg(stkMsg, pool.chain.CurrentBlock().Number())
```
rlan35: This few steps is a copy of the actual applyStakingTransaction, better to refactor them into a single function. Otherwise, there can be consistency issue in the future.

core/tx_pool.go
```diff
@@ -675,14 +718,282 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
+
+	return w.SanityCheck()
+}
+
+func (pool *TxPool) validateEditValidatorMsg(stkMsg *staking.EditValidator) error {
+	if !pool.currentState.IsValidator(stkMsg.ValidatorAddress) {
+		return errValidatorNotExist
+	}
+
+	validatorWrapper := pool.currentState.GetStakingInfo(stkMsg.ValidatorAddress)
```
rlan35: same here. Should refactor all these logic into a single function that can be called directly.

core/tx_pool.go
```diff
@@ -675,14 +718,282 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
+func (pool *TxPool) validateDelegateMsg(stkMsg *staking.Delegate) error {
+	if stkMsg.Amount.Sign() == -1 {
+		return errNegativeAmount
+	}
+
+	if !pool.currentState.IsValidator(stkMsg.ValidatorAddress) {
+		return errValidatorNotExist
+	}
+
+	validatorWrapper := pool.currentState.GetStakingInfo(stkMsg.ValidatorAddress)
```
rlan35: same here.

core/tx_pool.go
```diff
@@ -675,14 +718,282 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
+func (pool *TxPool) validateUndelegateMsg(stkMsg *staking.Undelegate) error {
+	if stkMsg.Amount.Sign() == -1 {
+		return errNegativeAmount
+	}
+
+	if !pool.currentState.IsValidator(stkMsg.ValidatorAddress) {
+		return errValidatorNotExist
+	}
+
+	validatorWrapper := pool.currentState.GetStakingInfo(stkMsg.ValidatorAddress)
```
rlan35: same

core/tx_pool.go
```diff
@@ -675,14 +718,282 @@ func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
...
+	return nil
+}
+
+func (pool *TxPool) validateCollectRewardsMsg(stkMsg *staking.CollectRewards) error {
+	chain, ok := pool.chain.(*BlockChain)
+	if !ok {
+		return nil // for testing, chain could be testing blockchain
+	}
+
+	delegations, err := chain.ReadDelegationsByDelegator(stkMsg.DelegatorAddress)
```
rlan35: same

core/tx_pool.go
```diff
@@ -1057,10 +1381,6 @@ func (pool *TxPool) promoteExecutables(accounts []common.Address) {
 		for _, tx := range list.Forward(nonce) {
 			hash := tx.Hash()
 			logger.Warn().Str("hash", hash.Hex()).Msg("Removed old queued transaction")
-			if pool.chain.CurrentBlock().Transaction(hash) == nil {
```
rlan35: why are these checks removed?

core/tx_pool.go
```diff
@@ -1242,10 +1557,6 @@ func (pool *TxPool) demoteUnexecutables() {
 		for _, tx := range list.Forward(nonce) {
 			hash := tx.Hash()
 			logger.Warn().Str("hash", hash.Hex()).Msg("Removed old pending transaction")
-			if pool.chain.CurrentBlock().Transaction(hash) == nil {
```
rlan35: same here?

core/tx_pool.go
Daniel-VDM: Error report not needed here since error would be caught when tx is added to tx-pool in the first place (from the `validateTx` fn). Also looking up a tx is a linear time op in the number of txs in the block.

core/tx_pool.go
Daniel-VDM: addressed in other msg

node/node.go
```diff
@@ -282,50 +279,48 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
...
-	// TODO: everyone should record staking txns, not just leader
-	if node.Consensus.IsLeader() &&
-		node.NodeConfig.ShardID == shard.BeaconChainShardID {
+func (node *Node) AddPendingStakingTransaction(newStakingTx *staking.StakingTransaction) {
+	if node.NodeConfig.ShardID == shard.BeaconChainShardID {
 		node.addPendingStakingTransactions(staking.StakingTransactions{newStakingTx})
-	} else {
-		utils.Logger().Info().Str("Hash", newStakingTx.Hash().Hex()).Msg("Broadcasting Staking Tx")
-		node.tryBroadcastStaking(newStakingTx)
+		if node.NodeConfig.Role() != nodeconfig.ExplorerNode {
```
rlan35: The original logic is still correct, we want any node that's not the leader to broadcast the staking transaction (not just the explorer node, it's a single point of failure)

Daniel-VDM: discussed in person, will also change regular tx logic so that everyone broadcasts in this PR

core/tx_pool.go
Daniel-VDM: refactored the staking tx check and tx pool now calls it

core/staking_verifier.go
```diff
@@ -0,0 +1,265 @@
...
+	wrapper.Snapshot.Epoch = epoch
+	wrapper.Snapshot.NumBlocksSigned = zero
+	wrapper.Snapshot.NumBlocksToSign = zero
+	return wrapper.SanityCheck()
+}
+
+// VerifyEditValidatorStakingMsg verifies the edit validator message using
+// the stateDB and chainContext if they are provided.
+// It is not necessary to provide the stateDB and/or chainContext; validation that
+// does not using said components will still run.
```
rlan35: "does not use". Please correct everywhere.

Daniel-VDM: corrected.

core/state_transition.go
```diff
@@ -438,167 +432,94 @@ func (st *StateTransition) applyEditValidatorTx(
...
-	}
 	wrapper := st.state.GetStakingInfo(delegate.ValidatorAddress)
 	if wrapper == nil {
 		return errValidatorNotExist
 	}
-
-	stateDB := st.state
-	delegatorExist := false
-	for i := range wrapper.Delegations {
-		delegation := &wrapper.Delegations[i]
```
rlan35: Use this original way of referencing delegation. Your way gets a copy of the delegation, and it is a source of problems

core/state_transition.go
```diff
@@ -314,54 +311,63 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
 		if err = rlp.DecodeBytes(msg.Data(), stkMsg); err != nil {
 			return 0, err
 		}
-		utils.Logger().Info().Msgf("[DEBUG STAKING] staking type: %s, txn: %+v", msg.Type(), stkMsg)
+		utils.Logger().Info().Msgf("[DEBUG STAKING] staking type: %s, gas: %d, txn: %+v", msg.Type(), gas, stkMsg)
 		if msg.From() != stkMsg.ValidatorAddress {
 			return 0, errInvalidSigner
 		}
+		if err := VerifyCreateValidatorStakingMsg(st.state, st.evm.EpochNumber, stkMsg); err != nil {
```
rlan35: Better put the Verify.... inside apply........, so the apply can never mistakenly commit an invalid message.

core/staking_verifier.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+	if !stateDB.IsValidator(msg.ValidatorAddress) {
+		return errValidatorNotExist
+	}
+	wrapper := stateDB.GetStakingInfo(msg.ValidatorAddress)
+	if wrapper == nil {
+		return errValidatorNotExist
+	}
+	delegatorExist := false
+	for _, delegation := range wrapper.Delegations {
```
rlan35: same here for the reference.

core/staking_verifier.go
```diff
@@ -0,0 +1,265 @@
...
+			return errors.Wrapf(
+				errInsufficientBalanceForStake,
+				"total-delegated %s own-current-balance %s amount-to-delegate %s",
+				totalInUndelegation.String(),
+				balance.String(),
+				msg.Amount.String(),
+			)
+		}
+	}
+	if !delegatorExist {
```
rlan35: why not remove delegatorExist here, but in the apply....?

core/state_transition.go
Daniel-VDM: oh alright, changed back to original way

core/state_transition.go
Daniel-VDM: sure, moved into apply....

core/staking_verifier.go
Daniel-VDM: oh, this can be removed. Done in latest

core/staking_verifier.go
```diff
@@ -0,0 +1,262 @@
...
+// VerifyDelegateStakingMsg verifies the delegate message using the stateDB if
+// it is provided. Otherwise, it will only check components that don't require it.
+//
+// Note that these checks never update the stateDB, it only reads from stateDB.
+func VerifyDelegateStakingMsg(stateDB vm.StateDB, msg *staking.Delegate) error {
+	if msg.Amount.Sign() == -1 {
+		return errNegativeAmount
+	}
+	if stateDB == nil {
+		return nil
```
rlan35: If there is no state, then should return error. Otherwise, it can be a false negative.

core/state_transition.go
```diff
@@ -373,19 +367,13 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
...
-	if val := createValidator.ValidatorAddress; st.state.IsValidator(val) {
-		return errors.Wrapf(errValidatorExist, common2.MustAddressToBech32(val))
-	}
-
-	if !CanTransfer(st.state, createValidator.ValidatorAddress, createValidator.Amount) {
-		return errInsufficientBalanceForStake
+func (st *StateTransition) verifyAndApplyCreateValidatorTx(
+	createValidator *staking.CreateValidator, blockNum *big.Int,
+) error {
+	err := VerifyCreateValidatorStakingMsg(st.state, st.evm.EpochNumber, createValidator)
```
rlan35: You can return the created validator from VerifyCreateValidatorStakingMsg (maybe rename to VerifyAndCreateValidatorStakingMsg), and directly commit with that validator. Then you don't need to redo the whole wrapper.validator = .... ; wrapper.snapshot = .... thing. They are duplicate code.

core/state_transition.go
```diff
@@ -410,13 +398,13 @@ func (st *StateTransition) applyCreateValidatorTx(createValidator *staking.Creat
 	return nil
 }
 
-func (st *StateTransition) applyEditValidatorTx(
+func (st *StateTransition) verifyAndApplyEditValidatorTx(
 	editValidator *staking.EditValidator, blockNum *big.Int,
 ) error {
-	if !st.state.IsValidator(editValidator.ValidatorAddress) {
-		return errValidatorNotExist
+	err := VerifyEditValidatorStakingMsg(st.state, st.bc, editValidator)
```
rlan35: Same here. Can be VerifyAndEdit...... and return the edited validator object, so you don't have to do it again after.

core/state_transition.go
```diff
@@ -438,167 +426,98 @@ func (st *StateTransition) applyEditValidatorTx(
...
 
-func (st *StateTransition) applyDelegateTx(delegate *staking.Delegate) error {
-	if delegate.Amount.Sign() == -1 {
-		return errNegativeAmount
-	}
-
-	if !st.state.IsValidator(delegate.ValidatorAddress) {
-		return errValidatorNotExist
+func (st *StateTransition) verifyAndApplyDelegateTx(delegate *staking.Delegate) error {
+	err := VerifyDelegateStakingMsg(st.state, delegate)
```
rlan35: similar. We basically don't want to duplicate the complicated validator object manipulation logic twice, it's error-prone. Just do it once and use the modified object here to commit.

core/state_transition.go
```diff
@@ -438,167 +426,98 @@ func (st *StateTransition) applyEditValidatorTx(
...
+	if err == nil {
+		st.state.SubBalance(delegate.DelegatorAddress, delegate.Amount)
 	}
+	return err
+}
 
-	if !st.state.IsValidator(undelegate.ValidatorAddress) {
-		return errValidatorNotExist
+func (st *StateTransition) verifyAndApplyUndelegateTx(undelegate *staking.Undelegate) error {
+	err := VerifyUndelegateStakingMsg(st.state, st.evm.EpochNumber, undelegate)
```
rlan35: Same.

core/state_transition.go
```diff
@@ -438,167 +426,98 @@ func (st *StateTransition) applyEditValidatorTx(
...
-			if delegation.Reward.Cmp(big.NewInt(0)) > 0 {
-				totalRewards.Add(totalRewards, delegation.Reward)
-			}
-
-			delegation.Reward.SetUint64(0)
-		}
-
-		err = st.state.UpdateStakingInfo(wrapper.Validator.Address, wrapper)
-		if err != nil {
+	updatedValidatorWrappers, totalRewards, err := PreProcessCollectRewardsFromDelegation(st.state, delegations)
```
rlan35: Yup. this is how it should work. Just do the modification once.

core/types/tx_pool.go
```diff
@@ -0,0 +1,102 @@
...
+	Hash() common.Hash
+	Nonce() uint64
+	ChainID() *big.Int
+	ToShardID() uint32
```
rlan35: where is ShardID() getter?

rlan35: For staking txn, there is no shard for it. But it implicitly means shard 0.

core/staking_verifier.go
Daniel-VDM: okay will return an error

core/types/tx_pool.go
Daniel-VDM: yes, though it was used in tx pool, but just check and it does not. Will remove this requirement of the interface and remove the dummy implementation of the method in `StakingTransaction`

core/state_transition.go
Daniel-VDM: alright, refactored the checks to return components that need to be updated in the state

core/state_transition.go
```diff
@@ -373,232 +366,68 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
...
-
-	if newRate.GT(wrapper.Validator.MaxRate) {
-		return errCommissionRateChangeTooHigh
-	}
-
-	if err := st.state.UpdateStakingInfo(wrapper.Address, wrapper); err != nil {
+	if err := st.state.UpdateStakingInfo(wrapper.Validator.Address, wrapper); err != nil {
 		return err
 	}
+	st.state.SubBalance(delegate.DelegatorAddress, delegate.Amount)
```
rlan35: This shouldn't be reducing by the whole delegate.Amount. The logic is this: to get the token for delegation, we first go into the list of undelegated tokens and use them first. If undelegated token are all used for delegation, then we use the account balance for the rest. So here it should be whatever is left in the "delegateBalance" in the original logic.

core/staking_verifier.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	return updatedValidatorWrappers, totalRewards, nil
+}
+
+// getPendingEpoch returns the epoch of the next block given a chain
+func getPendingEpoch(chain interface{ CurrentBlock() *types.Block }) *big.Int {
+	currentBlock := chain.CurrentBlock()
+	pendingBlockNumber := new(big.Int).Add(currentBlock.Number(), big.NewInt(1))
+	blocksPerEpoch := new(big.Int).SetUint64(
+		shard.Schedule.InstanceForEpoch(currentBlock.Epoch()).BlocksPerEpoch(),
```
rlan35: actually, don't add custom func like this. Just use the existing util "shard.Schedule.IsLastBlock". It will tell whether it's the last block of an epoch. From IsLastBlock, you can see the epoch calculation is not that straightforward due to some legacy nuances.

core/types/tx_pool.go
rlan35: but don't you still need to check the shardID in txPool and remove those txns that's not for my shard?

core/state_transition.go
Daniel-VDM: Oh I see, okay will return the balance to be deducted from the delegator in `VerifyAndDelegateFromMsg`

core/staking_verifier.go
Daniel-VDM: oh alright, corrected for it.

core/types/tx_pool.go
rlan35: How is this resolved? (Don't just resolve a comment without a clear closure).

Daniel-VDM: the current tx pool doesn't need shardIDs. Will implement this (and possibly Cx as well) in future PR.

PR #2244, opened 2020-02-11
State prune

rlan35:
Add system signal handling and node shut down process to dump state in memory to db.

====== WALLET BALANCES ======
    Balance in Shard 0:  71.809313809, nonce: 10
    Balance in Shard 1:  77.142647142, nonce: 10
====== RESULTS ======
PASS
^CGot interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...
Got interrupt signal. Dumping state to DB...


cmd/harmony/main.go
```diff
@@ -636,5 +637,25 @@ func main() {
 		go currentNode.CollectMetrics()
 	}
 
+	// Prepare for graceful shutdown from os signals
+	osSignal := make(chan os.Signal)
+	signal.Notify(osSignal, os.Interrupt, os.Kill)
```
LeoHChen: let's handle syscall.SIGTERM as well.

core/blockchain.go
```diff
@@ -555,6 +557,11 @@ func (bc *BlockChain) repair(head **types.Block) error {
 				Msg("Rewound blockchain to past state")
 			return nil
 		}
+		// Repair last commit sigs
```
LeoHChen: any reason of doing this?

node/node.go
```diff
@@ -711,3 +711,13 @@ func (node *Node) SetSyncFreq(syncFreq int) {
 func (node *Node) SetBeaconSyncFreq(syncFreq int) {
 	node.beaconSyncFreq = syncFreq
 }
+
+// ShutDown gracefully shut down the node server and dump the in-memory blockchain state into DB.
+func (node *Node) ShutDown() {
```
LeoHChen: do we need to flush the disk?

cmd/harmony/main.go
rlan35: Done.

core/blockchain.go
rlan35: Because when the chain is rolled back (repair) the last commit sig needs to be updated accordingly for consensus to work.

node/node.go
rlan35: The Blockchain().Stop() will flush the state to disk, and also wait for all blockchain operations to fully finish (there is a wait group)

cmd/harmony/main.go
```diff
@@ -636,5 +638,25 @@ func main() {
 		go currentNode.CollectMetrics()
 	}
 
+	// Prepare for graceful shutdown from os signals
+	osSignal := make(chan os.Signal)
+	signal.Notify(osSignal, os.Interrupt, os.Kill, syscall.SIGTERM)
```
denniswon: In addition to signals, os.Interrupt, os.Kill, syscall.SIGTERM, there are:
`syscall.SIGHUP, syscall.SIGQUIT, syscall.SIGTERM, syscall.SIGINT` (from this example of handling graceful exit https://github.com/apache/dubbo-go/issues/86)
Do we also want to include these?

core/blockchain.go
```diff
@@ -180,7 +180,7 @@ func NewBlockChain(db ethdb.Database, cacheConfig *CacheConfig, chainConfig *par
 	if cacheConfig == nil {
 		cacheConfig = &CacheConfig{
 			TrieNodeLimit: 256 * 1024 * 1024,
-			TrieTimeLimit: 5 * time.Minute,
+			TrieTimeLimit: 2 * time.Minute,
```
denniswon: why decreased to 2 minutes?
This means we flush to disk more frequently than before, thus less aggressive pruning.
Why not do more aggressive pruning with larger trie time limit since we have less number of transactions and states history changes less?

PR #2249, opened 2020-02-12
multi-bls key support for staking

gupadhyaya:


consensus/leader.go
```diff
@@ -55,21 +61,20 @@ func (consensus *Consensus) announce(block *types.Block) {
...
-	if err := consensus.prepareBitmap.SetKey(
-		consensus.PubKey, true,
-	); err != nil {
-		consensus.getLogger().Warn().Err(err).Msg(
-			"[Announce] Leader prepareBitmap SetKey failed",
+	for i, key := range consensus.PubKey.PublicKey {
+		consensus.Decider.SubmitVote(
+			quorum.Prepare,
+			key,
+			consensus.priKey.PrivateKey[i].SignHash(consensus.blockHash[:]),
```
rlan35: how can we make sure the consensus.PubKey.PublicKey and consensus.priKey.PrivateKey are consistent in terms of orders? Better refactor them into a class? Or just always derive pubkey from priKey.

consensus/quorum/one-node-one-vote.go
```diff
@@ -120,10 +120,12 @@ func (v *uniformVoteWeight) AmIMemberOfCommitee() bool {
 	}
 	identity, _ := pubKeyFunc()
 	everyone := v.DumpParticipants()
-	myVoterID := identity.SerializeToHexStr()
```
rlan35: AmIMemberOfCommitee us called for every message received. Better make it efficient.

node/node_metrics.go
```diff
@@ -43,12 +43,14 @@ func (node *Node) UpdateTxPoolSizeForMetrics(txPoolSize uint64) {
 
 // UpdateBalanceForMetrics uppdates node balance for metrics service.
 func (node *Node) UpdateBalanceForMetrics() {
-	curBalance, err := node.GetBalanceOfAddress(node.Consensus.SelfAddress)
-	if err != nil {
-		return
+	for _, addr := range node.Consensus.SelfAddress {
```
rlan35: SelfAddress => SelfAddresses

consensus/leader.go
gupadhyaya: when we read the multi-bls keys, we append private key and corresponding public key at once, hence order is unlikely to mess-up unless `GetPublicKey` from private key that is being loaded panics. after loading, the struct is never modified and only read, hence it is safe to assume that indices based access will always be consistent. Further, it may not be a good idea to derive pub key from private every time, as it an expensive c call to bls library. do you agree with my reasoning?

consensus/quorum/one-node-one-vote.go
gupadhyaya: you mean store the serialized hex str also part of multi-bls struct for public key?

consensus/leader.go
rlan35: I agree with the expensive part. But assuming the order is fine may not be good. Let's add a TODO to enforce the order by making pri/pub key a tuple.

consensus/quorum/one-node-one-vote.go
rlan35: It's a double for loop. If len(everyone) = 250. It's very expensive.  Better just use map lookup. At least avoid nested loop

cmd/harmony/main.go
```diff
@@ -114,7 +116,7 @@ var (
 	// Use a separate log file to log libp2p traces
 	logP2P = flag.Bool("log_p2p", false, "log libp2p debug info")
 
-	initialAccount = &genesis.DeployAccount{}
+	initialAccounts = make([]*genesis.DeployAccount, 0)
```
fxfactorial: why this change, not needed

fxfactorial: rather, make the literal slice rather than make

cmd/harmony/main.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/harmony-one/harmony/internal/genesis"
 	hmykey "github.com/harmony-one/harmony/internal/keystore"
 	"github.com/harmony-one/harmony/internal/memprofiling"
+	"github.com/harmony-one/harmony/internal/multibls"
```
fxfactorial: no, let's not put this into internal

cmd/harmony/main.go
```diff
@@ -101,6 +102,7 @@ var (
 	enableMemProfiling = flag.Bool("enableMemProfiling", false, "Enable memsize logging.")
 	enableGC           = flag.Bool("enableGC", true, "Enable calling garbage collector manually .")
 	blsKeyFile         = flag.String("blskey_file", "", "The encrypted file of bls serialized private key by passphrase.")
+	blsFolder          = flag.String("blsfolder", ".hmy/blskeys", "The folder that stores the bls keys; same blspass is used to decrypt all bls keys; all bls keys mapped to same shard")
```
fxfactorial: hmy is a unix world binary, please use directory, not folder

cmd/harmony/main.go
```diff
@@ -230,39 +252,80 @@ func setupStakingNodeAccount() error {
...
-	pubKey := consensusPriKey.GetPublicKey()
 
-	// Consensus keys are the BLS12-381 keys used to sign consensus messages
-	nodeConfig.ConsensusPriKey, nodeConfig.ConsensusPubKey = consensusPriKey, consensusPriKey.GetPublicKey()
-	if nodeConfig.ConsensusPriKey == nil || nodeConfig.ConsensusPubKey == nil {
-		fmt.Println("error to get consensus keys.")
-		os.Exit(100)
+	for _, blsKeyFile := range blsKeyFiles {
+		if filepath.Ext(blsKeyFile.Name()) != ".key" {
+			continue
```
fxfactorial: print out what it found instead

cmd/harmony/main.go
```diff
@@ -230,39 +252,80 @@ func setupStakingNodeAccount() error {
...
+		}
+		blsKeyFilePath := path.Join(*multiBlsKeyDir, blsKeyFile.Name())
+		consensusPriKey, err := blsgen.LoadBlsKeyWithPassPhrase(blsKeyFilePath, blsPassphrase) // uses the same bls passphrase for multiple bls keys
+		if err != nil {
+			return err
+		}
+		multibls.AppendPriKey(consensusMultiBlsPriKey, consensusPriKey)
```
fxfactorial: call should just be called Append and it takes the union of these two calls

cmd/harmony/main.go
```diff
@@ -230,39 +252,80 @@ func setupStakingNodeAccount() error {
...
+		multibls.AppendPubKey(consensusMultiBlsPubKey, consensusPriKey.GetPublicKey())
+	}
+
+	return nil
+}
+
+func setupConsensusKey(nodeConfig *nodeconfig.ConfigType) multibls.PublicKey {
+	var consensusMultiPriKey multibls.PrivateKey
```
fxfactorial: prefer literal, `consensusMultiPriKey := &multibls.PrivateKey{}`

cmd/harmony/main.go
```diff
@@ -317,12 +380,14 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
 	currentConsensus.Decider.SetShardIDProvider(func() (uint32, error) {
```
fxfactorial: we should refactor this dependency injection into a proper interface, not cbs

cmd/harmony/main.go
```diff
@@ -317,12 +380,14 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
...
 	})
-	currentConsensus.Decider.SetMyPublicKeyProvider(func() (*bls.PublicKey, error) {
+	currentConsensus.Decider.SetMyPublicKeyProvider(func() (*multibls.PublicKey, error) {
 		return currentConsensus.PubKey, nil
 	})
 
-	if initialAccount.Address != "" { // staking validator doesn't have to specify ECDSA address
-		currentConsensus.SelfAddress = common.ParseAddr(initialAccount.Address)
+	// staking validator doesn't have to specify ECDSA address
+	currentConsensus.SelfAddress = make(map[string]ethCommon.Address)
```
fxfactorial: prefer literal, `map[string]ethCommon.Address{}`

consensus/consensus.go
```diff
@@ -173,11 +174,26 @@ func (consensus *Consensus) GetBlockReward() *big.Int {
...
 }
 
+// GetLeaderPrivateKey returns leader private key if node is the leader
+func (consensus *Consensus) GetLeaderPrivateKey(leaderKey *bls.PublicKey) (*bls.SecretKey, error) {
+	for i, key := range consensus.PubKey.PublicKey {
+		if key.IsEqual(leaderKey) {
+			return consensus.priKey.PrivateKey[i], nil
+		}
+	}
+	return nil, fmt.Errorf("leader private key not found")
```
fxfactorial: why this Errorf if not args, just make package level value of error proper , then wrap it with errors.Wrap with the extra info being which leader key was requested, being looked for

consensus/consensus_service.go
```diff
@@ -61,13 +62,10 @@ var (
 )
 
 // Signs the consensus message and returns the marshaled message.
-func (consensus *Consensus) signAndMarshalConsensusMessage(
-	message *msg_pb.Message,
-) ([]byte, error) {
-	if err := consensus.signConsensusMessage(message); err != nil {
+func (consensus *Consensus) signAndMarshalConsensusMessage(message *msg_pb.Message, priKey *bls.SecretKey) ([]byte, error) {
```
fxfactorial: I prefer it how it was, in genearl try to limit wide to 100 columns max

consensus/consensus_service_test.go
```diff
@@ -37,9 +38,8 @@ func TestPopulateMessageFields(t *testing.T) {
 			Consensus: &msg_pb.ConsensusRequest{},
 		},
 	}
-	consensusMsg := consensus.populateMessageFields(
-		msg.GetConsensus(), consensus.blockHash[:],
-	)
+
+	consensusMsg := consensus.populateMessageFields(msg.GetConsensus(), consensus.blockHash[:], blsPriKey.GetPublicKey())
```
fxfactorial: same, prefer this style rather than long, over 100 cols

consensus/consensus_v2.go
```diff
@@ -109,8 +109,13 @@ func (consensus *Consensus) finalizeCommits() {
 		Int64("NumCommits", consensus.Decider.SignersCount(quorum.Commit)).
 		Msg("[Finalizing] Finalizing Block")
 	beforeCatchupNum := consensus.blockNum
+	leaderPriKey, err := consensus.GetConsensusLeaderPrivateKey()
+	if err != nil {
+		consensus.getLogger().Warn().Err(err).Msg("[FinalizeCommits] leader not found")
```
fxfactorial: `.Warn` does not make sense here, its `.Error()`

cmd/harmony/main.go
gupadhyaya: what would be the appropriate location?

cmd/harmony/main.go
gupadhyaya: need to check with @LeoHChen if he has assumed default to be `.hmy/blskeys` for the current mainnet upgrade.

cmd/harmony/main.go
gupadhyaya: will move it under `github.com/harmony-one/harmony/multibls`

consensus/quorum/one-node-one-vote.go
gupadhyaya: This code won't be used after staking, instead `func (v *stakedVoteWeight) AmIMemberOfCommitee() bool ` which is already made efficient. Do we still need to update this code?

consensus/leader.go
gupadhyaya: also, consensus does not expose the priKey, only PubKey is exposed. Making them into a tuple may leak information?

cmd/harmony/main.go
gupadhyaya: not changing for now as @LeoHChen deploy scripts hard code this.

cmd/harmony/main.go
gupadhyaya: can you elaborate?

gupadhyaya:
@rlan35 @fxfactorial @LeoHChen ready to merge?

PR #2253, opened 2020-02-12
[double-sign] Provide proof of double sign in slash record sent to beaconchain

fxfactorial:
Large PR which implements slashing as result of a double-signing event.


cmd/harmony/main.go
```diff
@@ -92,7 +92,7 @@ var (
 	// beaconSyncFreq indicates beaconchain sync frequency
 	beaconSyncFreq = flag.Int("beacon_sync_freq", 60, "unit in seconds")
 	// blockPeriod indicates the how long the leader waits to propose a new block.
-	blockPeriod    = flag.Int("block_period", 8, "how long in second the leader waits to propose a new block.")
+	blockPeriod    = flag.Int("block_period", 2, "how long in second the leader waits to propose a new block.")
```
rlan35: revert this when deploy to network.

consensus/leader.go
```diff
@@ -179,67 +194,94 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
...
+	// NOTE let it handle its own log
+	if !consensus.isRightBlockNumAndViewID(recvMsg) {
 		return
 	}
 
 	consensus.mutex.Lock()
 	defer consensus.mutex.Unlock()
 
+	if key := (bls.PublicKey{}); consensus.couldThisBeADoubleSigner(recvMsg) {
+		if alreadyCastBallot := consensus.Decider.ReadBallot(
```
rlan35: This is just checking whether there is a matching commit in the current round, right? What if the double sign is on a previous block that's already in the chain.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,251 @@
...
+	slashTrack *Application,
+) *staking.ValidatorWrapper {
+	for _, delegator := range wrapper.Delegations {
+		discounted := numeric.NewDecFromBigInt(
+			delegator.Amount,
+		).Mul(rate).TruncateInt()
+		half := numeric.NewDecFromBigInt(
+			new(big.Int).Sub(delegator.Amount, discounted),
+		).Mul(fiftyPercent).TruncateInt()
+		state.SubBalance(wrapper.Address, half)
```
rlan35: The actual slashing should not be  from the validator's balance, but from the validator's Delegation.Amount and Undelegation.Amount. The calculation of how much token to slash should be based on the staked token at the epoch of double signing. So it should be based on the validator snapshot of that specific epoch when the double sign happened.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,251 @@
...
+	wrapper *staking.ValidatorWrapper,
+	rate numeric.Dec,
+	state *state.DB,
+	reporter common.Address,
+	slashTrack *Application,
+) *staking.ValidatorWrapper {
+	// Kick them off forever
+	wrapper.Banned = true
+	// What is the post reduced amount
+	discounted := numeric.NewDecFromBigInt(
```
rlan35: We don't need to do the slashing separately for delegator and validator, it's generic process for all of them. Validator's stake is the first Delegation (wrapper.delegations[0]). We can just apply all the changes together. Also The MinSelfDelegation is not relevant here.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,301 @@
...
+		// that trusted the validator with some one token, they must also
+		// be slashed so they have skin in the game
+		rest := snapshot.Delegations[mustBeValidatorsDelegationToSelf:]
+		if err := delegatorSlashApply(
+			rest, rate, state, slash.Reporter, slash.Evidence.Epoch, slashDiff,
+		); err != nil {
+			return nil, err
+		}
+		// ...and I want those that I have not seen before
+		sinceSnapshotDelegations :=
```
rlan35: Why slash the new delegators in current epoch? Their stake are not used in the voting power yet. So their tokens are not dirty.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,301 @@
...
+	doubleSignEpoch *big.Int,
+	slashTrack *Application,
+) error {
+	for _, delegator := range delegations {
+		amt := numeric.NewDecFromBigInt(
+			delegator.Amount,
+		)
+		discounted := amt.Mul(rate).TruncateInt()
+		half := numeric.NewDecFromBigInt(
+			new(big.Int).Sub(delegator.Amount, discounted),
```
rlan35: This seems half of what's left after slashing. I think you want the half of the amount to slash.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,301 @@
...
+		).Mul(fiftyPercent)
+
+		paidOff := numeric.ZeroDec()
+
+		// NOTE Get the token trying to escape
```
rlan35: You wan to first check on the delegation.Amount and see if it's enough to slash. If not enough, then go to undelegation for the rest.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,301 @@
...
+			// such that epoch>= doubleSignEpoch should be slashable
+			if undelegate.Epoch.Cmp(doubleSignEpoch) == -1 {
+				continue
+			}
+			if paidOff.Equal(half) {
+				break
+			}
+			undelegate.Amount.SetInt64(0)
+			slashTrack.TotalSlashed.Add(slashTrack.TotalSlashed, undelegate.Amount)
+			paidOff = paidOff.Add(numeric.NewDecFromBigInt(undelegate.Amount))
```
rlan35: the undelegate.Amount is already set to 0 before.

rlan35: Also, the amount to add to paidOff shouldn't make paidOff greater than the amount to slash. Which means there should be some logic on the amount to add to paidOff, rather than simply using undelegate.Amount for the last part.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,301 @@
...
+		}
+
+		nowAccountState := state.GetStakingInfo(slash.Offender)
+
+		if nowAccountState == nil {
+			return nil, errors.New("cannot be nil code field in account state")
+		}
+
+		const mustBeValidatorsDelegationToSelf = 1
+		// Handle validator's delegation explicitly for sake of clarity
```
rlan35: don't need to handle validator's stake separately. Just do slashing on all Delegations[] together.

staking/slash/double-sign.go
fxfactorial: Why not use up as much undelegation as possible ? Otherwise it’s token that is escaping punishment

staking/slash/double-sign.go
fxfactorial: But I remember we had a conversation about the snapshot delegations vs current delegations. Can you elaborate again please what that issue was

staking/slash/double-sign.go
fxfactorial: I see, I need to use the delegation.Amount of now, not from snapshot

staking/slash/double-sign.go
```diff
@@ -1,52 +1,291 @@
...
+func delegatorSlashApply(
+	delegatedToValidator *staking.ValidatorWrapper,
+	rate numeric.Dec,
+	state *state.DB,
+	reporter common.Address,
+	doubleSignEpoch *big.Int,
+	slashTrack *Application,
+) error {
+	for _, delegationSnapshot := range delegatedToValidator.Delegations {
+		postSlashBal, slashDebt, halfOfSlashDebt := postSlashBalanceCurrentDelegator(
```
rlan35: This is not correct. when you calculate how much to slash, you should use the snapshot. Not the current state.

rlan35: And here, since you are using the current state, and applying the rate on it, postSlashBal will always be positive.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,291 @@
...
+			state, delegatedToValidator.Address,
+			delegationSnapshot.DelegatorAddress, rate,
+		)
+		if postSlashBal.Sign() == -1 {
```
rlan35: The right logic is to calculate the amount to slash based on the snapshot. And try to slash that amount in the current ValidatorWrapper in the current state. Don't modify the snapshot anyway.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,291 @@
...
+			stillOwe := new(big.Int).Sub(slashDebt, delegationSnapshot.Amount)
+			if wrapper := state.GetStakingInfo(
+				delegatedToValidator.Address,
+			); wrapper != nil {
+				for _, delegationNow := range wrapper.Delegations {
+					if delegationNow.Hash() == delegationSnapshot.Hash() {
+						for _, undelegate := range delegationNow.Undelegations {
+							// the epoch matters, only those undelegation
+							// such that epoch>= doubleSignEpoch should be slashable
+							if undelegate.Epoch.Cmp(doubleSignEpoch) == -1 {
```
rlan35: should be >=     or    == 1

staking/slash/double-sign.go
```diff
@@ -1,52 +1,291 @@
...
+								continue
+							}
+							if stillOwe.Cmp(common.Big0) <= 0 {
+								// paid off the slash debt
+								break
+							}
+							amtPostSlash, amtSlash, halfOfSlashAmt := applySlashRate(
```
rlan35: Don't need to calculate slash rate again, simply take out the amount up to stillOwe, one by one over each undelegation.

staking/slash/double-sign.go
fxfactorial: fixed

staking/slash/double-sign.go
```diff
@@ -145,66 +145,70 @@ var (
...
+				)
+				const (
+					haveEnoughToPayOff               = 1
+					paidOffExact                     = 0
+					debtCollectionsRepoUndelegations = -1
+				)
+				switch d := new(big.Int).Sub(nowAmt, halfOfSlashDebt); d.Sign() {
+				case haveEnoughToPayOff, paidOffExact:
+					slashTrack.TotalSlashed.Add(slashTrack.TotalSlashed, halfOfSlashDebt)
+					nowAmt.Sub(nowAmt, halfOfSlashDebt)
```
rlan35: should slash the whole debt amount. And the slashDebt == 0

staking/slash/double-sign.go
```diff
@@ -145,66 +145,70 @@ var (
...
+				state.AddBalance(reporter, halfOfSlashDebt)
+				slashTrack.TotalSnitchReward.Add(
+					slashTrack.TotalSnitchReward, halfOfSlashDebt,
+				)
+				const (
+					haveEnoughToPayOff               = 1
+					paidOffExact                     = 0
+					debtCollectionsRepoUndelegations = -1
+				)
+				switch d := new(big.Int).Sub(nowAmt, halfOfSlashDebt); d.Sign() {
```
rlan35: Should compare to slashDebt. halfOfSlashDebt is only used to credit reporter. Other than that, it's all about whole slashDebt.

fxfactorial: I see, so credit reporter is getting token out of thin air, the wording on slashing doc was a little ambiguous to me.

cmd/harmony/main.go
```diff
@@ -217,9 +217,21 @@ func setupLegacyNodeAccount() error {
 		_, initialAccount = genesisShardingConfig.FindAccount(pubKey.SerializeToHexStr())
 	}
 
-	if initialAccount == nil {
+	const crtValdidatorBLS = "be23bc3c93fe14a25f3533feee1cff1c60706845a4907c5df58bc19f5d1760bfff06fe7c9d1f596b18fdf529e0508e0a"
```
rlan35: clean up

cmd/trigger-double-sign/main.go
```diff
@@ -0,0 +1,25 @@
+package main
```
rlan35: put into /test if you want to keep it.

consensus/consensus.go
```diff
@@ -129,6 +129,8 @@ type Consensus struct {
 	lastBlockReward *big.Int
 	// Have a dedicated reader thread pull from this chan, like in node
 	SlashChan chan slash.Record
+	// only during testing
+	DoDoubleSign bool
```
rlan35: cleanup

consensus/consensus_service.go
```diff
@@ -213,14 +195,25 @@ func (consensus *Consensus) ResetState() {
...
+
+	return fmt.Sprintf(
+		"[Duty:%s Pub:%s ShardID:%v Header:%s Num:%d View:%d Shard:%d Epoch:%d DoubleSign:%v]",
+		duty,
+		consensus.PubKey.SerializeToHexStr(),
+		consensus.ShardID,
+		hex.EncodeToString(consensus.blockHeader),
+		consensus.blockNum,
+		consensus.viewID,
+		consensus.ShardID,
```
rlan35: shardID logged twice.

consensus/construct.go
```diff
@@ -33,11 +45,22 @@ func (consensus *Consensus) construct(
...
+	// then let the blockhash be the default off of consensus
+	if len(payloadForSignOverride) == 0 {
+		consensusMsg = consensus.populateMessageFields(
+			message.GetConsensus(), consensus.blockHash[:],
+		)
+	} else {
+		consensusMsg = consensus.populateMessageFields(
+			message.GetConsensus(), blockHash[:],
+		)
+	}
```
rlan35: should you also change the "case msg_pb.MessageType_COMMIT:" part?

core/blockchain.go
```diff
@@ -2211,36 +2211,44 @@ func (bc *BlockChain) ReadShardLastCrossLink(shardID uint32) (*types.CrossLink,
...
+	cls := slash.Records{}
 	if err := rlp.DecodeBytes(bytes, &cls); err != nil {
 		utils.Logger().Error().Err(err).Msg("Invalid pending slashing candidates RLP decoding")
 		return nil, err
 	}
 	return cls, nil
 }
 
+var (
+	// ErrPreStakingCRUDSlash ..
```
rlan35: why called CRUD?

cmd/trigger-double-sign/main.go
fxfactorial: how do you build into binary if put under test? i mean, will go_exectuable_build.sh build it and put into bin?

consensus/construct.go
fxfactorial: i think it currently is handled correctly

hmy/api_backend.go
```diff
@@ -398,6 +400,9 @@ func (b *APIBackend) GetTotalStakingSnapshot() *big.Int {
 	stakes := big.NewInt(0)
 	for i := range candidates {
 		validator, _ := b.hmy.BlockChain().ReadValidatorInformation(candidates[i])
+		if !effective.IsEligibleForEPOSAuction(validator) {
```
rlan35: Why add this condition here? What is "GetTotalStakingSnapshot" used for?

internal/params/protocol_params.go
```diff
@@ -6,7 +6,7 @@ const (
 	// MinGasLimit ...
 	MinGasLimit uint64 = 5000 // Minimum the gas limit may ever be.
 	// GenesisGasLimit ...
-	GenesisGasLimit uint64 = 4712388 // Gas limit of the Genesis block.
+	GenesisGasLimit uint64 = 10712388 // Gas limit of the Genesis block.
```
rlan35: if not reverted. Better add a separate TestGenesisGasLimit because the original gasLimit can be changed for the mainnet genesis.

hmy/api_backend.go
fxfactorial: because otherwise you are counting the stake of banned or active=false validators for the median stake

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
-	r.Signed.Header = signedHeader
-	r.Signed.Signature = signedSignature
-	r.DoubleSigned.Header = doubleSignedHeader
-	r.DoubleSigned.Signature = doubleSignedSignature
-	r.Beneficiary = beneficiary
-	return r
-}
-
-// TODO(Edgar) Implement Verify and Apply
+	ConflictingBallots
```
rlan35: ConflictingBallots is part of evidence

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+			errShardIDNotKnown, "given shardID %d", candidate.Evidence.ShardID,
+		)
+	}
+
+	if _, err := subCommittee.AddressForBLSKey(second.SignerPubKey); err != nil {
+		return err
+	}
+	// candidate.ConflictingBallots
+
+	// TODO Why this one printng have 00000000 for signature? something wrong earlier
```
rlan35: If these todos are launch blocking, better create issues for better tracking.

rlan35: put under "Staking" project.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+			log.Err(errSlashDebtNotFullyAccountedFor).
+				Str("snapshot-rlp", hex.EncodeToString(x1)).
+				Str("current-rlp", hex.EncodeToString(x2))
+			return errors.Wrapf(errSlashDebtNotFullyAccountedFor, "amt %v", slashDebt)
+		}
+	}
 	return nil
 }
+
+// TODO Need to keep a record in off-chain db of all the slashes?
```
rlan35: please create an issue under staking project.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+	chain staking.ValidatorSnapshotReader, state *state.DB,
+	slashes Records, rate numeric.Dec,
+) (*Application, error) {
+	slashDiff := &Application{big.NewInt(0), big.NewInt(0)}
+	for _, slash := range slashes {
+		// TODO Probably won't happen but we probably should
+		// be expilict about reading the right epoch validator snapshot,
+		// because it needs to be the epoch of which the double sign
+		// occurred
+		snapshot, err := chain.ReadValidatorSnapshot(
```
rlan35: With my latest code ReadValidatorSnapshot will be epoch specific.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+		current := state.GetStakingInfo(slash.Offender)
+		if current == nil {
+			addr := common2.MustAddressToBech32(slash.Offender)
+			return nil, errors.Wrapf(
+				errValidatorNotFoundDuringSlash, "lookup %s", addr,
+			)
+		}
+		// NOTE invariant: first delegation is the validators own
+		// stake, rest are external delegations.
+		// Bottom line: everyone must have have skin in the game,
```
rlan35: "have have", better => everyone will be slashed under the same rule.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+			if nowAmt := delegationNow.Amount; delegationNow.DelegatorAddress == snapshotAddr {
+				state.AddBalance(reporter, halfOfSlashDebt)
+				// NOTE only need to pay snitch here
+				slashTrack.TotalSnitchReward.Add(
+					slashTrack.TotalSnitchReward, halfOfSlashDebt,
+				)
+				// NOTE handle validator self delegation special case
+				if i == validatorsOwnDel && j == validatorsOwnDel {
+					// say my slash debt is 1.6, and my current amount is 1.2
+					// then while I can't drop below 1, I could still pay 0.2
```
rlan35: We should just change the sanityCheck to take into account the Banned field. Having complicated logic here to stay above 1 or minSelfDelegation is not necessary.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+
+const (
+	haveEnoughToPayOff               = 1
+	paidOffExact                     = 0
+	debtCollectionsRepoUndelegations = -1
+	validatorsOwnDel                 = 0
 )
 
+// invariant assumes snapshot, current can be rlp.EncodeToBytes
+func mustStayAboveZeroAfterDebtReduce(
```
rlan35: rename to "payDebtWithSanityCheck" or just "payDebt". Better use verb for a func that make changes.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+								undelegate.Amount.SetInt64(0)
+							case haveEnoughToPayOff:
+								undelegate.Amount.Sub(undelegate.Amount, slashDebt)
+								mustStayAboveZeroAfterDebtReduce(
+									snapshot, current, slashDebt, slashDebt, slashTrack,
+								)
+								undelegate.Amount.Set(newBal)
+							case -1:
+								// But do have enough to pay off something at least
+								partialPayment := new(big.Int).Sub(slashDebt, new(big.Int).Abs(newBal))
```
rlan35: This partial payment is just the whole undelegation.Amount. No need for the calculation.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+								break
+							}
+
+							switch newBal := new(big.Int).Sub(
+								// My slash debt is 1.6 and my undelegate amount is 1.0
+								// so if (1.6 - 1.0) > 0
+								undelegate.Amount,
+								slashDebt); newBal.Cmp(common.Big0) {
+							case paidOffExact:
+								undelegate.Amount.SetInt64(0)
```
rlan35: Better add the "mustStayAboveZeroAfterDebtReduce" here too. And basically, paidOffExact and haveEnoughToPayOff, can be merged into the same logic to reduce complexity in the code.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+					for _, undelegate := range delegationNow.Undelegations {
+						// the epoch matters, only those undelegation
+						// such that epoch>= doubleSignEpoch should be slashable
+						if undelegate.Epoch.Cmp(doubleSignEpoch) >= 0 {
+							if slashDebt.Cmp(common.Big0) <= 0 {
+								// paid off the slash debt
+								break
+							}
+
+							if undelegate.Amount.Cmp(common.Big0) == 1 &&
```
rlan35: The logic here is oversimplied. The newBal could be negative. Actually the one above for validator is correct.

rlan35: And I think we don't need to separate out validator/delegator slashing. After taking care of Banned field in sanity check, the two cases can be handled with the same logic.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,385 @@
...
+									snapshot, current, slashDebt, slashDebt, slashTrack,
+								)
+							}
+						}
+					}
+				}
+			}
+		}
+
+		if slashDebt.Cmp(common.Big0) != 0 {
```
rlan35: The amount of token slashable may not be enough for slashing. So the actual slashed amount can be < slashDebt. In that case, we should take the actual slashed amount for the truth, both for accounting and for the reporter awards.

staking/types/validator.go
```diff
@@ -85,25 +125,27 @@ type ValidatorStats struct {
...
-	MinSelfDelegation *big.Int
+	MinSelfDelegation *big.Int `json:"min-self-delegation"`
 	// maximum total delegation allowed
-	MaxTotalDelegation *big.Int
-	// Is the validator active in participating committee selection process or not
-	Active bool
+	MaxTotalDelegation *big.Int `json:"max-total-delegation"`
+	// Is the validator active in participating
+	// committee selection process or not
+	Active bool `json:"eligible-for-epos"`
```
rlan35: let's use the original wording to be consistent.

staking/types/validator.go
```diff
@@ -85,25 +125,27 @@ type ValidatorStats struct {
...
 	// commission parameters
 	Commission
 	// description for the validator
 	Description
 	// CreationHeight is the height of creation
-	CreationHeight *big.Int
+	CreationHeight *big.Int `json:"creation-height"`
 	// Banned records whether this validator is banned from the network because they double-signed
-	Banned bool
+	Banned bool `json:"banned-from-network"`
```
rlan35: "banned" is good enough.

staking/types/validator.go
fxfactorial: but the active field was making people confused and it doesn't capture the semantic implication that active is with respect to being able to be in epos auction

staking/slash/double-sign.go
```diff
@@ -1,52 +1,341 @@
...
+	if subCommittee == nil {
+		return errors.Wrapf(
+			errShardIDNotKnown, "given shardID %d", candidate.Evidence.ShardID,
+		)
+	}
+
+	if _, err := subCommittee.AddressForBLSKey(second.SignerPubKey); err != nil {
+		return err
+	}
+	// TODO need to finish this implementation
```
rlan35: Also need to check the validator is not slashed/banned already, to prevent repeated acceptance of same slash.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,341 @@
...
+	snapshot, current *staking.ValidatorWrapper,
+	rate numeric.Dec,
+	state *state.DB,
+	reporter common.Address,
+	doubleSignEpoch *big.Int,
+	slashTrack *Application,
+) error {
+	for _, delegationSnapshot := range snapshot.Delegations {
+		slashDebt := applySlashRate(delegationSnapshot.Amount, rate)
+		halfOfSlashDebt := new(big.Int).Div(slashDebt, common.Big2)
```
rlan35: Again, this halfOfSlashDebt should the half of what's actually taken away from the stake, not the calculated amount to pay, because they may not have enough to pay.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,341 @@
...
+				for _, undelegate := range delegationNow.Undelegations {
+					// the epoch matters, only those undelegation
+					// such that epoch>= doubleSignEpoch should be slashable
+					if undelegate.Epoch.Cmp(doubleSignEpoch) >= 0 {
+						if slashDebt.Cmp(common.Big0) <= 0 {
+							// paid off the slash debt
+							break
+						}
+
+						if undelegate.Amount.Cmp(common.Big0) == 1 &&
```
rlan35: This logic is still wrong. There should be two cases. One is debt >= undelegation.amount, then take out the whole amount and remove it, go to next undelegation. Otherwise, take the "debt" amount on the undelegation and break.

consensus/leader.go
```diff
@@ -173,73 +183,105 @@ func (consensus *Consensus) onPrepare(msg *msg_pb.Message) {
...
+
+						now := big.NewInt(time.Now().UnixNano())
+
+						go func(reporter common.Address) {
+							evid := slash.Evidence{
+								ConflictingBallots: slash.ConflictingBallots{
+									*alreadyCastBallot,
+									votepower.Ballot{
+										offender,
+										recvMsg.BlockHash,
```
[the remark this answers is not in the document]
fxfactorial: right, for next PR

node/node_genesis.go
```diff
@@ -89,7 +90,7 @@ func (node *Node) SetupGenesisBlock(db ethdb.Database, shardID uint32, myShardSt
 	default: // all other types share testnet config
 		// Test accounts
 		node.AddTestingAddresses(genesisAlloc, TestAccountNumber)
-
+		gasLimit = params.TestGenesisGasLimit
```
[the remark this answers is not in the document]
fxfactorial: This one you told me to add, it’s only for testnets, see in main

consensus/leader.go
```diff
@@ -173,73 +181,105 @@ func (consensus *Consensus) onPrepare(msg *msg_pb.Message) {
...
+										common.Hex2Bytes(doubleSign.SerializeToHexStr()),
+									}},
+								Moment: slash.Moment{
+									Epoch:        curHeader.Epoch(),
```
rlan35: make sure the epoch/height/viewid/shardID are the exact ones of the double sign. Recording the current ones is not useful. So maybe use those from "signed" and "recvMsg"

core/blockchain.go
```diff
@@ -2014,20 +2016,27 @@ func (bc *BlockChain) WritePendingCrossLinks(crossLinks []types.CrossLink) error
...
+		if err != nil {
+			return 0, err
+		}
 		return 1, err
 	}
-	cls = append(cls, *candidate)
-	err = bc.WritePendingSlashingCandidates(cls)
-	return len(cls), err
 
+	cls = append(cls, candidates...)
```
rlan35: no dedup?

core/staking_verifier.go
```diff
@@ -55,9 +55,10 @@ func VerifyAndCreateValidatorFromMsg(
 	wrapper.Delegations = []staking.Delegation{
 		staking.NewDelegation(v.Address, msg.Amount),
 	}
-	wrapper.Snapshot.Epoch = epoch
-	wrapper.Snapshot.NumBlocksSigned = big.NewInt(0)
-	wrapper.Snapshot.NumBlocksToSign = big.NewInt(0)
+	zero := big.NewInt(0)
+	wrapper.LastEpochInCommittee = epoch
```
rlan35: don't change LastEpochInCommittee. It's wrong here. LastEpochInCommittee is 0 if the validator is never selected in committee (new validator)

internal/chain/engine.go
```diff
@@ -291,37 +291,111 @@ func (e *engineImpl) Finalize(
...
+		superCommittee, err := chain.ReadShardState(chain.CurrentHeader().Epoch())
+
+		if err != nil {
+			return nil, nil, errors.New("could not read shard state")
+		}
+
+		totalCountStakedValidatrs, _, _, _ := superCommittee.StakedValidators()
+		// Apply the slashes, invariant: assume been verified as legit slash by this point
+		var slashApplied *slash.Application
+		rate := slash.Rate(caught, totalCountStakedValidatrs)
```
rlan35: The rate should be num(double_signing_bls_keys) / num(total_external_bls_keys). Not on the number of validators.

node/node_genesis.go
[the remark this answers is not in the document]
rlan35: cool

node/worker/worker.go
```diff
@@ -300,10 +306,56 @@ func (w *Worker) IncomingReceipts() []*types.CXReceiptsProof {
 	return w.current.incxs
 }
 
+// CollectAndVerifySlashes ..
+func (w *Worker) CollectAndVerifySlashes() error {
+	allSlashing, err := w.chain.ReadPendingSlashingCandidates()
```
rlan35: where do you remove the one that's proposed from pending?

staking/slash/double-sign.go
rlan35: This is not done yet.

staking/slash/double-sign.go
```diff
@@ -1,52 +1,403 @@
...
+				// NOTE Assume did as much as could above, now check the undelegations
+				for _, undelegate := range delegationNow.Undelegations {
+					// the epoch matters, only those undelegation
+					// such that epoch>= doubleSignEpoch should be slashable
+					if undelegate.Epoch.Cmp(doubleSignEpoch) >= 0 {
+						if slashDebt.Cmp(common.Big0) <= 0 {
+							l.Msg("paid off the slash debt")
+							break
+						}
+						nowAmt := undelegate.Amount
```
rlan35: add a todo to remove the delegation that has 0 amount after paying debt.

node/worker/worker.go
fxfactorial: i remove them all in writeblockwithstate, 

```
			if err := bc.UpdateBlockRewardAccumulator(
				batch, payout, block.Number().Uint64(),
			); err != nil {
				return NonStatTy, err
			}
			if err := bc.DeletePendingSlashingCandidates(); err != nil {
				return NonStatTy, err
			}
```

staking/slash/double-sign.go
fxfactorial: like doing an undelegate?

internal/chain/engine.go
fxfactorial: ah, right right

core/blockchain.go
fxfactorial: I think technically not needed since I clear out all the slashing candidates in end writeblockwithstate

cmd/harmony/main.go
```diff
@@ -338,18 +349,7 @@ func setupConsensusAndNode(nodeConfig *nodeconfig.ConfigType) *node.Node {
...
-		config, err := slash.NewDoubleSignWebHooksFromPath(p)
-		if err != nil {
-			fmt.Fprintf(
-				os.Stderr, "ERROR provided yaml path %s but yaml found is illegal", p,
-			)
-			os.Exit(1)
-		}
-		currentNode.NodeConfig.WebHooks.DoubleSigning = config
-	}
+	currentNode := node.New(myHost, currentConsensus, chainDBFactory, blacklist, true)
```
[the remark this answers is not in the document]
stephen-tse: https://github.com/harmony-one/harmony/pull/2514/files @fxfactorial

PR #2261, opened 2020-02-13
Add fork logic for added receipt logs

rlan35:
The receipt log was added and released to mainnet, which changed the bloom filter and receipt root calculation and it's not compatible with existing mainnet blocks.

Tested on mainnet node.


LeoHChen:
not needed.

PR #2267, opened 2020-02-13
Shutdown the node program for sys interrupt too

rlan35:
Our validators are used to use Ctrl+c to shut down the node. We should keep that behavior as is.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -645,7 +645,7 @@ func main() {
 		for {
 			select {
 			case sig := <-osSignal:
-				if sig == os.Kill || sig == syscall.SIGTERM || sig == os.Interrupt{
+				if sig == os.Kill || sig == syscall.SIGTERM || sig == os.Interrupt {
 					fmt.Printf("Got %s signal. Gracefully shutting down...\n", sig)
 					currentNode.ShutDown()
 				}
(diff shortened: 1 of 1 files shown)


cmd/harmony/main.go
```diff
@@ -645,15 +645,10 @@ func main() {
 		for {
 			select {
 			case sig := <-osSignal:
-				if sig == os.Kill || sig == syscall.SIGTERM {
+				if sig == os.Kill || sig == syscall.SIGTERM || sig == os.Interrupt {
 					fmt.Printf("Got %s signal. Gracefully shutting down...\n", sig)
```
fxfactorial: Very little time available in interrupt , do the print after

rlan35: Interrupt won't kill the program either way. Better to print the message early to indicate what's happening.

fxfactorial: the trying to catch on os.Kill is wrong. You can't handle a signal handler on os.kill (signal -9), doesn't work on linux.

rlan35:
> better add a user defined signal to handle the node shutdown gracefully. SO, during the rolling update, we can send the USER signal to node program before we stop it. That can leave enough time for node to handle the disk flush.

Good idea. Can you add an issue for it, combined with the script change.

LeoHChen:
> > better add a user defined signal to handle the node shutdown gracefully. SO, during the rolling update, we can send the USER signal to node program before we stop it. That can leave enough time for node to handle the disk flush.
> 
> Good idea. Can you add an issue for it, combined with the script change.

https://github.com/harmony-one/harmony/issues/2271

@Daniel-VDM

PR #2279, opened 2020-02-18
Refactor offchain data commit; Make block onchain/offchain commit atomic

rlan35:
====== WALLET BALANCES ======
    Balance in Shard 0:  71.38074238, nonce: 10
    Balance in Shard 1:  76.714075714, nonce: 10
====== RESULTS ======
PASS


core/rawdb/accessors_chain_test.go
```diff
@@ -145,11 +145,11 @@ func TestBlockStorage(t *testing.T) {
 	} else if types.DeriveSha(types.Transactions(entry.Transactions())) != types.DeriveSha(block.Transactions()) || types.CalcUncleHash(entry.Uncles()) != types.CalcUncleHash(block.Uncles()) {
 		t.Fatalf("Retrieved body mismatch: have %v, want %v", entry, block.Body())
 	}
-	if actual, err := ReadEpochBlockNumber(db, big.NewInt(0)); err != nil {
-		t.Fatalf("Genesis epoch block number not found, error=%#v", err)
-	} else if expected := big.NewInt(0); actual.Cmp(expected) != 0 {
-		t.Fatalf("Genesis epoch block number mismatch: have %v, want %v", actual, expected)
-	}
+	//if actual, err := ReadEpochBlockNumber(db, big.NewInt(0)); err != nil {
```
gupadhyaya: why are these tests not required?

core/blockchain.go
```diff
@@ -2432,58 +2132,11 @@ func (bc *BlockChain) CXMerkleProof(toShardID uint32, block *types.Block) (*type
 	return proof, nil
 }
 
-// LatestCXReceiptsCheckpoint returns the latest checkpoint
-func (bc *BlockChain) LatestCXReceiptsCheckpoint(shardID uint32) uint64 {
```
gupadhyaya: There are no longer needed?

rlan35: Yea, checked with Chao. it's not actually used after it's firstly designed. We went to another route.

core/rawdb/accessors_chain_test.go
rlan35: Because I removed the epoch number DB commit. It's originally added for resharding, but it's not actually used anywhere yet.

PR #2292, opened 2020-02-21
S3 receiptlog epoch

LeoHChen:


core/block_validator.go
```diff
@@ -32,9 +32,6 @@ import (
 	"github.com/harmony-one/harmony/internal/params"
 )
 
-// Beacon chain block 1213181 is a one-off block with empty receipts which is expected to be non-empty.
-const beaconBadBlock = 1213181
```
denniswon: can you explain why removed?

rlan35:
Thanks picking this up. Can you add the epoch fork logic in the chain config, the same way as CrossLink or Staking? We should put all forking in the same place. The sharding config should only be used for shard structures from now on.

LeoHChen:
> Thanks picking this up. Can you add the epoch fork logic in the chain config, the same way as CrossLink or Staking? We should put all forking in the same place. The sharding config should only be used for shard structures from now on.

Yes. Just changed.

core/block_validator.go
LeoHChen: use a new way.

PR #2308, opened 2020-02-26
only broadcast valid transactions

rlan35:
Previously we blindly re-broadcast all received txns from RPC no matter it's valid or spammer txns. This pr correct that to only rebroadcast the txn if the current node think it's valid (not being a aid to spammers)

Tested locally, it shielded the spamming effect from other nodes.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -278,12 +278,12 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
 }
 
 // Add new transactions to the pending transaction list.
-func (node *Node) addPendingTransactions(newTxs types.Transactions) {
+func (node *Node) addPendingTransactions(newTxs types.Transactions) []error {
 	poolTxs := types.PoolTransactions{}
 	for _, tx := range newTxs {
 		poolTxs = append(poolTxs, tx)
 	}
-	node.TxPool.AddRemotes(poolTxs)
+	errs := node.TxPool.AddRemotes(poolTxs)
 
 	pendingCount, queueCount := node.TxPool.Stats()
 	utils.Logger().Info().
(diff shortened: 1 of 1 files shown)


LeoHChen:
@rlan35 , Please cherry-pick this change to s3 as well.

node/node.go
```diff
@@ -278,56 +278,69 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
...
 		Int("length of newTxs", len(newTxs)).
 		Int("totalPending", pendingCount).
 		Int("totalQueued", queueCount).
 		Msg("Got more transactions")
+	return errs
 }
 
 // Add new staking transactions to the pending staking transaction list.
-func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTransactions) {
+func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTransactions) []error {
```
denniswon: why aren't we not broadcasting txns here?

node/node.go
```diff
@@ -278,56 +278,69 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
...
 }
 
 // AddPendingStakingTransaction staking transactions
 func (node *Node) AddPendingStakingTransaction(newStakingTx *staking.StakingTransaction) {
 	if node.NodeConfig.ShardID == shard.BeaconChainShardID {
-		node.addPendingStakingTransactions(staking.StakingTransactions{newStakingTx})
+		errs := node.addPendingStakingTransactions(staking.StakingTransactions{newStakingTx})
+		for i := range errs {
+			if errs[i] != nil {
+				return
```
denniswon: if there is at least one error, we don't broadcast any txns?

denniswon: ah nevermind. was confused because was a slice of errors, but its just one transaction.
Better to use if errs[0] != nil { return }

node/node.go
```diff
@@ -278,56 +278,69 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
...
-	node.tryBroadcastStaking(newStakingTx)
 }
 
 // AddPendingTransaction adds one new transaction to the pending transaction list.
 // This is only called from SDK.
 func (node *Node) AddPendingTransaction(newTx *types.Transaction) {
 	if newTx.ShardID() == node.NodeConfig.ShardID {
-		node.addPendingTransactions(types.Transactions{newTx})
+		errs := node.addPendingTransactions(types.Transactions{newTx})
+		for i := range errs {
```
denniswon: ditto Better to use if errs[0] != nil { return }

node/node.go
```diff
@@ -278,56 +278,69 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
...
+			if errs[i] != nil {
+				return
+			}
+		}
+		utils.Logger().Info().Str("Hash", newTx.Hash().Hex()).Msg("Broadcasting Tx")
```
denniswon: we should put this log line above before early error return so that log also prints for error case.

node/node.go
```diff
@@ -278,56 +278,69 @@ func (node *Node) tryBroadcastStaking(stakingTx *staking.StakingTransaction) {
...
 		for _, tx := range newStakingTxs {
 			poolTxs = append(poolTxs, tx)
 		}
-		node.TxPool.AddRemotes(poolTxs)
+		errs := node.TxPool.AddRemotes(poolTxs)
 		pendingCount, queueCount := node.TxPool.Stats()
 		utils.Logger().Info().
 			Int("length of newStakingTxs", len(poolTxs)).
 			Int("totalPending", pendingCount).
 			Int("totalQueued", queueCount).
```
denniswon: nit: add err txns counts for log print

node/node.go
denniswon: well I guess addPendingStakingTransactions() is only used from addPendingStakingTransaction, so no need. nvm. checked the code.

denniswon:
@rlan35 good catch. My comments are all nits. No need to update as a followup. can just take care of them on other related commits.

PR #2313, opened 2020-02-26
Do not rebroadcast if the txn is not accepted into txpool

rlan35:
Cherry pick from master.


LeoHChen:
we need to do a stress test for this change.  @Daniel-VDM , @SebastianJ , maybe use LRTN or another STN for s3.

PR #2315, opened 2020-02-27
change active flag to pointer to string instead of bool

gupadhyaya:
active flag was pointer to bool which could not be edited. changing it to pointer to string. since this is a data structure change, a new binary needs to be released so that go-sdk can be updated.

### DIFF
--- a/staking/types/validator.go
+++ b/staking/types/validator.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"strconv"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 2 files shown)


staking/types/validator.go
```diff
@@ -480,8 +481,12 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
 		}
 	}
 
-	if edit.Active != nil {
-		validator.Active = *edit.Active
+	if edit.Active != nil && *edit.Active != "" {
+		val, err := strconv.ParseBool(*edit.Active)
```
fxfactorial: Log this error and what was the crappy input

PR #2316, opened 2020-02-27
Fix last continuous crosslink logic

rlan35:
The last continuous crosslink logic doesn't work anymore after the batch update of offchain data.

Fixed by enforcing the rolling up of last continuous crosslink.

Tested locally

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1873,7 +1873,7 @@ func (bc *BlockChain) LastContinuousCrossLink(batch rawdb.DatabaseWriter, shardI
 		return err
 	}
 	newLink := oldLink
-	for i := oldLink.BlockNum() + 1;; i++ {
+	for i := oldLink.BlockNum() + 1; ; i++ {
 		tmp, err := bc.ReadCrossLink(shardID, i)
 		if err == nil && tmp != nil && tmp.BlockNum() == i {
 			newLink = tmp
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1873,6 +1873,7 @@ func (bc *BlockChain) LastContinuousCrossLink(batch rawdb.DatabaseWriter, shardI
 		return err
 	}
 	newLink := oldLink
+	// Starting from last checkpoint, keeping reading immediate next crosslink until there is a gap
 	for i := oldLink.BlockNum() + 1; ; i++ {
 		tmp, err := bc.ReadCrossLink(shardID, i)
 		if err == nil && tmp != nil && tmp.BlockNum() == i {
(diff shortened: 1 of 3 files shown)


core/blockchain.go
```diff
@@ -1863,38 +1863,28 @@ func (bc *BlockChain) ReadCrossLink(shardID uint32, blockNum uint64) (*types.Cro
...
+	if oldLink == nil || err != nil {
 		return err
 	}
-	newCheckpoint := uint64(0)
-	for i := cl0.BlockNum() + 1; i < cl.BlockNum(); i++ {
-		tmp, err := bc.ReadCrossLink(cl.ShardID(), i)
-		if err != nil || tmp == nil {
-			newCheckpoint = i - 1
+	newLink := oldLink
+	for i := oldLink.BlockNum() + 1; ; i++ {
```
fxfactorial: please comment explaining this loop logic

core/offchain.go
```diff
@@ -164,13 +164,21 @@ func (bc *BlockChain) CommitOffChainData(
...
+				rawdb.WriteShardLastCrossLink(batch, crossLink.ShardID(), crossLink.Serialize())
+			}
 		}
 
 		//clean/update local database cache after crosslink inserted into blockchain
 		num, err := bc.DeleteCommittedFromPendingCrossLinks(*crossLinks)
 		utils.Logger().Debug().Msgf("DeleteCommittedFromPendingCrossLinks, crosslinks in header %d,  pending crosslinks: %d, error: %+v", len(*crossLinks), num, err)
 	}
+	// Roll up latest crosslinks
+	for i := uint32(0); i < shard.Schedule.InstanceForEpoch(epoch).NumShards(); i++ {
```
fxfactorial: not sure if `i < shard.Schedule.InstanceForEpoch(epoch).NumShards()` will be called just once (otherwise many needless calls), maybe put as variable right above

internal/params/config.go
```diff
@@ -37,7 +37,7 @@ var (
 	TestnetChainConfig = &ChainConfig{
 		ChainID:         TestnetChainID,
 		CrossTxEpoch:    big.NewInt(0),
-		CrossLinkEpoch:  big.NewInt(3),
+		CrossLinkEpoch:  big.NewInt(4),
 		StakingEpoch:    big.NewInt(4),
```
gupadhyaya: Should we have shorter PreStaking and Staking Epoch gap? With this configuration, we have to wait for ~1 hour for staking logic to kick in? May be have similar configuration as localnet?

fxfactorial: testnet block per epoch is 75, so about 30 min

core/blockchain.go

core/offchain.go
rlan35: It's a network param. It's already a variable behind the call.

rlan35:
> and of course the epoch param changes undo

The epoch change is intended so crosslink and staking epoch is the same.

PR #2323, opened 2020-02-28
[WIP] Unit test for core/staking_verifier.go

Renewwen:
### Unit test for `harmony/core/staking_verifier.go`
Already tested:
- CV1
- CV3
- CV4

### Reference:
staking test plan:
https://docs.google.com/spreadsheets/d/17kSRWWIqacO-trshVrW2EgH16eVjQdj1Tnx7G2Tx_GI/edit?ts=5e307ae1#gid=0

### DIFF
--- a/core/staking_verifier_test.go
+++ b/core/staking_verifier_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/harmony-one/harmony/crypto/hash"
 	common2 "github.com/harmony-one/harmony/internal/common"
 	"github.com/harmony-one/harmony/numeric"
+	"github.com/harmony-one/harmony/internal/ctxerror"
 	"github.com/harmony-one/harmony/shard"
 	staking "github.com/harmony-one/harmony/staking/types"
 )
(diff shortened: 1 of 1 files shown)


harmony-ek:

core/staking_verifier_test.go
```diff
@@ -0,0 +1,149 @@
...
+		MaxTotalDelegation: maxTotalDel,
+		SlotPubKeys:        slotPubKeys,
+		SlotKeySigs:        slotKeySigs,
+		Amount:             amount,
+	}
+	return &v
+}
+
+// Test CV1: create validator
+func Test_CV1(t *testing.T) {
```
fxfactorial: This is not idiomatic go, remove the underscores

core/staking_verifier_test.go
```diff
@@ -0,0 +1,149 @@
...
+	statedb.AddBalance(msg.ValidatorAddress, big.NewInt(5e18))
+	identitylengthError := ctxerror.New("[EnsureLength] Exceed Maximum Length", "have", len(msg.Identity), "maxIdentityLen", staking.MaxIdentityLength)
+	if _, err := VerifyAndCreateValidatorFromMsg(statedb, big.NewInt(0), big.NewInt(0), msg); err.Error() != identitylengthError.Error() {
+		t.Error("expected", identitylengthError, "got", err)
+	}
+}
+
+// Test CV6: website longer than 140 characters
+func Test_CV6(t *testing.T) {
+	statedb, _ := state.New(common.Hash{}, state.NewDatabase(ethdb.NewMemDatabase()))
```
fxfactorial: This is repeated over and over , write a function to create state and do the initial validator

Renewwen: I have told with ganesha about this issue. he thought these tasks could run parallel, so I need to create statedb and validator for every test to prevent concurrency things, even it looks like redundant.

core/staking_verifier_test.go
Renewwen: will fix that!

core/staking_verifier_test.go
gupadhyaya: if we want to run the tests in parallel and make the tests self contained, it is better to create the statedb freshly. @Renewwen you could create state once and do Reset whenever you need fresh state.

fxfactorial:
Jenkins, test this please

PR #2324, opened 2020-02-28
fix gas cost check for tx pool

gupadhyaya:
fixing the transaction pool cost estimation. before it was wrongly estimating the cost and leading to no error message for the cli.

### DIFF
--- a/staking/types/transaction.go
+++ b/staking/types/transaction.go
@@ -172,8 +172,31 @@ func (tx *StakingTransaction) GasPrice() *big.Int {
 }
 
 // Cost ..
-func (tx *StakingTransaction) Cost() *big.Int {
-	return new(big.Int).Mul(tx.data.Price, new(big.Int).SetUint64(tx.data.GasLimit))
+func (tx *StakingTransaction) Cost() (*big.Int, error) {
+	total := new(big.Int).Mul(tx.data.Price, new(big.Int).SetUint64(tx.data.GasLimit))
+	switch tx.StakingType() {
+	case DirectiveCreateValidator:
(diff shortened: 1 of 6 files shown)


staking/types/transaction.go
```diff
@@ -173,7 +173,30 @@ func (tx *StakingTransaction) GasPrice() *big.Int {
 
 // Cost ..
 func (tx *StakingTransaction) Cost() *big.Int {
-	return new(big.Int).Mul(tx.data.Price, new(big.Int).SetUint64(tx.data.GasLimit))
+	total := new(big.Int).Mul(tx.data.Price, new(big.Int).SetUint64(tx.data.GasLimit))
+	switch tx.StakingType() {
+	case DirectiveCreateValidator:
+		msg, err := RLPDecodeStakeMsg(tx.Data(), DirectiveCreateValidator)
+		if err != nil {
```
fxfactorial: Why give back nil, change signature of cost to have error as well

gupadhyaya: It requires changing the `Cost()` interface. Which is also consumed by the normal transaction that does not throw any error. Hence, relied on nil check, anyways the proper error is generated by the consumer of Cost.

fxfactorial: Then change the interface and in plain txn case it just won’t hit that case. Point of interface is to abstract well. As it is, it’s not idiomatic go

fxfactorial: But I understand time pressure and whatnot so maybe make issue or we let Daniel know to do it along the way for something else

gupadhyaya: updated.

PR #2342, opened 2020-02-28
[availability] Count availability for shard chain, more log, remove error of div by zero as error condition 

fxfactorial:
Fix active, inactive not working


internal/chain/engine.go
```diff
@@ -304,30 +307,24 @@ func (e *engineImpl) Finalize(
...
 		if staked.CountStakedValidator > 0 {
-			l.RawJSON("external", []byte(staked.StateSubset.String())).
-				Msg("have non-zero external")
+			l.Msg("have non-zero external")
 
 			if err != nil {
 				return nil, nil, err
 			}
 
 			l.Msg("bumping validator signing counts")
```
rlan35: Remove this log.

staking/availability/measure.go
```diff
@@ -126,6 +128,17 @@ func bumpCount(
 			continue
 		}
 
+		snapshot, err := bc.ReadValidatorSnapshot(addr)
+		if err != nil {
+			return err
+		}
+
+		if snapshot.LastEpochInCommittee.Cmp(common.Big0) == 0 {
```
rlan35: The snapshot is what the validator was from last epoch. If it's new validator in last epoch, it doesn't affect that we should still count his signing in this epoch if it is already selected in the committee.

staking/availability/measure.go
```diff
@@ -220,10 +239,10 @@ func SetInactiveUnavailableValidators(
 			continue
 		}
 
-		if d := new(big.Int).Sub(now, snapEpoch); d.Cmp(common.Big1) != 0 {
+		if d := new(big.Int).Sub(nowEpoch, snapEpoch); d.Cmp(common.Big1) != 0 {
```
rlan35: again, new validator could have snapEpoch==0. What is the reason to do this check?

staking/availability/measure.go
fxfactorial: Because I noticed that it started counting one epoch too early because you have all validators written in snapshot , and maybe did have this epoch as 0, so I prefer keeping this guard

staking/availability/measure.go
fxfactorial: Because it’s all elected validators but that isn’t strictly the ones in committee selection

staking/availability/measure.go
fxfactorial: actually you're right

internal/chain/engine.go
```diff
@@ -304,56 +307,42 @@ func (e *engineImpl) Finalize(
...
-			if err != nil {
-				return nil, nil, err
-			}
-
-			l.Msg("bumping validator signing counts")
-			if err := availability.IncrementValidatorSigningCounts(
-				chain, header, header.ShardID(), state, staked.LookupSet,
+		if isNewEpoch && staked.CountStakedValidator > 0 {
+			l.Msg("in new epoch (aka last block), apply availability check for activity")
+			if err := availability.SetInactiveUnavailableValidators(
```
rlan35: Seems it's too late to set inactive, those ones who should be kicked out are already proposed in shardState in the header. We should set inactive before shardState proposal

Daniel-VDM:
Jenkins, test this please

LeoHChen:
Jenkins, test this please.

Issue #2227, opened 2020-02-08
Explorer for staking network is not displaying the transaction history


sophoah:
**Describe the bug**
Accessing an account on the explorer doesn't show the transaction history https://explorer.os.hmny.io/#/address/one1mcverl6yrcsy0xdrmf3mltm28kk6kux3nwtvxd

**To Reproduce**
1. make transaction in and out from an account
2. access that account on the explorer : https://explorer.os.hmny.io/#/address/one1mcverl6yrcsy0xdrmf3mltm28kk6kux3nwtvxd

**Expected behavior**
see the list of transaction like in the current dev testnet https://explorer.testnet.harmony.one/#/address/one1zksj3evekayy90xt4psrz8h6j2v3hla4qwz4ur

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS: win10

**Additional context**
N/A

rlan35:
@mikedoan can you work with Vlad to make sure staking transactions are displayed in explorer.

denniswon:
@flicker-harmony what is the RPC endpoint to use for fetching staking txns? (we already have this right?) so just need to additionally fetch not just regular txns, but also staking txns to be displayed on the explorer side

denniswon:
done

PR #2343, opened 2020-02-28
Add and fix missing txn/staking txn rpc apis

denniswon:
## Issue

Fixes issue https://github.com/harmony-one/harmony/issues/2227
Explorer for staking network is not displaying the transaction history

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    NO

    NO. (There is no change in API name or args. Just addition of new api funcs)

    YES

    Harmony doc should be updated with the new api functions with explanations.
    Also, fixed some api function comments correcting the functionality/purpose of the function.
    cc: @Renewwen @Cem-Harmony could you help me update these once merged and released?

    NO

## TODO

    Check and add/fix missing or incorrect staking related RPC apis if needed.


node/node_explorer.go
```diff
@@ -176,3 +176,34 @@ func (node *Node) GetTransactionsHistory(address, txType, order string) ([]commo
...
+	if order == "DESC" {
+		sort.Slice(addressData.TXs[:], func(i, j int) bool {
+			return addressData.TXs[i].Timestamp > addressData.TXs[j].Timestamp
+		})
+	} else {
+		sort.Slice(addressData.TXs[:], func(i, j int) bool {
+			return addressData.TXs[i].Timestamp < addressData.TXs[j].Timestamp
+		})
+	}
+	hashes := make([]common.Hash, 0)
```
fxfactorial: prefer literal when don't have dynamic size, aka `hashes := []common.Hash{}`

hmy/api_backend.go
```diff
@@ -241,6 +240,12 @@ func (b *APIBackend) GetTransactionsHistory(address, txType, order string) ([]co
 	return hashes, err
 }
 
+// GetStakingTransactionsHistory returns list of staking transactions hashes of address.
+func (b *APIBackend) GetStakingTransactionsHistory(address, txType, order string) ([]common.Hash, error) {
+	hashes, err := b.hmy.nodeAPI.GetStakingTransactionsHistory(address, txType, order)
```
fxfactorial: might as well just `return b.hmy`.... the signatures are the same

denniswon: could you clarify your comment here?

fxfactorial: the signature of `b.hmy.nodeAPI.GetStakingTransactionsHistory` is the same as the RPC `GetStakingTransactionsHistory` so no point to hold onto intermediate variables

node/node_explorer.go
```diff
@@ -176,3 +176,34 @@ func (node *Node) GetTransactionsHistory(address, txType, order string) ([]commo
 	}
 	return hashes, nil
 }
+
+// GetStakingTransactionsHistory returns list of staking transactions hashes of address.
+func (node *Node) GetStakingTransactionsHistory(address, txType, order string) ([]common.Hash, error) {
+	addressData := &explorer.Address{}
```
gupadhyaya: If the code is for this function is exactly same as `GetTransactionsHistory`, can't you just call it?

fxfactorial: yes, no need for even further code duplication

hmy/api_backend.go
denniswon: i see. so no possibility they will/could become different between regular vs. staking txns, correct?

fxfactorial: of course its possible, its an open source project, anything can change. But  we can change when needed, rather than adds lines of code now.

LeoHChen:
Jenkins, test this please.

denniswon:
@flicker-harmony this PR fixes the open staking launch blocker issue #2227.
Need to merge this asap.
Will merge this first, and review your PR.

@fxfactorial @gupadhyaya please review.

denniswon:
"There was this issue on master which RJ said is alright if it tx history works on s3
On master it doesnt work because of staking changes in
node/node_explorer.go
Where quorum gets checked
It fails and so no blocks are committed to explorer storage
Edgar said he’d look into it because it was changes from his side
But seems like he didnt"

From @flicker-harmony 

@rlan35 @fxfactorial context please? so should I rebase to s3 or t3 instead? cc: @LeoHChen

PR #2356, opened 2020-03-02
Remove invalid crosslinks from pending; optimize block timer

rlan35:
Remove invalid/obsolete crosslinks from pending during block proposal.

Also optimize block interval timer so it includes the time spent on block proposal.


node/node_newblock.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // Constants of proposing a new block
 const (
-	PeriodicBlock         = 200 * time.Millisecond
+	PeriodicBlock         = 20 * time.Millisecond
```
gupadhyaya: What is the basis of reducing the period to 20ms from 200ms?

node/node_newblock.go
```diff
@@ -156,14 +160,20 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
-
+		invalidToDelete := []types.CrossLink{}
 		if err == nil {
 			for _, pending := range allPending {
-				if err = node.VerifyCrossLink(pending); err != nil {
-					continue
-				}
 				exist, err := node.Blockchain().ReadCrossLink(pending.ShardID(), pending.BlockNum())
 				if err == nil || exist != nil {
+					invalidToDelete = append(invalidToDelete, pending)
```
gupadhyaya: do we have some number on how many invalid cross links we are seeing from the log?

PR #2363, opened 2020-03-03
[build] hardcode libcrypto in md5sum.txt

LeoHChen:


SebastianJ:
Seems to be working:
```
$ make all
$ cat bin/md5sum.txt | grep libcrypto
771150db04267126823190c873a96e48  libcrypto.so.10```

PR #2365, opened 2020-03-03
[project] Never hold reference to global logger

fxfactorial:


denniswon:
@fxfactorial 
Screen Shot 2020-03-03 at 12 30 05 PM

Many different places where we hold logger ref. Should we change these too?

fxfactorial:
no, that's fine because that's the singleton, the issue (i think) is the call to Logger() must be fresh

PR #2372, opened 2020-03-04
[availability] Adjust design to set active field on each block

fxfactorial:
...after convo with RJ, I had to pull over at a starbucks in millbrae to bang out this much more elegant solution.

### DIFF
--- a/internal/chain/engine.go
+++ b/internal/chain/engine.go
@@ -20,7 +20,6 @@ import (
 	"github.com/harmony-one/harmony/multibls"
 	"github.com/harmony-one/harmony/shard"
 	"github.com/harmony-one/harmony/shard/committee"
-	"github.com/harmony-one/harmony/staking/availability"
 	"github.com/harmony-one/harmony/staking/slash"
 	staking "github.com/harmony-one/harmony/staking/types"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


internal/chain/engine.go
```diff
@@ -292,6 +296,7 @@ func (e *engineImpl) Finalize(
 				)
 				state.AddBalance(delegation.DelegatorAddress, totalWithdraw)
 			}
+			wrapper.LastEpochInCommittee = newShardState.Epoch
```
rlan35: This is not the right way. LastEpochInCommittee should only be set if the validator is in the shardState.

staking/availability/measure.go
```diff
@@ -186,14 +186,15 @@ type Reader interface {
...
 
-// SetInactiveUnavailableValidators sets the validator to
+// ComputeAndRecord sets the validator to
 // inactive and thereby keeping it out of
 // consideration in the pool of validators for
 // whenever committee selection happens in future, the
 // signing threshold is 66%
-func SetInactiveUnavailableValidators(
-	bc Reader, state *state.DB, addrs []common.Address,
+func ComputeAndRecord(
```
rlan35: Why not move the logic here into the IncrementValidatorSigningCounts? Then you don't need to do Read and UpdateValidatorWrapper twice.

internal/chain/engine.go
fxfactorial: Hmm, that’s how it was before I merged big slashing PR

staking/availability/measure.go
fxfactorial: Thought about it but it’s a separation of concerns. But sure can fold them both together

internal/chain/engine.go
rlan35: No, it's not, just see the deleted code on the left side, it was ranging over shardState.validators. Now it's ranging over all validators.

fxfactorial: You're right, just looked a bit closer on older version code: 

```
 git show 801b4b83e277bfc50600d448fed3774cd4864650~1:internal/chain/engine.go
```

PR #2385, opened 2020-03-05
[reward][slash] Skip over banned validators even if they signed blocks

fxfactorial:
As things now, say a validator does a double sign in block 2 of a new epoch, then they have a whole rest of blocks during the epoch under which they can get block reward.

This PR changes rewarding logic to skip over banned validators for rewards.


fxfactorial:
Actually we need resharding to properly address this issue

PR #2391, opened 2020-03-05
[auditing] Add more logs; checks before processing staking txns

rlan35:


node/node_newblock.go
```diff
@@ -89,23 +93,30 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
 	)
 
-	node.Worker.GetCurrentHeader().SetCoinbase(coinbase)
-
+	header := node.Worker.GetCurrentHeader()
 	// After staking, all coinbase will be the address of bls pub key
-	if header := node.Worker.GetCurrentHeader(); node.Blockchain().Config().IsStaking(header.Epoch()) {
-		addr := common.Address{}
+	if node.Blockchain().Config().IsStaking(header.Epoch()) {
 		blsPubKeyBytes := node.Consensus.LeaderPubKey.GetAddress()
```
gupadhyaya: This will fail in case of multi-bls scenario, when the node has multiple bls keys and corresponding addresses. Should we fetch the address from node.Consensus.SelfAddresses map?

node/node_newblock.go
```diff
@@ -145,19 +158,18 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
 	receiptsList := node.proposeReceiptsProof()
 	if len(receiptsList) != 0 {
 		if err := node.Worker.CommitReceipts(receiptsList); err != nil {
-			utils.Logger().Error().Err(err).Msg("[proposeNewBlock] cannot commit receipts")
+			return nil, err
```
gupadhyaya: what is the reason for this new behavior? failed to commit receipts stop from proposing new block?

node/worker/worker.go
```diff
@@ -317,6 +338,7 @@ func (w *Worker) CollectAndVerifySlashes() error {
 		// "could not verify slash", which should not return as err
 		// and therefore stop the block proposal
 		if allSlashing, err = w.VerifyAll(d); err != nil {
+			// TODO(audit): very slashes individually; do not return err if verify fails
```
gupadhyaya: typo "very" => "verify"

node/worker/worker.go
```diff
@@ -130,23 +130,44 @@ func (w *Worker) CommitTransactions(
...
+			// Check whether the tx is replay protected. If we're not in the EIP155 hf
+			// phase, start ignoring the sender until we do.
+			if tx.Protected() && !w.config.IsEIP155(w.current.header.Number()) {
+				utils.Logger().Info().Str("hash", tx.Hash().Hex()).Str("eip155Epoch", w.config.EIP155Epoch.String()).Msg("Ignoring reply protected transaction")
+				txs.Pop()
+				continue
+			}
+
+			// Start executing the transaction
+			w.current.state.Prepare(tx.Hash(), common.Hash{}, len(w.current.txs))
```
gupadhyaya: there is another Prepare() call above right? why is this needed?

node/node_newblock.go
rlan35: nop

node/node_newblock.go
rlan35: yep, if anything fail to commit, there is no point of continuing.

node/worker/worker.go
rlan35: It is needed for every transactions to setup the context.

node/worker/worker.go
rlan35: Correction coming in next PR.

PR #2396, opened 2020-03-05
[staking][validation][protocol] 

fxfactorial:
1. Limit BLS keys to make dynamic as 1/3 of external validator slots available
2. Remove poor designed API from last fall in quorum 
3. Implement leftover TODO from verification of slash during block proposal 
4. Fold active and banned field of validator wrapper into single field which is an enumeration 
5. Add extra data points in log for downstream data analysis of delegation scalability issue
6. More noisy logs removed
7. Explicit upper bound on pending slashes 
8. Remove invalid slashes 
9. Factor out inlined code with method calls
10. Keep ValidatorInformation RPC as meaningful as possible
11. Never allow banned validator to change EPOS status via edit-validator
12. Expand on many comments
13. Give back caller their chain ID when it disagrees in txn via RPC


fxfactorial:
Jenkins, test this please

consensus/consensus_service.go
```diff
@@ -533,7 +533,7 @@ func (consensus *Consensus) UpdateConsensusInformation() Mode {
 
 	// update public keys in the committee
 	oldLeader := consensus.LeaderPubKey
-	pubKeys := committee.WithStakingEnabled.GetCommitteePublicKeys(
+	pubKeys, _ := committee.WithStakingEnabled.GetCommitteePublicKeys(
```
LeoHChen: is there any error we shall log?

core/blockchain.go
```diff
@@ -1906,56 +1901,38 @@ func (bc *BlockChain) ReadShardLastCrossLink(shardID uint32) (*types.CrossLink,
...
 		return err
 	}
 	if err := rawdb.WritePendingSlashingCandidates(bc.db, bytes); err != nil {
 		return err
 	}
-	bc.pendingSlashingCandidates.Add(pendingSCCacheKey, bytes)
 	return nil
 }
 
+// DeleteFromPendingSlashingCandidates ..
```
LeoHChen: better comment please for this public function

core/genesis.go
```diff
@@ -310,9 +310,13 @@ func (g *Genesis) MustCommit(db ethdb.Database) *types.Block {
 		panic(err)
 	}
 	rawdb.WriteBlockRewardAccumulator(db, big.NewInt(0), 0)
-	data, _ := rlp.EncodeToBytes([]slash.Record{})
-	rawdb.WritePendingSlashingCandidates(db, data)
-
+	data, err := rlp.EncodeToBytes(slash.Records{})
```
LeoHChen: can we write a log and then panic? that will help people knows the issue.

core/offchain.go
```diff
@@ -130,24 +131,34 @@ func (bc *BlockChain) CommitOffChainData(
...
+				Msg("[insertChain/crosslinks] cross links are not sorted")
 			return NonStatTy, errors.New("proposed cross links are not sorted")
 		}
 		for _, crossLink := range *crossLinks {
 			// Process crosslink
-			if err := bc.WriteCrossLinks(batch, types.CrossLinks{crossLink}); err == nil {
-				utils.Logger().Info().Uint64("blockNum", crossLink.BlockNum()).Uint32("shardID", crossLink.ShardID()).Msg("[insertChain/crosslinks] Cross Link Added to Beaconchain")
+			if err := bc.WriteCrossLinks(
+				batch, types.CrossLinks{crossLink},
+			); err == nil {
```
LeoHChen: is this log info or debug? we have flooded the log already.

node/node.go
```diff
@@ -578,17 +578,16 @@ func New(host p2p.Host, consensusObj *consensus.Consensus,
 	// Broadcast double-signers reported by consensus
 	if node.Consensus != nil {
 		go func() {
-
 			for {
 				select {
 				case doubleSign := <-node.Consensus.SlashChan:
-					l := utils.Logger().Info().RawJSON("double-sign", []byte(doubleSign.String()))
-
+					utils.Logger().Info().
```
LeoHChen: Is it Info or Warn/Error?

node/worker/worker.go
```diff
@@ -306,31 +306,28 @@ func (w *Worker) IncomingReceipts() []*types.CXReceiptsProof {
 	return w.current.incxs
 }
 
-// CollectAndVerifySlashes ..
-func (w *Worker) CollectAndVerifySlashes() error {
-	allSlashing, err := w.chain.ReadPendingSlashingCandidates()
-	if err != nil {
-		return err
+// CollectVerifiedSlashes ..
+func (w *Worker) CollectVerifiedSlashes() error {
```
LeoHChen: better comment please

shard/committee/assignment.go
```diff
@@ -123,17 +125,47 @@ func eposStakedCommittee(
...
+	shardState.Shards = make([]shard.Committee, shardCount)
+	hAccounts := s.HmyAccounts()
+	shardHarmonyNodes := s.NumHarmonyOperatedNodesPerShard()
+
+	for i := 0; i < shardCount; i++ {
+		shardState.Shards[i] = shard.Committee{uint32(i), shard.SlotList{}}
+		for j := 0; j < shardHarmonyNodes; j++ {
+			index := i + j*shardCount
+			pub := &bls.PublicKey{}
+			if err := pub.DeserializeHexStr(hAccounts[index].BlsPublicKey); err != nil {
```
LeoHChen: is it possible index is out of bound? do you need to check it?

shard/committee/assignment.go
```diff
@@ -123,17 +125,47 @@ func eposStakedCommittee(
...
+			shardState.Shards[i].Slots = append(shardState.Shards[i].Slots, shard.Slot{
+				common2.ParseAddr(hAccounts[index].Address),
+				pubKey,
+				nil,
+			})
+		}
+	}
+
+	if stakedSlotsCount == 0 {
+		utils.Logger().Info().
```
LeoHChen: is it debug or info?

shard/committee/assignment.go
```diff
@@ -123,17 +125,47 @@ func eposStakedCommittee(
...
 		validator, err := stakerReader.ReadValidatorInformation(candidates[i])
 		if err != nil {
 			return nil, err
 		}
-		if !effective.IsEligibleForEPOSAuction(validator) {
-			utils.Logger().Info().
-				Int("staked-candidates", len(candidates)).
-				RawJSON("candidate", []byte(validator.String())).
-				Msg("validator not eligible for epos")
+		if validator.EPOSStatus != effective.Active {
```
LeoHChen: use a function to warp this check?  like you have IsEligibleForEPOSAuction before?

shard/committee/assignment.go
```diff
@@ -291,5 +300,12 @@ func (def partialStakingEnabled) Compute(
 	}
 	// Set the epoch of shard state
 	shardState.Epoch = big.NewInt(0).Set(epoch)
+	staked := shardState.StakedValidators()
+	utils.Logger().Info().
```
LeoHChen: Info or debug only?

shard/shard_state.go
```diff
@@ -397,13 +397,14 @@ func (c *Committee) BLSPublicKeys() ([]BlsPublicKey, error) {
 var (
 	// ErrValidNotInCommittee ..
 	ErrValidNotInCommittee = errors.New("slot signer not this slot's subcommittee")
-	errCommitteeNil        = errors.New("subcommittee is nil pointer")
+	// ErrCommitteeNil ..
```
LeoHChen: better comment please

staking/effective/eligible.go
```diff
@@ -1,10 +1,19 @@
 package effective
 
-import (
-	staking "github.com/harmony-one/harmony/staking/types"
-)
+// Eligibility ..
```
LeoHChen: better comment please for public type

staking/slash/double-sign.go
```diff
@@ -84,7 +86,8 @@ type Record struct {
 
 // Application ..
```
LeoHChen: better comment please.

staking/types/validator.go
```diff
@@ -78,6 +81,19 @@ type ValidatorWrapper struct {
 	Counters    counters
 }
 
+// Computed ..
```
LeoHChen: better comment.

staking/types/validator.go
```diff
@@ -277,7 +297,8 @@ func (w *ValidatorWrapper) SanityCheck() error {
 			errInvalidSelfDelegation, "no self delegation given at all",
 		)
 	default:
-		if !w.Banned && w.Delegations[0].Amount.Cmp(w.Validator.MinSelfDelegation) < 0 {
+		if w.EPOSStatus != effective.Banned &&
```
LeoHChen: can you a comment on this logic here to make it more understandble

staking/types/validator.go
```diff
@@ -168,7 +187,8 @@ func (v *Validator) SanityCheck() error {
 	}
 
 	// MinSelfDelegation must be >= 1 ONE
-	if !v.Banned && v.MinSelfDelegation.Cmp(big.NewInt(denominations.One)) < 0 {
+	if v.EPOSStatus != effective.Banned &&
```
LeoHChen: do we have a realistic number of MinDelegation?

staking/types/validator.go
```diff
@@ -159,6 +172,12 @@ func (v *Validator) SanityCheck() error {
 		return errNeedAtLeastOneSlotKey
 	}
 
+	if c := len(v.SlotPubKeys); c > MaxNumBLSKeys {
+		return errors.Wrapf(
+			errExcessiveBLSKeys, "have: %d allowed: %d", c, MaxNumBLSKeys,
```
LeoHChen: better wording in the error message?

shard/committee/assignment.go
fxfactorial: Ya I did but it became a circular import and annoying

staking/types/validator.go
fxfactorial: One one

consensus/consensus_service.go
```diff
@@ -504,8 +501,9 @@ func (consensus *Consensus) UpdateConsensusInformation() Mode {
 	if len(curHeader.ShardState()) > 0 {
 		// increase curEpoch by one if it's the last block
 		consensus.SetEpochNum(curEpoch.Uint64() + 1)
-		consensus.getLogger().Info().Uint64("headerNum", curHeader.Number().Uint64()).
-			Msg("[UpdateConsensusInformation] Epoch updated for nextEpoch curEpoch")
+		consensus.getLogger().Info().
+			Uint64("headerNum", curHeader.Number().Uint64()).
+			Msg("[Epoch updated for nextEpoch curEpoch")
```
rlan35: clean [

consensus/threshold.go
```diff
@@ -42,12 +44,16 @@ func (consensus *Consensus) didReachPrepareQuorum() error {
 	// so by this point, everyone has committed to the blockhash of this block
 	// in prepare and so this is the actual block.
 	for i, key := range consensus.PubKey.PublicKey {
-		consensus.Decider.SubmitVote(
+		if _, err := consensus.Decider.SubmitVote(
```
rlan35: We should be lenient on this. If the leader self vote is not successful, he should still try continue  the consensus with other validators.

core/offchain.go
rlan35: it is info. This log will be O(1) as hoorays

core/offchain.go
```diff
@@ -198,8 +203,14 @@ func (bc *BlockChain) CommitOffChainData(
 			); err != nil {
 				return NonStatTy, err
 			}
-			if err := bc.DeletePendingSlashingCandidates(); err != nil {
-				return NonStatTy, err
+			records := slash.Records{}
+			if s := header.Slashes(); len(s) > 0 {
+				if err := rlp.DecodeBytes(s, &records); err != nil {
```
rlan35: should be lenient on here too. If the slash can't be deleted, they should be deleted at next proposal anyway as they will be invalid/duplicate.

shard/committee/assignment.go
fxfactorial: no, this is original same code as been running from ~july 2019

shard/committee/assignment.go
fxfactorial: just info

consensus/threshold.go
fxfactorial: The only way it fails is on the Phase being unknown, which has never failed before

shard/committee/assignment.go
```diff
@@ -159,20 +159,22 @@ func eposStakedCommittee(
...
 		if err != nil {
 			return nil, err
 		}
-		if !effective.IsEligibleForEPOSAuction(validator) {
+
+		if validator.EPOSStatus != effective.Active {
 			continue
 		}
-		if err := validator.SanityCheck(); err != nil {
+		if err := validator.SanityCheck(maxBLSKey); err != nil {
```
rlan35: This input of MaxBLSKey and everywhere is too verbose. Why not directly add the check inside SanityCheck using "stakedSlots :=
		(instance.NumNodesPerShard() - instance.NumHarmonyOperatedNodesPerShard()) *
			int(instance.NumShards())" ? 

Then we don't need all these wiring.

core/offchain.go
fxfactorial: Hmm, so remove error return on both decode & DeleteFromPending?

shard/committee/assignment.go
fxfactorial: I thought about that too but its very awkward to grab a handle on the instance since I don't have the epoch , so I must provide some function parameter

staking/effective/eligible.go
```diff
@@ -1,10 +1,19 @@
...
+type Eligibility byte
 
-// IsEligibleForEPOSAuction ..
-func IsEligibleForEPOSAuction(v *staking.ValidatorWrapper) bool {
-	return v.Active && !v.Banned
-}
+const (
+	// Active means allowed in epos auction
+	Active Eligibility = iota
```
rlan35: Will 0 byte work in the case of no-op EditValidator. Basically how to differentiate no-op vs setting to Active?

fxfactorial: thought about that, just addressed: 6c19abe8ff689958c7ac6a697b405408b33c75a9

node/node.go
fxfactorial: info

core/offchain.go
rlan35: yup.

shard/committee/assignment.go
rlan35: ok

staking/effective/eligible.go
```diff
@@ -1,10 +1,22 @@
...
+// that occurs just once an epoch on beaconchain
+type Eligibility byte
 
-// IsEligibleForEPOSAuction ..
-func IsEligibleForEPOSAuction(v *staking.ValidatorWrapper) bool {
-	return v.Active && !v.Banned
-}
+const (
+	// Unknown is a default state that represents bad input
+	Unknown Eligibility = iota
```
rlan35: It shouldn't be a bad input. Better call it "Nil" to represent the no-op the nil pointer previously is meant to achieve.

staking/types/validator.go
```diff
@@ -510,8 +541,18 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
...
-		validator.Active = *edit.Active
+	switch validator.EPOSStatus {
+	case effective.Banned:
+		return errCannotChangeBannedTaint
+	default:
+		switch edit.EPOSStatus {
+		case effective.Active, effective.Inactive:
+			validator.EPOSStatus = edit.EPOSStatus
+		default:
+			return errors.Wrapf(
```
rlan35: When the value is 0. It should mean no-op (as in nil pointer before). Don't return error.

PR #2400, opened 2020-03-05
Check for EditValidator.active null or not

coolcottontail:
## Issue
In RLP encoding, If input is non-value (uint(0), []byte{}, string(“”), empty pointer …), RLP encoding is 0x80. 

Scalar 0 :  big-endian interpretation of 0 consists of zeros - so after trimming, we get empty byte array. Empty byte array has length < 56, so it's length is encoded as 0+128 which is 0x80 in hex.

For EditValidator Active Field, if the value is pointer to false, then this field will be encoded to 0x80.
You can try the code below:

 var isActive bool = false
    msgEditValidator := EditValidator{...}; 
    msgEditValidator.isActive = &isActive;
  //encode to RLP
   rlpBytes, err := rlp.EncodeToBytes(msgEditValidator)
// RLP encoded value == 128    for *bool

 fmt.Println(rlpBytes[len(rlpBytes)-1])

 During decoding, the 0x80 will be decoded to nil.

So  the following code will fail  if edit.Active is actually mean to "false".
the RLP decoded validator.Active will be nil, instead of a pointer to "false"

if edit.Active != nil {
        validator.Active = *edit.Active
    }

### DIFF
--- a/staking/types/validator.go
+++ b/staking/types/validator.go
@@ -495,7 +495,11 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
 	}
 
 	if edit.Active != nil {
-		validator.Active = *edit.Active
+		if *edit.Active {
+			validator.Active = true
+		}
+	} else {
+		validator.Active = false
 	}
 
 	return nil


staking/types/validator.go
```diff
@@ -495,7 +495,11 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
 	}
 
 	if edit.Active != nil {
-		validator.Active = *edit.Active
+		if *edit.Active {
+			validator.Active = true
+		}
+	} else {
```
gupadhyaya: this is wrong right? if I don't want to set the Active field, it will reset to false

coolcottontail: The RLP encoded EditValidator.Active only will have two value 0x1 or 0x80:

1. default is 0x80,   
2. 0x80 or if *Active == false
3. 0x1 if  *Active == true , you want to set true.

the default value is same as the false value...

Unless we change to *int,  otherwise *bool can't have 3 stage value ...

fxfactorial:
I think I mentioned before , active needs to be an enumeration of possibility , not bool

fxfactorial:
will be addressed in https://github.com/harmony-one/harmony/pull/2396

PR #2404, opened 2020-03-06
Top addresses

flicker-harmony:
## Issue

Updated explorer service to collect top addresses and API /top for it.

## Test

Screenshot 2020-03-07 at 00 19 51

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

   NO

    NO

    NO

### DIFF
--- a/api/service/explorer/service.go
+++ b/api/service/explorer/service.go
@@ -4,16 +4,23 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"net"
 	"net/http"
+	"sort"
 	"strconv"
+	"time"
 
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/gorilla/mux"
 
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
+	internal_common "github.com/harmony-one/harmony/internal/common"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/p2p"
+	"github.com/harmony-one/harmony/shard"
 )
 
 // Constants for explorer service.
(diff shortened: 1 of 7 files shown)


api/service/explorer/service.go
```diff
@@ -127,6 +150,108 @@ func (s *Service) GetAddresses(w http.ResponseWriter, r *http.Request) {
...
+		if err != nil {
+			utils.Logger().Error().Err(err).Msg("error reading epoch")
+			continue
+		}
+		epoch := big.NewInt(0)
+		if err = rlp.DecodeBytes(bytes, &epoch); err != nil {
+			utils.Logger().Error().Err(err).Msg("cannot decode epoch")
+			continue
+		}
+		for curEpoch := prevEpoch + 1; curEpoch < epoch.Int64(); curEpoch++ {
```
rlan35: don't need to loop over shard state validator addresses. Just loop through all addresses in the trie.

flicker-harmony: @rlan35 what trie? Which method to use?

flicker-harmony:
Screenshot 2020-03-09 at 00 52 22

rlan35:
Jenkins, test this please.

api/service/explorer/service.go
```diff
@@ -127,6 +143,66 @@ func (s *Service) GetAddresses(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// UpdateTopAddresses updates 20 top addresses by balance from storage each 10 minutes.
+func (s *Service) UpdateTopAddresses() {
+	ticker := time.NewTicker(10 * time.Minute)
```
fxfactorial: you never close the ticker and always leak the resource.

api/service/explorer/service.go
```diff
@@ -127,6 +143,66 @@ func (s *Service) GetAddresses(w http.ResponseWriter, r *http.Request) {
...
+	utils.Logger().Info().Msg("start updating top addresses")
+	for range ticker.C {
+		db, err := s.State()
+		if err != nil {
+			utils.Logger().Error().Err(err).Msg("fetch state error")
+		}
+		it := trie.NewIterator(db.GetTrie().NodeIterator(nil))
+		topAddresses := make([]AddrBalance, 0)
```
fxfactorial: no purpose to use make when could use literal when size is 0

PR #2418, opened 2020-03-06
adding checks to prevent malicious leader behaviors

gupadhyaya:
1. validator in ViewChangeMode should not trigger another view change on conflicting announce message
2. Do not process blocks that are further than max block number from the current block
3. Do not process view change that have view id further than max view id number from the current view id
4. Add more checks to onNewView for handling receive message view id and sender key verification for new leader
5. force block header verification for syncing mode
6. refactor onViewChange and onNewView sanity checks into checks.go

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -7,6 +7,9 @@ import (
 	"github.com/harmony-one/harmony/shard"
 )
 
+// MaxBlockNumDiff limits the received block number to only 100 further from the current block number
+const MaxBlockNumDiff = 100
+
 func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
 	senderKey, err := consensus.verifySenderKey(msg)
 	if err != nil {
@@ -21,7 +24,10 @@ func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
 
 	if !senderKey.IsEqual(consensus.LeaderPubKey) &&
 		consensus.current.Mode() == Normal && !consensus.ignoreViewIDCheck {
-		consensus.getLogger().Warn().Msg("[OnPrepared] SenderKey not match leader PubKey")
+		consensus.getLogger().Warn().Msgf(
+			"[%s] SenderKey not match leader PubKey",
+			msg.GetType().String(),
+		)
 		return false
 	}
 
@@ -39,17 +45,22 @@ func (consensus *Consensus) leaderSanityChecks(msg *msg_pb.Message) bool {
 	senderKey, err := consensus.verifySenderKey(msg)
 	if err != nil {
 		if err == shard.ErrValidNotInCommittee {
-			consensus.getLogger().Info().Msg(
-				"[OnAnnounce] sender key not in this slot's subcommittee",
+			consensus.getLogger().Info().Msgf(
+				"[%s] sender key not in this slot's subcommittee",
+				msg.GetType().String(),
 			)
 		} else {
-			consensus.getLogger().Error().Err(err).Msg("[OnAnnounce] erifySenderKey failed")
+			consensus.getLogger().Error().Err(err).Msgf(
+				"[%s] verifySenderKey failed",
+				msg.GetType().String(),
+			)
 		}
 		return false
 	}
 	if err = verifyMessageSig(senderKey, msg); err != nil {
-		consensus.getLogger().Error().Err(err).Msg(
-			"[OnPrepare] Failed to verify sender's signature",
+		consensus.getLogger().Error().Err(err).Msgf(
+			"[%s] Failed to verify sender's signature",
+			msg.GetType().String(),
 		)
 		return false
 	}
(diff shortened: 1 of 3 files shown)


consensus/checks.go
```diff
@@ -166,3 +175,65 @@ func (consensus *Consensus) onPreparedSanityChecks(
 
 	return true
 }
+
+func (consensus *Consensus) viewChangeSanityCheck(msg *msg_pb.Message) bool {
+	senderKey, err := consensus.verifyViewChangeSenderKey(msg)
```
fxfactorial: Error handle before next line

consensus/checks.go
```diff
@@ -166,3 +175,65 @@ func (consensus *Consensus) onPreparedSanityChecks(
...
+	msgTypeName := msg.GetType().String()
+	if err != nil {
+		consensus.getLogger().Error().Err(err).Msgf("[%s] VerifySenderKey Failed", msgTypeName)
+		return false
+	}
+	if err = verifyMessageSig(senderKey, msg); err != nil {
```
fxfactorial: Always err :

consensus/checks.go
```diff
@@ -166,3 +184,70 @@ func (consensus *Consensus) onPreparedSanityChecks(
...
+			"[%s] Failed To Verify Sender's Signature",
+			msg.GetType().String(),
+		)
+		return false
+	}
+	return true
+}
+
+func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
+	// TODO: if difference is only one, new leader can still propose the same committed block to avoid another view change
```
fxfactorial: how relevant is this todo , what about associated formal github issue

gupadhyaya: this was old TODO, I didn't add. It is still relevant though.

fxfactorial: If still relevant then let’s make explicit issue for it

consensus/checks.go
```diff
@@ -166,3 +184,70 @@ func (consensus *Consensus) onPreparedSanityChecks(
...
+	}
+	if consensus.current.Mode() == ViewChanging &&
+		consensus.current.ViewID() > recvMsg.ViewID {
+		consensus.getLogger().Warn().
+			Uint64("MyViewChangingID", consensus.current.ViewID()).
+			Uint64("MsgViewChangingID", recvMsg.ViewID).
+			Msg("[onViewChange] ViewChanging ID Is Low")
+		return false
+	}
+	if recvMsg.ViewID-consensus.current.ViewID() > MaxViewIDDiff {
```
rlan35: is this the only new condition added?

gupadhyaya: for `onViewChange` sanity check, yes. I also added several other conditions:
* `onNewViewSanityCheck` is new
* moved common checks into `viewChangeSanityCheck`
* added `if consensus.current.Mode() == Normal || consensus.current.Mode() == Syncing {` in `onPreparedSanityChecks`
* added `} else if recvMsg.BlockNum-consensus.blockNum > MaxBlockNumDiff {` in `isRightBlockNumCheck`
* added 
```if consensus.current.Mode() == ViewChanging {
				consensus.getLogger().Debug().Msg(
					"[OnAnnounce] Already in ViewChanging mode, conflicing announce, doing noop",
				)
			}``` in `onAnnounceSanityChecks`
* and some parameterized log messages based on message types.

File: consensus/checks.go [hunk 5]
```diff
@@ -146,7 +164,7 @@ func (consensus *Consensus) onPreparedSanityChecks(
 			Msg("[OnPrepared] BlockHash not match")
 		return false
 	}
-	if consensus.current.Mode() == Normal {
+	if consensus.current.Mode() == Normal || consensus.current.Mode() == Syncing {
```
rlan35: @chaosma does this change look good to this issue? https://github.com/harmony-one/security-audit/issues/14

consensus/checks.go
```diff
@@ -166,3 +184,70 @@ func (consensus *Consensus) onPreparedSanityChecks(
...
+
+func (consensus *Consensus) onNewViewSanityCheck(recvMsg *FBFTMessage) bool {
+	if recvMsg.ViewID < consensus.current.ViewID() {
+		consensus.getLogger().Warn().
+			Uint64("MyViewChangingID", consensus.current.ViewID()).
+			Uint64("MsgViewChangingID", recvMsg.ViewID).
+			Msg("[onNewView] ViewChanging ID Is Low")
+		return false
+	}
+	if !recvMsg.SenderPubkey.IsEqual(consensus.LeaderPubKey) {
```
rlan35: this may not work because for the node that's out of sync with others in ViewID, he won't be able to accept a valid newView msg.

gupadhyaya: we can remove this check I think. As the code below handles the case where the claimed leader (newLeader or some validator) has not achieved quorum. 
```
if !consensus.Decider.IsQuorumAchievedByMask(m3Mask) {
		consensus.getLogger().Warn().
			Msgf("[onNewView] Quorum Not achieved")
		return
	}
```

rlan35: Yea, I actually asked Peckshield to revisit the bug. The solution could actually be to only accept newView when in viewChange mode.

gupadhyaya: removed the senderKey check against consensus.LeaderPubKey check.

consensus/checks.go
chaosma: @rlan35  In this case, it seems that we could add sanity check for all modes including view change. Also, according to the security audit, we could store the received message according to hash of the content instead of address.

PR #2423, opened 2020-03-07
[webhook][rpc][availability] Hooks, current signing, remove median-stake cached values

fxfactorial:
Further non-protocol changing improvements for OSTN

1. Do webhook call for when node drops in availabiltity signing
2. Remove accidential infinite recursion 
3. Never ignore errors in legacy RPCs
4. Expose bad block as an RPC, watchdog as downstream consumer
5. Remove signers that missed signing threshold from validatorlist, which is suspected of causing caching problems for median-stake.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -259,7 +259,7 @@ func (bc *BlockChain) ValidateNewBlock(block *types.Block) error {
 
 	// NOTE Order of mutating state here matters.
 	// Process block using the parent state as reference point.
-	receipts, cxReceipts, _, usedGas, _, err := bc.processor.Process(
+	receipts, cxReceipts, _, usedGas, _, _, err := bc.processor.Process(
 		block, state, bc.vmConfig,
 	)
 	if err != nil {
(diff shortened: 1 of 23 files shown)


core/blockchain.go
```diff
@@ -2442,6 +2489,20 @@ func (bc *BlockChain) UpdateStakingMetaData(
...
+		if active := editVal.Active; active != nil && *active {
+			list, err := bc.ReadValidatorList()
+			if err != nil {
+				return err
+			}
+			if list == nil {
+				list = []common.Address{}
+			}
+			list = utils.AppendIfMissing(list, editVal.ValidatorAddress)
+			if err := bc.WriteValidatorList(batch, list); err != nil {
```
rlan35: No removal when it turns to inactive? Then this part of the code is always no-ops, because the ValidatorList contains all the validators.

rlan35: nvm, I see it later.

core/offchain.go
```diff
@@ -187,3 +214,20 @@ func (bc *BlockChain) CommitOffChainData(
...
+		return err
+	}
+	eligible := []common.Address{}
+	for _, addr := range addrs {
+		if _, mia := missedThreshold[addr]; !mia {
+			eligible = append(eligible, addr)
+		}
+	}
+
+	return bc.WriteValidatorList(bc.db, eligible)
```
rlan35: If we remove those validators from the list, then their is no wait to iterate/list all validators anymore. Then the explorer won't be able to show them. Why do we need to change the ValidatorList? Can we do filtering instead on reading ValidatorList?

fxfactorial: because median stake is reading from this (as well as committee asssignment), the inactive ones are coming from the block and I remove the inactive ones from the list.

fxfactorial: if I do filtering, then it means you want me to read from state each time any median stake computation happens, either via RPC or commitee assignment. Then you defeat the purpose of this cache in the first place.

PR #2424, opened 2020-03-07
[profiling] Add support for pprof in the harmony binary + node.sh

SebastianJ:
This PR adds support for starting a pprof server using the specified args `-pprof ADDRESS` (for the harmony binary) or `-R ADDRESS` (for node.sh) 

Using the updated node.sh file, a node can be started with a pprof enabled server using e.g:
```
./node.sh -k bls.key -N staking -z -D -S -R localhost:6060
```

The above command will start the pprof server at http://localhost:6060 and you can interact with various endpoints (e.g. using a web browser) via http://localhost:6060/debug/pprof

Here are the endpoints that can be used:

**Memory / goroutines / misc**

- http://localhost:6060/debug/pprof/allocs - A sampling of all past memory allocations
- http://localhost:6060/debug/pprof/block - Stack traces that led to blocking on synchronization primitives
- http://localhost:6060/debug/pprof/cmdline - The command line invocation of the current program
- http://localhost:6060/debug/pprof/goroutine - Stack traces of all current goroutines
- http://localhost:6060/debug/pprof/heap - A sampling of memory allocations of live objects. You can specify the gc GET parameter to run GC before taking the heap sample.
- http://localhost:6060/debug/pprof/mutex - Stack traces of holders of contended mutexes
- http://localhost:6060/debug/pprof/threadcreate - Stack traces that led to the creation of new OS threads

**CPU:**

- http://localhost:6060/debug/pprof/profile?seconds=60 - CPU profile. You can specify the duration in the seconds GET parameter. After you get the profile file, use the go tool pprof command to investigate the profile.
- http://localhost:6060/debug/pprof/trace?seconds=60 - A trace of execution of the current program. You can specify the duration in the seconds GET parameter. After you get the trace file, use the go tool trace command to investigate the trace.

To export the data from these endpoints as PDF docs, you can use:
`go tool pprof --pdf http://localhost:6060/debug/pprof/profile?seconds=60 > cpu.pdf`

The PDF export functionality requires having graphviz installed.

Here's an example for how a heap profile can look: [Heap profile 5 - 252.22 mb allocated](https://docs.google.com/viewer?url=http://tools.harmony.one.s3.amazonaws.com/pprof/with-payload/profile005-spammers_active.pdf)

### Unit Test Coverage

Before:

```
$ go test -cover
?   	github.com/harmony-one/harmony/cmd/harmony	[no test files]
```

After:

```
$ go test -cover
?   	github.com/harmony-one/harmony/cmd/harmony	[no test files]
```

### Test/Run Logs

```
$ ./node.sh -k BLS_KEY.key -N staking -z -D -S -R http://localhost:6060
node.sh: using manually specified BLS key file: BLS_KEY.key
node.sh: binaries did not change
node.sh: public IP address autodetected: xxx.xxx.xxx.xxx
Enter passphrase for the BLS key file BLS_KEY.key:
node.sh: ############### Running Harmony Process ###############
Staking mode; node key BLS_KEY -> shard 3
```

```
$ curl http://localhost:6060/debug/pprof/allocs > allocs
$ ls -lh allocs
-rw-rw-r-- 1 deploy deploy 392K Mar  7 14:18 allocs
```

## Operational Checklist

    **NO**

    **NO**

    **YES**

## NOTE
I know that there's already some profiling support in the Harmony binary - but the existing profiling doesn't start a pprof server, and pprof + PDF exports of memory and CPU data is super handy for figuring out where the code allocates a lot of memory, uses a lot of CPU etc etc.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -6,6 +6,8 @@ import (
 	"io/ioutil"
 	"math/big"
 	"math/rand"
+	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"os/signal"
 	"path"
@@ -70,6 +72,7 @@ var (
 	freshDB          = flag.Bool("fresh_db", false, "true means the existing disk based db will be removed")
 	profile          = flag.Bool("profile", false, "Turn on profiling (CPU, Memory).")
 	metricsReportURL = flag.String("metrics_report_url", "", "If set, reports metrics to this URL.")
+	pprof            = flag.String("pprof", "", "what address and port the pprof profiling server should listen on")
 	versionFlag      = flag.Bool("version", false, "Output version info")
 	onlyLogTps       = flag.Bool("only_log_tps", false, "Only log TPS if true")
 	dnsZone          = flag.String("dns_zone", "", "if given and not empty, use peers from the zone (default: use libp2p peer discovery instead)")
@@ -140,6 +143,11 @@ var (
 )
 
 func initSetup() {
+
+	// Setup pprof
+	if addr := *pprof; addr != "" {
+		go func() { http.ListenAndServe(addr, nil) }()
+	}
 
 	// maybe request passphrase for bls key.
 	passphraseForBls()
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -125,6 +125,7 @@ options:
    -M             support multi-key mode (default: off)
    -A             enable archival node mode (default: off)
    -B blacklist   specify file containing blacklisted accounts as a newline delimited file (default: ./.hmy/blacklist.txt)
+   -R address     start a pprof profiling server listening on the specified address
    -I             use statically linked Harmony binary
 
 examples:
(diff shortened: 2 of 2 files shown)


cmd/harmony/main.go
```diff
@@ -6,6 +6,8 @@ import (
 	"io/ioutil"
 	"math/big"
 	"math/rand"
+	"net/http"
+	_ "net/http/pprof"
```
fxfactorial: does this import do side effects? will it generate slower code?

SebastianJ: `_ "net/http/pprof"` simply registers the /debug/pprof/ endpoints via pprof's init() function.

"net/http" already gets pulled in by internal/memprofiling/lib.go, internal/profiler/profiler.go and staking/webhooks/yaml.go - so "net/http" will be included in the binary anyways.

If the imports in those files haven't slowed down the binary so far, then I assume the imports for this PR won't slow it down either.

Issue #2398, opened 2020-03-05
Support different passwords for multi BLS keys


yelllowsin:
Currently latest `node.sh` version only supports one password to load all BLS keys using parameter -M. As discussed on Discord @gupadhyaya, please add support for a different password on each BLS key. Command used to run the node:

`./node.sh -M -N staking -z -S`

Node Version:

```
./node.sh -V
Harmony (C) 2020. harmony, version v5637-master-20191208.0-260-g285e1423 (jenkins@ 2020-03-04T19:09:09+0000)
```

BrunoMarshall:
have the same issue. I´m doing different passwords for different BLS keys.
Running the node is only asking 1 password.

**Currently ALL BLS keys inside the folder need to have the same password**
- Solution founded was to create new BLS keys with the same password. 

Question: do we really need password for BLS keys? Based in the fact that the wallet creation has the option to add password

yelllowsin:
Minor design improvement. When this is implemented, if you could also break the line for every BLS keys used when running the node for better visualization it would be nice:

![image](https://user-images.githubusercontent.com/7217595/76337977-ddaf5480-62d6-11ea-9cae-14fc8258e01b.png)

PR #2437, opened 2020-03-09
adding facility to read passphrase for individual multi-bls keys

gupadhyaya:
Fixes https://github.com/harmony-one/harmony/issues/2398

* Multi-bls users need to put all their keys and passphrases under `.hmy/blskeys`
* Keys need to have `<blskey>.key` extension
* Passphrases need to have `<blskey>.pass` extension
* Error if any file that has different extension than `.key` or `.pass` is found
* For any `<blskey>.key` if no passphrase file is available, it will use the default specified when running the node e.g., `./node.sh -p blspass.txt`

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -105,7 +105,7 @@ var (
 	enableMemProfiling = flag.Bool("enableMemProfiling", false, "Enable memsize logging.")
 	enableGC           = flag.Bool("enableGC", true, "Enable calling garbage collector manually .")
 	blsKeyFile         = flag.String("blskey_file", "", "The encrypted file of bls serialized private key by passphrase.")
-	blsFolder          = flag.String("blsfolder", ".hmy/blskeys", "The folder that stores the bls keys; same blspass is used to decrypt all bls keys; all bls keys mapped to same shard")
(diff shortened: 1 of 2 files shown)


cmd/harmony/main.go
```diff
@@ -272,19 +272,53 @@ func setupStakingNodeAccount() error {
...
+		ext := filepath.Ext(fullName)
+		if ext == ".key" {
+			blsKeyFiles = append(blsKeyFiles, info)
+		} else if ext == ".pass" {
+			passFileName := "file:" + path
+			passphrase, err := utils.GetPassphraseFromSource(passFileName)
+			if err != nil {
+				return err
+			}
+			name := fullName[0 : len(fullName)-len(ext)]
```
fxfactorial: not necessary i think to say 0 explicit

cmd/harmony/main.go
```diff
@@ -272,19 +272,53 @@ func setupStakingNodeAccount() error {
...
+			keyPasses[name] = passphrase
+		} else {
+			return errors.Errorf(
+				"[Multi-BLS] found file: %s that does not have .key or .pass file extension",
+				path,
+			)
+		}
+		return nil
+	}); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr,
```
fxfactorial: don't even hold onto either _, _ since not using either

yelllowsin:
Thanks @gupadhyaya , worked like a charm.

PR #2448, opened 2020-03-10
[audit] Various fixes and todos for auditing; add log for collect rewards

rlan35:
Local test works.


consensus/leader.go
```diff
@@ -311,6 +311,7 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 		return
 	}
 
+	// TODO(audit): verify signature on hash+blockNum+viewID (add a hard fork)
```
fxfactorial: why is it a hard fork though, this is just at consensus level, i guess then it needs to be driven down to the blockchain level too

core/staking_verifier.go
```diff
@@ -98,10 +101,9 @@ func VerifyAndEditValidatorFromMsg(
 		return nil, errCommissionRateChangeTooHigh
 	}
 
-	// TODO: make sure we are reading from the correct snapshot
 	snapshotValidator, err := chainContext.ReadValidatorSnapshot(wrapper.Address)
```
fxfactorial: so we don't need to ensure its at right snapshot epoch?

core/staking_verifier.go
```diff
@@ -172,9 +174,12 @@ func VerifyAndDelegateFromMsg(
 				); err != nil {
 					return nil, nil, err
 				}
-				// Return remaining balance to be deducted for delegation
 				if delegateBalance.Cmp(big.NewInt(0)) < 0 {
-					return nil, nil, errInsufficientBalanceForStake // shouldn't really happen
+					return nil, nil, errNegativeAmount // shouldn't really happen
+				}
+				// Return remaining balance to be deducted for delegation
+				if !CanTransfer(stateDB, msg.DelegatorAddress, delegateBalance) {
```
fxfactorial: can `CanTransfer` be private ?

core/staking_verifier.go
```diff
@@ -172,9 +174,12 @@ func VerifyAndDelegateFromMsg(
...
+					return nil, nil, errInsufficientBalanceForStake
```
fxfactorial: would be nice , `errors.Wrapf(errInsufficientBalanceForStake, "had %v, tried to stake %v", balance, stakeWanted)`

core/state_processor.go
```diff
@@ -119,14 +119,14 @@ func (p *StateProcessor) Process(
 		err := ApplyIncomingReceipt(p.config, statedb, header, cx)
 		if err != nil {
 			return nil, nil,
-				nil, 0, nil, ctxerror.New("cannot apply incoming receipts").WithCause(err)
+				nil, 0, nil, ctxerror.New("[Process] Cannot apply incoming receipts").WithCause(err)
```
fxfactorial: but why put in `[Process]`, anyway you will find out what function it was via line number always included in log

core/state_processor.go
```diff
@@ -253,11 +253,8 @@ func ApplyStakingTransaction(
 
 	// Apply the transaction to the current state (included in the env)
 	gas, err = ApplyStakingMessage(vmenv, msg, gp, bc)
-	utils.Logger().Info().Msgf("ApplyStakingMessage: usedGas: %v, err: %v, stakingTxn:", gas, err)
-
-	// even there is error, we charge it
 	if err != nil {
-		return nil, gas, err
+		return nil, 0, err
```
fxfactorial: why gas is always 0? was the previous variable always 0 as well?

core/state_transition.go
```diff
@@ -396,11 +406,10 @@ func (st *StateTransition) verifyAndApplyDelegateTx(delegate *staking.Delegate)
...
 		return err
 	}
-	if err := st.state.UpdateValidatorWrapper(wrapper.Validator.Address, wrapper); err != nil {
-		return err
-	}
+
 	st.state.SubBalance(delegate.DelegatorAddress, balanceToBeDeducted)
-	return nil
+
+	return st.state.UpdateValidatorWrapper(wrapper.Validator.Address, wrapper)
```
fxfactorial: you can just say `wrapper.Address`

core/types/block.go
```diff
@@ -294,12 +294,19 @@ func NewBlock(
...
+
 	b := &Block{header: CopyHeader(header)}
 
 	if len(receipts) != len(txs)+len(stks) {
+		utils.Logger().Error().
+			Int("receiptsLen", len(receipts)).
+			Int("txnsLen", len(txs)).
+			Int("stakingTxnsLen", len(stks)).
+			Msg("Length of receipts doesn't match length of transactions")
 		return nil
```
fxfactorial: why it doesn't return error value?

fxfactorial: ah, i see, API like that, should be explicit error

core/types/block.go
```diff
@@ -294,12 +294,19 @@ func NewBlock(
...
 	}
 
+	// Put transactions into block
```
fxfactorial: this comment confused me, says put transactions into block but next line is when nothing to add at all

staking/types/validator.go
```diff
@@ -180,6 +180,7 @@ func (v *Validator) SanityCheck(oneThirdExtrn int) error {
 		return err
 	}
 
+	// TODO(audit): add limit on the number of bls keys one can have.
```
fxfactorial: implemented in https://github.com/harmony-one/harmony/pull/2396

staking/types/validator.go
```diff
@@ -200,10 +201,8 @@ func (v *Validator) SanityCheck(oneThirdExtrn int) error {
 		return errNilMaxTotalDelegation
 	}
 
-	// if I'm not banned, then I must
-	// ensure that MinSelfDelegation >= 1 ONE
-	if v.EPOSStatus != effective.Banned &&
-		v.MinSelfDelegation.Cmp(big.NewInt(denominations.One)) < 0 {
+	// MinSelfDelegation must be >= 1 ONE
+	if v.MinSelfDelegation.Cmp(big.NewInt(denominations.One)) < 0 {
```
fxfactorial: are you sure to take away ` v.EPOSStatus != effective.Banned`

consensus/leader.go
rlan35: Because the signed data changed, then the verification and all need to be changed too.

core/staking_verifier.go
rlan35: It's already ensured by the snapshot code being correct. The todo was added when the feature was first implemented.

core/staking_verifier.go
rlan35: It's EVM's code. we are just reusing it.

core/staking_verifier.go
rlan35: done

core/state_processor.go
rlan35: it's easier to visually search for.

core/state_processor.go
rlan35: The previous logic is wrong. We shouldn't charge gas when the staking txn failed.

core/state_transition.go
rlan35: done.

core/types/block.go
rlan35: put empty txns is also put :)

staking/types/validator.go
rlan35: yea. basically MinSelfDelegation rule should still apply.  Slashing doesn't change it

core/state_processor.go
fxfactorial: Okay wow this mistake lasted a long time.

staking/types/validator.go
fxfactorial: You don’t want to do anything with banned in sanity check ?

rlan35: Banned is used to avoid checking stake vs MinSelfStake. But it's not needed for checking minSelfStake vs 1ONE.

Issue #2260, opened 2020-02-13
Offchain data should be committed consistently with onchain data


rlan35:
Many of the offchain data such as activeValidatorList is critical to the onchain data processing logic. Even though offchain data can be derived from onchain history, it requires reprocessing the whole blockchain history to recreate the data.

We need to make sure the offchain data is always consistent with the onchain data in the block commit phase. This include:

1. Make sure the block commit process is atomic.
2. Make it possible to revert the offchain data as the blocks are reverted.

PR #2479, opened 2020-03-11
[Audit] Refactor offchain commit and fix consistency issue due to batch write

rlan35:
Previously the change to batch write to DB would cause issue if there are multiple createValidator or delegation happens in the same block for the same address. 

The reason is that the batch write of the metadata only take effect at the end of all the changes. But it won't reflect it if reading in the middle of the operations, which may cause race condition.

This change take care of the issue by adding a cache layer to accumulate all the changes and batch write them at the end.

Tested locally.


denniswon:
@rlan35 is this for https://github.com/harmony-one/harmony/issues/2260? or just refactoring?

rlan35:
> @rlan35 is this for #2260? or just refactoring?

Yea, it's related.

rlan35:
Jenkins, test this please.

PR #2484, opened 2020-03-11
[rpc][availability][apr] Richer validator information, implement APR, unify EPoS computation, remove fall 2019 tech debt

fxfactorial:
See change log for description of changes in this PR

https://gist.github.com/fxfactorial/4ce5cdc1b63291c27534bb88df63e024


staking/types/validator.go
```diff
@@ -87,16 +86,53 @@ type ValidatorWrapper struct {
...
+	TotalDelegated *big.Int `json:"total-delegation"`
 }
 
 // ValidatorRPCEnchanced contains extra information for RPC consumer
 type ValidatorRPCEnchanced struct {
-	Wrapper                  ValidatorWrapper `json:"validator"`
-	CurrentSigningPercentage Computed         `json:"current-epoch-signing-percent"`
-	CurrentVotingPower       []VotePerShard   `json:"current-epoch-voting-power"`
+	Wrapper     ValidatorWrapper        `json:"validator"`
+	Performance CurrentEpochPerformance `json:"current-epoch-performance"`
```
stephen-tse: @fxfactorial @rlan35 let's tag backend data changes for review with @mikedoan @potvik on dashboard frontend. edgar, do you have a sample json for the new format?

fxfactorial: not yet, I know to ping when ready.

mikedoan:
For the stasking dashboard, please just give me in-advance the format like. 
```
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "current-epoch-signing-percent": {
      "current-epoch-signed": 0,
      "current-epoch-to-sign": 0,
      "percentage": "0.000000000000000000"
    },
    "current-epoch-voting-power": [],
    "validator": {
      "address": "one14438psd5vrjes7qm97jrj3t0s5l4qff5j5cn4h",
      "availability": {
        "num-blocks-signed": 0,
        "num-blocks-to-sign": 0
      },
      "bls-public-keys": [
        "67336532c04545afc5c1c979f58b5c301af406eaa0f4c900dcd3004189936c7213ee126d9591026f65248e5f25278f02"
      ],
      "creation-height": 224,
      "delegations": [
        {
          "amount": 3.46e+19,
          "delegator-address": "one14438psd5vrjes7qm97jrj3t0s5l4qff5j5cn4h",
          "reward": 0,
          "undelegations": []
        },
        {
          "amount": 1.5e+18,
          "delegator-address": "one1nqevvacj3y5ltuef05my4scwy5wuqteur72jk5",
          "reward": 0,
          "undelegations": []
        }
      ],
      "details": "none",
      "epos-eligibility-status": 1,
      "identity": "test_account2",
      "last-epoch-in-committee": 0,
      "max-change-rate": "0.052212761523253600",
      "max-rate": "0.179184469782137200",
      "max-total-delegation": 8e+19,
      "min-self-delegation": 3.2e+18,
      "name": "_Test_key_validator2",
      "rate": "0.127983520183826780",
      "security-contact": "Edgar-VDM",
      "update-height": 224,
      "website": "harmony.one"
    }
  }
}
```

fxfactorial:
On course Minh, just like I have for every single change.

sophoah:
Let us know when changed. That proposed json is changing the epos-eligibility-status value (now active/inactive). Need to update docs...

sophoah:
Ok just clarified with Edgar, no change in the epos-eligibility-status

core/blockchain.go
```diff
@@ -2257,47 +2257,60 @@ func (bc *BlockChain) writeValidatorSnapshots(
...
+		stats := rawdb.ReadValidatorStats(bc.db, key)
+		stats.TotalEffectiveStake = value.TotalEffectiveStake
+		stats.VotingPowerPerShard = value.VotingPower
+		stats.BLSKeyPerShard = value.BLSPublicKeysOwned
+		if err := rawdb.WriteValidatorStats(batch, key, stats); err != nil {
 			return err
 		}
 	}
 
+	// TODO Need to remove the set difference of those
```
rlan35: use TODO(audit):

core/offchain.go
```diff
@@ -187,18 +194,41 @@ func (bc *BlockChain) CommitOffChainData(
 		utils.Logger().Debug().Msgf(msg, len(*crossLinks), num)
 	}
 	// Roll up latest crosslinks
-	for i := uint32(0); i < shard.Schedule.InstanceForEpoch(epoch).NumShards(); i++ {
-		bc.LastContinuousCrossLink(batch, i)
+	for i, c := uint32(0), shard.Schedule.InstanceForEpoch(
+		epoch,
+	).NumShards(); i < c; i++ {
+		if err := bc.LastContinuousCrossLink(batch, i); err != nil {
+			// TODO RJ is this a critical error?
```
rlan35: it's not, but let's log the error for sure.

internal/chain/reward.go
```diff
@@ -201,53 +283,53 @@ func AccumulateRewards(
...
-				common.Address
-				*big.Int
-			}{common2.MustAddressToBech32(receipient), receipient, amount},
-			)
-		},
-	)
+	totalAmount := big.NewInt(0)
+
+	{
+		last := big.NewInt(0)
```
rlan35: This is just copy and paste from the old code right?

fxfactorial: yes, see from that deleted Award, itself a copy paste from orginal code

staking/apr/compute.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	// avg_reward_per_block = total_reward_per_epoch / blocks_per_epoch
+	// estimated_reward_per_year = avg_reward_per_block * blocks_per_year
+	// estimated_reward_per_year / total_stake
+	perEpoch := new(big.Int).SetUint64(blocksPerEpoch)
+	avgRewardPerBlock := new(big.Int).Div(stats.BlockReward, perEpoch)
+	estimatedBlocksPerYear, err := expectedNumBlocksPerYear(
+		nil, nil, blocksPerEpoch,
+	)
+	estimatedRewardPerYear := new(big.Int).Div(
```
rlan35: this should be Mul()

core/blockchain.go
fxfactorial: implemented

core/blockchain.go
```diff
@@ -2263,41 +2263,55 @@ func (bc *BlockChain) ReadValidatorStats(
...
 			return err
 		}
 	}
 
+	existing, replacing :=
+		currentEpochSuperCommittee.StakedValidators(),
+		newEpochSuperCommittee.StakedValidators()
+	for currentValidator := range existing.LookupSet {
+		if _, keptSlot := replacing.LookupSet[currentValidator]; !keptSlot {
+			rawdb.DeleteValidatorSnapshot(
```
rlan35: why delete snapshot? They should be kept in db all the time.

internal/chain/reward.go
```diff
@@ -27,22 +28,53 @@ func ballotResultBeaconchain(
...
+) (*votepower.Roster, error) {
+	key := fmt.Sprintf("%s-%d", epoch.String(), subComm.ShardID)
+	results, err, _ := votingPowerCache.Do(
+		key, func() (interface{}, error) {
+			votingPower, err := votepower.Compute(subComm)
+			if err != nil {
+				return nil, err
+			}
+			go func() {
+				time.Sleep(votepowerCacheTimeLimit)
```
rlan35: why forget every 3 minutes? We need the voting power for the whole epoch right?

staking/apr/compute.go
```diff
@@ -0,0 +1,73 @@
...
+	blocksPerEpoch uint64,
+) (*big.Int, error) {
+	snapshot, err := bc.ReadValidatorSnapshot(wrapper.Address)
+	if err != nil {
+		return nil, err
+	}
+	// avg_reward_per_block = total_reward_per_epoch / blocks_per_epoch
+	// estimated_reward_per_year = avg_reward_per_block * blocks_per_year
+	// estimated_reward_per_year / total_stake
+	// TODO use the blockreward diff like in measure
```
rlan35: so the APR is not finished yet?

staking/effective/eligible.go
```diff
@@ -7,28 +7,43 @@ type Eligibility byte
 const (
 	// Nil is a default state that represents a no-op
 	Nil Eligibility = iota
-	// Active means allowed in epos auction
-	Active
-	// Inactive means validator did not sign enough over 66%
+	// FirstTimeCandidate ..
+	FirstTimeCandidate
```
rlan35: I think these new statuses are too much and may cause more potential mistakes in coding. I don't see the extra benefit outweigh the potential issues.

staking/types/validator.go
```diff
@@ -555,6 +574,16 @@ func UpdateValidatorFromEditMsg(validator *Validator, edit *EditValidator) error
 	return nil
 }
 
+// IsEligibleForEPoSAuction ..
+func IsEligibleForEPoSAuction(validator *ValidatorWrapper) bool {
+	switch validator.EPOSStatus {
+	case effective.FirstTimeCandidate, effective.Candidate:
```
rlan35: For example, I don't see where the validator turn from other status e.g. InCommitteeAndSigning to Candidate? Seems elected validators won't get elected in new epochs. That's the complexity of state transition I am talking about after adding too many different status.

staking/apr/compute.go
fxfactorial: It’s there, I’m just not sure about the computation . All the data points should be available

internal/chain/reward.go
fxfactorial: Just picking some period till you help me pick a better value

core/blockchain.go
fxfactorial: Because they aren’t in committee anymore so why hold onto that data

staking/types/validator.go
fxfactorial: It’s true but the status are fleshing out the overloaded way we are using this field , downstream on CLI people were misunderstanding the meaning of active / inactive

core/blockchain.go
fxfactorial: Otherwise we will have an explosion of validator snapshots

staking/types/validator.go
fxfactorial: That first scenario you mention is done in edit validator as default case

staking/effective/eligible.go
rlan35: Ok, then I think we can remove firstTimeCandidate, it's not necessary. And basically it's two bit, candidate/not, incommittee/not.

We can do 

1. Candidate
2. NotCandidate
3. ElectedAndSigning
4. ElectedAndFellBelowThreshold

consensus/quorum/one-node-staked-vote.go
```diff
@@ -97,15 +97,16 @@ func (v *stakedVoteWeight) computeCurrentTotalPower(p Phase) (*numeric.Dec, erro
...
 	for i := range members {
-		w.FromLibBLSPublicKey(members[i])
+		if err := w.FromLibBLSPublicKey(members[i]); err != nil {
+			return nil, err
+		}
 		if _, didVote := ballot.voters[w]; !didVote &&
 			v.ReadBallot(p, members[i]) != nil {
-			err := w.FromLibBLSPublicKey(members[i])
-			if err != nil {
+			if err := w.FromLibBLSPublicKey(members[i]); err != nil {
```
rlan35: existing problem, but duplicated "w.FromLibBLSPublicKey"

consensus/quorum/quorum.go
```diff
@@ -125,13 +124,13 @@ type Decider interface {
 
 // Registry ..
 type Registry struct {
-	Deciders      map[uint32]Decider `json:"quorum-deciders"`
+	Deciders      map[string]Decider `json:"quorum-deciders"`
```
rlan35: why change from uint32 to string? It's harder to address the key now.

fxfactorial: This was only used for super committee RPC and makes key easier to pull for downstream consumer watchdog

consensus/quorum/one-node-staked-vote.go
fxfactorial: Only moved line to scope the error

consensus/votepower/roster.go
```diff
@@ -61,81 +62,95 @@ func (b Ballot) String() string {
...
+	s, _ := json.Marshal(v)
+	return string(s)
+}
+
+type topLevelRegistry struct {
 	OurVotingPowerTotalPercentage   numeric.Dec
 	TheirVotingPowerTotalPercentage numeric.Dec
-	RawStakedTotal                  numeric.Dec
-	HmySlotCount                    int64
+	RawStakedTotal                  *big.Int
```
rlan35: So because RawStakedTotal is not related to protocol. You changed it from dec to int for simplicity?

fxfactorial: Because I realized that stake can only be integer because token is only integer and leaving as big int reduces memory usage and your point as well

hmy/api_backend.go
```diff
@@ -333,88 +334,63 @@ func (b *APIBackend) GetValidatorInformation(
...
-
-	instance := shard.Schedule.InstanceForEpoch(b.CurrentBlock().Epoch())
-	stakedSlots :=
-		(instance.NumNodesPerShard() - instance.NumHarmonyOperatedNodesPerShard()) *
-			int(instance.NumShards())
-	median, _ := effective.Compute(essentials, stakedSlots)
-	return median.TruncateInt(), nil
+	// NOTE We do this because for RPC presentation, we index from 1 instead of 0
+	// like how committee assignment does.
+	round.MaximumExternalSlotAllowed++
```
rlan35: why? this is just a count right? Like 320.

fxfactorial: Because like note mentions, otherwise it comes out off by 1 . Aka 39 open slots instead of 40. So the 320 would look like 319 without the ++

internal/chain/engine.go
```diff
@@ -273,6 +265,25 @@ func (e *engineImpl) Finalize(
 	isBeaconChain := header.ShardID() == shard.BeaconChainShardID
 	isNewEpoch := len(header.ShardState()) > 0
 	inStakingEra := chain.Config().IsStaking(header.Epoch())
+	if isBeaconChain && inStakingEra {
```
rlan35: These code can be packed into state.AddReward and stay inside the AccumulateReward func. Doesn't have to be another separate loop update.

internal/chain/reward.go
```diff
@@ -27,34 +27,65 @@ func ballotResultBeaconchain(
...
+				return nil, err
+			}
+			return votingPower, nil
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	if new(big.Int).Sub(currentEpoch, epoch).Cmp(common.Big3) == 1 {
```
rlan35: If the epoch is not looked up. even if it passes 3 epoch old, it will never be cleared from cache?

fxfactorial: Right but via crosslinks you’ll usually end up hitting this case. We could have a time based approach as well or instead as a fall back or replacement of logic

shard/committee/assignment.go
```diff
@@ -35,11 +32,94 @@ type Reader interface {
...
+		validator, err := stakedReader.ReadValidatorInformation(
+			candidates[i],
+		)
+		if err != nil {
+			return nil, err
+		}
+		if !staking.IsEligibleForEPoSAuction(validator) {
+			continue
+		}
+		if err := validator.SanityCheck(staking.DoNotEnforceMaxBLS); err != nil {
```
rlan35: no need to do another sanity check here. The validator in db is already valid.

shard/committee/assignment.go
```diff
@@ -151,107 +222,32 @@ func eposStakedCommittee(
 		}
 	}
 
-	if stakedSlotsCount == 0 {
+	completedEPoSRound, err := NewEPoSRound(stakerReader)
+
+	if completedEPoSRound.MaximumExternalSlotAllowed == 0 {
```
rlan35: MaximumExternalSlotAllowed is never set to 0 even if there is no external validators elected.

staking/apr/compute.go
```diff
@@ -0,0 +1,101 @@
...
+type Reader interface {
+	GetHeaderByNumber(number uint64) *block.Header
+	ReadValidatorSnapshotAtEpoch(
+		epoch *big.Int,
+		addr common.Address,
+	) (*staking.ValidatorWrapper, error)
+}
+
+const (
+	nanoSecondsInYear = time.Nanosecond * 3.154e+16
```
rlan35: why use nanoseconds? The header.Time() is in Unix seconds.

staking/apr/compute.go
```diff
@@ -0,0 +1,101 @@
...
+
+	estimatedRewardPerYear, err := expectedRewardPerYear(
+		bc.GetHeaderByNumber(blockNumAtOneEpochAgo),
+		bc.GetHeaderByNumber(blockNumAtTwoEpochAgo),
+		oneSnapshotAgo,
+		twoSnapshotAgo,
+		blocksPerEpoch,
+	)
+
+	return new(big.Int).Div(
```
rlan35: big int doesn.t work, as it's fractional, use Dec.

staking/apr/compute.go
fxfactorial: I thought I saw it used nano

staking/availability/measure.go
```diff
@@ -234,29 +249,29 @@ func compute(
...
+	switch computed.IsBelowThreshold {
 	case missedTooManyBlocks:
-		wrapper.EPOSStatus = effective.Inactive
+		wrapper.Status = effective.Inactive
 		utils.Logger().Info().
 			Str("threshold", measure.String()).
 			Msg("validator failed availability threshold, set to inactive")
 	default:
-		wrapper.EPOSStatus = effective.Active
+		wrapper.Status = effective.Active
```
rlan35: please add a TODO： we need to take care of the situation when a validator wants to stop validating, but if he turns his validator to inactive and his node is still running, the status will be turned back to active automatically.  Don't need to address in this PR.

staking/effective/calculate.go
```diff
@@ -143,19 +130,21 @@ func Compute(
...
+	*big.Int, []SlotPurchase,
+) {
 	median, picks := Compute(shortHand, pull)
 	for i := range picks {
-		picks[i].Dec = effectiveStake(median, picks[i].Dec)
+		picks[i].Stake = effectiveStake(median, picks[i].Stake)
 	}
 
-	return picks
+	return median.TruncateInt(), picks
```
rlan35: why truncate median?

fxfactorial: Good point , it is a decimal

hmy/api_backend.go
rlan35: then the func which returns it should do the job. Because "MaximumExternalSlotAllowed" naming meant a count, not an index. The user of it expect a count.

staking/apr/compute.go
```diff
@@ -22,11 +23,11 @@ type Reader interface {
 }
 
 const (
-	nanoSecondsInYear = time.Nanosecond * 3.154e+16
+	secondsInYear = time.Second * 3.154e+7
```
rlan35: you should double check when you make the changes. Underlying, It's the same number Int counting nanoseconds. We don't need the time. package. Just use int64 is fine.

fxfactorial: You mean just do int64(3.154e+7) directly? (I like this way because it is semantically clear in intent, I liked named constants)

consensus/quorum/one-node-one-vote.go
```diff
@@ -109,11 +89,10 @@ func (v *uniformVoteWeight) AmIMemberOfCommitee() bool {
 		return false
 	}
 	identity, _ := pubKeyFunc()
-	everyone := v.DumpParticipants()
+	everyone := v.Participants()
 	for _, key := range identity.PublicKey {
-		myVoterID := key.SerializeToHexStr()
 		for i := range everyone {
-			if everyone[i] == myVoterID {
+			if everyone[i] == key {
```
rlan35: This is comparing pointers. We shouldn't compare pointers. it's error prone.

fxfactorial: Why, it’s comparing a fixed size buffer in value. , would be point if was plain [] byte. It’s like comparing common.Address, allowed

consensus/votepower/roster.go
```diff
@@ -144,78 +159,66 @@ func AggregateRosters(rosters []RosterPerShard) map[common.Address]Staker {
...
+
 	for i := range staked {
-		if staked[i].EffectiveStake == nil {
-			roster.HmySlotCount++
-		} else {
-			roster.RawStakedTotal = roster.RawStakedTotal.Add(
-				*staked[i].EffectiveStake,
+		if e := staked[i].EffectiveStake; e != nil {
+			roster.RawStakedTotal.Add(
+				roster.RawStakedTotal, e.TruncateInt(),
```
rlan35: RawStakedTotal is a bad name. It's the TotalEffectiveStake, and it's composed of Decs. We should keep using dec rather than int.

consensus/quorum/one-node-one-vote.go
fxfactorial: Its shard.BlsPublickKey, not *bls.PublickKey

consensus/votepower/roster.go
fxfactorial: But the stake itself is token and token can only be integer, right? So decimal only needed in actual computations when floats arrive, like the epos computation itself, but summing up what people actually stake is only integer token, no?

staking/types/validator.go
stephen-tse: @mikedoan see https://gist.github.com/fxfactorial/6c518ac0e90c3978f8420d414c3cf611 for json samples/changes, and release notes https://gist.github.com/fxfactorial/4ce5cdc1b63291c27534bb88df63e024

consensus/quorum/one-node-one-vote.go
rlan35: Participants() returns []*bls.PublicKey, pubKeyFunc also returns []*bls.PublicKey. Comparing pointer won't be correct.

fxfactorial: Ah you're right, my mistake. I forgot to unify this usage to be shard.BlsPublicKey like I did else where, easy fix.

core/blockchain.go
```diff
@@ -2263,39 +2265,68 @@ func (bc *BlockChain) ReadValidatorStats(
 
 // UpdateValidatorVotingPower writes the voting power for the committees
 func (bc *BlockChain) UpdateValidatorVotingPower(
-	batch rawdb.DatabaseWriter, state *shard.State,
+	batch rawdb.DatabaseWriter,
+	newEpochSuperCommittee, currentEpochSuperCommittee *shard.State,
+	// NOTE Do not update this state, only read from
```
rlan35: FYI, you can do a copy() on state.

shard/committee/assignment.go
```diff
@@ -35,11 +34,134 @@ type Reader interface {
...
+		essentials[validator.Address] = &effective.SlotOrder{
+			validatorStake,
+			validator.SlotPubKeys,
+			tempZero,
+		}
+	}
+	totalStakedDec := numeric.NewDecFromBigInt(totalStaked)
+
+	for _, value := range essentials {
+		value.Percentage = numeric.NewDecFromBigInt(value.Stake).Quo(totalStakedDec)
```
rlan35: What is this Percentage used for ?

shard/committee/assignment.go
```diff
@@ -35,11 +34,134 @@ type Reader interface {
...
+			Validator: key,
+		}
+		i++
+	}
+
+	return &CompletedEPoSRound{
+		MedianStake: median,
+		// NOTE + 1 because we use it as index in assigment,
+		// but as count in semantics
+		MaximumExternalSlot: maxExternalSlots + 1,
```
rlan35: I checked maxExternalSlots = (instance.NumNodesPerShard() -
			instance.NumHarmonyOperatedNodesPerShard()) *
			int(instance.NumShards()) 

it's real count. Not index starting from 0. Don't need to + 1.

staking/apr/compute.go
rlan35: yea. because time.Second = 1000000000 in int (it counts in nano seconds in a second). secondsInYear is not actually in seconds, it's still in nano seconds.

staking/types/validator.go
```diff
@@ -77,26 +78,64 @@ type counters struct {
...
 
 // Computed represents current epoch
 // availability measures, mostly for RPC
 type Computed struct {
-	Signed     *big.Int    `json:"current-epoch-signed"`
-	ToSign     *big.Int    `json:"current-epoch-to-sign"`
-	Percentage numeric.Dec `json:"percentage"`
+	Signed            *big.Int    `json:"current-epoch-signed"`
+	ToSign            *big.Int    `json:"current-epoch-to-sign"`
+	BlocksLeftInEpoch uint64      `json:"current-epoch-blocks-left"`
```
rlan35: BlocksLeftInEpoch remove?

shard/committee/assignment.go
fxfactorial: because the function prepares orders, people downstream been asking in epos, what % of the total state did I have during auction, so that median-stake RPC gets more feature rich: 

https://gist.github.com/fxfactorial/6c518ac0e90c3978f8420d414c3cf611#file-median-stake-json

staking/types/validator.go
fxfactorial: Its useful, people downstream asking for it and good to know how much blocks left for me to go

staking/apr/compute.go
```diff
@@ -23,7 +22,7 @@ type Reader interface {
 }
 
 const (
-	secondsInYear = time.Second * 3.154e+7
+	secondsInYear = int64(3.154e+7)
```
stephen-tse: not sure why the google top result gives the wrong answer, but 31557600 
 is a better estimate? https://www.rapidtables.com/calc/time/seconds-in-year.html

fxfactorial: Sure will address in next pr

PR #2495, opened 2020-03-13
[RPC] Add GetLastCrossLinks RPC

janet-harmony:
# Issue

Add RPC for https://github.com/harmony-one/harmony-ops/issues/463

### Test/Run Logs

From Postman:
```
{"jsonrpc":"2.0", "method":"hmy_getLastCrossLinks", "params":[], "id":1}

{
    "jsonrpc": "2.0",
    "id": 1,
    "result": [
        {
            "parentHash": "0xfb7b0bae66dd10780ec7aa6baa52a552ef7fc86ac15f5fb7b6aedbca5b434ad6",
            "blockNumber": 28,
            "signature": "1ad743b56a9915564dbe8102f0cbdaa218f855aa0d0d922205b675ababd18aa7d49b6c355a279db08889f5c33d5dc20c2a074275a593bf4a90cb8992c3feac4a124425441ca84a5f4fc674a868a3fa695d1a49f459c842a82e6b890cfeafdb0b",
            "signatureBitmap": "3f",
            "shardId": 1,
            "epochNumber": 4
        }
    ]
}
```

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/core/types/crosslink.go
+++ b/core/types/crosslink.go
@@ -1,6 +1,7 @@
 package types
 
 import (
+	"encoding/hex"
 	"math/big"
 	"sort"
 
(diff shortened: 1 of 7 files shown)


core/types/crosslink.go
```diff
@@ -23,6 +24,16 @@ type CrossLink struct {
 	EpochF       *big.Int
 }
 
+// RPCCrossLink is used for returning data to RPC calls, converting byte arrays to hex encoded strings
+type RPCCrossLink struct {
```
fxfactorial: Not needed to make this data type wrapper, instead just define a marshaljson method on the crosslink type itself , return a crosslink type in rpc

janet-harmony: Made the struct, so Signature & SignatureBitmaps would return something as a hex encoded string, instead of a byte array, to match latestHeader.

fxfactorial: I know, you can control that in marshaljson method of cross link since the marshaljson method controls presentation

PR #2503, opened 2020-03-14
[config] make STN 2 shards

Daniel-VDM:
STN should focus more on CPU analysis than network size.


LeoHChen:
Jenkins, test this please.

PR #2522, opened 2020-03-18
[apr] Traverse till find right epochs or bottom out 

fxfactorial:
... because GetHeaderByNumber doesn't work on epoch -1 for whatever (existing and known as quirky) reason

### DIFF
--- a/staking/apr/compute.go
+++ b/staking/apr/compute.go
@@ -6,6 +6,7 @@ import (
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/harmony-one/harmony/block"
 	"github.com/harmony-one/harmony/core/state"
+	"github.com/harmony-one/harmony/internal/params"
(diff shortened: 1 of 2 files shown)


staking/apr/compute.go
```diff
@@ -57,6 +56,57 @@ func expectedRewardPerYear(
 	return new(big.Int).Mul(expectedValue, oneYear), nil
 }
 
+func pastTwoEpochHeaders(
+	bc Reader,
+) (*block.Header, *block.Header, error) {
+	current := bc.CurrentHeader()
+	epochNow := current.Epoch()
+	oneEpochAgo, twoEpochAgo :=
+		new(big.Int).Sub(epochNow, common.Big2),
```
rlan35: sub big1.

stephen-tse: let's write unit tests for all new functions such as `pastTwoEpochHeaders`

PR #2528, opened 2020-03-18
add test code for delegations reward scalability

rlan35:
simple scalability test program.

### DIFF
--- a/test/chain/reward/main.go
+++ b/test/chain/reward/main.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"fmt"
+	"math/big"
+	"math/rand"
+	"time"
+
+	common2 "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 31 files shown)


test/chain/reward/main.go
```diff
@@ -0,0 +1,109 @@
...
+		})
+	}
+
+	statedb.UpdateValidatorWrapper(validator.Address, validator)
+
+	startTime := time.Now()
+	validator, _ = statedb.ValidatorWrapper(msg.ValidatorAddress)
+
+	fmt.Printf("Total num delegations: %d\n", len(validator.Delegations))
+	statedb.AddReward(validator, big.NewInt(1000))
```
gupadhyaya: probably go benchmarking is better.

rlan35: Other similar tests are all in /test. Let's keep it as is.

Issue #2520, opened 2020-03-18
current-epoch-blocks-left overflow error


mindstyle85:
**Describe the bug**

"current-epoch-signing-percent": {
        **"current-epoch-blocks-left": 18446744073709552000,**
        "current-epoch-signed": 44,
        "current-epoch-signing-percentage": "0.458333333333333333",
        "current-epoch-to-sign": 96

mikedoan:
can you please elaborate on this issue?

mindstyle85:
you dont see the abnormally high (bolded) curren-epoch-blocks-left number? @fxfactorial can explain better, but that number is wrong and WAY too high, @fxfactorial said its an overflow issue

mindstyle85:
it also shows too small of a number for blocks to sign:

"current-epoch-performance": {
      "current-epoch-signing-percent": {
        **"current-epoch-blocks-left": 20,**
        "current-epoch-signed": 18,
        "current-epoch-signing-percentage": "1.000000000000000000",
        "current-epoch-to-sign": 18

after that counter hits zero, the above mentioned abnormally high number appears

@mikedoan

mikedoan:
@mindstyle85 got. it. this should be assigned to @fxfactorial

PR #2533, opened 2020-03-19
Check signing threshold and set EPoS status only at last epoch

rlan35:
Do ComputeAndMutateEPOSStatus at the last epoch, instead of at every block.

In EPoS election candidacy, validators in current committee who doesn't sign enough blocks will be kicked out no matter what. For those who signed enough, the candidacy is determined by the Status. 

Tested locally

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -356,13 +356,10 @@ func (b *APIBackend) GetValidatorInformation(
 		return defaultReply, nil
 	}
 
-	computed, err := availability.ComputeCurrentSigning(
-		snapshot, wrapper, shard.Schedule.BlocksPerEpoch(),
+	computed := availability.ComputeCurrentSigning(
+		snapshot, wrapper,
 	)
-
-	if err != nil {
-		return defaultReply, nil
-	}
+	computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - computed.ToSign.Uint64()
 
 	stats, err := b.hmy.BlockChain().ReadValidatorStats(addr)
 	if err != nil {
@@ -402,8 +399,9 @@ func (b *APIBackend) GetTotalStakingSnapshot() *big.Int {
 	}
 	stakes := big.NewInt(0)
 	for i := range candidates {
+		snapshot, _ := b.hmy.BlockChain().ReadValidatorSnapshot(candidates[i])
 		validator, _ := b.hmy.BlockChain().ReadValidatorInformation(candidates[i])
-		if !staking.IsEligibleForEPoSAuction(validator) {
+		if !committee.IsEligibleForEPoSAuction(snapshot, validator) {
 			continue
 		}
 		for i := range validator.Delegations {
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -494,10 +494,11 @@ func (node *Node) PostConsensusProcessing(
 				if err != nil {
 					return
 				}
-				computed, err :=
-					availability.ComputeCurrentSigning(
-						snapshot, wrapper, shard.Schedule.BlocksPerEpoch(),
-					)
+				computed := availability.ComputeCurrentSigning(
+					snapshot, wrapper,
+				)
+				computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - computed.ToSign.Uint64()
+
 				if err != nil && computed.IsBelowThreshold {
 					url := h.Availability.OnDroppedBelowThreshold
 					go func() {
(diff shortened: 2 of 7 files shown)


staking/availability/measure.go
```diff
@@ -71,25 +69,8 @@ func BlockSigners(
 // BallotResult returns
 // (parentCommittee.Slots, payable, missing, err)
 func BallotResult(
-	bc engine.ChainReader, header *block.Header, shardID uint32,
+	parentHeader, header *block.Header, parentShardState *shard.State, shardID uint32,
 ) (shard.SlotList, shard.SlotList, shard.SlotList, error) {
-	// TODO ek – retrieving by parent number (blockNum - 1) doesn't work,
-	//  while it is okay with hash.  Sounds like DB inconsistency.
-	//  Figure out why.
-	parentHeader := bc.GetHeaderByHash(header.ParentHash())
```
rlan35: these are moved out to break import cycle.

staking/availability/measure.go
```diff
@@ -193,68 +166,63 @@ func ComputeCurrentSigning(
 	signed, toSign :=
 		new(big.Int).Sub(statsNow.NumBlocksSigned, snapSigned),
 		new(big.Int).Sub(statsNow.NumBlocksToSign, snapToSign)
-	leftToGo := blocksPerEpoch - toSign.Uint64()
 
 	computed := staking.NewComputed(
-		signed, toSign, leftToGo, numeric.ZeroDec(), true,
+		signed, toSign, 0, numeric.ZeroDec(), true,
```
fxfactorial: hmm, why take it out, people wanted it, will affect RPC

node/node_handler.go
```diff
@@ -494,10 +494,11 @@ func (node *Node) PostConsensusProcessing(
...
 					return
 				}
-				computed, err :=
-					availability.ComputeCurrentSigning(
-						snapshot, wrapper, shard.Schedule.BlocksPerEpoch(),
-					)
+				computed := availability.ComputeCurrentSigning(
+					snapshot, wrapper,
+				)
+				computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - computed.ToSign.Uint64()
```
fxfactorial: ah I see, you set the blocks left her

staking/availability/measure.go
rlan35: i moved it outside the func. it's only for RPCs. for protocol calculation, it's not needed.

staking/availability/measure.go
fxfactorial: what was the cycle? (interfaces also great way to break cycles)

staking/availability/measure.go
rlan35: to avoid wiring unnecessary things into the protocol code.

shard/committee/assignment.go
```diff
@@ -160,6 +168,28 @@ func prepareOrders(
 	return essentials, nil
 }
 
+// IsEligibleForEPoSAuction ..
+func IsEligibleForEPoSAuction(snapshot, validator *staking.ValidatorWrapper) bool {
+	if snapshot.Counters.NumBlocksToSign.Cmp(validator.Counters.NumBlocksToSign) != 0 {
```
fxfactorial: so we assume that snapshot of wrapper should have happened already by now, correct? That's only way these will match

hmy/api_backend.go
```diff
@@ -356,13 +356,10 @@ func (b *APIBackend) GetValidatorInformation(
...
-	computed, err := availability.ComputeCurrentSigning(
-		snapshot, wrapper, shard.Schedule.BlocksPerEpoch(),
+	computed := availability.ComputeCurrentSigning(
+		snapshot, wrapper,
 	)
-
-	if err != nil {
-		return defaultReply, nil
-	}
+	computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - computed.ToSign.Uint64()
```
gupadhyaya: can you also fix this here: https://github.com/harmony-one/harmony/issues/2520

staking/availability/measure.go
rlan35: availability to engine, and engine to availability.

shard/committee/assignment.go
rlan35: yea, snapshot for elected validator is gauranteed to exist.

staking/availability/measure.go
fxfactorial: Ya this cycle hit me so many times too. So annoying

PR #2535, opened 2020-03-19
suppress prompt to the action to clean up old db and logs

AndyBoWu:
read the src and understood the logic, the risk is low to remove this prompt. 

Two safety checks for this action (already there):

* `-c` enabled
* !mainnet network

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -605,7 +605,9 @@ done
 if [[ "${start_clean}" == "true" && "${network_type}" != "mainnet" ]]
 then
    msg "cleaning up old database/logs (-c)"
-   read -rp "Remove old database/logs? (Y/n) " yesno
+   # set a 2s timeout, and set its default return value to Y (true)
+   read -rp -t 2 "Remove old database/logs? (Y/n) " yesno
+   yesno=${yesno:-Y}
    echo
    if [[ "$yesno" == "y" || "$yesno" == "Y" ]]; then
       unset -v backup_dir now


scripts/node.sh
```diff
@@ -604,17 +604,12 @@ done
...
-      backup_dir=$(mktemp -d "backups/${now}.XXXXXX")
-      mv -f harmony_db_* latest .dht* "${backup_dir}/" 2>/dev/null || :
-      rm -rf latest
-   fi
+   unset -v backup_dir now
+   now=$(date -u +%Y-%m-%dT%H:%M:%SZ)
+   mkdir -p backups; rm -rf backups/*
+   backup_dir=$(mktemp -d "backups/${now}.XXXXXX")
+   mv -f harmony_db_* latest .dht* "${backup_dir}/" 2>/dev/null || :
+   rm -rf latest
```
LeoHChen: I suggest to use read with a timeout option.

https://stackoverflow.com/questions/9483633/how-to-use-bash-read-with-a-timeout

AndyBoWu: decent solution, but a single timeout won't solve the problem I believe.

`
-t timeout: cause read to time out and return failure if a complete line of input is not read within timeout seconds.
`

We can set a short time out, like 2 seconds, if timed out, set the default value to true.

read -rp "Remove old database/logs? (Y/n) " yesno
yesno=${yesno:-Y}

This is the ideal solution I can propose at the moment.

LeoHChen: agree.

LeoHChen:
cherry-pick it to t3 please.

PR #2544, opened 2020-03-20
Adding minimum stake for create validator and delegation

gupadhyaya:
Restrictions imposed for create validator and delegation:
* create validator must have 10,000 ONE
* delegation must have at least 1000 ONE

### DIFF
--- a/core/staking_verifier_test.go
+++ b/core/staking_verifier_test.go
@@ -20,6 +20,8 @@ import (
 var (
 	validatorAddress = common.Address(common2.MustBech32ToAddress("one1pdv9lrdwl0rg5vglh4xtyrv3wjk3wsqket7zxy"))
 	postStakingEpoch = big.NewInt(200)
+	tenK             = new(big.Int).Mul(big.NewInt(10000), big.NewInt(1e18))
+	twelveK          = new(big.Int).Mul(big.NewInt(12000), big.NewInt(1e18))
 )
 
 func generateBlsKeySigPair() (shard.BlsPublicKey, shard.BLSSignature) {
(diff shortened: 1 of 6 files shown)


staking/types/validator.go
```diff
@@ -327,6 +336,12 @@ func (w *ValidatorWrapper) SanityCheck(
 				"have %s want %s", w.Delegations[0].Amount.String(), w.Validator.MinSelfDelegation,
 			)
 		}
+		if w.Delegations[0].Amount.Cmp(minimumStake) < 0 {
```
rlan35: just check the MinSelfDelegation is >= 10000ONE, it is checking against 1 ONE now. inside Validator.SanityCheck.

gupadhyaya: updated.

core/staking_verifier_test.go
```diff
@@ -543,21 +545,21 @@ func TestCV32(t *testing.T) {
...
-	// MinSelfDelegation < 1 ONE
-	msg.MinSelfDelegation = big.NewInt(1e18 - 1)
+	statedb.AddBalance(msg.ValidatorAddress, tenK)
+	// MinSelfDelegation < 10,000 ONE
+	msg.MinSelfDelegation = big.NewInt(1e18)
 	if _, err := VerifyAndCreateValidatorFromMsg(
 		statedb, postStakingEpoch, big.NewInt(0), msg,
 	); err == nil {
-		t.Error("expected", "delegation-given 999999999999999999: min_self_delegation has to be greater than 1 ONE", "got", nil)
+		t.Error("expected", "delegation-given 1000000000000000000: min_self_delegation has to be greater than 10,000 ONE", "got", nil)
```
stephen-tse: we can allow 10k OR greater? so that it's easier to create a validator at 10k self delegation, instead of 10k+1

stephen-tse: @john-harmony we should update https://docs.harmony.one/home/validators/first-time-setup/creating-a-validator to have 10k+1? currently, 

```
--validator-addr [ONE ADDRESS] --amount 10 \
--min-self-delegation has to be at least 10,000 ONE.
```

PR #2548, opened 2020-03-20
limiting max bls keys per node to 4 and some node.sh enhancements

gupadhyaya:
* Limits max bls keys per node to 4
* adds feature to read passphrase for bls keys in node.sh
* adds feature to pass bls key folder as input to harmony binary in node.sh

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -110,6 +110,7 @@ var (
 	blsFolder          = flag.String("blsfolder", ".hmy/blskeys", "The folder that stores the bls keys and corresponding passphrases; e.g. <blskey>.key and <blskey>.pass; all bls keys mapped to same shard")
 	blsPass            = flag.String("blspass", "", "The file containing passphrase to decrypt the encrypted bls file.")
 	blsPassphrase      string
+	maxBlsKeysPerNode  = flag.Int("max_bls_keys_per_node", 4, "maximum number of bls keys allowed per node (default 4)")
 	// Sharding configuration parameters for devnet
 	devnetNumShards   = flag.Uint("dn_num_shards", 2, "number of shards for -network_type=devnet (default: 2)")
 	devnetShardSize   = flag.Int("dn_shard_size", 10, "number of nodes per shard for -network_type=devnet (default 10)")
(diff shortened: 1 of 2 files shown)


scripts/node.sh
```diff
@@ -736,6 +746,19 @@ if ! ${multi_key}; then
...
+   for f in ${blsfolder}/*.key 
+   do
+      if [ ! -f $f ]; then
+         err 10 "could not find bls key file: $f"
+      fi
+      passfile=${blsfolder}/$(basename "${f%.*}").pass
+      if [ ! -f "$passfile" ]; then
+         unset -v passphrase
+         read -rsp "Enter passphrase for the BLS key file $f: " passphrase
+         echo ${passphrase} | tee $passfile
```
LeoHChen: this is very risky to save the passphrase to a local file w/o user's awareness.  You need to WARN user about this.  Ask user to confirm. Also, think about removing the file after the node is started.

cmd/harmony/main.go
```diff
@@ -273,6 +273,9 @@ func setupStakingNodeAccount() error {
 	return nil
 }
 
+// MaxNumBLSKeysPerNode specifies the maximum number of bls keys that a node can run
+const MaxNumBLSKeysPerNode = 4
```
LeoHChen: don't hardcode the number here.  This configuration variable could be network specific, we may use different number for testnet/devnet, etc.  So, I think you should add it as a network configuration parameter.

gupadhyaya: you mean in `internal/configs/sharding` or `internal/params/config.go` ?

scripts/node.sh
gupadhyaya: @LeoHChen added notification for user. Is this sufficient? To implement automatic delete of these files, I need to do that inside harmony program, which is not good. Any other way to do it as part of node.sh?

cmd/harmony/main.go
gupadhyaya: @LeoHChen ?

rlan35: as in a main.go flag.

scripts/node.sh
LeoHChen: add a parallel background process to delete the file after started harmony program.

gupadhyaya: done.

scripts/node.sh
```diff
@@ -804,6 +846,7 @@ do
    *) ld_path_var=LD_LIBRARY_PATH;;
    esac
    run() {
+      (sleep 10 && rm_bls_pass)&
```
LeoHChen: use a variable to replace 10 here, I suggest default to 30 sec then.

scripts/node.sh
```diff
@@ -700,6 +710,31 @@ kill_node() {
...
+      if [ ! -f "$passfile" ]; then
+         unset -v passphrase
+         read -rsp "Enter passphrase for the BLS key file $f: " passphrase
+         echo ${passphrase} | tee $passfile
+         echo "Passphrase is temporarely saved to: $passfile, which will be deleted after the node is launched successfully!!!"
+      fi
+   done
+}
+
+rm_bls_pass() {
```
LeoHChen: we can give an option to users that they don't have to remove the passfile if they choose to. then we just give them a warning then.

scripts/node.sh
```diff
@@ -812,5 +855,9 @@ do
    esac || msg "node process finished with status $?"
    ${loop} || break
    msg "restarting in 10s..."
+   rm_bls_pass
    sleep 10
+   if ${multi_key}; then
```
LeoHChen: this won't work, the program will run in line 853 and 854.

gupadhyaya: this is only for the failure case. non failure case is already considered.

scripts/node.sh
gupadhyaya: done.

PR #2557, opened 2020-03-20
T3

john-harmony:
Update T3 with latest master branch for OSTN P2 prep

### DIFF
--- a/shard/shard_state.go
+++ b/shard/shard_state.go
@@ -7,14 +7,17 @@ import (
 	"errors"
 	"math/big"
 	"sort"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/harmony-one/bls/ffi/go/bls"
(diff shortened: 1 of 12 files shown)


shard/shard_state.go
```diff
@@ -382,23 +385,52 @@ func (c *Committee) DeepCopy() Committee {
...
+				if err := c.Slots[j].BlsPublicKey.ToLibBLSPublicKey(
+					committerKey,
+				); err != nil {
+					return nil, err
+				}
+				slice[j] = committerKey
+			}
+			// Only made once
+			go func() {
+				time.Sleep(25 * time.Minute)
```
LeoHChen: why 25 minutes hardcoded?

rlan35: it's just a reasonable time to keep the cached result.

LeoHChen: opened a ticket to keep track of this issue which has to be fixed later.

LeoHChen: https://github.com/harmony-one/harmony/issues/2559

PR #2564, opened 2020-03-21
Fix node sync issue which caused failure to sign and node being kicked out

rlan35:
Node in listening mode doesn't process any consensus message, which make them less likely to stay sync'ed and have to always rely on state sync to happen once per 1 min. Also there wasn't a channel for consensus to trigger state sync in time.

This PR includes the following changes:
1. Always set consensus mode after UpdateConsenusInformation()
2. Still process consensus messages even in Listening mode
3. Don't send back consensus replies in Listening mode
4. Trigger state sync immediate if receiving a committed message that have higher height.
5. Do not print BINGO if the validator is in Listening mode

Tested in OSTN for case of 1) new validator joining consensus; 2) existing validator drop out and join again immediately

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -11,7 +11,7 @@ import (
 const MaxBlockNumDiff = 100
 
 func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
-	consensus.getLogger().Info().
+	consensus.getLogger().Debug().
 		Uint64("blockNum", msg.GetConsensus().BlockNum).
 		Uint64("viewID", msg.GetConsensus().ViewId).
 		Str("msgType", msg.Type.String()).
@@ -47,7 +47,7 @@ func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
 }
 
 func (consensus *Consensus) leaderSanityChecks(msg *msg_pb.Message) bool {
-	consensus.getLogger().Info().
+	consensus.getLogger().Debug().
 		Uint64("blockNum", msg.GetConsensus().BlockNum).
 		Uint64("viewID", msg.GetConsensus().ViewId).
 		Str("msgType", msg.Type.String()).
@@ -196,7 +196,7 @@ func (consensus *Consensus) onPreparedSanityChecks(
 }
 
 func (consensus *Consensus) viewChangeSanityCheck(msg *msg_pb.Message) bool {
-	consensus.getLogger().Info().
+	consensus.getLogger().Debug().
 		Uint64("blockNum", msg.GetConsensus().BlockNum).
 		Uint64("viewID", msg.GetConsensus().ViewId).
 		Str("msgType", msg.Type.String()).
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -174,7 +174,7 @@ func (consensus *Consensus) onPreparedSanityChecks(
 			Msg("[OnPrepared] BlockHash not match")
 		return false
 	}
-	if consensus.current.Mode() == Normal || consensus.current.Mode() == Syncing {
+	if consensus.current.Mode() == Normal {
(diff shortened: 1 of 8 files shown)


consensus/checks.go
```diff
@@ -164,7 +174,7 @@ func (consensus *Consensus) onPreparedSanityChecks(
 			Msg("[OnPrepared] BlockHash not match")
 		return false
 	}
-	if consensus.current.Mode() == Normal {
+	if consensus.current.Mode() == Normal || consensus.current.Mode() == Syncing {
```
rlan35: This bring back the peckshield audit fix. It wasn't this which caused the signing issue.

core/offchain.go
```diff
@@ -192,13 +192,16 @@ func (bc *BlockChain) CommitOffChainData(
...
 	}
-	// Roll up latest crosslinks
-	for i, c := uint32(0), shard.Schedule.InstanceForEpoch(
-		epoch,
-	).NumShards(); i < c; i++ {
-		if err := bc.LastContinuousCrossLink(batch, i); err != nil {
-			utils.Logger().Info().
-				Err(err).Msg("could not batch process last continuous crosslink")
+
+	if isBeaconChain {
```
rlan35: simple fix so shard chain doesn't execute this part which should belong to only beacon chain.

consensus/checks.go
```diff
@@ -11,6 +11,11 @@ import (
 const MaxBlockNumDiff = 100
 
 func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
+	consensus.getLogger().Info().
```
LeoHChen: is it for debug only? if yes, we can use Debug().

consensus/checks.go
```diff
@@ -42,6 +47,11 @@ func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
 }
 
 func (consensus *Consensus) leaderSanityChecks(msg *msg_pb.Message) bool {
+	consensus.getLogger().Info().
```
LeoHChen: same thing, debug only?

core/offchain.go
```diff
@@ -192,13 +192,16 @@ func (bc *BlockChain) CommitOffChainData(
...
+		// Roll up latest crosslinks
+		for i, c := uint32(0), shard.Schedule.InstanceForEpoch(
+			epoch,
+		).NumShards(); i < c; i++ {
+			if err := bc.LastContinuousCrossLink(batch, i); err != nil {
+				utils.Logger().Info().
```
LeoHChen: Use Error() instead of Info()

node/node_syncing.go
```diff
@@ -203,55 +203,62 @@ func (node *Node) DoBeaconSyncing() {
...
-
+	ticker := time.NewTicker(time.Duration(node.syncFreq) * time.Second)
 	// TODO ek – infinite loop; add shutdown/cleanup logic
-SyncingLoop:
 	for {
-		if node.stateSync == nil {
-			node.stateSync = syncing.CreateStateSync(node.SelfPeer.IP, node.SelfPeer.Port, node.GetSyncID())
-			utils.Logger().Debug().Msg("[SYNC] initialized state sync")
+		select {
+		case <-ticker.C:
```
gupadhyaya: may be combine these two cases?

rlan35: There is no easy way to combine select cases.

core/offchain.go
rlan35: Though it's returned in error, but logically it's not a error for the system.. So i will keep as it.

consensus/checks.go
rlan35: done.

consensus/checks.go
rlan35: done.

PR #2579, opened 2020-03-21
Merge master to t3 0321

LeoHChen:
To bring latest master to t3


consensus/checks.go
```diff
@@ -197,6 +197,9 @@ func (consensus *Consensus) onPreparedSanityChecks(
 
 func (consensus *Consensus) viewChangeSanityCheck(msg *msg_pb.Message) bool {
 	consensus.getLogger().Debug().
+		Uint64("blockNum", msg.GetConsensus().BlockNum).
```
rlan35: We should remove this part from master. I only fixed it in t3.

LeoHChen: This PR is for t3, so we are good.

LeoHChen: fixed

Issue #2494, opened 2020-03-13
[node] use viper to support better command line option


LeoHChen:
## Summary

harmony node program use basic flag to define command line option support. We need to revamp it using latest library https://github.com/spf13/viper to support both command line option, plus configuration, and all kinds of configuration options.

## Current Design

Currently, we use basic flags support.

## Problems

No configuration file support, no environmental variable override support etc.

## Proposal

Use https://github.com/spf13/viper to rewrite the configuration option support, better be backward compatible with existing options.

PR #2581, opened 2020-03-22
support config file and environment variable by the viper lib

LeiWWSccc:
## Issue
#2494  : harmony node program use basic flag to define command line option support. We need to revamp it using latest library https://github.com/spf13/viper to support both command line option, plus configuration, and all kinds of configuration options.

Currently, it supports 1) config file like "config.json" to read and overwrite the command line options, also support config sections. 2) environment variables(ENV) to overwrite the command line options, the format is like "HMY_SECTION_FLAG".

Please notice the order of precedence: ENV > config_file value >  command line option > default value. The config value will be _0/ false/ ""_ if the config file or ENV is missing.

### TODO
add more test cases

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -29,6 +29,7 @@ import (
 	"github.com/harmony-one/harmony/internal/common"
 	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
 	shardingconfig "github.com/harmony-one/harmony/internal/configs/sharding"
+	viperconfig "github.com/harmony-one/harmony/internal/configs/viper"
 	"github.com/harmony-one/harmony/internal/genesis"
 	hmykey "github.com/harmony-one/harmony/internal/keystore"
 	"github.com/harmony-one/harmony/internal/memprofiling"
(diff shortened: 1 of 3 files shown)


LeoHChen:
Jenkins, test this please.

cmd/harmony/main.go
```diff
@@ -574,6 +575,68 @@ func setupBlacklist() (map[ethCommon.Address]struct{}, error) {
 	return addrMap, nil
 }
 
+func setupViperConfig() {
+
+	// read from environment
+	envViper := viperconfig.CreateEnvViper()
+
+	//read from config file
+	configFileViper := viperconfig.CreateConfFileViper("config", "json")
```
LeoHChen: is this the config filename?  Can we create the config file in .hmy directory? 
like .hmy/nodeconfig.json

LeiWWSccc: yes, it's the config filename. The new commit can set the config directory and it's in ./hmy/nodeconfig.json

LeoHChen:
Jenkins, test this please.

cmd/harmony/main.go
```diff
@@ -574,6 +575,68 @@ func setupBlacklist() (map[ethCommon.Address]struct{}, error) {
 	return addrMap, nil
 }
 
+func setupViperConfig() {
+
+	// read from environment
+	envViper := viperconfig.CreateEnvViper()
+
+	//read from config file
+	configFileViper := viperconfig.CreateConfFileViper("$GOPATH/src/github.com/harmony-one/harmony/.hmy", "nodeconfig", "json")
```
LeoHChen: don't use GOPATH, just use the current directory .hmy subfolder.

LeiWWSccc: yes, changed to relative path .hmy and if it miss the default value is "", so it won't trigger some errors.

PR #2582, opened 2020-03-23
Fix nil pointer error

rlan35:
same has been done in t3.

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -197,9 +197,6 @@ func (consensus *Consensus) onPreparedSanityChecks(
 
 func (consensus *Consensus) viewChangeSanityCheck(msg *msg_pb.Message) bool {
 	consensus.getLogger().Debug().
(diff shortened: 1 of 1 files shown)


consensus/checks.go
```diff
@@ -197,9 +197,6 @@ func (consensus *Consensus) onPreparedSanityChecks(
 
 func (consensus *Consensus) viewChangeSanityCheck(msg *msg_pb.Message) bool {
 	consensus.getLogger().Debug().
-		Uint64("blockNum", msg.GetConsensus().BlockNum).
-		Uint64("viewID", msg.GetConsensus().ViewId).
-		Str("msgType", msg.Type.String()).
```
gupadhyaya: you can still use the msg.GetType().String() right?

rlan35: yea, but it's not very necessary. Just to be safe, let's make it same as T3.

PR #2585, opened 2020-03-24
adding log to print out the current view id

gupadhyaya:
* Enhanced log to print out the consensus current view id
* Also make sure to check only if the current view id is not zero. Looks like when the consensus restarts the current view id is set to zero.

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -242,8 +242,8 @@ func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
 	if recvMsg.ViewID-consensus.current.ViewID() > MaxViewIDDiff {
 		consensus.getLogger().Debug().
 			Uint64("MsgViewID", recvMsg.ViewID).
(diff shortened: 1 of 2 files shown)


consensus/checks.go
```diff
@@ -239,11 +239,11 @@ func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
 			Msg("[onViewChange] ViewChanging ID Is Low")
 		return false
 	}
-	if recvMsg.ViewID-consensus.current.ViewID() > MaxViewIDDiff {
+	if consensus.current.ViewID() != 0 && recvMsg.ViewID-consensus.current.ViewID() > MaxViewIDDiff {
```
rlan35: better fix in the right way. set current.viewID along with consensus.viewID's setting. In here:

// SetViewID set the viewID to the height of the blockchain
func (consensus *Consensus) SetViewID(height uint64) {
	consensus.viewID = height
}

PR #2586, opened 2020-03-24
Fix APR's nill header issue; Compute APR based on last epoch

rlan35:
The issue why previously the header of last block of one epoch ago is always nil, is because the last epoch is actually set to current epoch whose last block is actually the current block to commit, thus it can not be found in blockchain.

Also hit an issue that when external node uses same BLS keys are harmony nodes, the logic is messed up, external nodes may get reward because harmony node signed the block.

Tested locally.

### DIFF
--- a/consensus/votepower/roster.go
+++ b/consensus/votepower/roster.go
@@ -3,6 +3,7 @@ package votepower
 import (
 	"encoding/hex"
 	"encoding/json"
+	"github.com/harmony-one/harmony/internal/utils"
 	"math/big"
 	"sort"
 
@@ -207,6 +208,8 @@ func Compute(subComm *shard.Committee, epoch *big.Int) (*Roster, error) {
 		// TODO: make sure external user's BLS key can be same as harmony's bls keys
 		if _, ok := roster.Voters[staked[i].BlsPublicKey]; !ok {
 			roster.Voters[staked[i].BlsPublicKey] = &member
+		} else {
+			utils.Logger().Debug().Str("blsKey", staked[i].BlsPublicKey.Hex()).Msg("Duplicate BLS key found")
 		}
 	}
 
--- a/consensus/votepower/roster.go
+++ b/consensus/votepower/roster.go
@@ -204,7 +204,10 @@ func Compute(subComm *shard.Committee, epoch *big.Int) (*Roster, error) {
 			ourPercentage = ourPercentage.Add(member.OverallPercent)
 		}
 
-		roster.Voters[staked[i].BlsPublicKey] = &member
+		// TODO: make sure external user's BLS key can be same as harmony's bls keys
+		if _, ok := roster.Voters[staked[i].BlsPublicKey]; !ok {
+			roster.Voters[staked[i].BlsPublicKey] = &member
+		}
 	}
 
 	// NOTE Enforce voting power sums to one,
--- a/staking/apr/compute.go
+++ b/staking/apr/compute.go
@@ -1,11 +1,12 @@
 package apr
 
 import (
+	"github.com/harmony-one/harmony/core/types"
+	"github.com/harmony-one/harmony/shard"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/harmony-one/harmony/block"
-	"github.com/harmony-one/harmony/core/state"
 	"github.com/harmony-one/harmony/internal/params"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/numeric"
@@ -15,6 +16,7 @@ import (
 
 // Reader ..
 type Reader interface {
+	GetHeaderByNumber(number uint64) *block.Header
 	Config() *params.ChainConfig
 	GetHeaderByHash(hash common.Hash) *block.Header
 	// GetHeader retrieves a block header from the database by hash and number.
@@ -27,7 +29,7 @@ type Reader interface {
 }
 
 const (
-	secondsInYear = int64(31_557_600)
+	secondsInYear = int64(31557600)
 )
 
 var (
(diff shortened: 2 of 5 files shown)


stephen-tse:
looks good. let's inline these insights into the code?

"The issue why previously the header of last block of one epoch ago is always nil, is because the last epoch is actually set to current epoch whose last block is actually the current block to commit, thus it can not be found in blockchain.

Also hit an issue that when external node uses same BLS keys are harmony nodes, the logic is messed up, external nodes may get reward because harmony node signed the block.

consensus/votepower/roster.go
```diff
@@ -204,7 +204,10 @@ func Compute(subComm *shard.Committee, epoch *big.Int) (*Roster, error) {
 			ourPercentage = ourPercentage.Add(member.OverallPercent)
 		}
 
-		roster.Voters[staked[i].BlsPublicKey] = &member
+		// TODO: make sure external user's BLS key can be same as harmony's bls keys
+		if _, ok := roster.Voters[staked[i].BlsPublicKey]; !ok {
```
LeoHChen: do you need to print out a log  if the abnormal situation is found?

staking/apr/compute.go
```diff
@@ -27,22 +29,21 @@ type Reader interface {
 }
 
 const (
-	secondsInYear = int64(31_557_600)
+	secondsInYear = int64(31557600)
```
LeoHChen: why change this line?

staking/apr/compute.go
```diff
@@ -55,152 +56,81 @@ func expectedRewardPerYear(
...
-
-	// result := numeric.NewDecFromBigInt(estimatedRewardPerYear).Quo(
-	// 	total,
-	// )
-	// return &result, nil
+	oneEpochAgo, zero :=
+		new(big.Int).Sub(block.Epoch(), common.Big1),
+		numeric.ZeroDec()
+
+	utils.Logger().Info().
```
LeoHChen: debug?

staking/apr/compute.go
```diff
@@ -55,152 +56,81 @@ func expectedRewardPerYear(
...
-	// total := numeric.NewDecFromBigInt(validatorNow.TotalDelegation())
-	// if total.IsZero() {
-	// 	return nil, errors.New("zero total delegation will cause div by zero")
-	// }
-
-	// result := numeric.NewDecFromBigInt(estimatedRewardPerYear).Quo(
-	// 	total,
-	// )
-	// return &result, nil
+	oneEpochAgo, zero :=
```
LeoHChen: do you need to check if oneEpochAgo is greater than 0?

staking/apr/compute.go
```diff
@@ -55,152 +56,81 @@ func expectedRewardPerYear(
...
+	)
+
+	if err != nil {
+		return &zero, nil
+	}
+
+	blockNumAtOneEpochAgo := shard.Schedule.EpochLastBlock(oneEpochAgo.Uint64())
+
+	headerOneEpochAgo := bc.GetHeaderByNumber(blockNumAtOneEpochAgo)
+	if block.Header() == nil || headerOneEpochAgo == nil || err != nil {
```
gupadhyaya: `err != nil` already checked?

staking/apr/compute.go
```diff
@@ -55,152 +56,81 @@ func expectedRewardPerYear(
...
+		Uint64("one-epoch-ago", oneEpochAgo.Uint64()).
+		Msg("apr - begin compute for validator ")
+
+	oneSnapshotAgo, err := bc.ReadValidatorSnapshotAtEpoch(
+		oneEpochAgo,
+		validatorNow.Address,
+	)
+
+	if err != nil {
+		return &zero, nil
```
gupadhyaya: why discard the err?

staking/apr/compute.go
```diff
@@ -55,152 +56,81 @@ func expectedRewardPerYear(
...
+		RawJSON("one-epoch-ago-header", []byte(headerOneEpochAgo.String())).
+		Msg("headers used for apr computation")
+
+	estimatedRewardPerYear, err := expectedRewardPerYear(
+		block.Header(), headerOneEpochAgo,
+		validatorNow, oneSnapshotAgo,
+	)
+
+	if err != nil {
+		return nil, err
```
gupadhyaya: why return `nil` over `&zero` ? I am bit confused about the return value specification.

staking/apr/compute.go
```diff
@@ -27,22 +29,21 @@ type Reader interface {
...
+	now, oneEpochAgo *block.Header,
+	curValidator, snapshotLastEpoch *staking.ValidatorWrapper,
 ) (*big.Int, error) {
-	oneTAgo, twoTAgo := oneEpochAgo.Time(), twoEpochAgo.Time()
+	timeNow, oneTAgo := now.Time(), oneEpochAgo.Time()
 	diffTime, diffReward :=
-		new(big.Int).Sub(twoTAgo, oneTAgo),
-		new(big.Int).Sub(twoSnapshotAgo.BlockReward, oneSnapshotAgo.BlockReward)
+		new(big.Int).Sub(timeNow, oneTAgo),
+		new(big.Int).Sub(curValidator.BlockReward, snapshotLastEpoch.BlockReward)
```
gupadhyaya: do we need to take the absolute value here? as `diffReward` can be negative?

consensus/votepower/roster.go
rlan35: done.

staking/apr/compute.go
rlan35: it doesn't compile well in my IDE. Guess the style is not recognized universally

staking/apr/compute.go
rlan35: done

staking/apr/compute.go
rlan35: no, it should be and this only happens in staking epoch. And if not, the next part of the code to fetch snapshot will return nothing and it just returns.

staking/apr/compute.go
rlan35: good catch.

staking/apr/compute.go
rlan35: because this error should be a non-blocking error. We don't want to fail the block commitment because of this.

rlan35: yea, we shouldn't. Fixed.

staking/apr/compute.go
rlan35: fixed. the errors should be returned.

staking/apr/compute.go
rlan35: diffReward can not. because reward is increasing only.

gupadhyaya: true. I got confused with block reward.

PR #2593, opened 2020-03-24
[slashing][audit] Refactor slashing code and fix audit todos

rlan35:
Refactor slash detection code.

Fix previous audit todos:
 1. Group slash records by blocks and apply separately.
 2. Use the epoch at double sign rather than at current epoch.

### DIFF
--- a/consensus/double_sign.go
+++ b/consensus/double_sign.go
@@ -105,13 +105,13 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) {
 func (consensus *Consensus) couldThisBeADoubleSigner(
 	recvMsg *FBFTMessage,
 ) bool {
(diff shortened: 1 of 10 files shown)


consensus/double_sign.go
```diff
@@ -0,0 +1,119 @@
...
+			}
+		}
+		return
+	}
+}
+
+func (consensus *Consensus) couldThisBeADoubleSigner(
+	recvMsg *FBFTMessage,
+) bool {
+	num, hash, now := consensus.blockNum, recvMsg.BlockHash, consensus.blockNum
```
gupadhyaya: `num` and `now` both not needed.

rlan35: removed now. num is needed still.

PR #2610, opened 2020-03-26
Return if potential double sign

rlan35:
This is consistent with the previous logic.

Since the check didn't return on the OnCommit(). The double signed msg pass through to the code below and actually got accepted into the votes and overwritten the original good signature, causing the final aggregated signature to be invalid and the validator won't be able to verify it.

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -203,6 +203,11 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 	consensus.mutex.Lock()
 	defer consensus.mutex.Unlock()
 
+	// Check for potential double signing
+	if consensus.checkDoubleSign(recvMsg) {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


consensus/leader.go
```diff
@@ -203,6 +203,11 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 	consensus.mutex.Lock()
 	defer consensus.mutex.Unlock()
 
+	// Check for potential double signing
+	if consensus.checkDoubleSign(recvMsg) {
+		return
+	}
+
```
gupadhyaya: can you briefly explain why accepting the double signed validator's vote brings the shard down?

rlan35: because since the check didn't return on the OnCommit(). The double signed msg pass through to the code below and actually got accepted into the votes and overwritten the original good signature, causing the final aggregated signature to be invalid and the validator won't be able to verify it.

PR #2612, opened 2020-03-26
Fix the validator auction eligibility condition

rlan35:
This original condition to check whether a validator is in last committee depends on whether the counter of NumBlocksToSign increased in last epoch. This condition is not stable because cross-links may arrive after the epoch ends and it still gets counted into the NumBlocksToSign, making this condition to be true after the new epoch even when the validator is actually not in committee.

### DIFF
--- a/shard/committee/assignment.go
+++ b/shard/committee/assignment.go
@@ -127,7 +127,7 @@ func prepareOrders(
 		if err != nil {
 			return nil, err
 		}
-		if !IsEligibleForEPoSAuction(snapshot, validator) {
+		if !IsEligibleForEPoSAuction(snapshot, validator, stakedReader.CurrentBlock().Epoch()) {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


shard/committee/assignment.go
```diff
@@ -169,8 +169,14 @@ func prepareOrders(
 }
 
 // IsEligibleForEPoSAuction ..
-func IsEligibleForEPoSAuction(snapshot, validator *staking.ValidatorWrapper) bool {
-	if snapshot.Counters.NumBlocksToSign.Cmp(validator.Counters.NumBlocksToSign) != 0 {
+func IsEligibleForEPoSAuction(snapshot, validator *staking.ValidatorWrapper, curEpoch *big.Int) bool {
+	// This original condition to check whether a validator is in last committee is not stable
```
stephen-tse: awesome to inline more discussion on design choices and context!

PR #2621, opened 2020-03-28
Set block reward to flat-rate 28 ONE

rlan35:
Simply removed the logic to do the dynamic adjustment and use a flat rate reward of 28 ONE per block.


gupadhyaya:
Is there an impact of this elsewhere? clients?

rlan35:
> @rlan35 Is there an impact of this elsewhere? clients?

no other impact other than the reward amount is different now.

Issue #2620, opened 2020-03-21
wrong error message for trying to undelegate fully from your validator


scraido:
**Describe the bug**

EDITED by mindstyle: this user tried to undelegate his full amount from his validator, which ofcourse isnt working as there is a min-self delegation that cannot be undelegated.

**Expected behavior**
it should when you can't do unstake the message is related to why you couldn't do it if you have balance staked, instead of show that you didn't have the balance that you own and want to unstake

**Screenshots**
here the CLI picture update, using enter, enter and without passphrase
![with-without](https://user-images.githubusercontent.com/44559893/77221640-f4c72f80-6b7d-11ea-9758-cfad2ca398e4.png)

but dashboard still show it;
![dashboardshow](https://user-images.githubusercontent.com/44559893/77221707-bbdb8a80-6b7e-11ea-8a74-0498ccc829d7.png)

**Additional context**
This really a serious issue, how if that happen to your 10mm? when you try to undelegate using CLI, it show messages; have 0 want 1xxxx, it's can make heart attack, how to report if this happen on mainnet, lucky on my case we still used testnet token and also on dashboard it still show the messages

thankyou

mindstyle85:
@janet-harmony can we close this one? seems that with the newer versions, validators are able to undelegate fully anyways?

janet-harmony:
Confirmed that full undelegation works on latest OSTN version.

PR #2622, opened 2020-03-28
Update undelegate error message & fix overflow for 'current-epoch-blocks-left'

janet-harmony:
Fixes #2530 & #2620 .

Tested locally, but unable to reproduce overflow error on localnet.

New error message:
```
[Undelegate] min_self_delegation 10000000000000000000000, after undelegation amount 0: self delegation can not be less than min_self_delegation -- 2020-03-27 15:48:40 -0700 PDT
```

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -370,7 +370,8 @@ func (b *APIBackend) GetValidatorInformation(
 	computed := availability.ComputeCurrentSigning(
 		snapshot, wrapper,
 	)
-	computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - computed.ToSign.Uint64()
+	beaconChainBlocks := uint64(b.hmy.BeaconChain().CurrentBlock().Header().Number().Int64()) % shard.Schedule.BlocksPerEpoch()
+	computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - beaconChainBlocks
 
 	stats, err := bc.ReadValidatorStats(addr)
 	if err != nil {
--- a/staking/types/validator.go
+++ b/staking/types/validator.go
@@ -331,7 +331,8 @@ func (w *ValidatorWrapper) SanityCheck(
 			w.Delegations[0].Amount.Cmp(w.Validator.MinSelfDelegation) < 0 {
 			return errors.Wrapf(
 				errInvalidSelfDelegation,
-				"have %s want %s", w.Delegations[0].Amount.String(), w.Validator.MinSelfDelegation,
+				"min_self_delegation %s, after undelegation amount %s",
+				w.Validator.MinSelfDelegation, w.Delegations[0].Amount.String(),
 			)
 		}
 	}
--- a/staking/types/validator.go
+++ b/staking/types/validator.go
@@ -331,7 +331,7 @@ func (w *ValidatorWrapper) SanityCheck(
 			w.Delegations[0].Amount.Cmp(w.Validator.MinSelfDelegation) < 0 {
 			return errors.Wrapf(
 				errInvalidSelfDelegation,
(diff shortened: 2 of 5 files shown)


staking/types/validator.go
```diff
@@ -95,7 +95,7 @@ type ValidatorWrapper struct {
 type Computed struct {
 	Signed            *big.Int    `json:"current-epoch-signed"`
 	ToSign            *big.Int    `json:"current-epoch-to-sign"`
-	BlocksLeftInEpoch uint64      `json:"current-epoch-blocks-left"`
+	BlocksLeftInEpoch uint64      `json:"num-beacon-blocks-until-next-epoch"`
```
gupadhyaya: why is this name change? from `current-epoch-blocks-left` to `num-beacon-blocks-until-next-epoch`? I don't like the new name.

hmy/api_backend.go
```diff
@@ -370,7 +370,8 @@ func (b *APIBackend) GetValidatorInformation(
 	computed := availability.ComputeCurrentSigning(
 		snapshot, wrapper,
 	)
-	computed.BlocksLeftInEpoch = shard.Schedule.BlocksPerEpoch() - computed.ToSign.Uint64()
+	beaconChainBlocks := uint64(b.hmy.BeaconChain().CurrentBlock().Header().Number().Int64()) % shard.Schedule.BlocksPerEpoch()
```
gupadhyaya: I think the overflow specifically happens for the multi-bls case where number of blocks to sign is more than schedule.BlocksPerEpoch(). I am not sure how this will fix the problem.

staking/types/validator.go
```diff
@@ -331,7 +331,8 @@ func (w *ValidatorWrapper) SanityCheck(
 			w.Delegations[0].Amount.Cmp(w.Validator.MinSelfDelegation) < 0 {
 			return errors.Wrapf(
 				errInvalidSelfDelegation,
-				"have %s want %s", w.Delegations[0].Amount.String(), w.Validator.MinSelfDelegation,
+				"min_self_delegation %s, after undelegation amount %s",
```
gupadhyaya: message should not say `undelegation` right? as `SanityCheck` is also called in other staking messages like Delegate, Create, Edit, and CollectReward. May be just print min_self_delegation, and amount.

hmy/api_backend.go
gupadhyaya: may be multiply `shard.Schedule.BlocksPerEpoch()` with number of bls keys? how about `len(wrapper.Validator.SlotPubKeys)`?

gupadhyaya: why not just get the blockchain current block header?

staking/types/validator.go
janet-harmony: Changing the name to reflect the new meaning of the field.
Blocks left per epoch doesn't mean anything for non-beacon chain, since the number of blocks on auxiliary shards are dependent on the speed of beacon chain relative to the other shards.
Calculating number of blocks left on Beaconchain until next epoch instead of how many left depending on blocks signed by the validator.

hmy/api_backend.go
janet-harmony: Want to compute blocks left on Beaconchain until next epoch.

staking/types/validator.go
janet-harmony: Changed.

staking/types/validator.go
gupadhyaya: would this impact staking explorer? @mikedoan do you agree on this change?

LeoHChen: generally, we shouldn't change the API like naming once we have downstream users like dashboard/explorer unless consumers agreed and made corresponding changes.

rlan35: this field is just an informational field for validators. It's not used for any downstream programs like UI yet.

PR #2626, opened 2020-03-29
[state] Do not add reward to banned validator

fxfactorial:
Early return if validator is banned

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -29,8 +29,10 @@ import (
 	"github.com/ethereum/go-ethereum/trie"
 	"github.com/harmony-one/harmony/core/types"
 	common2 "github.com/harmony-one/harmony/internal/common"
+	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/numeric"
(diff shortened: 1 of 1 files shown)


core/state/statedb.go
```diff
@@ -748,6 +750,16 @@ func (db *DB) IsValidator(addr common.Address) bool {
 
 // AddReward distributes the reward to all the delegators based on stake percentage.
 func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int) error {
+	wrapper, err := db.ValidatorWrapper(snapshot.Address)
```
gupadhyaya: there is a duplicate check below, which can be moved up here right?

core/state/statedb.go
```diff
@@ -748,6 +750,16 @@ func (db *DB) IsValidator(addr common.Address) bool {
...
+	if err != nil {
+		return err
+	}
+
+	if wrapper.Status == effective.Banned {
+		utils.Logger().Info().Msg("cannot add reward to banned validator")
```
gupadhyaya: provide more information, may be validator address?

PR #2627, opened 2020-03-29
[slash] Quick hash sanity check for conflicting ballots

fxfactorial:
From audit-1, another check on slash

### DIFF
--- a/staking/slash/double-sign.go
+++ b/staking/slash/double-sign.go
@@ -177,7 +177,9 @@ func Verify(
 		)
 	}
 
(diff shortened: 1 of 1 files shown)


staking/slash/double-sign.go
```diff
@@ -217,6 +219,15 @@ func Verify(
 		return err
 	}
 
+	// last ditch check
+	if hash.FromRLPNew256(
+		candidate.Evidence.AlreadyCastBallot,
+	) == hash.FromRLPNew256(
+		candidate.Evidence.DoubleSignedBallot,
+	) {
+		return errBallotsNotDiff
```
gupadhyaya: may be useful to wrap it with bls key that conflict?

PR #2636, opened 2020-03-30
[validator] Cache harmony keys in a closure

fxfactorial:
Can easily avoid this loop, hide state in closure. 

// TODO Clear out the cache map of epochs more than 5 ago, or something like that.

### DIFF
--- a/staking/types/validator.go
+++ b/staking/types/validator.go
@@ -471,31 +471,55 @@ func VerifyBLSKey(pubKey *shard.BlsPublicKey, pubKeySig *shard.BLSSignature) err
 	return nil
 }
 
-func containsHarmonyBlsKeys(blsKeys []shard.BlsPublicKey,
-	hmyAccounts []genesis.DeployAccount) error {
+func containsHarmonyBLSKeys(
+	blsKeys []shard.BlsPublicKey,
+	hmyAccounts []genesis.DeployAccount,
+	epoch *big.Int,
+) error {
 	for i := range blsKeys {
-		if err := matchesHarmonyBlsKey(&blsKeys[i], hmyAccounts); err != nil {
+		if err := matchesHarmonyBLSKey(
+			&blsKeys[i], hmyAccounts, epoch,
+		); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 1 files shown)


staking/types/validator.go
```diff
@@ -471,31 +471,55 @@ func VerifyBLSKey(pubKey *shard.BlsPublicKey, pubKeySig *shard.BLSSignature) err
...
-				"slot key %#v conflicts with internal keys",
-				hmyAccounts[i].BlsPublicKey,
+func matchesHarmonyBLSKey(
+	blsKey *shard.BlsPublicKey,
+	hmyAccounts []genesis.DeployAccount,
+	epoch *big.Int,
+) error {
+	type publicKeyAsHex = string
+	cache := map[string]map[publicKeyAsHex]struct{}{}
+	return func() error {
```
gupadhyaya: do we need more anonymous functions?

staking/types/validator.go
```diff
@@ -471,31 +471,55 @@ func VerifyBLSKey(pubKey *shard.BlsPublicKey, pubKeySig *shard.BLSSignature) err
...
-func containsHarmonyBlsKeys(blsKeys []shard.BlsPublicKey,
-	hmyAccounts []genesis.DeployAccount) error {
+func containsHarmonyBLSKeys(
+	blsKeys []shard.BlsPublicKey,
+	hmyAccounts []genesis.DeployAccount,
+	epoch *big.Int,
+) error {
 	for i := range blsKeys {
-		if err := matchesHarmonyBlsKey(&blsKeys[i], hmyAccounts); err != nil {
+		if err := matchesHarmonyBLSKey(
```
gupadhyaya: how about also cache across this outer loop?

staking/types/validator.go
fxfactorial: yes, to use the closure

staking/types/validator.go
fxfactorial: wrong to cache on those because they came from the validator off the street, otherwise cache will explode. Our cache will be always finite and known size

PR #2639, opened 2020-03-30
[slash] Adjust rate as function of voting power of double-signers rat…

fxfactorial:
Instead of being a flat % for the slashing rate now the slashing rate is a summation of all double signer's computed voting power.

### DIFF
--- a/internal/chain/engine.go
+++ b/internal/chain/engine.go
@@ -428,6 +428,7 @@ func applySlashes(
 		return false
 	})
 
+	beaconCurrentEpoch := chain.CurrentHeader().Epoch()
 	// Do the slashing by groups in the sorted order
 	for _, key := range sortedKeys {
 		records := groupedRecords[key]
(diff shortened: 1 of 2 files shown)


staking/slash/double-sign.go
```diff
@@ -494,16 +494,24 @@ func Apply(
...
 	}
-	switch doubleSignerCount {
-	case 1:
-		return oneDoubleSignerRate
-	default:
-		return numeric.NewDec(
-			int64(doubleSignerCount),
-		).Quo(numeric.NewDec(int64(committeeSize)))
+
+	if rate.Equal(zero) {
```
rlan35, line 512: This condition should be rate < oneDoubleSignerRate

internal/chain/engine.go
```diff
@@ -428,6 +428,7 @@ func applySlashes(
 		return false
 	})
 
+	beaconCurrentEpoch := chain.CurrentHeader().Epoch()
```
rlan35, line 431: this is not right. Should use the key.epoch for the doublesigner groups.

PR #2649, opened 2020-03-31
[Core + API + Node] Add raw tx size limitation

Daniel-VDM:
Implementation of https://github.com/harmony-one/harmony/pull/2382 on s3

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -171,7 +171,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	AccountQueue: 64,
 	GlobalQueue:  1024,
 
-	Lifetime: 3 * time.Hour,
+	Lifetime: 30 * time.Minute,
 
 	Blacklist: &map[common.Address]struct{}{},
 }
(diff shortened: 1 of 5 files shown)


node/node_handler.go
```diff
@@ -211,6 +219,14 @@ func (node *Node) transactionMessageHandler(msgPayload []byte) {
 }
 
 func (node *Node) stakingMessageHandler(msgPayload []byte) {
+	if len(msgPayload) >= types.MaxEncodedPoolTransactionSize {
+		utils.Logger().Warn().Err(core.ErrOversizedData).Msgf("encoded tx size: %d", len(msgPayload))
+		return
+	}
+	if len(msgPayload) < 1 {
```
denniswon: Len equal to 0 check is better as uint

core/tx_pool.go
```diff
@@ -623,8 +623,8 @@ func (pool *TxPool) local() map[common.Address]types.Transactions {
 // validateTx checks whether a transaction is valid according to the consensus
 // rules and adheres to some heuristic limits of the local node (price and size).
 func (pool *TxPool) validateTx(tx *types.Transaction, local bool) error {
-	// Heuristic limit, reject transactions over 32KB to prevent DOS attacks
-	if tx.Size() > 32*1024 {
+	// For DOS prevention, reject excessively large transactions.
+	if tx.Size() >= types.MaxPoolTransactionDataSize {
```
denniswon: Strictly Bigger than

PR #2653, opened 2020-03-31
[committee] Use epoch aware snapshot during median-stake computation,…

fxfactorial:


fxfactorial:
actually, nvm not needed

PR #2659, opened 2020-04-01
[node.sh] enable p2p tracefile using -R option

LeoHChen:


LeoHChen:
cherry-pick the change from s3 branch.  @SebastianJ , I have to change pprof option to use '-r'.

PR #2692, opened 2020-04-03
[go.mod][pubsub][test] Update libp2p, use MaxMessage at p2p level protection, remove dead code 

fxfactorial:


fxfactorial:
They are not used or outdated or misleading

PR #2693, opened 2020-04-03
[offchain] fix earned-reward per key, aggregate and write to batch once 

gupadhyaya:


core/offchain.go
```diff
@@ -12,6 +12,7 @@ import (
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/shard"
 	"github.com/harmony-one/harmony/staking/slash"
+	types2 "github.com/harmony-one/harmony/staking/types"
```
fxfactorial, line 15: namespace this as `staking`, not `types2`

core/offchain.go
```diff
@@ -214,36 +215,40 @@ func (bc *BlockChain) CommitOffChainData(
...
 						}
-					} else {
-						utils.Logger().Info().Err(err).
-							Str("bls-earning-key", paid[i].EarningKey.Hex()).
-							Msg("could not read validator stats to update for earning per key")
 					}
+
+				}
+			}
+			for addr, stats := range tempValidatorStats {
```
fxfactorial, line 245: I know that order doesn't matter here in principle but I want to always enforce reproduciblity so please make a slice of vlaidaotr slice, map to slice, sort.SliceStable the slice, then loop over and write

PR #2724, opened 2020-04-04
[state] ValidatorWrapper div zero panic fix - 

fxfactorial:
I think could be 0% commision rate and better safe than sorry error on 0 total delegation

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -749,19 +749,15 @@ func (db *DB) IsValidator(addr common.Address) bool {
 }
 
 var (
-	zero                      = numeric.ZeroDec()
-	errTotalDelegationDivZero = errors.New("total delegation cannot be 0")
-	errZeroReward             = errors.New("reward cannot be zero")
+	zero = numeric.ZeroDec()
 )
 
 // AddReward distributes the reward to all the delegators based on stake percentage.
 func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int) error {
 	if reward.Cmp(common.Big0) == 0 {
-		return errors.Wrapf(
-			errZeroReward,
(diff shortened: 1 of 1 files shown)


core/state/statedb.go
```diff
@@ -748,8 +748,22 @@ func (db *DB) IsValidator(addr common.Address) bool {
...
 
+var (
+	zero                      = numeric.ZeroDec()
+	errTotalDelegationDivZero = errors.New("total delegation cannot be 0")
+	errZeroReward             = errors.New("reward cannot be zero")
+)
+
 // AddReward distributes the reward to all the delegators based on stake percentage.
 func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int) error {
+	if reward.Cmp(common.Big0) == 0 {
```
rlan35, line 759: Don't return error. If it's 0. it's 0. just do nothing.

core/state/statedb.go
```diff
@@ -765,15 +779,27 @@ func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int) error {
...
+		rewardPool.Sub(rewardPool, commissionInt)
+	}
 	totalRewardForDelegators := big.NewInt(0).Set(rewardPool)
 	// Payout each delegator's reward pro-rata
 	totalDelegationDec := numeric.NewDecFromBigInt(snapshot.TotalDelegation())
 	for i := range snapshot.Delegations {
 		delegation := snapshot.Delegations[i]
 		// NOTE percentage = <this_delegator_amount>/<total_delegation>
+		if totalDelegationDec.Equal(zero) {
+			return errors.Wrapf(
```
rlan35, line 797: Same here. If the delegations are all 0. Just do nothing.

PR #2737, opened 2020-04-05
fix earned-reward by writing the stats only once, remove from UpdateV…

gupadhyaya:
remove WriteValidatorStats from UpdateValidatorVotingPower and write the stats only once while updating reward.

### DIFF
--- a/core/offchain.go
+++ b/core/offchain.go
@@ -134,26 +134,6 @@ func (bc *BlockChain) CommitOffChainData(
 		}
 	}
 
-	// Update voting power of validators for all shards
-	if isNewEpoch && isBeaconChain {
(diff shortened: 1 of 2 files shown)


core/offchain.go
```diff
@@ -213,6 +193,29 @@ func (bc *BlockChain) CommitOffChainData(
 		}
 	}
 
+	// Update voting power of validators for all shards
+	tempValidatorStats := map[common.Address]*staking.ValidatorStats{}
```
fxfactorial, line 197: Why did this block move?

gupadhyaya: just to keep it closer to the two places where the stats are used.

Issue #2741, opened 2020-04-06
View change fail due to voting power always being 0


rlan35:
{"level":"info","port":"9000","ip":"3.19.143.176","policy":"SuperMajorityStake","phase":"viewChange","threshold":"0.666666666666666667","total-power-of-signers":"0.000000000000000000","caller":"/home/leochen/go/src/github.com/harmony-one/harmony/consensus/quorum/one-node-staked-vote.go:70","time":"2020-04-05T04:56:59.220972237Z","message":"Attempt to reach quorum"}

When view change happens, The new leader receives ViewChange messages, I expected the voting-power should increase. but it's always 0

PR #2758, opened 2020-04-06
[consensus][viewchange] fix total power of signers in IsQuorumAchieved for view change

gupadhyaya:
fixes https://github.com/harmony-one/harmony/issues/2741

didn't want to completely get rid of quorum.ViewChange, as signers total voting power information could be useful. Further, removing the option  quorum.ViewChange may lead to regression.

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -357,7 +357,7 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 			binary.LittleEndian.PutUint64(blockNumBytes[:], consensus.blockNum)
 			commitPayload := append(blockNumBytes[:], consensus.blockHash[:]...)
 			if _, err := consensus.Decider.SubmitVote(
-				quorum.Commit,
+				quorum.ViewChange,
 				newLeaderKey,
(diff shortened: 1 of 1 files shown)


fxfactorial:
This is a change of original view change logic. Dig deeper into the git commit history of initial usage of submitvote to see this was original value used. I also was confused by this initially but didn't want to deviate from original code.

rlan35:
if consensus.Decider.IsQuorumAchieved(quorum.ViewChange) {
            Int64("have", consensus.Decider.SignersCount(quorum.ViewChange)).
            Int64("need", consensus.Decider.TwoThirdsSignersCount()).
            Str("validatorPubKey", recvMsg.SenderPubkey.SerializeToHexStr()).
            Msg("[onViewChange] Received Enough View Change Messages")
        return
    }

this part was a no-op, because it never reach quorum. But it wasn't a problem because after that it's checked by Mask.
So to me, there wasn't a problem maybe. Just in the last time view change failed (maybe due to not enough votes). And when I look at the log, I saw the 0 voting power thing in IsQuorumAchieved. But it isn't the cause of view change failure.

PR #2761, opened 2020-04-06
fix block proposal ordering; other offchain commits change

rlan35:
Move EPoS Status check before the AccumulateRewards to be consistent with block proposal state.

Remove some dead code on delete validator snapshot.

Fix InCommittee condition check.

Add div by 0 safety guard to avoid crash.

Do no compute uptime for EPoS Booted status. It's not consistent with that for block proposal. Instead just use Inactive.

More cleanup on common.Big0 usage.

### DIFF
--- a/core/offchain.go
+++ b/core/offchain.go
@@ -92,26 +92,17 @@ func (bc *BlockChain) CommitOffChainData(
 	//	}
 	//}
 
-	// Do bookkeeping for new staking txns
-	newVals, err := bc.UpdateStakingMetaData(
-		batch, block.StakingTransactions(), state, epoch, isNewEpoch,
-	)
-	if err != nil {
-		utils.Logger().Err(err).Msg("UpdateStakingMetaData failed")
-		return NonStatTy, err
-	}
-
+	newEpoch := new(big.Int).Add(header.Epoch(), common.Big1)
 	// Shard State and Validator Update
 	if isNewEpoch {
 		// Write shard state for the new epoch
-		epoch := new(big.Int).Add(header.Epoch(), common.Big1)
 		shardState, err := block.Header().GetShardState()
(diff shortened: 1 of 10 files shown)


staking/availability/measure.go
```diff
@@ -185,6 +185,11 @@ func ComputeCurrentSigning(
 	}
 
 	s1, s2 := numeric.NewDecFromBigInt(signed), numeric.NewDecFromBigInt(toSign)
+	if s2.IsNil() || s2.IsZero() {
```
fxfactorial, line 188: Should remove this, if this is violated, then we should find out asap and I did from the stack trace last time (that something was iffy)

rlan35: ok sure.

core/offchain.go
```diff
@@ -187,13 +186,12 @@ func (bc *BlockChain) CommitOffChainData(
 		for i, c := uint32(0), shard.Schedule.InstanceForEpoch(
 			epoch,
 		).NumShards(); i < c; i++ {
-			if err := bc.LastContinuousCrossLink(batch, i); err != nil {
-				utils.Logger().Info().
-					Err(err).Msg("could not batch process last continuous crosslink")
-			}
+			bc.LastContinuousCrossLink(batch, i)
```
fxfactorial, line 189: if its not even worth logging out an error, then remove the error return value

rlan35: it may be useful for other cases. Noted though.

fxfactorial: then keep the error value and leave a NOTE , remove the log. leaving the error unhandled code with no comemnt explaining is that was intentional or not is ambigious for tomorrrow.

rlan35: https://github.com/harmony-one/harmony/issues/2762. Don't want to restart the jenkins.

Issue #2696, opened 2020-04-03
Expand error messages when bls keys aren’t all on one shard 


fxfactorial:
Right now if user puts in bls keys across multiple shards then they get this error (dumped by node.sh)

node.sh: ############### Running Harmony Process ###############
cannot set up node account: bls keys do not belong to the same shard
node.sh: node process finished with status 1
node.sh: restarting in 10s...

This is not helpful, we should tell the user which keys they gave are on what shard so they immediately know which bls keys is the Odd one out, otherwise they have to use hmy utility shard-for-bls and how would they be expected to know that

PR #2768, opened 2020-04-07
enhance multikey validation error message

gupadhyaya:
fixes https://github.com/harmony-one/harmony/issues/2696

### DIFF
--- a/internal/configs/node/config.go
+++ b/internal/configs/node/config.go
@@ -6,6 +6,7 @@ package nodeconfig
 import (
 	"fmt"
 	"math/big"
+	"strings"
 	"sync"
 
 	"github.com/harmony-one/bls/ffi/go/bls"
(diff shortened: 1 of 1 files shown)


gupadhyaya:
output:
```
cannot set up node account: bls keys do not belong to same shard
key: 2258b3b7b2b95d528467ddbfd25d8b551f1917d0589ee64dc13b90c3364a62627f3ec139a7f76fb82d0832cb404a478c, shard id: 0
key: c167fc725b898faabf54fc30ad073419f38b94a4cadeced6957a65cc3810fa631003b19f165101b61f0e2bfa027c6e11, shard id: 1
key: cc7d8215460868af2731d01a0ecf0a3159c3f0117bc26f7fd977c92542c725d6c1201a8031ef6581e9bf1e2564ffa091, shard id: 1
```

internal/configs/node/config.go
```diff
@@ -251,37 +252,45 @@ func SetShardingSchedule(schedule shardingconfig.Schedule) {
...
 	epoch := conf.networkType.ChainConfig().StakingEpoch
 	numShards := conf.shardingSchedule.InstanceForEpoch(epoch).NumShards()
 	shardID := new(big.Int).Mod(pubKey.Big(), big.NewInt(int64(numShards)))
 	return uint32(shardID.Uint64()), nil
 }
 
+// ShardIDFromConsensusKey returns the shard ID statically determined from the
+// consensus key.
+func (conf *ConfigType) ShardIDFromConsensusKey() (uint32, error) {
+	return conf.ShardIDFromKey(conf.ConsensusPubKey.PublicKey[0])
```
fxfactorial, line 272: this [0] will never explode?

internal/configs/node/config.go
```diff
@@ -251,37 +252,45 @@ func SetShardingSchedule(schedule shardingconfig.Schedule) {
...
+		if shardID != sID {
+			isSameShard = false
 		}
+		keyShardStrs = append(
+			keyShardStrs,
+			fmt.Sprintf("key: %s, shard id: %d", key.SerializeToHexStr(), shardID),
+		)
+	}
+	if !isSameShard {
+		return fmt.Errorf("bls keys do not belong to same shard\n%s", strings.Join(keyShardStrs, "\n"))
```
fxfactorial, line 293: lets get rid of all `fmt.Errof` and only use `pkg/error`, therefore `errors.Errof`

internal/configs/node/config.go
gupadhyaya: nope. at least one bls key is always provided, including explorer for which default dummy bls is used.

PR #2774, opened 2020-04-07
[apr] fix the validator stats not showing up for two epochs issue

gupadhyaya:
minor mistake. should fix the issue.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2336,24 +2336,21 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 				return nil, err
 			}
 
-			aprComputed, err := apr.ComputeForValidator(
-				bc, block, wrapper,
-			)
-			if err == nil && aprComputed != nil {
-				a := *aprComputed
-				utils.Logger().Info().
-					Str("return-rate", a.String()).
-					Uint64("new-epoch", newEpochSuperCommittee.Epoch.Uint64()).
-					Msg("apr computed")
-				stats.APR = a
-			}
-
-			if err != nil {
-				utils.Logger().Debug().Err(err).Msg("issue with compute of apr")
-			}
-
-			if err != nil {
-				return nil, err
+			stats.APR = numeric.ZeroDec()
+			if wrapper.Delegations[0].Amount.Cmp(common.Big0) > 0 {
+				if aprComputed, err := apr.ComputeForValidator(
+					bc, block, wrapper,
+				); err != nil {
+					if errors.Cause(err) == apr.ErrInsufficientEpoch {
+						utils.Logger().Info().Err(err).Msg("apr could not be computed")
+					} else {
+						return nil, err
+					}
+				} else {
+					stats.APR = *aprComputed
+				}
+			} else {
+				utils.Logger().Info().Msg("zero total delegation, skipping apr computation")
 			}
 
 			if _, wasBooted := bootedFromSuperCommittee[wrapper.Address]; wasBooted {
--- a/staking/apr/compute.go
+++ b/staking/apr/compute.go
@@ -15,6 +15,13 @@ import (
 	"github.com/pkg/errors"
 )
 
+var (
+	// ErrInsufficientEpoch is returned when insufficient past epochs for apr computation
+	ErrInsufficientEpoch = errors.New("insufficient past epochs to compute apr")
+	// ErrCouldNotRetreiveHeaderByNumber is returned when fail to retrieve header by number
+	ErrCouldNotRetreiveHeaderByNumber = errors.New("could not retrieve header by number")
+)
+
 // Reader ..
 type Reader interface {
 	GetHeaderByNumber(number uint64) *block.Header
@@ -87,26 +94,30 @@ func ComputeForValidator(
 	)
 
 	if err != nil {
-		return &zero, err
+		return nil, errors.Wrapf(
+			ErrInsufficientEpoch,
+			"current epoch %d, one-epoch-ago %d",
+			block.Epoch().Uint64(),
+			oneEpochAgo.Uint64(),
+		)
 	}
 
 	blockNumAtOneEpochAgo := shard.Schedule.EpochLastBlock(oneEpochAgo.Uint64())
-
 	headerOneEpochAgo := bc.GetHeaderByNumber(blockNumAtOneEpochAgo)
-	if block.Header() == nil || headerOneEpochAgo == nil {
+
+	if headerOneEpochAgo == nil {
 		utils.Logger().Debug().
 			Msgf("apr compute headers epochs ago %+v %+v %+v",
 				oneEpochAgo,
 				blockNumAtOneEpochAgo,
 				headerOneEpochAgo,
 			)
-		return &zero, errors.New("can't get headers for APR computation")
+		return nil, errors.Wrapf(
+			ErrCouldNotRetreiveHeaderByNumber,
+			"num header wanted %d",
+			blockNumAtOneEpochAgo,
+		)
 	}
-
-	utils.Logger().Debug().
-		RawJSON("current-epoch-header", []byte(bc.CurrentHeader().String())).
-		RawJSON("one-epoch-ago-header", []byte(headerOneEpochAgo.String())).
-		Msg("headers used for apr computation")
 
 	estimatedRewardPerYear, err := expectedRewardPerYear(
 		block.Header(), headerOneEpochAgo,
@@ -122,10 +133,6 @@ func ComputeForValidator(
 	}
 
 	total := numeric.NewDecFromBigInt(validatorNow.TotalDelegation())
-	if total.IsZero() {
-		return nil, errors.New("zero total delegation will cause div by zero")
-	}
-
 	result := numeric.NewDecFromBigInt(estimatedRewardPerYear).Quo(
 		total,
 	)


core/blockchain.go
```diff
@@ -2352,10 +2352,6 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 				utils.Logger().Debug().Err(err).Msg("issue with compute of apr")
 			}
 
-			if err != nil {
```
fxfactorial, line 2355: I don't like this because now we are ignoring an error but its okay to ignore it because of the business logic carves out this way. 

The root problem is the APR return value semantics, let's fix that here as well since now we know that is correct & stable (the getting of the correct `*block.Header`) which was the original reason why the return error and value was so accompanding.

gupadhyaya: right now it doesn't even get to the header, it is not able to get the `ReadValidatorSnapshotAtEpoch` `oneEpochAgo`. Which is a valid case when the validator is only created in the current epoch and does not exists in the previous epoch. The return value seems to be right below:
```
oneSnapshotAgo, err := bc.ReadValidatorSnapshotAtEpoch(
		oneEpochAgo,
		validatorNow.Address,
	)

	if err != nil {
		return &zero, err
	}
```

core/blockchain.go
```diff
@@ -2336,24 +2336,26 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
...
+					} else {
+						return nil, err
+					}
+				}
 				a := *aprComputed
 				utils.Logger().Info().
 					Str("return-rate", a.String()).
 					Uint64("new-epoch", newEpochSuperCommittee.Epoch.Uint64()).
 					Msg("apr computed")
 				stats.APR = a
```
fxfactorial, line 2356: `a` not needed, but this point `stats.APR = &aprComputed`

fxfactorial: and we can get rid of this log

gupadhyaya: `aprComputed` is a dec pointer. Also, need it for logging?

staking/apr/compute.go
```diff
@@ -117,15 +124,7 @@ func ComputeForValidator(
 		return nil, err
 	}
 
-	if estimatedRewardPerYear.Cmp(common.Big0) == 0 {
```
fxfactorial, line 120: leave this one, we can have specific or other business logic for `expectedRewardPerYear`

staking/apr/compute.go
```diff
@@ -87,25 +94,25 @@ func ComputeForValidator(
...
 				headerOneEpochAgo,
 			)
-		return &zero, errors.New("can't get headers for APR computation")
+		return nil, ErrCouldNotRetreiveHeaderByNumber
 	}
 
 	utils.Logger().Debug().
-		RawJSON("current-epoch-header", []byte(bc.CurrentHeader().String())).
-		RawJSON("one-epoch-ago-header", []byte(headerOneEpochAgo.String())).
+		Str("current-epoch-header %s", bc.CurrentHeader().String()).
```
fxfactorial, line 114: no, `.String()` will still call MarshalJSON. Remove this log altogether

core/blockchain.go
fxfactorial: log not needed, just assign the pointer itself, and even in this case, what was the point of the pointer deref assignment, Str("return-rate", aprComputed.String()) would work fine too

staking/apr/compute.go
```diff
@@ -87,25 +94,25 @@ func ComputeForValidator(
...
 	if err != nil {
-		return &zero, err
+		return &zero, ErrInsufficientEpoch
 	}
 
 	blockNumAtOneEpochAgo := shard.Schedule.EpochLastBlock(oneEpochAgo.Uint64())
-
 	headerOneEpochAgo := bc.GetHeaderByNumber(blockNumAtOneEpochAgo)
+
 	if block.Header() == nil || headerOneEpochAgo == nil {
```
fxfactorial, line 103: pretty sure have invariant to assume that `.Header() != nil` by caller, can prob drop.

staking/apr/compute.go
```diff
@@ -87,25 +94,25 @@ func ComputeForValidator(
...
 		utils.Logger().Debug().
 			Msgf("apr compute headers epochs ago %+v %+v %+v",
 				oneEpochAgo,
 				blockNumAtOneEpochAgo,
 				headerOneEpochAgo,
 			)
-		return &zero, errors.New("can't get headers for APR computation")
+		return nil, ErrCouldNotRetreiveHeaderByNumber
```
fxfactorial, line 110: errors.Wrapf(ErrCouldNotRetreiveHeaderByNumber, "num header wanted %d", blockNumAtOneEpochAgo)

fxfactorial: in general, the error message should be adding info making it easy to fix/debug. Because this error will be propagated upward  as well. At each level propagation of error, make it easy to fix and see where things went wrong.

staking/apr/compute.go
```diff
@@ -87,25 +94,25 @@ func ComputeForValidator(
 	)
 
 	if err != nil {
-		return &zero, err
+		return &zero, ErrInsufficientEpoch
```
fxfactorial, line 97: wrap(..., which epoch is current, which epoch was asked for)

core/blockchain.go
```diff
@@ -2336,24 +2336,25 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
...
-			)
-			if err == nil && aprComputed != nil {
-				a := *aprComputed
+			stats.APR = numeric.ZeroDec()
+			if wrapper.TotalDelegation().Cmp(common.Big0) > 0 {
+				aprComputed, err := apr.ComputeForValidator(
+					bc, block, wrapper,
+				)
+				if err != nil {
+					if err == apr.ErrInsufficientEpoch {
```
fxfactorial, line 2345: but now its wrapped right, so need the .Contain, and in the log, utils.Logger().Err(err).Info().Msg("..")

fxfactorial: the err being on the plain err, not the contained one

gupadhyaya: oops.

gupadhyaya: `errors.Cause(` ?

staking/apr/compute.go
```diff
@@ -87,27 +94,31 @@ func ComputeForValidator(
 	)
 
 	if err != nil {
-		return &zero, err
+		return &zero, errors.Wrapf(
```
fxfactorial, line 97: don't need to accomdate the bad API from this package anymore, return nil,  the_wrapped_error. 

That's why we handled this, at callsite, in blockchain.go

core/blockchain.go
```diff
@@ -2336,24 +2336,25 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
...
+						utils.Logger().Info().Msg("insufficient past epochs to compute apr")
+					} else {
+						return nil, err
+					}
+				}
 				utils.Logger().Info().
```
fxfactorial, line 2351: please remove this whole call, its not adding useful info because you don't know to whom the apr is computed to, so its just writing that it computed apr but we can't debug and say to whom, so its not useful.

staking/apr/compute.go
gupadhyaya: fall through will try to dereference if we return nil.

core/blockchain.go
```diff
@@ -2336,24 +2336,25 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 				return nil, err
 			}
 
-			aprComputed, err := apr.ComputeForValidator(
-				bc, block, wrapper,
-			)
-			if err == nil && aprComputed != nil {
-				a := *aprComputed
+			stats.APR = numeric.ZeroDec()
+			if wrapper.TotalDelegation().Cmp(common.Big0) > 0 {
```
fxfactorial, line 2340: actually this TotalDelegation is an expensive for-loop. Really we should just check if wrapper.Delegations[0].Cmp() > 0 

then its constant time. Let's think careful if can use this assumption

core/blockchain.go
fxfactorial: Ya, like you used in another place before, I think ya, .Cause (I rememberde something with C)

staking/apr/compute.go
fxfactorial: then deal with it there. I want one place of ad-hoc code and that should be at call site, since we are doing it there on error. In this packages, like under staking/*, should assume the functions are pure in intent of computation asked, not business logic what should APR be when don't enough epoch

core/blockchain.go
gupadhyaya: if we use .Err(err) and don't return, sounds weird... ?

fxfactorial: No, you're not understanidng the logger API (I was confused by this as well), 

It would be weird if we had 

utils.Logger().Error().Err(err).Msg()

and didn't return, 

but this is `.Info()`

fxfactorial: we're aware that there is an error but the business logic handles it as an acceptable error. Error is just a value and we can branch on its value as well (although usually shouldn't)

staking/apr/compute.go
gupadhyaya: there is still one case left, leave it as is? 
```
	if estimatedRewardPerYear.Cmp(common.Big0) == 0 {
		return &zero, nil
	}
```

core/blockchain.go
gupadhyaya: done.

PR #2775, opened 2020-04-07
optimize leader signature collection strategy; Clean up logs

rlan35:
Leader sig collection strategy:

**before**: once 2/3 voting power/sig collected, wait for another 2s to collect more sigs; or if 90% reached, just commit.

**now**: 
    1. always wait for the full block time (8s) and collect as much sigs as possible. New block won't be proposed before full block time.
    2. once 2/3 voting power/sig collected, wait for another 2s or more if full block time not reached.
    3. If 100% sig collected, commit the block immediately
    4. Both 2 and 3 works under constraint of 1

This change will help improve the chance of slow node getting their signatures collected in block

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -181,10 +181,10 @@ func (consensus *Consensus) finalizeCommits() {
 		Int("numStakingTxns", len(block.StakingTransactions())).
 		Msg("HOORAY!!!!!!! CONSENSUS REACHED!!!!!!!")
 
-	if time.Now().Before(consensus.NextBlockDue) {
+	if n := time.Now(); n.Before(consensus.NextBlockDue) {
 		// Sleep to wait for the full block time
 		consensus.getLogger().Debug().Msg("[finalizeCommits] Waiting for Block Time")
-		time.Sleep(consensus.NextBlockDue.Sub(time.Now()))
+		time.Sleep(consensus.NextBlockDue.Sub(n))
 	}
 	// Send signal to Node to propose the new block for consensus
 	consensus.ReadySignal <- struct{}{}
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -367,13 +367,13 @@ func (consensus *Consensus) Start(
 				consensus.SetViewID(consensus.ChainReader.CurrentHeader().ViewID().Uint64() + 1)
 				mode := consensus.UpdateConsensusInformation()
 				consensus.current.SetMode(mode)
-				consensus.getLogger().Info().Str("Mode", mode.String()).Msg("Node is in sync")
+				consensus.getLogger().Info().Str("Mode", mode.String()).Msg("Node is IN SYNC")
 
 			case <-consensus.syncNotReadyChan:
 				consensus.getLogger().Debug().Msg("[ConsensusMainLoop] syncNotReadyChan")
 				consensus.SetBlockNum(consensus.ChainReader.CurrentHeader().Number().Uint64() + 1)
 				consensus.current.SetMode(Syncing)
-				consensus.getLogger().Info().Msg("Node is out of sync")
+				consensus.getLogger().Info().Msg("[ConsensusMainLoop] Node is OUT OF SYNC")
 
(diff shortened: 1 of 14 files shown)


consensus/consensus_v2.go
```diff
@@ -473,12 +470,19 @@ func (consensus *Consensus) Start(
 
 			case viewID := <-consensus.commitFinishChan:
 				consensus.getLogger().Debug().Msg("[ConsensusMainLoop] commitFinishChan")
+
 				// Only Leader execute this condition
 				func() {
+					if time.Now().Before(blockTime) {
```
fxfactorial, line 476:
```
if n := time.Now(); n.Before(blockTime) {

time.Sleep(blockTime.Sub(n)) 

}
```

unless you are intentionallyy making a fresh `time.Now()`?

fxfactorial:
This PR actually has quite some nuanced logic in consensus and thresholding. Can we test on STN before merge to master?

consensus/consensus_v2.go
rlan35: good point. done.

rlan35:
> This PR actually has quite some nuanced logic in consensus and thresholding. Can we test on STN before merge to master?

I've already tested locally, and consensus works fine. Won't hurt to merge and test with STN.

fxfactorial:
> > This PR actually has quite some nuanced logic in consensus and thresholding. Can we test on STN before merge to master?
> 
> I've already tested locally, and consensus works fine. Won't hurt to merge and test with STN.

fair - I agree

consensus/quorum/one-node-one-vote.go
```diff
@@ -52,9 +52,9 @@ func (v *uniformVoteWeight) QuorumThreshold() numeric.Dec {
 	return numeric.NewDec(v.TwoThirdsSignersCount())
 }
 
-// RewardThreshold ..
-func (v *uniformVoteWeight) IsRewardThresholdAchieved() bool {
-	return v.SignersCount(Commit) >= (v.ParticipantsCount() * 9 / 10)
+// IsAllSigsCollected ..
+func (v *uniformVoteWeight) IsAllSigsCollected() bool {
+	return v.SignersCount(Commit) == v.ParticipantsCount()
```
fxfactorial, line 57: but this will make it easier for consensus to break and p2p hasn't been worked on yet, why bump it up prematurely?

rlan35:
> in principle approve but the bump up from 90% to 100% i think will cause problems

The 90% or 100% is a second tier criteria. The first tier is 2/3 and wait for 2s for any more sigs.

100% won't matter when first tier works. So should be a pure win situation.

LeoHChen:
Sorry, I'm late for reviewing this PR. After reviewing the PR, I think this is the right direction to collection signatures.  If you don't mind, I still have a few questions even though the PR is merged.
1. do we always have to wait for the full block time before proposing the next block? what if all external nodes have signed the block already, then we don't need to wait. Noted, our own nodes have 2s commit delay already.
1. can we get @gupadhyaya review and approve this PR?
1. What would be a good test plan for this consensus update and block period update in STN? any corner case we shall cover?  we need to properly test it. @john-harmony , @SebastianJ

rlan35:
> Sorry, I'm late for reviewing this PR. After reviewing the PR, I think this is the right direction to collection signatures. If you don't mind, I still have a few questions even though the PR is merged.
> 
> 1. do we always have to wait for the full block time before proposing the next block? what if all external nodes have signed the block already, then we don't need to wait. Noted, our own nodes have 2s commit delay already.
> 2. can we get @gupadhyaya review and approve this PR?
> 3. What would be a good test plan for this consensus update and block period update in STN? any corner case we shall cover?  we need to properly test it. @john-harmony , @SebastianJ

1. Yes, we still always wait for 8 seconds block time. It's same as before. So no change on this logic.
2. Yes, both Leo and Ganesha can take a extra look
3. This thing just needs to live through more test time with very test situation we do currently. It's hard to craft specific test case for it as it's distributed system thing. As long as there is no regression on block time or consensus stuck in next STN deploy, it's good to go.

PR #2776, opened 2020-04-08
[consensus] rejecting malformed messages before processing

gupadhyaya:
checked locally by trying to run a node on ostn. Currently with master code (@fxfactorial undo of protobuf changes) this crash is reproducible as ostn does not have this undo. It was convenient to test this fix.

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -11,6 +11,10 @@ import (
 const MaxBlockNumDiff = 100
 
 func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
+	if msg.GetConsensus() == nil {
+		consensus.getLogger().Warn().Msg("[validatorSanityChecks] malformed message")
+		return false
+	}
 	consensus.getLogger().Debug().
 		Uint64("blockNum", msg.GetConsensus().BlockNum).
 		Uint64("viewID", msg.GetConsensus().ViewId).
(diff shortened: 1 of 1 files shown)


consensus/consensus_v2.go
```diff
@@ -29,6 +29,12 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
 		return
 	}
 
+	// handle malformed messages to prevent crash
+	if msg.GetConsensus() == nil {
```
fxfactorial, line 33: not sure if this is the right place because there is Mode logic and ViewChangeing logic as well, and also its checked here: 

https://github.com/harmony-one/harmony/blob/a1e75767e01fa4fb88c616ff0dfa7a7b6d70c82b/consensus/consensus_v2.go#L58

Maybe let's keep the PR as a hotfix - that is, add this only in the explding validatorSanityCheck , keep our change local in scope.

gupadhyaya: agreed. fixed to handle malformed messages locally. plz review.

consensus/checks.go
```diff
@@ -180,6 +188,9 @@ func (consensus *Consensus) onPreparedSanityChecks(
 }
 
 func (consensus *Consensus) viewChangeSanityCheck(msg *msg_pb.Message) bool {
+	if msg.GetViewchange() == nil {
+		consensus.getLogger().Warn().Msg("[viewChangeSanityCheck] malformed message")
```
fxfactorial, line 192: Where the early return

gupadhyaya: done.

PR #2780, opened 2020-04-08
[project] Remove over complicated packages, useless tests, dead functions

fxfactorial:
No business logic changes 

1. Remove overcomplicated `ctxerror`
2. Remove overcomplicated mockgen, no one used it, just gives false sense of testing & slows down testing, iteration (always have to update code for type changes/name changes)
3. Remove dead code from `internal/utils`

PR #2783, opened 2020-04-08
-is_genesis is no longer used

SebastianJ:
## Issue

`-is-genesis` is no longer part of cmd/harmony/main.go - so need to remove it

E.g:
```
node.sh: ############### Running Harmony Process ###############
flag provided but not defined: -is_genesis
```


fxfactorial:
done in https://github.com/harmony-one/harmony/pull/2780

Issue #2759, opened 2020-04-06
hmy_getTransactionsCount logic should be separated into regular txns vs staking txns


denniswon:
The current hmy_getTransactionsCount uses account nonce to figure out the txns count for an account. As nonce is incremented both by regular and staking txns, using nonce is not the right way to return the regular and staking txns count separately.

Harmony block explorer displays txns and staking txns separately. 
https://github.com/harmony-one/harmony-dashboard-backend/issues/35

Instead of nonce, hmy_getTransactionsCount and hmy_getStakingTransactionsCount will loop through the txns histories to count the counts for each. (For now, seems most probable approach)

cc: @fxfactorial

fxfactorial:
true, but i think we can limit the changes just to this one RPC which will just give back results for both. Just ignore the field you don't need

PR #2788, opened 2020-04-09
[rpc] separate transaction counts rpc into regular vs staking

denniswon:
## Issue

https://github.com/harmony-one/harmony/issues/2759

## Test

launched local network, sent staking and regular txns using hmy cli, from shard 0 with bls key in shard 0. then, saw hmy_getAccountNonce, hmy_getTransactionCount, hmy_getStakingTransactionCount all works correctly. 

```
curl --location --request POST 'http://localhost:9500' --header 'Content-Type: application/json' --data-raw '{
    "jsonrpc": "2.0",
    "method": "hmyv2_getStakingTransactionCount",
    "params": ["one1grlw2qw7t9ulhms38w4ea0gssz0dk8vvsl7jum", "ALL"}],
    "id": 1
}'

curl --location --request POST 'http://localhost:9500' --header 'Content-Type: application/json' --data-raw '{
    "jsonrpc": "2.0",
    "method": "hmyv2_getTransactionCount",
    "params": ["one1grlw2qw7t9ulhms38w4ea0gssz0dk8vvsl7jum", "ALL"}],
    "id": 1
}'

curl --location --request POST 'http://localhost:9500' --header 'Content-Type: application/json' --data-raw '{
    "jsonrpc": "2.0",
    "method": "hmyv2_getAccountNonce",
    "params": ["one1grlw2qw7t9ulhms38w4ea0gssz0dk8vvsl7jum", "latest"}],
    "id": 1
}'
```

```
dennis.won@Jongs-MacBook-Pro:~/harmony-one/harmony (txs_count) $ go test ./node
ok  	github.com/harmony-one/harmony/node	1.674s
```

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -241,26 +241,22 @@ func (b *APIBackend) GetBalance(ctx context.Context, address common.Address, blo
 
 // GetTransactionsHistory returns list of transactions hashes of address.
 func (b *APIBackend) GetTransactionsHistory(address, txType, order string) ([]common.Hash, error) {
-	hashes, err := b.hmy.nodeAPI.GetTransactionsHistory(address, txType, order)
-	return hashes, err
+	return b.hmy.nodeAPI.GetTransactionsHistory(address, txType, order)
 }
 
 // GetStakingTransactionsHistory returns list of staking transactions hashes of address.
 func (b *APIBackend) GetStakingTransactionsHistory(address, txType, order string) ([]common.Hash, error) {
-	hashes, err := b.hmy.nodeAPI.GetStakingTransactionsHistory(address, txType, order)
-	return hashes, err
+	return b.hmy.nodeAPI.GetStakingTransactionsHistory(address, txType, order)
 }
 
 // GetTransactionsCount returns the number of regular transactions of address.
 func (b *APIBackend) GetTransactionsCount(address, txType string) (uint64, error) {
-	count, err := b.hmy.nodeAPI.GetTransactionsCount(address, txType)
-	return count, err
+	return b.hmy.nodeAPI.GetTransactionsCount(address, txType)
 }
 
 // GetStakingTransactionsCount returns the number of staking transactions of address.
 func (b *APIBackend) GetStakingTransactionsCount(address, txType string) (uint64, error) {
-	count, err := b.hmy.nodeAPI.GetStakingTransactionsCount(address, txType)
-	return count, err
+	return b.hmy.nodeAPI.GetStakingTransactionsCount(address, txType)
 }
 
 // NetVersion returns net version
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -251,15 +251,15 @@ func (b *APIBackend) GetStakingTransactionsHistory(address, txType, order string
 	return hashes, err
 }
 
-// GetTransactionCount returns the number of regular transactions of address.
-func (b *APIBackend) GetTransactionCount(address, txType string) (uint64, error) {
-	count, err := b.hmy.nodeAPI.GetTransactionCount(address, txType)
+// GetTransactionsCount returns the number of regular transactions of address.
+func (b *APIBackend) GetTransactionsCount(address, txType string) (uint64, error) {
+	count, err := b.hmy.nodeAPI.GetTransactionsCount(address, txType)
 	return count, err
 }
 
-// GetStakingTransactionCount returns the number of staking transactions of address.
-func (b *APIBackend) GetStakingTransactionCount(address, txType string) (uint64, error) {
-	count, err := b.hmy.nodeAPI.GetStakingTransactionCount(address, txType)
+// GetStakingTransactionsCount returns the number of staking transactions of address.
+func (b *APIBackend) GetStakingTransactionsCount(address, txType string) (uint64, error) {
+	count, err := b.hmy.nodeAPI.GetStakingTransactionsCount(address, txType)
 	return count, err
 }
 
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -220,6 +220,16 @@ func (b *APIBackend) GetPoolTransactions() (types.PoolTransactions, error) {
 	return txs, nil
 }
 
+// GetAccountNonce returns the nonce value of the given address for the given block number
+func (b *APIBackend) GetAccountNonce(
+	ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (uint64, error) {
+	state, _, err := b.StateAndHeaderByNumber(ctx, blockNr)
+	if state == nil || err != nil {
+		return 0, err
+	}
+	return state.GetNonce(address), state.Error()
+}
+
 // GetBalance returns balance of an given address.
 func (b *APIBackend) GetBalance(ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (*big.Int, error) {
 	state, _, err := b.StateAndHeaderByNumber(ctx, blockNr)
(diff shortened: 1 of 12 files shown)


hmy/api_backend.go
```diff
@@ -220,6 +220,16 @@ func (b *APIBackend) GetPoolTransactions() (types.PoolTransactions, error) {
 	return txs, nil
 }
 
+// GetAccountNonce returns the nonce value of the given address for the given block number
+func (b *APIBackend) GetAccountNonce(
+	ctx context.Context, address common.Address, blockNr rpc.BlockNumber) (uint64, error) {
```
gupadhyaya, line 225: No need to worry about the pending transactions anymore?

internal/hmyapi/apiv1/transactionpool.go
```diff
@@ -152,26 +152,6 @@ func (s *PublicTransactionPoolAPI) GetStakingTransactionByBlockHashAndIndex(ctx
 	return nil
 }
 
-// GetTransactionCount returns the number of transactions the given address has sent for the given block number
```
gupadhyaya, line 155: we don't need GetTransactionCount and GetStakingTransactionCount for apiv1?

scripts/api_test.sh
```diff
@@ -670,7 +670,7 @@ function run_tests() {
 	API_netPeerCount_test
 	API_getBalance_test
 	API_getStorageAt_test
-	API_getTransactionCount_test
+	API_getAccountNonce_test
 	API_sendRawTransaction_test
```
gupadhyaya, line 674: Is it possible to add api tests for transaction and staking transaction counts as well?

gupadhyaya:
Generally looks good. As, this PR has impact on the clients like js-sdk that uses getTransactionCount for nonce, how do you plan to update the client?

hmy/api_backend.go
denniswon: There is getPoolNonce that does the same job

internal/hmyapi/apiv1/transactionpool.go
denniswon: Just deeted the self explanatory comment. Technically dont need to add on v1 but this pr just adds to v1 as well for consistency

scripts/api_test.sh
denniswon: sure

hmy/api_backend.go
gupadhyaya: I mean to say if there is a pending transaction shouldn't you be returning that number instead of just the committed transactions?

denniswon: ah i see. now understood. yeah, I guess we do need to consider pending txns here.

denniswon: @gupadhyaya actually, isn't pool nonce >= account nonce always? so getPoolNonce() is used for getting number of transactions including pending, and getAccountNonce() used for getting number of confirmed transactions?

internal/hmyapi/apiv1/transactionpool.go
denniswon: we actually do. this was why the api_test.sh was failing. Putting this back with comment.

denniswon:
@gupadhyaya updated the PR. we actually need to add getTransactionCount api back to v1. Added comment that it is legacy and for the new txns and staking txns counts apis, added them as getTransactionsCount and getStakingTransactionCount.

This was actually api_test.sh was failing. fixed the test by adding the getTransactionCount back to apiv1.

hmy/api_backend.go
gupadhyaya: what will happen if from the cli I made a transaction which is waiting in the pool and then I made the next transaction? Both of them will have the same nonce and only one will work?

hmy/api_backend.go
```diff
@@ -241,6 +251,18 @@ func (b *APIBackend) GetStakingTransactionsHistory(address, txType, order string
 	return hashes, err
 }
 
+// GetTransactionsCount returns the number of regular transactions of address.
+func (b *APIBackend) GetTransactionsCount(address, txType string) (uint64, error) {
+	count, err := b.hmy.nodeAPI.GetTransactionsCount(address, txType)
+	return count, err
```
gupadhyaya, line 257: simply return `b.hmy.nodeAPI.GetTransactionsCount(address, txType)`

hmy/api_backend.go
```diff
@@ -241,6 +251,18 @@ func (b *APIBackend) GetStakingTransactionsHistory(address, txType, order string
...
+}
+
+// GetStakingTransactionsCount returns the number of staking transactions of address.
+func (b *APIBackend) GetStakingTransactionsCount(address, txType string) (uint64, error) {
+	count, err := b.hmy.nodeAPI.GetStakingTransactionsCount(address, txType)
+	return count, err
```
gupadhyaya, line 263: same as above.

internal/hmyapi/apiv1/transactionpool.go
```diff
@@ -172,6 +174,34 @@ func (s *PublicTransactionPoolAPI) GetTransactionCount(ctx context.Context, addr
...
+	var err error
+	if !strings.HasPrefix(address, "one1") {
+		addr := internal_common.ParseAddr(address)
+		address, err = internal_common.AddressToBech32(addr)
+		if err != nil {
+			return 0, err
+		}
+	}
+	count, err := s.b.GetTransactionsCount(address, txType)
+	return count, err
```
gupadhyaya, line 188: `return s.b.GetTransactionsCount(address, txType)`

internal/hmyapi/apiv1/transactionpool.go
```diff
@@ -172,6 +174,34 @@ func (s *PublicTransactionPoolAPI) GetTransactionCount(ctx context.Context, addr
...
+	var err error
+	if !strings.HasPrefix(address, "one1") {
+		addr := internal_common.ParseAddr(address)
+		address, err = internal_common.AddressToBech32(addr)
+		if err != nil {
+			return 0, err
+		}
+	}
+	count, err := s.b.GetStakingTransactionsCount(address, txType)
+	return count, err
```
gupadhyaya, line 202: same as above.

internal/hmyapi/apiv2/transactionpool.go
```diff
@@ -200,28 +200,32 @@ func (s *PublicTransactionPoolAPI) GetStakingTransactionByHash(ctx context.Conte
...
 			return 0, err
 		}
-		return nonce, nil
 	}
-	// Resolve block number and use its state to ask for the nonce
-	state, _, err := s.b.StateAndHeaderByNumber(ctx, rpc.BlockNumber(blockNr))
-	if state == nil || err != nil {
-		return 0, err
+	count, err := s.b.GetTransactionsCount(address, txType)
+	return count, err
```
gupadhyaya, line 214: `return s.b.GetTransactionsCount(address, txType)`

internal/hmyapi/apiv2/transactionpool.go
```diff
@@ -200,28 +200,32 @@ func (s *PublicTransactionPoolAPI) GetStakingTransactionByHash(ctx context.Conte
...
+		addr := internal_common.ParseAddr(address)
+		address, err = internal_common.AddressToBech32(addr)
+		if err != nil {
+			return 0, err
+		}
 	}
-	nonce := state.GetNonce(address)
-	return nonce, state.Error()
+	count, err := s.b.GetStakingTransactionsCount(address, txType)
+	return count, err
```
gupadhyaya, line 228: same as above.

gupadhyaya:
@denniswon good to go after fixing the above mentioned minor things.

PR #2799, opened 2020-04-10
[node] move self address to node, make it updatable every epoch

gupadhyaya:
* Remove self address assignment while starting node
* Move the self address out of consensus to node
* Get the self addresses for bls keys run by the node using shard state 
* Updates the self address every epoch

### DIFF
--- a/consensus/double_sign.go
+++ b/consensus/double_sign.go
@@ -70,6 +70,14 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 
 						now := big.NewInt(time.Now().UnixNano())
 
+						leaderShardKey := *shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
+						leaderAddr, err := subComm.AddressForBLSKey(leaderShardKey)
+						if err != nil {
+							consensus.getLogger().Err(err).Str("msg", recvMsg.String()).
+								Msg("could not find address for leader bls key")
+							return true
+						}
+
 						go func(reporter common.Address) {
 							evid := slash.Evidence{
 								ConflictingBallots: slash.ConflictingBallots{
@@ -93,7 +101,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 								Offender: *addr,
 							}
 							consensus.SlashChan <- proof
-						}(consensus.SelfAddresses[consensus.LeaderPubKey.SerializeToHexStr()])
+						}(*leaderAddr)
 						return true
 					}
 				}
--- a/consensus/double_sign.go
+++ b/consensus/double_sign.go
@@ -54,6 +54,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 							consensus.getLogger().Error().
 								Str("msg", recvMsg.String()).
 								Msg("could not get shard key from sender's key")
+							return true
 						}
 						subComm, err := committee.FindCommitteeByID(
 							consensus.ShardID,
@@ -79,6 +80,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 							consensus.getLogger().Error().
 								Str("msg", recvMsg.String()).
 								Msg("could not get shard key from leader's key")
+							return true
 						}
 						leaderAddr, err := subComm.AddressForBLSKey(*leaderShardKey)
 						if err != nil {
--- a/consensus/double_sign.go
+++ b/consensus/double_sign.go
@@ -49,7 +49,12 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 								Msg("could not read shard state")
 							return true
 						}
-						offender := *shard.FromLibBLSPublicKeyUnsafe(recvMsg.SenderPubkey)
+						offender := shard.FromLibBLSPublicKeyUnsafe(recvMsg.SenderPubkey)
+						if offender == nil {
+							consensus.getLogger().Error().
+								Str("msg", recvMsg.String()).
+								Msg("could not get shard key from sender's key")
+						}
 						subComm, err := committee.FindCommitteeByID(
 							consensus.ShardID,
 						)
@@ -60,8 +65,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 							return true
 						}
 
-						addr, err := subComm.AddressForBLSKey(offender)
-
+						addr, err := subComm.AddressForBLSKey(*offender)
 						if err != nil {
 							consensus.getLogger().Err(err).Str("msg", recvMsg.String()).
 								Msg("could not find address for bls key")
@@ -70,8 +74,13 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 
 						now := big.NewInt(time.Now().UnixNano())
 
-						leaderShardKey := *shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
-						leaderAddr, err := subComm.AddressForBLSKey(leaderShardKey)
+						leaderShardKey := shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
+						if leaderShardKey == nil {
+							consensus.getLogger().Error().
+								Str("msg", recvMsg.String()).
+								Msg("could not get shard key from leader's key")
+						}
+						leaderAddr, err := subComm.AddressForBLSKey(*leaderShardKey)
 						if err != nil {
 							consensus.getLogger().Err(err).Str("msg", recvMsg.String()).
 								Msg("could not find address for leader bls key")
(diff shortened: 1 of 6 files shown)


consensus/double_sign.go
```diff
@@ -70,6 +70,14 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 
 						now := big.NewInt(time.Now().UnixNano())
 
+						leaderShardKey := *shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
```
rlan35, line 73: This FromLibBLSPublicKeyUnsafe could return nil. And deref it will crash. Better check nil first.

node/node.go
```diff
@@ -701,3 +708,73 @@ func (node *Node) ShutDown() {
...
+	if err != nil {
+		utils.Logger().Error().Err(err).
+			Int64("epoch", epoch.Int64()).
+			Uint32("shard-id", shardID).
+			Msg("[PopulateSelfAddresses] failed to find shard committee")
+		return
+	}
+
+	for _, blskey := range node.Consensus.PubKey.PublicKey {
+		shardkey := *shard.FromLibBLSPublicKeyUnsafe(blskey)
```
rlan35, line 737: same here.

node/node.go
```diff
@@ -701,3 +708,73 @@ func (node *Node) ShutDown() {
 	fmt.Print(msg)
 	os.Exit(0)
 }
+
+func (node *Node) populateSelfAddresses(epoch *big.Int) {
+	// reset the self addresses
+	node.SelfAddresses = map[string]common.Address{}
```
rlan35, line 714: this may not be thread safe to do. It's reset when the call uses a different epoch, which can be problematic.

node/node.go
```diff
@@ -701,3 +708,73 @@ func (node *Node) ShutDown() {
 	fmt.Print(msg)
 	os.Exit(0)
 }
+
+func (node *Node) populateSelfAddresses(epoch *big.Int) {
```
rlan35, line 712: Also, the mapping from BLS-> ecdsa address should never change. a BLS is always tied to the first registered validator. So it's safe to just set the mapping once at start.

consensus/double_sign.go
fxfactorial: yes, the Unsafe is only a convience when have invariant of never nil pointer

consensus/double_sign.go
```diff
@@ -70,6 +70,14 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
...
+						leaderAddr, err := subComm.AddressForBLSKey(leaderShardKey)
+						if err != nil {
+							consensus.getLogger().Err(err).Str("msg", recvMsg.String()).
```
fxfactorial, line 76: how often this case been happening, how possible to check with custom node against ostn ?

node/node.go
rlan35: Or we could just maintain the map without reset and keep updating it when new epoch changes.

node/node.go
```diff
@@ -181,6 +184,10 @@ type Node struct {
 		failedTxns        *ring.Ring
 	}
 	unixTimeAtNodeStart int64
+
+	// SelfAddresses holds the addresses of bls keys run by the node
+	SelfAddresses    map[string]common.Address
```
fxfactorial, line 189: this naming is really awkward in golang and always was in codebase, `Self`<anything> 

lets remove and make more meaningful, maybe like ValidatorAddrs, or something like that

node/node.go
```diff
@@ -701,3 +708,73 @@ func (node *Node) ShutDown() {
...
+// GetAddressForBLSKey retrieves the ECDSA address associated with bls key for epoch
+func (node *Node) GetAddressForBLSKey(blskey *bls.PublicKey, epoch *big.Int) common.Address {
+	// populate if first time setting or new epoch
+	if node.selfAddressEpoch == nil || epoch.Cmp(node.selfAddressEpoch) != 0 {
+		node.populateSelfAddresses(epoch)
+	}
+	blsStr := blskey.SerializeToHexStr()
+	addr, ok := node.SelfAddresses[blsStr]
+	if !ok {
+		return common.Address{}
```
fxfactorial, line 767: This is an error case right?, so lets give back error, not just empty addr

node/node.go
```diff
@@ -701,3 +708,73 @@ func (node *Node) ShutDown() {
...
+	}
+	return addr
+}
+
+// GetSelfAddresses retrieves all ECDSA addresses of the bls keys for epoch
+func (node *Node) GetSelfAddresses(epoch *big.Int) map[string]common.Address {
+	// populate if first time setting or new epoch
+	if node.selfAddressEpoch == nil || epoch.Cmp(node.selfAddressEpoch) != 0 {
+		node.populateSelfAddresses(epoch)
```
fxfactorial, line 776: prefer explicit value returned rather than side effect driven. populate could be a plain function that we use to assign to some node.<field>

node/node.go
gupadhyaya: it is a valid scenario to remove bls key from one validator and add it to another? especially when a staking partner has many validators?

rlan35: ok, you are right. Then let's make this part thread safe. Make sure it doesn't crash

consensus/double_sign.go
gupadhyaya: I don't have the double-sign setup to test this. can you?

node/node.go
gupadhyaya: sure. how about just `KeysToAddrs`? don't want to use `Validator...` as node is not concerned with validator. Any other suggestions?

node/node.go
gupadhyaya: semantics was to make it no-error, as the caller `proposeNewBlock` is handling it as empty address for error case.

gupadhyaya: Not sure whether to handle error locally or send it back to caller for proposeNewBlock?

node/node.go
gupadhyaya: since this map is only accessed linearly, will keep it thread unsafe. It is accessed by `proposeNewBlock` and `PostConsensusProcessing`.

node/node.go
gupadhyaya: using `KeysToAddrs`

node/node.go
gupadhyaya: as the map is never nil and the caller is basically using it to update the web hook, I would like to keep it side effect based and not return error.

consensus/double_sign.go
```diff
@@ -60,8 +65,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 							return true
 						}
 
-						addr, err := subComm.AddressForBLSKey(offender)
-
+						addr, err := subComm.AddressForBLSKey(*offender)
```
rlan35, line 68: it will blow up here too. Just return if offender is nil

consensus/double_sign.go
```diff
@@ -70,12 +74,25 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 
 						now := big.NewInt(time.Now().UnixNano())
 
+						leaderShardKey := shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
+						if leaderShardKey == nil {
+							consensus.getLogger().Error().
+								Str("msg", recvMsg.String()).
+								Msg("could not get shard key from leader's key")
+						}
+						leaderAddr, err := subComm.AddressForBLSKey(*leaderShardKey)
```
rlan35, line 83: same

node/node.go
```diff
@@ -701,3 +709,83 @@ func (node *Node) ShutDown() {
...
+			Str("bls-key", blsStr).
+			Str("address", common2.MustAddressToBech32(*addr)).
+			Msg("[PopulateSelfAddresses]")
+	}
+}
+
+// GetAddressForBLSKey retrieves the ECDSA address associated with bls key for epoch
+func (node *Node) GetAddressForBLSKey(blskey *bls.PublicKey, epoch *big.Int) common.Address {
+	// populate if first time setting or new epoch
+	if node.keysToAddrsEpoch == nil || epoch.Cmp(node.keysToAddrsEpoch) != 0 {
```
rlan35, line 772: Should lock in this func. Reading on a object that's changed by other thread also crash.

node/node.go
```diff
@@ -701,3 +709,83 @@ func (node *Node) ShutDown() {
...
+	if !ok {
+		return common.Address{}
+	}
+	return addr
+}
+
+// GetAddresses retrieves all ECDSA addresses of the bls keys for epoch
+func (node *Node) GetAddresses(epoch *big.Int) map[string]common.Address {
+	// populate if first time setting or new epoch
+	if node.keysToAddrsEpoch == nil || epoch.Cmp(node.keysToAddrsEpoch) != 0 {
```
rlan35, line 786: same

consensus/double_sign.go
gupadhyaya: oops. missed return.

PR #2807, opened 2020-04-10
[node.sh] detect hmy profile in rclone.conf

LeoHChen:


LeoHChen:
Tested with two conditions, [hmy] in rclone.conf and not in.

Issue #2808, opened 2020-04-11
MAINNET: node.sh panics


mindstyle85:
**Describe the bug**

OS: ubuntu 18.04 64bit
node.sh: version: v1 20200327.0

![photo_2020-04-11_02-12-59](https://user-images.githubusercontent.com/55639830/79030825-538b3200-7b9b-11ea-8d1c-d916b8fa7cbf.jpg)

gupadhyaya:
```
if currentBlock == nil {
		// Corrupt or empty database, init from scratch
		utils.Logger().Warn().Str("hash", head.Hex()).Msg("Head block missing, resetting chain")
		return bc.Reset()
```
exploding in `return bc.Reset()`, which can be fixed, @mindstyle85 did you try cleaning db?

mindstyle85:
hmm thats odd, it happened on a FN runner that just downloaded db via rclone, so only possibility is the snapshots were corrupted? I will check with him though and ask him to re-download the snapshots @gupadhyaya

mindstyle85:
confirmed it was a db problem, looks like the rclone copy yesterday was bad

im guessing it should still not explode right, but rather give out an error message and shut down?

gupadhyaya:
@mindstyle85 yes, this is also observed in go-ethereum: https://github.com/ethereum/go-ethereum/issues/19286. Will port their fix that initializes the `currentBlock` and `currentFastBlock` atomic to nil block of right type.

mindstyle85:
sure @gupadhyaya once the fix is ported in you can close this one, will be hard  to check unless we corrupt db on purpose hehe

PR #2812, opened 2020-04-11
port the fix from go-ethereum for interface conversion panic

gupadhyaya:
fixes https://github.com/harmony-one/harmony/issues/2808

problem is also observed in go-ethereum: https://github.com/ethereum/go-ethereum/issues/19286

this pr is simply porting their fix. basically, just make a nil pointer and initialize the `currentBlock` and `currentFastBlock` atomic with it.


LeoHChen:
Can you cherrypick this change to master branch as well? @gupadhyaya

PR #2816, opened 2020-04-12
port the fix from go-ethereum for interface conversion panic

gupadhyaya:


core/blockchain.go
```diff
@@ -240,6 +240,9 @@ func NewBlockChain(
 	if bc.genesisBlock == nil {
 		return nil, ErrNoGenesis
 	}
+	var nilBlock *types.Block
```
fxfactorial, line 243: What is context for this fix and why a nil pointer solves it ?

gupadhyaya: https://github.com/harmony-one/harmony/issues/2808

fxfactorial: I don't think this fixes it because this is still a nil pointer. The code will explode because loadLastState will probably explode via `currentBlock.Root()` or something else because many things will call something on block which will use header.

gupadhyaya: it is not a generic fix, but it will surely fix the exact stack trace mentioned in the above issue that happens due to interface conversion from nil to *types.Block.

PR #2817, opened 2020-04-12
[node] Redundant field removed - dead files & scripts , update deprecated libp2p package names, fold p2p code 

fxfactorial:
1. Some leftover commits from among the various branches I have open from current p2p adjustment work.
2. Update all deprecated module package naming
3. Update unidiomatic function signatures
4. Fold all p2p code into single file, needless packages, dead ideas,


node/node_handler.go
```diff
@@ -559,46 +554,30 @@ func (node *Node) pingMessageHandler(msgPayload []byte, sender libp2p_peer.ID) i
...
-		if numPeersNow == 0 {
-			utils.Logger().Info().
-				Int("numPeersNow", numPeersNow).
-				Msg("No peers, continue")
-			continue
-		} else if numPeersNow > lastPeerNum {
-			utils.Logger().Info().
-				Int("previousNumPeers", lastPeerNum).
-				Int("numPeersNow", numPeersNow).
-				Int("targetNumPeers", node.Consensus.MinPeers).
```
rlan35, line 598: This log is what we look for to see whether enough peer is obtained, which is critical in the bootstrap process. Usually when bootstrap fail. we need to double whether for this log whether it's num peer problem.

fxfactorial: I see, will add a log back but is the it incremental increase in peers that you want or the just if enough peers added or not?

fxfactorial: added a log

PR #2819, opened 2020-04-13
[p2p] Throttle receive with constant cost function

fxfactorial:
1. remove msgq
2. remove dead top level package values
3. remove unnecessary abstraction/interface/wrapper types
4. Only allow 100 message handlers at a time with a weighted semaphore.


node/node.go
```diff
@@ -374,27 +355,60 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
 
-func (node *Node) startRxPipeline(
-	receiver p2p.GroupReceiver, queue *msgq.Queue, numWorkers int,
-) {
-	// consumers
-	for i := 0; i < numWorkers; i++ {
-		go queue.HandleMessages(node)
+// Start kicks off the node message handling
+func (node *Node) Start() error {
+	allTopics := node.host.AllTopics()
```
rlan35, line 360: Where are the topics registered?

fxfactorial: They are already registered done by this point, same as before

rlan35: Ok, So we don't have topic specific worker pool anymore? Everything is competing for the same resources with equal weight?

fxfactorial: No, each one is having its own channel - same as before conceptually

fxfactorial: see the for loop

rlan35: ok, cool

LeoHChen:
so we expect this PR will prevent OOM issue? with only 300 messages being processed at the same time?

fxfactorial:
> so we expect this PR will prevent OOM issue? with only 300 messages being processed at the same time?

yes. the OOM issue also caused by the msgq impl which held onto references and wouldn't let go of memory. wrt amount of messages - its a throttle which I will make smarter later .

PR #2822, opened 2020-04-13
[core] fix validator boot status for outgoing validators

gupadhyaya:
Boot status was set inside the incoming/retained validator slots, instead should have been set for the outgoing validator slots.


fxfactorial:
Nvm , I saw files first before conversation tab

PR #2826, opened 2020-04-13
[consensus][view-change] onViewChange set keys and add signatures for all keys in case of multi-key new leader

gupadhyaya:
There were several bugs in OnViewChange that might have caused view change delays due to only adding signatures and setting keys of new leader's leader key, instead all keys must be used in the multi-key case. Note that, this will not fix the `total-power-of-signers` bug. Actually, the `IsQuorumAchieved` does not make sense for view change, as view changes used multiple signature maps (nilSigs, bhpSigs, viewIDSigs) per viewID. `IsQuorumAchieved` is not needed at this moment for view change to work correctly. 

This pr should fix the view change bugs.

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -182,8 +182,8 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 
 	// TODO: remove NIL type message
 	// add self m1 or m2 type message signature and bitmap
-	_, ok1 := consensus.nilSigs[recvMsg.ViewID][consensus.PubKey.SerializeToHexStr()]
-	_, ok2 := consensus.bhpSigs[recvMsg.ViewID][consensus.PubKey.SerializeToHexStr()]
+	_, ok1 := consensus.nilSigs[recvMsg.ViewID][newLeaderKey.SerializeToHexStr()]
+	_, ok2 := consensus.bhpSigs[recvMsg.ViewID][newLeaderKey.SerializeToHexStr()]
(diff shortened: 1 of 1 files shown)


consensus/view_change.go
```diff
@@ -356,22 +365,25 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
+				priKey := consensus.priKey.PrivateKey[i]
+				if _, err := consensus.Decider.SubmitVote(
+					quorum.Commit,
+					key,
+					priKey.SignHash(commitPayload),
+					common.BytesToHash(consensus.blockHash[:]),
+					consensus.blockNum,
+					recvMsg.ViewID,
+				); err != nil {
+					consensus.getLogger().Debug().Msg("submit vote on viewchange commit failed")
```
fxfactorial, line 378: .Err(err).Msg()

consensus/view_change.go
```diff
@@ -356,22 +365,25 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
+					return
+				}
 
-			if err := consensus.commitBitmap.SetKey(newLeaderKey, true); err != nil {
-				consensus.getLogger().Debug().
-					Msg("[OnViewChange] New Leader commit bitmap set failed")
-				return
+				if err := consensus.commitBitmap.SetKey(key, true); err != nil {
+					consensus.getLogger().Debug().
```
fxfactorial, line 383: same, .Err(err).Msg(

PR #2828, opened 2020-04-14
[Explorer Node] Explorer service StakingTransaction struct unification for explorer RPC level staking message data decoding

denniswon:
- make staking txn field values to be decoded on explorer node ldb storage
- Explorer Staking Transaction message struct unified to contain all fields of different staking txns (Create/EditValidator, (Un)Delegate, CollectReward) so that the go struct of StakingTransaction for Explorer node can use that merged Staking Transaction msg struct as the return type and return the staking txns as decoded
- do use to-address field for staking txns for explorer node storage so that fetching by sent vs received txns work too for staking txns. 

## Issue

Fixes issue: https://github.com/harmony-one/harmony-dashboard/issues/58 and https://github.com/harmony-one/harmony-dashboard/issues/69

## Test

Launched local net with an explorer node, query the explorer rpc for getStakingTransactionByHash and getStakingTransactionsHistory api to see that the issues above are fixed (received txns do get retrieved even for staking txns) and now the staking message data is returned decoded.

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -94,7 +94,6 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
 		if err != nil {
 			utils.Logger().Error().Err(err).Str("txHash", tx.Hash().String()).
 				Msg("Failed to get explorer StakingTransaction mapping")
-			continue
 		}
 		storage.UpdateStakingTxAddress(explorerTransaction, tx)
 	}
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -90,7 +90,12 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
 	}
 	// Store staking txns
 	for _, tx := range block.StakingTransactions() {
-		explorerTransaction := GetStakingTransaction(tx, block)
+		explorerTransaction, err := GetStakingTransaction(tx, block)
+		if err != nil {
+			utils.Logger().Error().Err(err).Str("txHash", tx.Hash().String()).
+				Msg("Failed to get explorer StakingTransaction mapping")
+			continue
+		}
 		storage.UpdateStakingTxAddress(explorerTransaction, tx)
 	}
 }
--- a/api/service/explorer/structs.go
+++ b/api/service/explorer/structs.go
@@ -1,15 +1,22 @@
 package explorer
 
 import (
+	"github.com/ethereum/go-ethereum/common"
+
+	"bytes"
 	"encoding/hex"
 	"math/big"
 	"strconv"
 
 	core2 "github.com/harmony-one/harmony/core"
 	"github.com/harmony-one/harmony/core/types"
-	"github.com/harmony-one/harmony/internal/common"
+	common2 "github.com/harmony-one/harmony/internal/common"
 	"github.com/harmony-one/harmony/internal/utils"
+	"github.com/harmony-one/harmony/numeric"
+	"github.com/harmony-one/harmony/shard"
+	"github.com/harmony-one/harmony/staking/effective"
 	staking "github.com/harmony-one/harmony/staking/types"
+	"github.com/pkg/errors"
 )
 
 /*
@@ -20,6 +27,15 @@ import (
 const (
 	Received = "RECEIVED"
 	Sent     = "SENT"
+)
+
+// Errors ...
+var (
+	// ErrInvalidMsgForStakingDirective is returned if a staking message does not
+	// match the related directive
+	ErrInvalidMsgForStakingDirective = errors.New("staking message does not match directive message")
+	// ErrInvalidSender is returned if a txn or staking txn message has incorrect from address
+	ErrInvalidSender = errors.New("staking message does not match directive message")
 )
 
 // Data ...
@@ -60,12 +76,12 @@ func GetTransaction(tx *types.Transaction, addressBlock *types.Block) *Transacti
 	gasFee = gasFee.Mul(tx.GasPrice(), new(big.Int).SetUint64(tx.Gas()))
 	to := ""
 	if msg.To() != nil {
-		if to, err = common.AddressToBech32(*msg.To()); err != nil {
+		if to, err = common2.AddressToBech32(*msg.To()); err != nil {
 			return nil
 		}
 	}
 	from := ""
-	if from, err = common.AddressToBech32(msg.From()); err != nil {
+	if from, err = common2.AddressToBech32(msg.From()); err != nil {
 		return nil
 	}
 	return &Transaction{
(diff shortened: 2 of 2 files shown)


gupadhyaya:
@denniswon not sure how https://github.com/harmony-one/harmony-dashboard/issues/58 is fixed? 
for https://github.com/harmony-one/harmony-dashboard/issues/69, fully decoding the staking transaction message part is an overkill, unless you intend to use all the fields. Also, can you not simply decode the staking message and populate the `to` field such that account-history for both validator and delegator addresses are correctly maintained? 
I don't see the requirement for fully decoding and having a unified staking tx type for rpc. Can you please explain?

denniswon:
@gupadhyaya 

1. harmony-one/harmony-dashboard#58 and harmony-one/harmony-dashboard#69 are related, caused by same issue. The missing txns not returned by transactions history api call are the received staking txns.

2. can you not simply decode the staking message and populate the to field such that account-history for both validator and delegator addresses are correctly maintained
--> populating the to field is the exact fix of the 2 issues above. Staking txns dont have to field in StakingTransaction struct, so the logic for saving txns instances as sent and received for regular txns was not working (or working only half, just saving as sent transaction).    

3. don't see the requirement for fully decoding and having a unified staking tx type for rpc. --> True, not a requirement. It's more of a design choice. Stuffing actual meaningful data all in msg type as json and let the clients handle, or decode at rpc layer. I thoght the latter would be more user-friendly. And with transactoin and staking txns unification later, this decoding would go away.

api/service/explorer/structs.go
```diff
@@ -129,4 +170,156 @@ func GetStakingTransaction(tx *staking.StakingTransaction, addressBlock *types.B
...
+		if err != nil {
+			return nil, err
+		}
+
+		if _, ok := stkMsg.(*staking.CreateValidator); !ok {
+			return nil, ErrInvalidMsgForStakingDirective
+		}
+
+		createValidatorMsg := stkMsg.(*staking.CreateValidator)
+		if bytes.Compare(msg.From().Bytes(), createValidatorMsg.ValidatorAddress.Bytes()) != 0 {
```
gupadhyaya, line 186: lets use `!bytes.Equal(msg.From()[:], createValidatorMsg.ValidatorAddress[:])`

api/service/explorer/structs.go
```diff
@@ -22,6 +29,15 @@ const (
 	Sent     = "SENT"
 )
 
+// Errors ...
+var (
+	// ErrInvalidMsgForStakingDirective is returned if a staking message does not
+	// match the related directive
+	ErrInvalidMsgForStakingDirective = errors.New("staking message does not match directive message")
```
gupadhyaya, line 36: no need to redefine these? available under core2?

api/service/explorer/structs.go
```diff
@@ -85,38 +101,63 @@ func GetTransaction(tx *types.Transaction, addressBlock *types.Block) *Transacti
...
+	Description        staking.Description     `json:"description"`
+	CommissionRates    staking.CommissionRates `json:"commission"`
+	MinSelfDelegation  *big.Int                `json:"min-self-delegation"`
+	MaxTotalDelegation *big.Int                `json:"max-total-delegation"`
+	SlotPubKeys        []shard.BLSPublicKey    `json:"slot-pub-keys"`
+	SlotKeySigs        []shard.BLSSignature    `json:"slot-key-sigs"`
+
+	// EditValidator
+
+	// ValidatorAddress common.Address        `json:"validator-address"`
```
gupadhyaya, line 118: cleanup please.

api/service/explorer/structs.go
```diff
@@ -85,38 +101,63 @@ func GetTransaction(tx *types.Transaction, addressBlock *types.Block) *Transacti
...
-	Data      string   `json:"data"`
-	GasFee    *big.Int `json:"gasFee"`
-	FromShard uint32   `json:"fromShard"`
-	ToShard   uint32   `json:"toShard"`
-	Type      string   `json:"type"`
+	Transaction
+
+	// CreateValidator
+
+	ValidatorAddress   common.Address          `json:"validator-address"`
```
gupadhyaya, line 108: I agree that by expanding fully, we can save a lot of decoding time during access, but this would mean that there will be lot of `null` fields in the output? How would you avoid the possible confusions? 
@Dewansahil?

api/service/explorer/storage.go
```diff
@@ -90,7 +90,12 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
 	}
 	// Store staking txns
 	for _, tx := range block.StakingTransactions() {
-		explorerTransaction := GetStakingTransaction(tx, block)
+		explorerTransaction, err := GetStakingTransaction(tx, block)
+		if err != nil {
+			utils.Logger().Error().Err(err).Str("txHash", tx.Hash().String()).
+				Msg("Failed to get explorer StakingTransaction mapping")
+			continue
```
gupadhyaya, line 97: why is it okay to continue here when error for some staking transaction?

api/service/explorer/structs.go
gupadhyaya: I still think, we could simply store the staking tx specific contents under data (json string) and only populate the to field. Anyways, the explorer is meant to be consumed for information purpose only. Please clarify, @denniswon and @Dewansahil

denniswon: @gupadhyaya agree about the empty fields. that was because of the nature of golang, where it does not allow parent struct to be used as the function return type and have the child struct instances to be returned. Any suggestion? @fxfactorial

denniswon: @gupadhyaya https://stackoverflow.com/questions/24216510/empty-or-not-required-struct-fields-in-golang
Found the solution. will using 'omitempty' tag

denniswon: Since this needs to be discussed further, I will revert this staking transaction fields unification change from this pr to unblock launch issues. Will come back to this discussion later.

api/service/explorer/storage.go
```diff
@@ -90,7 +90,11 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
 	}
 	// Store staking txns
 	for _, tx := range block.StakingTransactions() {
-		explorerTransaction := GetStakingTransaction(tx, block)
+		explorerTransaction, err := GetStakingTransaction(tx, block)
+		if err != nil {
+			utils.Logger().Error().Err(err).Str("txHash", tx.Hash().String()).
+				Msg("Failed to get explorer StakingTransaction mapping")
+		}
 		storage.UpdateStakingTxAddress(explorerTransaction, tx)
```
gupadhyaya, line 98: will explode if explorerTransaction is nil @denniswon

fxfactorial: giving back nil should have been caught in the err case @denniswon

Issue #2802, opened 2020-04-10
Invalid Bloom on beacon sync after crashed node got sync'ed again.


rlan35:
Nodes in shard 2 crashed without graceful shutdown, so the blocks are lost and need resync.

The node immediate see errors below in beacon sync after restart.

Similar node which didn't crash didn't see the error.

Suspect it's related to offchain data inconsistency due to crash.

\nError: invalid bloom (remote: 00000000000000000000000000000000000000000000000000000000000000000000000000008000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000200000000000000000000400000010000000000000000000000000000800000000000000000000000000000000000000000000000040000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000  local: 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000)\n##############################\n"}

LeoHChen:
The db snapshot of shard2 is uploaded here. s3://pub.harmony.one/ostn/crash/0409/hmydb.tgz

rlan35:
Fixed.

PR #2834, opened 2020-04-14
make delegation index consistent by block number

rlan35:
For delegationIndex used for collect rewards, when there is a crash, the offchain delegationIndex stay the same, but the latest blocks need to be re-processed to catch up

So when the node resync, it will use delegationIndex for the delegation that does not exist yet, which will lead to 0 rewards for the CollectReward txn which should have some rewards.

This PR add block number to the delegation index so the index can be used consistently with the current block.

Fixing: https://github.com/harmony-one/harmony/issues/2802

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2414,16 +2414,28 @@ func (bc *BlockChain) WriteValidatorList(
 // ReadDelegationsByDelegator reads the addresses of validators delegated by a delegator
 func (bc *BlockChain) ReadDelegationsByDelegator(
 	delegator common.Address,
-) (staking.DelegationIndexes, error) {
+) (m staking.DelegationIndexes, err error) {
+	rawResult := staking.DelegationIndexes{}
 	if cached, ok := bc.validatorListByDelegatorCache.Get(string(delegator.Bytes())); ok {
 		by := cached.([]byte)
-		m := []staking.DelegationIndex{}
-		if err := rlp.DecodeBytes(by, &m); err != nil {
+		if err := rlp.DecodeBytes(by, &rawResult); err != nil {
 			return nil, err
 		}
-		return m, nil
+	} else {
+		if rawResult, err = rawdb.ReadDelegationsByDelegator(bc.db, delegator); err != nil {
+			return nil, err
+		}
 	}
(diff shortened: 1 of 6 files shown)


core/blockchain.go
```diff
@@ -2414,16 +2414,28 @@ func (bc *BlockChain) WriteValidatorList(
...
 			return nil, err
 		}
-		return m, nil
+	} else {
+		if rawResult, err = rawdb.ReadDelegationsByDelegator(bc.db, delegator); err != nil {
+			return nil, err
+		}
+	}
+	for _, index := range rawResult {
+		if index.BlockNum.Cmp(bc.CurrentBlock().Number()) <= 0 {
```
fxfactorial, line 2430: Don't need to need `bc.CurrentBlock().Number()` over & over

core/blockchain.go
```diff
@@ -2567,7 +2580,7 @@ func (bc *BlockChain) prepareStakingMetaData(
 				}
 			}
 			if delegations, err = bc.addDelegationIndex(
-				delegations, delegate.DelegatorAddress, delegate.ValidatorAddress, state,
+				delegations, delegate.DelegatorAddress, delegate.ValidatorAddress, state, block.Number(),
```
fxfactorial, line 2583: same, don't need .Number over & over, right

multibls/multibls.go
```diff
@@ -23,7 +23,7 @@ func (multiKey *PublicKey) SerializeToHexStr() string {
 	}
 	var builder strings.Builder
 	for _, pubKey := range multiKey.PublicKey {
-		builder.WriteString(pubKey.SerializeToHexStr())
+		builder.WriteString(pubKey.SerializeToHexStr() + ";")
```
fxfactorial, line 26: will this break anything existing?

rlan35: it's only for printouts. No logical change.

core/blockchain.go
rlan35: right. fixed.

core/blockchain.go
rlan35: fixed.

PR #2839, opened 2020-04-16
Optimize delegation reward distribution with caching

rlan35:
1. Make validator snapshot tied with the epoch with a new struct ValidatorSnapshot
2. Use caching for the calculation of shares of delegators in the reward distribution time. This will save around 1/3 of the time in AddReward()
3. Fix the cache cleanup logic to always cleanup data 3 epochs ago.

Did more scalability testing, it turns out the loading (rlp decode) of the validator takes similar amount of time of calculating the shares of all its delegators.

```
Time required to read validator: 0.057869 seconds
Time required to calc percentage 100001 delegations: 0.064249 seconds
Time required to reward a validator with 100001 delegations: 0.122621 seconds
```

So we should next optimize the code to avoid reading and writing validator repeated for different operations.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -753,7 +753,7 @@ var (
 )
 
 // AddReward distributes the reward to all the delegators based on stake percentage.
-func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int) error {
+func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int, shareLookup map[common.Address]numeric.Dec) error {
 	if reward.Cmp(common.Big0) == 0 {
 		utils.Logger().Info().RawJSON("validator", []byte(snapshot.String())).
 			Msg("0 given as reward")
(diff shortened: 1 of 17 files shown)


internal/chain/reward.go
```diff
@@ -42,11 +45,12 @@ func ballotResultBeaconchain(
 }
 
 var (
-	votingPowerCache singleflight.Group
+	votingPowerCache   singleflight.Group
```
denniswon, line 48: wow singleflight is cool. good learning :)

core/state/statedb.go
```diff
@@ -783,19 +783,17 @@ func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int) error {
...
-		// NOTE percentage = <this_delegator_amount>/<total_delegation>
-		if totalDelegationDec.Equal(zero) {
-			utils.Logger().Info().
-				RawJSON("validator-snapshot", []byte(snapshot.String())).
-				Msg("zero total delegation during AddReward delegation payout")
-			return nil
+		percentage, ok := shareLookup[delegation.DelegatorAddress]
+
+		if !ok {
+			continue
```
fxfactorial, line 794: this happening is an oddity, so we should have logged at least something

shard/committee/assignment.go
```diff
@@ -184,18 +184,18 @@ func prepareOrders(
...
-func IsEligibleForEPoSAuction(snapshot, validator *staking.ValidatorWrapper, curEpoch *big.Int) bool {
+func IsEligibleForEPoSAuction(snapshot *staking.ValidatorSnapshot, validator *staking.ValidatorWrapper) bool {
 	// This original condition to check whether a validator is in last committee is not stable
 	// because cross-links may arrive after the epoch ends and it still got counted into the
 	// NumBlocksToSign, making this condition to be true when the validator is actually not in committee
 	//if snapshot.Counters.NumBlocksToSign.Cmp(validator.Counters.NumBlocksToSign) != 0 {
 
 	// Check whether the validator is in current committee
-	if validator.LastEpochInCommittee.Cmp(curEpoch) == 0 {
+	if validator.LastEpochInCommittee.Cmp(snapshot.Epoch) == 0 {
```
fxfactorial, line 194: why this changed?

Issue #2831, opened 2020-04-14
Explorer node restart will accumulate redundant txn history data


rlan35:
1. Explorer goes through all blocks and process them again everytime explorer node is restarted.
2. it was fine before because there is no cummulative data to store, but now it doesn't work because we added txn history data, which append new txns to the existing data.
3. If we don't remove the explorer db , then every restart will double the txn history data on the addresses. Making it more and more likely to OOM

Request:
1. We need to add checkpoint of processed blocks so as to not process them again
2. Need to change the txn history logic to avoid adding duplicate txns.
3. We need to store the txns data separately. rather than appending raw txn data together for txn history.

AndyBoWu:
for @john-harmony , the explorer DB is located in the home dir with name pattern like `explorer_storage_<ip>`. you can remove the old data by issue the cmd `sudo rm -rd explorer_storage_<ip>`. This db folder is NOT the harmony db folder we refer to. 

LMK if you have any questions.

PR #2842, opened 2020-04-16
explorer node. Skip dump for redundant blocks with lower block number than the checkpoint block number

denniswon:
## Issue
Point 1. Explorer goes through all blocks and process them again everytime explorer node is restarted.
from https://github.com/harmony-one/harmony/issues/2831

## Test
Launched explorer node from the binary built with this change, and checked the log for block dumps being skipped for blocks that are already saved on the node ldb.

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -2,6 +2,7 @@ package explorer
 
 import (
 	"fmt"
+	"math/big"
 	"os"
 	"path"
 	"sync"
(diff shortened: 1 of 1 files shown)


api/service/explorer/storage.go
```diff
@@ -97,6 +108,9 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
 		}
 		storage.UpdateStakingTxAddress(explorerTransaction, tx)
 	}
+
+	// save checkpoint of block dumped
+	storage.GetDB().Put([]byte(DumpCheckpointKey), blockNum.Bytes(), nil)
```
rlan35, line 113: if check point is 10, and a block with 20 is dumped. Then the checkpoint is set to 20 and 11-19 blocks won't be dumped anymore. The checkpointing mechanism will be more complicated than this. I would recommend just add a key for each block to mark whether they are dumped or not. It will be much simpler, just a bit more db space needed.

denniswon: good point. I thought that was not possible, but checking the code, during state sync, its actually possible as dump doesn't necessarily dump blocks from genesis in order.

PR #2844, opened 2020-04-16
Add cache to state validator modification; refactor snapshot read/write

rlan35:
Add cache to state validator access. So other than staking txns, all other validator wrapper changes don't need to rlp decode and encode everytime the wrapper is changed.;

Validator wrapper change from staking txns still needs the rlp decode/encode because txn may fail and we don't have the mechanism to easily revert the changes on validator data.

Refactor snapshot read/write;

Directly cache on snapshot object, rather than the byte code.

Results:

Before this change:
Sync'ing  OSTN from block 0 to 32173 of shard 0 took around 1.5 hour. On Average, each block took **0.17s**, and the epoch block took **0.3-0.5s**. Definitely a lot of room to optimize.

After this change:
The same sync work took 1h24m. On average, a normal block took **0.156s** and an epoch block took **0.2-0.25s**. 

Not a huge gain because there isn't much delegations in the OSTN now. The expected gain will be at least 50% reduction on block processing time when the number of delegations is huge (e.g. 10k-100k delegations in total, which will happen eventually)

@SebastianJ can you do some comparison of block processing time with/without this change in STN after a lot of delegations is created?

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2203,8 +2203,8 @@ func (bc *BlockChain) ReadValidatorSnapshot(
 	return rawdb.ReadValidatorSnapshot(bc.db, addr, epoch)
 }
 
-// writeValidatorSnapshot writes the snapshot of provided validator
-func (bc *BlockChain) writeValidatorSnapshot(
+// WriteValidatorSnapshot writes the snapshot of provided validator
+func (bc *BlockChain) WriteValidatorSnapshot(
 	batch rawdb.DatabaseWriter, snapshot *staking.ValidatorSnapshot,
 ) error {
 	// Batch write the current data as snapshot
@@ -2367,7 +2367,7 @@ func (bc *BlockChain) UpdateValidatorSnapshots(
 		}
 
 		snapshot := &staking.ValidatorSnapshot{validator, epoch}
-		if err := bc.writeValidatorSnapshot(batch, snapshot); err != nil {
+		if err := bc.WriteValidatorSnapshot(batch, snapshot); err != nil {
 			return err
 		}
 	}
@@ -2485,13 +2485,13 @@ func (bc *BlockChain) UpdateStakingMetaData(
 				return newValidators, err
 			}
 
-			if err := rawdb.WriteValidatorSnapshot(batch, validator, epoch); err != nil {
+			if err := bc.WriteValidatorSnapshot(batch, &staking.ValidatorSnapshot{validator, epoch}); err != nil {
 				return newValidators, err
 			}
 			// For validator created at exactly the last block of an epoch, we should create the snapshot
 			// for next epoch too.
 			if newEpoch.Cmp(epoch) > 0 {
-				if err := rawdb.WriteValidatorSnapshot(batch, validator, newEpoch); err != nil {
+				if err := bc.WriteValidatorSnapshot(batch, &staking.ValidatorSnapshot{validator, newEpoch}); err != nil {
 					return newValidators, err
 				}
 			}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2366,8 +2366,7 @@ func (bc *BlockChain) UpdateValidatorSnapshots(
 			return err
 		}
 
-		copy := *validator
-		snapshot := &staking.ValidatorSnapshot{&copy, epoch}
+		snapshot := &staking.ValidatorSnapshot{validator, epoch}
 		if err := bc.writeValidatorSnapshot(batch, snapshot); err != nil {
 			return err
 		}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -148,7 +148,7 @@ type BlockChain struct {
 	lastCommitsCache              *lru.Cache
 	epochCache                    *lru.Cache    // Cache epoch number → first block number
 	randomnessCache               *lru.Cache    // Cache for vrf/vdf
-	validatorCache                *lru.Cache    // Cache for validator info
+	validatorSnapshotCache        *lru.Cache    // Cache for validator snapshot
 	validatorStatsCache           *lru.Cache    // Cache for validator stats
 	validatorListCache            *lru.Cache    // Cache of validator list
 	validatorListByDelegatorCache *lru.Cache    // Cache of validator list by delegator
@@ -217,7 +217,7 @@ func NewBlockChain(
 		lastCommitsCache:              commitsCache,
 		epochCache:                    epochCache,
 		randomnessCache:               randomnessCache,
-		validatorCache:                validatorCache,
+		validatorSnapshotCache:        validatorCache,
 		validatorStatsCache:           validatorStatsCache,
 		validatorListCache:            validatorListCache,
 		validatorListByDelegatorCache: validatorListByDelegatorCache,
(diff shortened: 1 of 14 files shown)


core/staking_verifier.go
```diff
@@ -313,11 +313,6 @@ func VerifyAndCollectRewardsFromDelegation(
 				Msg("Delegation index out of bound")
 			return nil, nil, errors.New("Delegation index out of bound")
 		}
-		if err := wrapper.SanityCheck(
```
gupadhyaya, line 316: why sanity check not required?

core/state/statedb.go
```diff
@@ -592,6 +597,11 @@ func (db *DB) GetRefund() uint64 {
 // Finalise finalises the state by removing the db destructed objects
 // and clears the journal as well as the refunds.
 func (db *DB) Finalise(deleteEmptyObjects bool) {
```
gupadhyaya, line 599: what should be the behavior for graceful shutdown? flush the cache or inconsistent state is okay?

core/blockchain.go
```diff
@@ -2168,7 +2168,7 @@ func (bc *BlockChain) ReadValidatorInformationAt(
 	if err != nil || state == nil {
 		return nil, errors.Wrapf(err, "at root: %s", root.Hex())
 	}
-	wrapper, err := state.ValidatorWrapper(addr)
+	wrapper, err := state.ValidatorWrapper(addr, false)
```
fxfactorial, line 2171: Please hide this function and make two wrappers , one for true and false and put name of book intent right in function name

core/blockchain.go
```diff
@@ -2381,7 +2358,21 @@ func (bc *BlockChain) UpdateValidatorSnapshots(
 
 	allValidators = append(allValidators, newValidators...)
 
-	return bc.writeValidatorSnapshots(batch, allValidators, epoch, state)
+	// Read all validator's current data and snapshot them
+	for i := range allValidators {
```
fxfactorial, line 2362: We can do this concurrently

core/state/statedb.go
```diff
@@ -592,6 +597,11 @@ func (db *DB) GetRefund() uint64 {
...
+	// Commit validator changes in cache to stateObjects
+	for addr, val := range db.stateValidators {
```
fxfactorial, line 601: This should be atomic right

core/state/statedb.go
```diff
@@ -791,7 +820,7 @@ func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int, shareLo
 		percentage, ok := shareLookup[delegation.DelegatorAddress]
 
 		if !ok {
-			continue
+			return errors.Wrapf(err, "missing delegation shares for reward distribution")
```
fxfactorial, line 823: Why this and error now and not before

core/state_transition.go
```diff
@@ -397,8 +396,8 @@ func (st *StateTransition) verifyAndApplyCreateValidatorTx(
 	if err := st.state.UpdateValidatorWrapper(wrapper.Address, wrapper); err != nil {
 		return err
 	}
-	st.state.SetValidatorFlag(wrapper.Address)
-	st.state.SubBalance(wrapper.Address, createValidator.Amount)
+	st.state.SetValidatorFlag(createValidator.ValidatorAddress)
```
fxfactorial, line 399: Why this flip in choice

core/vm/interface.go
```diff
@@ -42,7 +42,7 @@ type StateDB interface {
 	SetCode(common.Address, []byte)
 	GetCodeSize(common.Address) int
 
-	ValidatorWrapper(common.Address) (*staking.ValidatorWrapper, error)
+	ValidatorWrapper(common.Address, bool) (*staking.ValidatorWrapper, error)
```
fxfactorial, line 45: We should take the three validator methods , assuming you do the two named book methods , and factor them out into separate interface , then embed into this one , then we can more easily test just staking validator related state changes

internal/chain/engine.go
```diff
@@ -360,19 +354,13 @@ func setLastEpochInCommittee(header *block.Header, state *state.DB) error {
...
 	for _, addr := range newShardState.StakedValidators().Addrs {
-		wrapper, err := state.ValidatorWrapper(addr)
+		wrapper, err := state.ValidatorWrapper(addr, false)
 		if err != nil {
 			return errors.New(
 				"[Finalize] failed to get validator from state to finalize",
 			)
 		}
 		wrapper.LastEpochInCommittee = newShardState.Epoch
-		if err := state.UpdateValidatorWrapper(
```
fxfactorial, line 370: I guess all these updates were redundant ?

node/worker/worker.go
```diff
@@ -461,11 +462,8 @@ func (w *Worker) FinalizeNewBlock(
 		}
 	}
 
-	state := w.current.state.Copy()
-	copyHeader := types.CopyHeader(w.current.header)
-	// TODO: feed coinbase into here so the proposer gets extra rewards.
 	block, _, err := w.engine.Finalize(
-		w.chain, copyHeader, state, w.current.txs, w.current.receipts,
```
fxfactorial, line 468: You made me add the header copy , now taking it out lol

staking/availability/measure.go
```diff
@@ -108,12 +108,6 @@ func bumpCount(
 					wrapper.Counters.NumBlocksSigned, common.Big1,
 				)
 			}
-
```
fxfactorial, line 111: Redudant write ?

staking/slash/double-sign.go
```diff
@@ -473,9 +474,7 @@ func Apply(
 			RawJSON("slash", []byte(slash.String())).
 			Msg("about to update staking info for a validator after a slash")
 
-		if err := state.UpdateValidatorWrapper(
-			snapshot.Validator.Address, current,
-		); err != nil {
+		if err := current.SanityCheck(staking.DoNotEnforceMaxBLS); err != nil {
```
fxfactorial, line 477: I think you turned off the check so it was intentional , then can remove the bls const hack

core/blockchain.go
rlan35: Leveldb batch is not thread safe https://github.com/google/leveldb/blob/master/doc/index.md#concurrency

core/staking_verifier.go
rlan35: it's redundant, because CollectReward doesn't affect any part of the sanity checks. Meaning it will never fail the check.

core/state/statedb.go
rlan35: The in-memory data in a state being processed is ephemeral data before the state is finalized and confirmed. So no need to worry about graceful shutdown (same for the existing stateObjects logic)

core/state/statedb.go
rlan35: This is not committing to DB. It's only committing to the in-memory cache of stateObject, which will only be committed to DB when state.Commit() is called. So stateValidators is only a layer of cache for validatorWrapper object so it doesn't have to rlp to stateObject.code. stateObjects cache will be eventually committed to db when state.Commit() is called.

core/state/statedb.go
rlan35: I was thinking to be lenient on the error to not stall the consensus, but after a second thought, this better be a explicit error to expose problems.

core/state_transition.go
rlan35: just to all use createValidator as the source of truth, basically it's the same.

staking/slash/double-sign.go
rlan35: what do you mean turn of the bls const hack? You mean the SanityCheck is not needed?

staking/availability/measure.go
rlan35: yea, for non-staking txn related validatorWrapper changes, there is no need to do UpdateValidatorWrapper. Just directly change on the live object using state.ValidatorWrapper(addr, false). Because if there is an error, the whole processing of state will error out. No need to revert changes.

node/worker/worker.go
rlan35: Yea, at first it's for safety. But after looking deeper, now can remove for speed.

internal/chain/engine.go
rlan35: unless the change may fail and need to revert (in staking txn case). Other cases can directly update on the live validatorWrapper object.

core/vm/interface.go
rlan35: clarify your suggestion? I don't get it.

core/blockchain.go
fxfactorial: interesting - how much benefit is there to the batch though vs concurrent write?

core/vm/interface.go
fxfactorial: like:

```go
type ValidatorUpdater interface {
  WrapperBoolTrue(common.Address)
  WrapperBoolFalse(common.Addres)
  UpsertValidator(common.Addrsss...)
} 


type StateDB struct {
  ValidatorUpdater
} 
```

Something like that, but not a big deal if not done

fxfactorial: where the booltrue/boolfalse are better named to reflect what the bool param does.

staking/slash/double-sign.go
fxfactorial: oh wait, I was wrong, I saw wrng on earlier comment by ganesh, I though this part of logic in the check itself was removed

core/blockchain.go
rlan35: Batch is not for mainly for speed, but for atomicity

core/vm/interface.go
rlan35: ok, agree.

core/blockchain.go
rlan35: done.

rlan35:
jenkins, test this please.

PR #2857, opened 2020-04-18
explorer node. handle get txn/staking txn being nil during block dump

denniswon:
## Issue

unknown

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -102,10 +102,9 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
 		if err != nil {
 			utils.Logger().Error().Err(err).Str("txHash", tx.Hash().String()).
 				Msg("[Explorer Storage] Failed to get GetTransaction mapping")
+			continue
 		}
-		if explorerTransaction != nil {
-			storage.UpdateTxAddress(explorerTransaction, tx)
-		}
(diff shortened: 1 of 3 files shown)


api/service/explorer/storage.go
```diff
@@ -98,17 +98,25 @@ func (storage *Storage) Dump(block *types.Block, height uint64) {
...
 	// Store txs
 	for _, tx := range block.Transactions() {
-		explorerTransaction := GetTransaction(tx, block)
-		storage.UpdateTxAddress(explorerTransaction, tx)
+		explorerTransaction, err := GetTransaction(tx, block)
+		if err != nil {
+			utils.Logger().Error().Err(err).Str("txHash", tx.Hash().String()).
+				Msg("[Explorer Storage] Failed to get GetTransaction mapping")
+		}
+		if explorerTransaction != nil {
```
gupadhyaya, line 106: not necessary to check for nil, as you err != nil already means `explorerTransaction` is nil. May be just `continue` when `err != nil`? Same for below.

denniswon: Yeah originally i had continue here from my original pr. But then during review removed (think had this discussion with you) continue was removed so that we can actually catch the case where txn is nil and err is also nil. But it seems now that continue is actually better here than seg fault as we can still investigate with err logs.

denniswon:
@gupadhyaya updated. :)

PR #2858, opened 2020-04-19
Explorer node dump opt

denniswon:
## Issue

related https://github.com/harmony-one/harmony/issues/2831

This PR optimizes the block dump process in terms of memory usage during dump and also the disk storage size of transactions history. Both transaction and staking transactions.

### Unit Test Coverage

```
dennis.won@Jongs-MacBook-Pro:~/harmony-one/harmony (explorer_node_dump_opt) $ go test ./api/service/explorer/
ok  	github.com/harmony-one/harmony/api/service/explorer	0.785s
dennis.won@Jongs-MacBook-Pro:~/harmony-one/harmony (explorer_node_dump_opt) $ go test ./node/node
node.go               node_cross_link.go    node_explorer.go      node_handler.go       node_newblock.go      node_test.go
node.md               node_cross_shard.go   node_genesis.go       node_handler_test.go  node_syncing.go       node_utils.go
dennis.won@Jongs-MacBook-Pro:~/harmony-one/harmony (explorer_node_dump_opt) $ go test ./node
ok  	github.com/harmony-one/harmony/node	1.408s
dennis.won@Jongs-MacBook-Pro:~/harmony-one/harmony (explorer_node_dump_opt) $
```

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -11,7 +11,6 @@ import (
 	"github.com/harmony-one/harmony/core/types"
 	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
 	"github.com/harmony-one/harmony/internal/utils"
-	staking "github.com/harmony-one/harmony/staking/types"
 	"github.com/syndtr/goleveldb/leveldb"
 	"github.com/syndtr/goleveldb/leveldb/filter"
 	"github.com/syndtr/goleveldb/leveldb/opt"
@@ -20,9 +19,10 @@ import (
 
 // Constants for storage.
 const (
-	AddressPrefix    = "ad"
-	CheckpointPrefix = "dc"
-	PrefixLen        = 3
+	AddressPrefix        = "ad"
+	CheckpointPrefix     = "dc"
+	PrefixLen            = 3
+	AdddressTxnsInitSize = 10
 )
 
 // GetAddressKey ...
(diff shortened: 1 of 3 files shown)


LeoHChen:
conflict files, please resolve it. thanks.

gupadhyaya:
@denniswon how urgent is this PR? can it wait few more days? my PR (unifying stx and tx) will conflict with this and some of the code will be removed anyways.

denniswon:
@gupadhyaya not super urgent, but don't think we should delay few more days. The txns and staking txns unification downstream update does require many different components on the code to be updated modularly, and explorer node usage is just one of them. And when explorer node update happens, the storage of explorer node is also just one of it. Instead of putting back all these downstream changes on hold, we should plan and execute with plan.

rlan35:
> @gupadhyaya not super urgent, but don't think we should delay few more days. The txns and staking txns unification downstream update does require many different components on the code to be updated modularly, and explorer node usage is just one of them. And when explorer node update happens, the storage of explorer node is also just one of it. Instead of putting back all these downstream changes on hold, we should plan and execute with plan.

Actually this PR still is not solving the root problem. The root problem is that the all the raw transactions of an address are concatenated and stored as a big list and stored in a single DB entry. When the number of txns grows larger (1000s) in a single address, it will need to rlp encode/decode 1000s txns in order add a single new txns. 

The best solution is definitely a SQL-like query approach where the txns are stored and indexed by address. You can do some search and see whether leveldb support it (I doubt it since it's no-sql db). 

The other clear optimization is to only store list of txn hashes for the address. And store the txn separately. This way instead of storing 1000s of raw txns together as one record, it's 1000s of hashes, which will cost much less memory when loaded to memory.

denniswon:
The root problem is that the all the raw transactions of an address are concatenated and stored as a big list and stored in a single DB entry. When the number of txns grows larger (1000s) in a single address, it will need to rlp encode/decode 1000s txns in order add a single new txns.

-> this PR resolves this by not storing raw txns anymore. It only saves the txRecord containing minimal info (hash, timetstamp, type --> the latter 2 needed for query ordering)

The best solution is definitely a SQL-like query approach where the txns are stored and indexed by address. You can do some search and see whether leveldb support it (I doubt it since it's no-sql db). This PR makes the update to actually not store raw transactions anywhere in the leveldb during dump. This is not needed as after getting the list of tx hashses, it calls getTransactionByHash() which reads directly from chain. https://github.com/harmony-one/harmony/blob/e0ccc7b71cc1aa51dfdd55573d1670ec190cb84c/internal/hmyapi/apiv2/transactionpool.go#L69

This itself will achieve major space/load optimization of explorer nodes. Please review this PR again more closely. cc: @rlan35

denniswon:
> Actually this PR still is not solving the root problem. The root problem is that the all the raw transactions of an address are concatenated and stored as a big list and stored in a single DB entry. When the number of txns grows larger (1000s) in a single address, it will need to rlp encode/decode 1000s txns in order add a single new txns.
> 

This PR clearly solves this issue you address here.
-- 1. It does NOT store raw txn anymore. It just store TxRecord for each txn (txRecord being a struct of hash, timestamp, and type). Think of TxRecord as an "index for chain db" to enable quick tx hashes search queries for transaction history query api. This is the first optimization done here.
-- 2. Note that I am constructing a map of address -> txns/staking txns "records" first per block and **for each address key, perform RLP decoding/encoding just once** instead of encode/decode 1000s times in order to add a single new txn. This is the second optimization done here.

>
> The best solution is definitely a SQL-like query approach where the txns are stored and indexed by address. You can do some search and see whether leveldb support it (I doubt it since it's no-sql db).
>

goleveldb does not natively support SQL like queries.
In fact, TxRecord introduced here is acting like such query index for blockchain db. It provides tx hashes list quickly for transaction history queries, and then for each of those hashes, each tx is looked up from the rawdb.

> 
> The other clear optimization is to only store list of txn hashes for the address. And store the txn separately. This way instead of storing 1000s of raw txns together as one record, it's 1000s of hashes, which will cost much less memory when loaded to memory.
>

Exactly. That is exactly what this PR is trying to bring.

api/service/explorer/storage.go
```diff
@@ -20,9 +19,12 @@ import (
 
 // Constants for storage.
 const (
-	AddressPrefix    = "ad"
-	CheckpointPrefix = "dc"
-	PrefixLen        = 3
+	AddressPrefix            = "ad"
+	CheckpointPrefix         = "dc"
+	TransactionPrefix        = "tx"
```
rlan35, line 24: What are StakingTransactionPrefix and TransactionPrefix for? Seems they are not used.

rlan35:
I see. Sorry I thought it will require a bigger change involving GetTransactionsHistory. But it turns out it's not necessary. Then it was a clear bad design before storing lots of useless data. This is the right solution..

api/service/explorer/storage.go
```diff
@@ -202,3 +159,84 @@ func (storage *Storage) GetAddresses(size int, prefix string) ([]string, error)
...
+
+func computeAccountsTransactionsMapForBlock(
+	block *types.Block,
+) (map[string][]*TxRecord, map[string][]*TxRecord) {
+	// mapping from account address to TxRecords for txns in the block
+	var acntsTxns map[string][]*TxRecord
+	// mapping from account address to TxRecords for staking txns in the block
+	var acntsStakingTxns map[string][]*TxRecord
+
+	blockTime := int(block.Time().Int64() * 1000)
```
rlan35, line 171: let's just use the raw timestamp then? Better not change the scale.

denniswon:
@rlan35 updated per review.

PR #2863, opened 2020-04-21
[p2p] Add libp2p_metrics into host and use bwCounter to record in/out…

jz0000:
… networking traffic in a node

## Issue

So far we don't have good visibility into what's going on in the p2p layer, to improve the performance of p2p layer, we need to have some metrics to be collected.  This is the first step, I will gradually add more metrics in more places as needed. 

The changes in the go.mod is automatically generated when I ran "make" in my workspace, I have retried a few times and it seems not harmful.  Please let me know if otherwise. Thanks.

### Unit Test Coverage

I have run go test -cover inside node directory, let me know if I need to run some other unit testing. 

Jians-MacBook-Pro:node jz$ go test -cover
PASS
coverage: 12.4% of statements
ok  	github.com/harmony-one/harmony/node	1.195s

### Test/Run Logs

Have run ./test/debug.sh and generated logs  like 

tmp_log/log-20200419-232519/zerolog-validator-127.0.0.1-9000.log, on my machine. Here I share a few examples of the generated messages in the log:

https://gist.github.com/jz0000/807652e4b0f05b91eae3319246652f6e

## Operational Checklist

NO

NO

There could be some minor increase in the log it generated, thus disk usage might increase a bit, but don't see this an issue.

### DIFF
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -59,6 +59,13 @@ func (node *Node) processSkippedMsgTypeByteValue(
 
 // HandleMessage parses the message and dispatch the actions.
 func (node *Node) HandleMessage(content []byte, sender libp2p_peer.ID) {
+	// log in-coming metrics
+	node.host.LogRecvMessage(content)
+	utils.Logger().Info().
+		Int64("TotalIn", node.host.GetBandwidthTotals().TotalIn).
+		Float64("RateIn", node.host.GetBandwidthTotals().RateIn).
+		Msg("Record Receiving Metrics!")
+
 	msgCategory, err := proto.GetMessageCategory(content)
 	if err != nil {
 		utils.Logger().Error().
@@ -447,6 +454,10 @@ func (node *Node) PostConsensusProcessing(
 
 	// Broadcast client requested missing cross shard receipts if there is any
 	node.BroadcastMissingCXReceipts()
+
+	// Clear metrics after one consensus cycle
+	node.host.ResetMetrics()
+	utils.Logger().Info().Msg("Metrics cleared after one consensus cycle")	
 
 	// Update consensus keys at last so the change of leader status doesn't mess up normal flow
 	if len(newBlock.Header().ShardState()) > 0 {
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -457,7 +457,7 @@ func (node *Node) PostConsensusProcessing(
 
 	// Clear metrics after one consensus cycle
 	node.host.ResetMetrics()
-	utils.Logger().Info().Msg("Metrics cleared after one consensus cycle")	
+	utils.Logger().Info().Msg("Metrics cleared after one consensus cycle")
 
 	// Update consensus keys at last so the change of leader status doesn't mess up normal flow
 	if len(newBlock.Header().ShardState()) > 0 {
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -64,7 +64,7 @@ func (node *Node) HandleMessage(content []byte, sender libp2p_peer.ID) {
 	utils.Logger().Info().
 		Int64("TotalIn", node.host.GetBandwidthTotals().TotalIn).
 		Float64("RateIn", node.host.GetBandwidthTotals().RateIn).
-		Msg("Record Receiving Metrics!")
+		Msg("[metrics][p2p] traffic in in bytes")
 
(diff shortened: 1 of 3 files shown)


p2p/host.go
```diff
@@ -88,14 +94,18 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
 
 	self.PeerID = p2pHost.ID()
 	subLogger := utils.Logger().With().Str("hostID", p2pHost.ID().Pretty()).Logger()
+
+	newMetrics := libp2p_metrics.NewBandwidthCounter()
```
denniswon, line 98: make this as an option param to node.sh? Might want to avoid having "all nodes" logging this metric.

fxfactorial:
Please let’s hold off on this until after my bigger PR finished

LeoHChen:
Jenkins, test this please.

p2p/host.go
LeoHChen: this is local log, should be fine.

p2p/host.go
```diff
@@ -161,7 +174,14 @@ func (host *HostV2) SendMessageToGroups(groups []nodeconfig.GroupID, msg []byte)
 			err = e
 			continue
 		}
+		// log out-going metrics
+		host.metrics.LogSentMessage(int64(len(msg)))
 	}
+	host.logger.Info().
+		Int64("TotalOut", host.GetBandwidthTotals().TotalOut).
+		Float64("RateOut", host.GetBandwidthTotals().RateOut).
+		Msg("Record Sending Metrics!")
```
LeoHChen, line 183: use the same prefix for metrics

p2p/host.go
jz0000: ok.

node/node_handler.go
```diff
@@ -448,6 +455,10 @@ func (node *Node) PostConsensusProcessing(
 	// Broadcast client requested missing cross shard receipts if there is any
 	node.BroadcastMissingCXReceipts()
 
+	// Clear metrics after one consensus cycle
+	node.host.ResetMetrics()
+	utils.Logger().Info().Msg("Metrics cleared after one consensus cycle")
```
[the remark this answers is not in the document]
jz0000: Unfortunately this is needed. Our data scientist will need to know when this counter resets, to be able to compute the sum correctly, if needed. 

Thought about removing the reset all together, but that might let counters wrap around after sometime, once the network becomes bigger.

node/node_handler.go
```diff
@@ -59,6 +59,13 @@ func (node *Node) processSkippedMsgTypeByteValue(
 
 // HandleMessage parses the message and dispatch the actions.
 func (node *Node) HandleMessage(content []byte, sender libp2p_peer.ID) {
+	// log in-coming metrics
+	node.host.LogRecvMessage(content)
+	utils.Logger().Info().
+		Int64("TotalIn", node.host.GetBandwidthTotals().TotalIn).
+		Float64("RateIn", node.host.GetBandwidthTotals().RateIn).
+		Msg("Record Receiving Metrics!")
```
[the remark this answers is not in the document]
jz0000: Sounds good.

p2p/host.go
jz0000: ok.

fxfactorial:
@jz0000 as I already wrote : let’s hold off on this PR because it will cause huge and needless merge conflicts when al this code is changing anyway under my p2p revamp

jz0000:
> @jz0000 as I already wrote : let’s hold off on this PR because it will cause huge and needless merge conflicts when al this code is changing anyway under my p2p revamp

Leo let me to merge to s3 branch, you are going to master correct?

fxfactorial:
> > @jz0000 as I already wrote : let’s hold off on this PR because it will cause huge and needless merge conflicts when al this code is changing anyway under my p2p revamp
> 
> Leo let me to merge to s3 branch, you are going to master correct?

ah, my bad - yes, I'm targetting master

jz0000:
> > > @jz0000 as I already wrote : let’s hold off on this PR because it will cause huge and needless merge conflicts when al this code is changing anyway under my p2p revamp
> > 
> > 
> > Leo let me to merge to s3 branch, you are going to master correct?
> 
> ah, my bad - yes, I'm targetting master

PR #2869, opened 2020-04-23
[core][node/worker] processing staking and normal transactions together

gupadhyaya:
* For block proposal, process pending PoolTransactions in the TransactionsByPriceAndNonce order (regardless of normal or staking)
* For block verification, pull and put normal and staking transactions into the same PoolTransactions list and apply and process in TransactionsByPriceAndNonce order

### DIFF
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -126,7 +126,6 @@ func (p *StateProcessor) Process(
 			err       error
 			receipt   *types.Receipt
 			cxReceipt *types.CXReceipt
-			receipts  types.Receipts
 		)
 
 		if plainTx, ok := tx.(*types.Transaction); ok {
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -91,67 +91,86 @@ func (p *StateProcessor) Process(
 
 	poolTransactions := map[common.Address]types.PoolTransactions{}
 
-	// Iterate over and add normal transactions to poolTransactions
-	for _, tx := range block.Transactions() {
-		from, err := types.Sender(signer, tx)
-		if err != nil {
-			return nil, nil, nil, 0, nil, err
-		}
-		poolTransactions[from] = append(poolTransactions[from], tx)
-	}
-
-	// Iterate over and add staking transactions to poolTransactions
-	for _, tx := range block.StakingTransactions() {
-		from, err := tx.SenderAddress()
-		if err != nil {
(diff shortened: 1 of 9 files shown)


node/worker/worker.go
```diff
@@ -86,15 +83,36 @@ func (w *Worker) CommitTransactions(
 			continue
 		}
 		// Start executing the transaction
-		w.current.state.Prepare(tx.Hash(), common.Hash{}, len(w.current.txs))
+		w.current.state.Prepare(tx.Hash(), common.Hash{}, dbIndex)
+		dbIndex = dbIndex + 1
```
rlan35, line 87: Here the index counter should only increment when the txn succeeded. Failed txns won't be included in the new block.

core/types/tx_pool.go
```diff
@@ -102,3 +103,97 @@ type PoolTxByNonce PoolTransactions
...
+	heads  PoolTxByPrice                       // Next transaction for each unique account (price heap)
+	signer Signer                              // Signer for the set of transactions
+}
+
+// NewPoolTransactionsByPriceAndNonce creates a transaction set that can retrieve
+// price sorted transactions in a nonce-honouring way.
+//
+// Note, the input map is reowned so the caller should not interact any more with
+// if after providing it to the constructor.
+func NewPoolTransactionsByPriceAndNonce(signer Signer, txs map[common.Address]PoolTransactions) *PoolTransactionsByPriceAndNonce {
```
rlan35, line 143: no functional changes on these code, right?

core/state_processor.go
```diff
@@ -79,41 +79,79 @@ func (p *StateProcessor) Process(
...
 	)
-	beneficiary, err := p.bc.GetECDSAFromCoinbase(header)
 
+	beneficiary, err := p.bc.GetECDSAFromCoinbase(header)
 	if err != nil {
 		return nil, nil, nil, 0, nil, err
 	}
 
-	// Iterate over and process the individual transactions
-	for i, tx := range block.Transactions() {
```
rlan35, line 89: need a forking condition on pre-staking epoch, have to use the old logic before that.

gupadhyaya: staking transactions won't be included in the block due to prestaking check in the `CommitTransactions` right?

core/types/tx_pool.go
gupadhyaya: yes...

PR #2870, opened 2020-04-23
Jz.libp2p metrics.0

jz0000:
## Issue

[This PR has been reviewed in its previous version for master branch https://github.com/harmony-one/harmony/pull/2863]

So far we don't have good visibility into what's going on in the p2p layer, to improve the performance of p2p layer, we need to have some metrics to be collected. This is the first step, I will gradually add more metrics in more places as needed.

The changes in the go.mod is automatically generated when I ran "make" in my workspace, I have retried a few times and it seems not harmful. Please let me know if otherwise. Thanks.
### Unit Test Coverage

Jians-MacBook-Pro:p2p jz$ go test
PASS
ok  	github.com/harmony-one/harmony/p2p	0.014s

Jians-MacBook-Pro:node jz$ go test
PASS
ok  	github.com/harmony-one/harmony/node	1.658s

### Test/Run Logs

Have run ./test/debug.sh and generated logs like

tmp_log/log-20200422-222845/zerolog-validator-127.0.0.1-9001.log

on my machine. Here I share a few examples of the generated messages in the log:
https://gist.github.com/jz0000/f81e3dcc3efb74ac4d2e5a9d0aa0c834

## Operational Checklist

NO

NO

There could be some minor increase in the log it generated, thus disk usage might increase a bit, but don't see this an issue.


LeoHChen:
Jenkins, test this please.

PR #2871, opened 2020-04-23
Add raw stake and median to shard state and roster

gupadhyaya:
reuse all the code from `GetMedianRawStakeSnapshot`, but only work with validator snapshot to recompute the results of previous election. the format of the rpc result is same as `GetMedianRawStakeSnapshot`.

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -587,6 +587,18 @@ func (b *APIBackend) GetSuperCommittees() (*quorum.Transition, error) {
 		quorum.NewRegistry(stakedSlotsThen),
 		quorum.NewRegistry(stakedSlotsNow)
 
+	then.MedianStake, then.SlotPurchases =
+		committee.ComputeMedianRawStakeForShardCommittee(
+			b.hmy.BlockChain(),
+			prevCommittee.StakedValidators().Addrs,
+		)
+
+	now.MedianStake, now.SlotPurchases =
+		committee.ComputeMedianRawStakeForShardCommittee(
+			b.hmy.BlockChain(),
+			nowCommittee.StakedValidators().Addrs,
+		)
+
 	for _, comm := range prevCommittee.Shards {
 		decider := quorum.NewDecider(quorum.SuperMajorityStake, comm.ShardID)
 		if _, err := decider.SetVoters(&comm, prevCommittee.Epoch); err != nil {
--- a/shard/committee/assignment.go
+++ b/shard/committee/assignment.go
@@ -84,13 +84,14 @@ func ComputeMedianRawStakeForShardCommittee(
 	if err != nil {
 		return numeric.NewDec(0), nil
 	}
+
 	maxExternalSlots := shard.ExternalSlotsAvailableForEpoch(
 		stakedReader.CurrentBlock().Epoch(),
 	)
-	median, winners := effective.Apply(
+
+	return effective.Apply(
 		candidateSlotOrders, maxExternalSlots,
 	)
-	return median, winners
 }
 
 // NewEPoSRound runs a fresh computation of EPoS using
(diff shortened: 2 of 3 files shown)


shard/committee/assignment.go
```diff
@@ -123,6 +142,7 @@ func NewEPoSRound(stakedReader StakingCandidatesReader) (
 
 func prepareOrders(
 	stakedReader StakingCandidatesReader,
+	snapshotOnly bool,
```
fxfactorial, line 145: no bool parameters please. Just make two functions for the branch. prepareOrdersSnapshot, prepareOrders

rlan35:
This is not 100% correct. Because when you recompute the election, you are doing it on the latest list of validators (even though you use the snapshot data).

The correct and simpler way is just in the GetSuperCommitte. Add a field of raw stake and compute the raw stake based on snapshot data. So you don't need to modify EPoS compute functions. Just add some data in RPC results.

shard/shard_state.go
```diff
@@ -33,8 +33,9 @@ const (
 
 // State is the collection of all committees
 type State struct {
-	Epoch  *big.Int    `json:"epoch"`
-	Shards []Committee `json:"shards"`
+	Epoch       *big.Int    `json:"epoch"`
+	Shards      []Committee `json:"shards"`
+	MedianStake numeric.Dec `json:"epos-median-stake" rlp:"nil"`
```
fxfactorial, line 38: we should not change this field for this

shard/committee/assignment.go
```diff
@@ -183,6 +202,61 @@ func prepareOrders(
...
+func prepareOrdersSnapshot(
+	stakedReader StakingCandidatesReader,
+	candidates []common.Address,
+) (map[common.Address]*effective.SlotOrder, error) {
+	blsKeys := map[shard.BLSPublicKey]struct{}{}
+	essentials := map[common.Address]*effective.SlotOrder{}
+	totalStaked, tempZero := big.NewInt(0), numeric.ZeroDec()
+
+	for i := range candidates {
+		snapshot, err := stakedReader.ReadValidatorSnapshot(
```
rlan35, line 214: Should use the right snapshot. Use ReadValidatorSnapshotAtEpoch(...

hmy/api_backend.go
```diff
@@ -587,6 +587,18 @@ func (b *APIBackend) GetSuperCommittees() (*quorum.Transition, error) {
...
 		quorum.NewRegistry(stakedSlotsNow)
 
+	then.MedianStake, then.SlotPurchases =
+		committee.ComputeMedianRawStakeForShardCommittee(
+			b.hmy.BlockChain(),
+			prevCommittee.StakedValidators().Addrs,
+		)
+
+	now.MedianStake, now.SlotPurchases =
+		committee.ComputeMedianRawStakeForShardCommittee(
```
rlan35, line 597: Actually no need to be this complicated. Just add a field in the final result:

 {
                            "is-harmony-slot": false,
                            "earning-account": "one1x8kgyyqgslzrm7jrlj4m7rpg285etdywjk7gez",
                            "bls-public-key": "5e8635085788492a87d3eab69732ae72a55087e95c210ad20a726dd169a69b5792cb94616f6eccc4ea0a54a4a77ca88c",
                            "voting-power-unnormalized": "0.026785021562679984",
                            "voting-power-%": "0.008571206900057595",
                            "effective-stake": "5920773850000000000000000.000000000000000000"
                            "raw-stake": NEW_FIELD,
                        },

The raw-stake is just validatorSnap.TotalStake / size(validatorSnap.BLSKeys). 

No need to add new func like: ComputeMedianRawStakeForShardCommittee. Just loop through all the validators in the decider result and for each validator address, fetch the snapshot and compute the raw-stake and put it in the result.

shard/committee/assignment.go
rlan35: And no need to add new function just for the single RPC.

gupadhyaya:
updated https://github.com/harmony-one/harmony/pull/2880

PR #2874, opened 2020-04-23
[audit] Fix and make exact 7 epoch for undelegation

rlan35:
Should be locking undelegation for only 7 epoch time. Previously to the user, it's actually locked for 8 epoch time.

### DIFF
--- a/staking/slash/double-sign.go
+++ b/staking/slash/double-sign.go
@@ -352,7 +352,8 @@ func delegatorSlashApply(
 				}
 
 				// NOTE Assume did as much as could above, now check the undelegations
-				for _, undelegate := range delegationNow.Undelegations {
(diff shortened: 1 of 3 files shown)


staking/slash/double-sign.go
```diff
@@ -352,7 +352,8 @@ func delegatorSlashApply(
 				}
 
 				// NOTE Assume did as much as could above, now check the undelegations
-				for _, undelegate := range delegationNow.Undelegations {
+				for i := range delegationNow.Undelegations {
+					undelegate := delegationNow.Undelegations[i]
```
rlan35, line 356: this is no-op, since undelegation.Amount is a pointer. But better be explicit on reference.

PR #2875, opened 2020-04-23
Make commit sig committed along with the block

rlan35:
Last commit wasn't consistently written to db as blocks. This change fixes it.


rlan35:
> so with this change, the last commit sig is only available after the block is added to chain?

yea, it should.

PR #2877, opened 2020-04-24
Add viewID to block signed content

### LINKED ISSUE #2630: Change block signing logic to include view ID in the signed content
Previously, the consensus signatures are signed on blockHeight+blockHash.

We need to include also viewID in the signed data, which is required by the new slashing logic.

Search for "signature on hash+blockNum+viewID (add a hard fork)" in the code base and find all the places where viewID needs to be added into the signature.

Need to add the new logic as a fork at staking epoch.

Daniel-VDM:
## Issue

This closes #2630 

### Unit Test Coverage

Before:

```
PASS
coverage: 7.4% of statements
ok  	github.com/harmony-one/harmony/consensus	1.380s
```
```
PASS
coverage: 21.5% of statements
ok  	github.com/harmony-one/harmony/core	6.803s
```
```
PASS
coverage: 47.3% of statements
ok  	github.com/harmony-one/harmony/staking/slash	0.243s
```

After:

```
PASS
coverage: 7.5% of statements
ok      github.com/harmony-one/harmony/consensus        1.265s
```
```
PASS
coverage: 21.7% of statements
ok      github.com/harmony-one/harmony/core     6.865s
```
```
PASS
coverage: 47.2% of statements
ok      github.com/harmony-one/harmony/staking/slash    0.304s
```

## Operational Checklist

    **YES**

   At the staking epoch fork, signing and verification will be done with the addition of the viewID in which the block is going to be/was committed.

   Tested on localnet and it works, got well past staking epoch. Tested on STN: got past staking epoch with external validator signing and earning. 

   Nothing new other than what is needed for staking launch.

    **NO**

    **NO**

### DIFF
--- a/consensus/signature/signature.go
+++ b/consensus/signature/signature.go
@@ -8,20 +8,21 @@ import (
 	"github.com/harmony-one/harmony/internal/params"
 )
 
-type configChainReader interface {
+type signatureChainReader interface {
 	Config() *params.ChainConfig
 }
 
 // ConstructCommitPayload returns the commit payload for consensus signatures.
 func ConstructCommitPayload(
-	chain configChainReader, blockHash common.Hash, epoch, blockNum, viewID uint64,
+	chain signatureChainReader, epoch *big.Int, blockHash common.Hash, blockNum, viewID uint64,
 ) []byte {
 	blockNumBytes := make([]byte, 8)
-	viewIDBytes := make([]byte, 8)
 	binary.LittleEndian.PutUint64(blockNumBytes, blockNum)
-	binary.LittleEndian.PutUint64(viewIDBytes, viewID)
-	if chain.Config().IsStaking(new(big.Int).SetUint64(epoch)) {
-		return append(blockNumBytes, append(viewIDBytes, blockHash.Bytes()...)...)
+	commitPayload := append(blockNumBytes, blockHash.Bytes()...)
+	if !chain.Config().IsStaking(epoch) {
+		return commitPayload
 	}
-	return append(blockNumBytes, blockHash.Bytes()...)
+	viewIDBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(viewIDBytes, viewID)
+	return append(commitPayload, viewIDBytes...)
 }
--- a/consensus/signature/signature.go
+++ b/consensus/signature/signature.go
@@ -1,0 +1,27 @@
+package signature
+
+import (
+	"encoding/binary"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/harmony-one/harmony/internal/params"
+)
+
+type configChainReader interface {
+	Config() *params.ChainConfig
+}
+
+// ConstructCommitPayload returns the commit payload for consensus signatures.
+func ConstructCommitPayload(
+	chain configChainReader, blockHash common.Hash, epoch, blockNum, viewID uint64,
+) []byte {
+	blockNumBytes := make([]byte, 8)
+	viewIDBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(blockNumBytes, blockNum)
+	binary.LittleEndian.PutUint64(viewIDBytes, viewID)
+	if chain.Config().IsStaking(new(big.Int).SetUint64(epoch)) {
+		return append(blockNumBytes, append(viewIDBytes, blockHash.Bytes()...)...)
+	}
+	return append(blockNumBytes, blockHash.Bytes()...)
+}
--- a/consensus/threshold.go
+++ b/consensus/threshold.go
@@ -1,11 +1,13 @@
 package consensus
 
 import (
-	"encoding/binary"
+	"encoding/hex"
+	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
 	"github.com/harmony-one/harmony/consensus/quorum"
+	"github.com/harmony-one/harmony/consensus/signature"
 	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/p2p"
@@ -37,10 +39,14 @@ func (consensus *Consensus) didReachPrepareQuorum() error {
 	consensus.aggregatedPrepareSig = aggSig
 	consensus.FBFTLog.AddMessage(FBFTMsg)
 	// Leader add commit phase signature
-	// TODO(audit): sign signature on hash+blockNum+viewID (add a hard fork)
-	blockNumHash := [8]byte{}
-	binary.LittleEndian.PutUint64(blockNumHash[:], consensus.blockNum)
-	commitPayload := append(blockNumHash[:], consensus.blockHash[:]...)
+	commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
+		new(big.Int).SetUint64(consensus.epoch), consensus.blockHash, consensus.blockNum, consensus.viewID)
+	// TODO: remove debug msg after STN testing
+	utils.Logger().Debug().
+		Uint64("epoch", consensus.epoch).
+		Uint64("block-number", consensus.blockNum).
+		Uint64("view-id", consensus.viewID).
+		Msgf("[COMMIT-PAYLOAD] didReachPrepareQuorum %v", hex.EncodeToString(commitPayload))
 
 	// so by this point, everyone has committed to the blockhash of this block
 	// in prepare and so this is the actual block.
--- a/consensus/threshold.go
+++ b/consensus/threshold.go
@@ -1,7 +1,6 @@
 package consensus
 
 import (
-	"encoding/hex"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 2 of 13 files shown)


api/proto/message/message.proto
```diff
@@ -82,11 +82,12 @@ message StakingRequest {
 message ConsensusRequest {
   uint64 view_id = 1;
   uint64 block_num = 2;
-  uint32 shard_id = 3;
-  bytes block_hash = 4;
-  bytes block = 5;
-  bytes sender_pubkey = 6;
-  bytes payload = 7;
+  uint64 epoch = 3;
```
fxfactorial, line 85: cannot change protobuf def -

Daniel-VDM: hmm, why not?

Daniel-VDM: actually, guess it would make rolling upgrades not possible.

rlan35: yup, can only add field new field number, can't change existing numbers.

api/proto/message/message.proto
```diff
@@ -113,4 +114,4 @@ message ViewChangeRequest {
   bytes m3_aggsigs = 11; // m3: |viewID|
   bytes m3_bitmap= 12;
 
-}
+}
```
rlan35, line 117: nit.

api/proto/message/message.proto
Daniel-VDM: cool latest is fine then?

api/proto/message/message.proto
Daniel-VDM: changed

consensus/signature/signature.go
```diff
@@ -0,0 +1,27 @@
...
+// ConstructCommitPayload returns the commit payload for consensus signatures.
+func ConstructCommitPayload(
+	chain configChainReader, blockHash common.Hash, epoch, blockNum, viewID uint64,
+) []byte {
+	blockNumBytes := make([]byte, 8)
+	viewIDBytes := make([]byte, 8)
+	binary.LittleEndian.PutUint64(blockNumBytes, blockNum)
+	binary.LittleEndian.PutUint64(viewIDBytes, viewID)
+	if chain.Config().IsStaking(new(big.Int).SetUint64(epoch)) {
+		return append(blockNumBytes, append(viewIDBytes, blockHash.Bytes()...)...)
```
rlan35, line 24: Let's put viewID after blockNum and blockHash to be consistent with before.

api/proto/message/message.proto
```diff
@@ -87,6 +87,7 @@ message ConsensusRequest {
   bytes block = 5;
   bytes sender_pubkey = 6;
   bytes payload = 7;
+  uint64 epoch = 8;
```
rlan35, line 90: why do you need the epoch here? Is there no other way to get epoch? We want to make consensus msg as light weight as possible because it's the majority of the traffic.

internal/chain/engine.go
```diff
@@ -227,10 +228,13 @@ func (e *engineImpl) VerifySeal(chain engine.ChainReader, header *block.Header)
 		}
 	}
 
-	// TODO(audit): verify signature on hash+blockNum+viewID (add a hard fork)
-	blockNumHash := make([]byte, 8)
-	binary.LittleEndian.PutUint64(blockNumHash, header.Number().Uint64()-1)
-	lastCommitPayload := append(blockNumHash, parentHash[:]...)
+	lastCommitPayload := signature.ConstructCommitPayload(e.Beaconchain(),
```
rlan35, line 231: use e.Blockchain(). Beaconchain may not exist for all nodes.

internal/chain/engine.go
```diff
@@ -543,10 +546,13 @@ func (e *engineImpl) VerifyHeaderWithSignature(chain engine.ChainReader, header
 			)
 		}
 	}
-	// TODO(audit): verify signature on hash+blockNum+viewID (add a hard fork)
-	blockNumHash := make([]byte, 8)
-	binary.LittleEndian.PutUint64(blockNumHash, header.Number().Uint64())
-	commitPayload := append(blockNumHash, hash[:]...)
+	commitPayload := signature.ConstructCommitPayload(e.Beaconchain(),
```
rlan35, line 549: use e.Blockchain(). Beaconchain may not exist for all nodes.

core/types/crosslink.go
```diff
@@ -46,6 +57,11 @@ func (cl CrossLink) Number() *big.Int {
 	return cl.BlockNumberF
 }
 
+// ID returns viewID with big.Int format
+func (cl CrossLink) ID() *big.Int {
```
rlan35, line 61: bad naming. Just use ViewID()

consensus/view_change.go
```diff
@@ -533,14 +539,19 @@ func (consensus *Consensus) onNewView(msg *msg_pb.Message) {
 	// TODO: check magic number 32
 	if len(recvMsg.Payload) > 32 {
 		// Construct and send the commit message
-		blockNumHash := make([]byte, 8)
-		binary.LittleEndian.PutUint64(blockNumHash, consensus.blockNum)
+		commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
+			consensus.blockHash, consensus.epoch, consensus.blockNum, consensus.viewID)
```
rlan35, line 543: @gupadhyaya please double check the view change logic.

rlan35: I think we should have use the recvMsg's data for ViewID

consensus/view_change.go
```diff
@@ -361,17 +363,21 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 			consensus.aggregatedPrepareSig = aggSig
 			consensus.prepareBitmap = mask
 			// Leader sign and add commit message
-			// TODO(audit): verify signature on hash+blockNum+viewID (add a hard fork)
-			blockNumBytes := [8]byte{}
-			binary.LittleEndian.PutUint64(blockNumBytes[:], consensus.blockNum)
-			commitPayload := append(blockNumBytes[:], consensus.blockHash[:]...)
+			commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
+				consensus.blockHash, consensus.epoch, consensus.blockNum, consensus.viewID)
```
rlan35, line 367: same here about the ViewID

api/proto/message/message.proto
Daniel-VDM: Yes, I considered other ways. But unsure about how to best handle `ExplorerMessageHandler`.
 
Could add commit sig check after explorer can get a block from FBFTLog?

consensus/validator.go
```diff
@@ -199,16 +200,21 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
 	if bytes.Equal(consensus.blockHash[:], emptyHash[:]) {
 		copy(consensus.blockHash[:], blockHash[:])
 	}
-	blockNumBytes := make([]byte, 8)
-	binary.LittleEndian.PutUint64(blockNumBytes, consensus.blockNum)
+	commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
+		consensus.blockHash, consensus.epoch, consensus.blockNum, consensus.viewID)
```
rlan35, line 204: use the viewID from the recvMsg. Local viewID may be not consistent with others.

consensus/threshold.go
```diff
@@ -37,10 +38,13 @@ func (consensus *Consensus) didReachPrepareQuorum() error {
 	consensus.aggregatedPrepareSig = aggSig
 	consensus.FBFTLog.AddMessage(FBFTMsg)
 	// Leader add commit phase signature
-	// TODO(audit): sign signature on hash+blockNum+viewID (add a hard fork)
-	blockNumHash := [8]byte{}
-	binary.LittleEndian.PutUint64(blockNumHash[:], consensus.blockNum)
-	commitPayload := append(blockNumHash[:], consensus.blockHash[:]...)
+	commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
+		consensus.blockHash, consensus.epoch, consensus.blockNum, consensus.viewID)
```
rlan35, line 42: For leader, it's ok to use consensus.viewID because leader is the master.

api/proto/message/message.proto
rlan35: Actually the signed content should always be consistent with those in the block. So you can always get the viewID from the block. Same for ExplorerMessageHandler

consensus/view_change.go
gupadhyaya: using `consensus.viewID` is ok here, as line 518 sets it `consensus.viewID = recvMsg.ViewID`

consensus/signature/signature.go
Daniel-VDM: sure, fixed in latest

consensus/validator.go
Daniel-VDM: changed in latest to use received viewID

core/types/crosslink.go
```diff
@@ -13,27 +13,38 @@ import (
...
 	SignatureF   [96]byte //aggregated signature
 	BitmapF      []byte   //corresponding bitmap mask for agg signature
 	ShardIDF     uint32   //will be verified with signature on |blockNumber|blockHash| is correct
 	EpochF       *big.Int
 }
 
 // NewCrossLink returns a new cross link object
-// epoch is the parentHeader's epoch
-func NewCrossLink(header *block.Header, epoch *big.Int) CrossLink {
+func NewCrossLink(header *block.Header, parentHeader *block.Header) CrossLink {
```
fxfactorial, line 30: New implies returns pointer , make it a pointer and all these crosslink methods should have pointer receiver

Daniel-VDM: Sure thing, changed in latest

PR #2878, opened 2020-04-24
[p2p][node] Use IPFS as P2P layer, fix > 7 data race conditions 

fxfactorial:
This PR is a subset of a larger body of work I have at the moment. 

1. Remove bootnodes, now use IPFS machines as default bootnode machines. Need to expose seednode addr for node.sh (maybe in this PR, maybe next)
2. Use berty project's service code
3. remove network info service
4. race condition on starting consensus
5. race condition on node.State
6. race condition on commit finish, grace period
7. Remove SendNewBlockToUnsync (pushing blocks)
8. race condition in grpc resp in syncing
9. remove unused field in consensus
10. remove discovery service
11. Remove mutex usage in favor of atomics in consensus current state
12. remove race condition logger 
13. race condition on IsLeader
14. remove internal ad-hoc timer impl, had race conditions
15. remove slow and orthogonal to this project unit tests (leftover old ones) 
16. go.mod network libraries bumped up 
17. remove leftover discovery data structures
18. remove dead fields in consensus
19. remove package level host handle
20. remove unused direct dependencies 

With no bootnodes, we create initial network connectivity via 4 multiaddrs functioning as public IPFS seed nodes. Of course we can still add our own seed nodes instead/in addition to using these four pubic IPFS addrs, will get to that in a later PR.


cmd/harmony/main.go
```diff
@@ -371,6 +365,36 @@ func setupConsensusKey(nodeConfig *nodeconfig.ConfigType) multibls.PublicKey {
 	return *consensusMultiPubKey
 }
 
+func setupHost(nodeConfig *nodeconfig.ConfigType) (p2p.Host, error) {
+	baseDS := datastore.NewMapDatastore()
+	const DevRendezVousPoint = "/ip4/167.99.223.55/tcp/4040/p2p/QmTo3RS6Uc8aCS5Cxx8EBHkNCe4C7vKRanbMEboxkA92Cn"
```
rlan35, line 370: Dev? Is this  the bootnode? Do we need to change bootnode setup for the network?

consensus/checks.go
```diff
@@ -12,36 +13,36 @@ const MaxBlockNumDiff = 100
 
 func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
 	if msg.GetConsensus() == nil {
-		consensus.getLogger().Warn().Msg("[validatorSanityChecks] malformed message")
+		utils.Logger().Warn().Msg("[validatorSanityChecks] malformed message")
```
rlan35, line 16: let's not change these loggers for now. Consensus logger logs the internal state of consensus, it's useful to troubleshoot hard bugs.

cmd/harmony/main.go
fxfactorial: no - there is no more bootnode as separate project binary. we link up using IPFS public nodes + a seed node of ours, this is a leftover seednode I was using - will update

consensus/checks.go
fxfactorial: no, this logger was causing data race conditions, not adding value, caller should add what they need to plain logger

consensus/consensus_service.go
```diff
@@ -308,7 +251,7 @@ func (consensus *Consensus) checkViewID(msg *FBFTMessage) error {
 		consensus.current.SetViewID(msg.ViewID)
 		consensus.LeaderPubKey = msg.SenderPubkey
```
rlan35, line 252: Better add a func for updating the leaderPubKey with the logic to update IsLeader. There are many places LeaderPubKey is updated without updated the IsLeader.

consensus/checks.go
rlan35: So it crashed on the consensus logger? Ok, never seen it before though. How did you reveal so many race conditions now?

fxfactorial: by using our amazing golang tools haha

consensus/validator.go
```diff
@@ -276,14 +277,13 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
...
 	if recvMsg.BlockNum-consensus.blockNum > consensusBlockNumBuffer {
-		consensus.getLogger().Debug().Uint64("MsgBlockNum", recvMsg.BlockNum).Msg("[OnCommitted] OUT OF SYNC")
+		utils.Logger().Debug().
+			Uint64("MsgBlockNum", recvMsg.BlockNum).
+			Msg("[OnCommitted] OUT OF SYNC")
 		go func() {
 			select {
 			case consensus.BlockNumLowChan <- struct{}{}:
 				consensus.current.SetMode(Syncing)
-				for _, v := range consensus.consensusTimeout {
```
rlan35, line 284: is it safe to remove this part?

consensus/consensus_service.go
fxfactorial: yes, I had that on bigger PR but then there are issues about Getconsensus private key and whatnot, will address in later PR

consensus/validator.go
fxfactorial: I think so - its based on a faulty assumption - I can put it back in, but I don't think its a make or break thing - in any case, I need a deeper redo of consensus interacting with state sync, which will get to in next PR  (part of my work in other branch from which this work taken from)

node/node.go
```diff
@@ -640,15 +605,15 @@ func (node *Node) InitConsensusWithValidators() (err error) {
...
-	for _, p := range peers {
-		key := fmt.Sprintf("%s:%s:%s", p.IP, p.Port, p.PeerID)
-		_, ok := node.Neighbors.LoadOrStore(key, *p)
-		if !ok {
-			// !ok means new peer is stored
-			node.host.AddPeer(p)
-			continue
-		}
-	}
+	// for _, p := range peers {
```
rlan35, line 608: dev code? Just delete the func?

node/service_setup.go
```diff
@@ -101,10 +75,17 @@ func (node *Node) StopServices() {
...
-	)
+// ForceJoiningTopics ..
+func (node *Node) ForceJoiningTopics() error {
+	groups := []nodeconfig.GroupID{
+		node.NodeConfig.GetShardGroupID(),
+		nodeconfig.NewClientGroupIDByShardID(shard.BeaconChainShardID),
+		node.NodeConfig.GetClientGroupID(),
+	}
+
+	// force the side effect of topic join
```
rlan35, line 86: elaborate a bit more? What side effect?

p2p/host.go
```diff
@@ -4,250 +4,228 @@ import (
...
-func (host *HostV2) GetP2PHost() libp2p_host.Host {
-	return host.h
-}
+// NewHost ..
+func NewHost(opts *Opts, own *Peer) (Host, error) {
+	var swarmAddresses []string
+
+	if opts.Port != 0 {
+		swarmAddresses = []string{
+			fmt.Sprintf("/ip4/127.0.0.1/tcp/%d", opts.Port),
```
rlan35, line 151: these hard coded value needs updates?

p2p/host.go
```diff
@@ -4,250 +4,228 @@ import (
...
+	api, node, err := ipfsutil.NewConfigurableCoreAPI(
+		ctx,
+		cfg,
+		ipfsutil.OptionMDNSDiscovery,
+	)
+	if err != nil {
+		fatal(err)
+	}
+
+	<-crouting
```
rlan35, line 209: Comment for this line?

p2p/ipfsutil/api_inmemory.go
```diff
@@ -0,0 +1,109 @@
...
+	routing := ipfs_libp2p.DHTClientOption
+	if opts.Routing != nil {
+		routing = opts.Routing
+	}
+
+	hostopts := ipfs_libp2p.DefaultHostOption
+	return &ipfs_node.BuildCfg{
+		Online:                      true,
+		Permanent:                   true,
+		DisableEncryptedConnections: false,
```
rlan35, line 54: Better add some comments to some of the config worth explaining. Like do we really need encryption? Will it slow things down (as in DisableEncryptedConnections)

node/service_setup.go
fxfactorial: there is a nuanced behavior of pubsub that under gossip sub, you must first publish before subscribe works (this is only when pubsub running as gossip sub)

p2p/ipfsutil/api_inmemory.go
fxfactorial: you always have encryption in libp2p , by default always been on

p2p/tinder/driver.go
```diff
@@ -0,0 +1,56 @@
+package tinder
```
rlan35, line 1: are p2p/tinder and p2p/ipfsutil mostly copied code?

fxfactorial: yes - I took from a project I found while looking for correct usages of libp2p (see point 2- Use berty project's service code) https://github.com/berty/berty

p2p/host.go
fxfactorial: sort of - I don't think strictly necessray but we do allow clinet to put in IP so will try - (although I think if we deploy on stn, it will work as it)

consensus/consensus_v2.go
```diff
@@ -119,88 +181,81 @@ func (consensus *Consensus) finalizeCommits() {
...
 
+	// if consensus.BlockNum()-beforeCatchupNum != 1 {
+	// 	utils.Logger().Warn().
+	// 		Uint64("beforeCatchupBlockNum", beforeCatchupNum).
+	// 		Msg("[FinalizeCommits] Leader cannot provide the correct block for committed message")
+	// 	return errors.New("leader cannot provide the correct block for committed message")
+	// }
+
 	// if leader success finalize the block, send committed message to validators
-	if err := consensus.msgSender.SendWithRetry(
```
rlan35, line 140: why remove the retry logic? I think the retry logic can same us from potential failures, even when the p2p layer is stronger than before, problems may still happen.

consensus/leader.go
```diff
@@ -12,154 +12,179 @@ import (
...
 			quorum.Prepare,
-			key,
-			consensus.priKey.PrivateKey[i].SignHash(consensus.blockHash[:]),
-			common.BytesToHash(consensus.blockHash[:]),
-			consensus.blockNum,
-			consensus.viewID,
+			consensus.PubKey.PublicKey[i],
+			consensus.priKey.PrivateKey[i].SignHash(h[:]),
+			d,
+			consensus.BlockNum(),
```
rlan35, line 75: load this before hand so not repeated.

consensus/consensus_v2.go
fxfactorial: it will/can come back - but its has its own mutexes that lock up - first need correctness, then easy to add retry on top if needed -

consensus/validator.go
```diff
@@ -1,188 +1,184 @@
...
-				groupID,
-				p2p.ConstructMessage(networkMessage.Bytes),
-			); err != nil {
-				consensus.getLogger().Warn().Err(err).Msg("[OnAnnounce] Cannot send prepare message")
-			} else {
-				consensus.getLogger().Info().
-					Str("blockHash", hex.EncodeToString(consensus.blockHash[:])).
-					Msg("[OnAnnounce] Sent Prepare Message!!")
-			}
+		if err := consensus.host.SendMessageToGroups(
```
rlan35, line 76: why is the listening mode check removed?

node/block-procesing.go
```diff
@@ -0,0 +1,190 @@
...
+
+		if err := node.Gossiper.AcceptedBlock(
+			node.Consensus.ShardID, newBlock,
+		); err != nil {
+			return err
+		}
+
+		// node.Gossiper.NewBeaconChainBlock(newBlock)
+
+		// if node.Consensus.ShardID == shard.BeaconChainShardID {
```
rlan35: why are these deleted?

fxfactorial: The topics ideas I changed up , the topics groups aren’t being used correctly , and currently it will cause deadlocks, it’s my next focus since getting foundation of sync working

PR #2880, opened 2020-04-24
Add raw stake and median to super committee

fxfactorial:


consensus/quorum/one-node-staked-vote.go
```diff
@@ -162,6 +162,11 @@ func (v *stakedVoteWeight) String() string {
 	return string(s)
 }
 
+// HACK later remove - unify votepower in UI (aka MarshalJSON)
+func (v *stakedVoteWeight) SetRawStake(key shard.BLSPublicKey, d numeric.Dec) {
+	v.roster.Voters[key].RawStake = d
```
rlan35, line 167: check for potential nil pointer .

rlan35:
> Maybe we should singleflight on this, the summation and div is waste to compute over and over and with watchdog , you have it guaranteed going off all the time

we can do it in next PR

PR #2883, opened 2020-04-25
Validator dedup on identity and bls keys

rlan35:
Add the dedup feature for validator identifying fields including identity and bls keys.

Fail the staking txn if dups are detected and return corresponding errors.

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -46,7 +46,6 @@ var (
 	errCommissionRateChangeTooHigh = errors.New("commission rate can not be higher than maximum commission rate")
 	errNoRewardsToCollect          = errors.New("no rewards to collect")
 	errNegativeAmount              = errors.New("amount can not be negative")
-	errDupName                     = errors.New("validator name exists")
 	errDupIdentity                 = errors.New("validator identity exists")
 	errDupBlsKey                   = errors.New("BLS key exists")
 )
@@ -389,13 +388,12 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
 	return st.gasUsed(), err
 }
 
-func (st *StateTransition) checkDuplicateFields(validator common.Address, name, identity string, blsKeys []shard.BLSPublicKey) error {
+func (st *StateTransition) checkDuplicateFields(validator common.Address, identity string, blsKeys []shard.BLSPublicKey) error {
 	addrs, err := st.bc.ReadValidatorList()
 	if err != nil {
 		return err
 	}
 
-	checkName := name != ""
 	checkIdentity := identity != ""
 	checkBlsKeys := len(blsKeys) != 0
 
@@ -412,9 +410,6 @@ func (st *StateTransition) checkDuplicateFields(validator common.Address, name, 
 				return err
 			}
 
-			if checkName && wrapper.Name == name {
-				return errDupName
-			}
 			if checkIdentity && wrapper.Identity == identity {
 				return errDupIdentity
 			}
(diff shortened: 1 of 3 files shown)


fxfactorial:
I don’t think we should dedup on anything but crypto based things. Otherwise we can have have lots of squatters eat up spots of unique names

core/state_transition.go
```diff
@@ -384,9 +389,57 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
...
+
+	for _, addr := range addrs {
+		if bytes.Compare(validator.Bytes(), addr.Bytes()) != 0 {
+			wrapper, err := st.state.ValidatorWrapperCopy(addr)
+
+			if err != nil {
+				return err
+			}
+
+			if checkName && wrapper.Name == name {
```
denniswon, line 415: Why cant the validators have same name and identity? Do they have to be unique? Blskey wise, how can there be duplicates when bls key is generated by the protocol amd unique per validator?

sophoah: @denniswon we have issues with node operator mistakenly re-using the same BLS for a different validator

sophoah: regarding the name and identify this is more to avoid people trying to impersonate another validator, though they could still trick a delegator with some character difference : "P-OPS validator| vs "P-0PS validator"

gupadhyaya:
Agree with @fxfactorial I also think dedup is not needed.

rlan35:
> Agree with @fxfactorial I also think dedup is not needed.

What about identity proof if we later want to support it? https://github.com/solana-labs/solana/issues/5187   For name, I think it's not a worry about name hoarding. There is a txn fee cost to that. Having duplicate names will really be potential trouble or attack to validators/delegators.

rlan35:
Jenkins, test this please

fxfactorial:
I would prefer then thinking slower and less adhoc than this way for the Solana idea , we could tolerate naming issues for now until figure out Solana way

rlan35:
Discussed offline. Removed the naming dedupl. But identity should by definition be unique.

core/state_transition.go
```diff
@@ -384,9 +388,52 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
...
+	checkIdentity := identity != ""
+	checkBlsKeys := len(blsKeys) != 0
+
+	blsKeyMap := map[shard.BLSPublicKey]struct{}{}
+	for _, key := range blsKeys {
+		blsKeyMap[key] = struct{}{}
+	}
+
+	for _, addr := range addrs {
+		if bytes.Compare(validator.Bytes(), addr.Bytes()) != 0 {
```
fxfactorial, line 406: Not compare, .equal

rlan35: done.

core/state_transition.go
```diff
@@ -384,9 +388,52 @@ func (st *StateTransition) StakingTransitionDb() (usedGas uint64, err error) {
...
+			}
+
+			if checkIdentity && wrapper.Identity == identity {
+				return errDupIdentity
+			}
+			if checkBlsKeys {
+				for _, existingKey := range wrapper.SlotPubKeys {
+					if _, ok := blsKeyMap[existingKey]; ok {
+						return errDupBlsKey
+					}
```
fxfactorial, line 420: Error.wrapf(errdupkey, “the dupled key

rlan35: done.

core/state_transition.go
denniswon: but I am not sure if we should make identity and name as unique at the protocol level. or should do at the sdk/cli level. idk, not sure if having more than bls key being unique to validator, other info do not necessarily have to be unique at the protocol level. _shrug_

PR #2886, opened 2020-04-25
[p3] sharding configuration for p3 launch

LeoHChen:

### DIFF
--- a/internal/configs/sharding/pangaea.go
+++ b/internal/configs/sharding/pangaea.go
@@ -15,8 +15,8 @@ var PangaeaSchedule pangaeaSchedule
 type pangaeaSchedule struct{}
 
 const (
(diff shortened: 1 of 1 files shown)


internal/configs/sharding/pangaea.go
```diff
@@ -81,4 +81,4 @@ var pangaeaReshardingEpoch = []*big.Int{
 }
 
 var pangaeaV0 = MustNewInstance(4, 60, 60, numeric.OneDec(), genesis.TNHarmonyAccounts, genesis.TNFoundationalAccounts, pangaeaReshardingEpoch, PangaeaSchedule.BlocksPerEpoch())
```
rlan35, line 83: if we reduce our node to 30, should change this part too.

PR #2889, opened 2020-04-27
[rpc] singleflight for all validator info, median, and super committee rpcs

gupadhyaya:
added singleflight for:
* GetAllValidatorInformation
* GetSuperCommittee
* GetMedianRawStakeSnapshot
apis

### DIFF
--- a/hmy/api_backend.go
+++ b/hmy/api_backend.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"math/big"
 	"sync"
-	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/math"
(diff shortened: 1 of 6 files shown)


hmy/api_backend.go
```diff
@@ -443,7 +460,21 @@ func (b *APIBackend) GetValidatorInformation(
 func (b *APIBackend) GetMedianRawStakeSnapshot() (
 	*committee.CompletedEPoSRound, error,
 ) {
-	return committee.NewEPoSRound(b.hmy.BlockChain())
+	res, err := b.SingleFlightRequest(
+		"median",
```
rlan35, line 464: This doesn't work because the result keep changing every block, And the key is always “median" it will give out stale data.

rlan35: This RPC can not be cached simply like this at this high level. 
Can only go deeper and cache on lower-level computations.

internal/hmyapi/apiv1/backend.go
```diff
@@ -30,6 +30,8 @@ type Backend interface {
 	NetVersion() uint64
 	ProtocolVersion() int
 	ChainDb() ethdb.Database
+	SingleFlightRequest(key string, fn func() (interface{}, error)) (interface{}, error)
```
denniswon, line 33: dont think we should add single flight at this api level.

hmy/api_backend.go
gupadhyaya: the cache is cleared below every block. why not cache the whole rpc? there are mainly three steps if we were to cache inner computations: 1) prepreOrder (preprocessing), which should be most computational, 2) followed by median which should be second most computational, and 3) post processing (lite one). Even the inner computations (1 & 2) will have to be cache block-level, which is same as above right?

internal/hmyapi/apiv1/backend.go
gupadhyaya: why? unless deeper level is more beneficial.

hmy/api_backend.go
rlan35: The right cache should include the block number of the current block in chain..

rlan35: I see you discarded the cache every 8 seconds. But let's do the correct way of adding block num to the key

gupadhyaya: done.

Issue #2887, opened 2020-04-26
ValidatorWrapper func() got "address not present in state" error


rlan35:
{"level":"error","port":"9000","ip":"35.166.89.131","error":"at root: 0xd207a625abff6e64667a0cb1f306693b0bb51e28311f18a87d7853b8bbbd278d: address not present in state","blockNum":111222,"shardID":0,"epoch":2926,"caller":"/home/leochen/go/src/github.com/harmony-one/harmony/node/node.go:608","time":"2020-04-26T01:04:24.702415948Z","message":"[InitConsensusWithValidators] Failed getting shard state"}
{"level":"warn","port":"9000","ip":"35.166.89.131","shardID":0,"error":"at root: 0xd207a625abff6e64667a0cb1f306693b0bb51e28311f18a87d7853b8bbbd278d: address not present in state","caller":"/home/leochen/go/src/github.com/harmony-one/harmony/cmd/harmony/main.go:523","time":"2020-04-26T01:04:24.702432638Z","message":"InitConsensusWithMembers failed"}

This happened after explorer node crashed during validator creation spamming.

Could be a potential data consistency issue on ValidatorList.

PR #2891, opened 2020-04-27
Roll back validator list when block is reverted

rlan35:
Fix:  https://github.com/harmony-one/harmony/issues/2887


core/blockchain.go
```diff
@@ -551,8 +548,41 @@ func (bc *BlockChain) repair(head **types.Block) error {
 		bc.WriteCommitSig((*head).NumberU64()-1, sigAndBitMap)
 
 		// Otherwise rewind one block and recheck state availability there
+		for _, stkTxn := range (*head).StakingTransactions() {
+			if stkTxn.StakingType() == staking.DirectiveCreateValidator {
+				if addr, err := stkTxn.SenderAddress(); err != nil {
```
gupadhyaya, line 553: should this be `err == nil` ?

PR #2899, opened 2020-04-29
Fix super committee bug and add raw-stake to per-key metrics

gupadhyaya:
* fixed super-committee rpc showing zero `epos-median-stake` 
* set raw-stake in per-key metric, such that getting validator information reflects raw-stake as well

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2354,6 +2354,19 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 		}
 		stats.MetricsPerShard = earningWrapping
 
+		// fetch raw-stake from snapshot and update per-key metrics
+		if snapshot, err := bc.ReadValidatorSnapshotAtEpoch(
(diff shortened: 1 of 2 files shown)


core/blockchain.go
```diff
@@ -2354,6 +2354,19 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 		}
 		stats.MetricsPerShard = earningWrapping
 
+		// fetch raw-stake from snapshot and update per-key metrics
+		if snapshot, err := bc.ReadValidatorSnapshotAtEpoch(
+			newEpochSuperCommittee.Epoch, key,
+		); err == nil {
+			wrapper := snapshot.Validator
+			spread := numeric.NewDecFromBigInt(wrapper.TotalDelegation()).
+				QuoInt64(int64(len(wrapper.SlotPubKeys)))
```
rlan35, line 2363: always check for div by 0

gupadhyaya: done.

Issue #2837, opened 2020-04-15
validator APR calculation should cover more time period


rlan35:
Right now the validator APR is simply the extrapolated estimate of APR based on the reward from last epoch. This makes the APR value fluctuate very much and unstable to user to rely on.

We should calculate the APR based on as much as possible the data points of the rewards of epochs (with an upper limit of e.g. 100) .

PR #2900, opened 2020-04-29
apr average over last hundred epochs

gupadhyaya:
fixes https://github.com/harmony-one/harmony/issues/2837
* store apr every epoch (as previously apr is overwritten every epoch)
* average over last 100 epochs when getting apr in get validator information
* average is cached for every epoch to avoid recomputation within epoch

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -2231,6 +2231,13 @@ func (bc *BlockChain) ReadValidatorSnapshotAtEpoch(
 	return rawdb.ReadValidatorSnapshot(bc.db, addr, epoch)
 }
 
+// ReadValidatorAPRAt read the snapshot APR for validator
+func (bc *BlockChain) ReadValidatorAPRAt(
+	epoch *big.Int, addr common.Address,
+) (numeric.Dec, error) {
+	return rawdb.ReadValidatorAPR(bc.db, addr, epoch)
+}
+
 // ReadValidatorSnapshot reads the snapshot staking information of given validator address
 func (bc *BlockChain) ReadValidatorSnapshot(
 	addr common.Address,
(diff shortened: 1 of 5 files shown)


core/blockchain.go
```diff
@@ -2373,6 +2380,19 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 					}
 				} else {
 					stats.APR = *aprComputed
```
rlan35, line 2382: why not just add the history APRs into the existing stats structure?

hmy/api_backend.go
```diff
@@ -437,7 +437,43 @@ func (b *APIBackend) GetValidatorInformation(
...
+	nowMinus100 := now.Sub(now, big.NewInt(100))
+	if nowMinus100.Cmp(epochFrom) > 0 {
+		epochFrom = nowMinus100
+	}
+
+	if avgAPR, err := b.SingleFlightRequest(
+		key, func() (interface{}, error) {
+			total := numeric.ZeroDec()
+			count := 0
+			for i := now.Int64(); i > nowMinus100.Int64(); i-- {
```
rlan35, line 458: This logic will have issue. If a validator is elected only at first and last epoch in the 100 epochs,  then the calculated APR looks normal, but actually to the delegators, the return is very little. 

better do this, start from nowMinus100, start counting (count+1) only after hitting the first present APR (to avoid new validator not having enough history), and from there, starting counting non-present APR too (the holes should be counted). finish until now.

core/blockchain.go
gupadhyaya: stats are overwritten. no snapshotting. didn't want to snapshot the whole stats, hence separately snapshot the APRs. any other alternatives?

hmy/api_backend.go
gupadhyaya: good point... agree.

core/blockchain.go
rlan35: add like a slice or map inside the stats?

gupadhyaya: changed to fixed length circular slice inside stats.

staking/types/validator.go
```diff
@@ -183,8 +185,10 @@ type VoteWithCurrentEpochEarning struct {
 
 // ValidatorStats to record validator's performance and history records
 type ValidatorStats struct {
-	// APR ..
-	APR numeric.Dec `json:"-"`
+	// APRs is the APR history
+	APRs []numeric.Dec `json:"-"`
```
rlan35, line 189: I think it's better to use a map from epoch to apr. This current way is not flexible in terms of usage.

gupadhyaya: done.

core/blockchain.go
```diff
@@ -2372,7 +2371,22 @@ func (bc *BlockChain) UpdateValidatorVotingPower(
 						return nil, err
 					}
 				} else {
-					stats.APR = *aprComputed
+					now := newEpochSuperCommittee.Epoch
```
rlan35, line 2374: should be currentEpochSuperCommittee.Epoch.

PR #2904, opened 2020-04-29
compute shardExternalShare based on payableSigners overallpercent

gupadhyaya:


internal/chain/reward.go
```diff
@@ -294,9 +294,12 @@ func AccumulateRewardsAndCountSigs(
 					return network.EmptyPayout, err
 				}
 
-				shardExternalShare := shard.Schedule.InstanceForEpoch(
-					epoch,
-				).ExternalVotePercent()
+				shardExternalShare := numeric.ZeroDec()
```
rlan35, line 297: rename to allSignersShare.  Also do the same for beacon chain rewarding "beaconExternalShare"

PR #2909, opened 2020-04-30
fix super committee median for current committee and raw-stake assignment

gupadhyaya:
* fix super committee bug that was overwriting previous median instead of current
* read from validator wrapper to update raw-stake instead of snapshot, as snapshot is still in batch and not flushed.


rlan35:
Jenkins, test this please.

rlan35:
Jenkins, test this please.

PR #2924, opened 2020-05-05
Fix error sink msg duplication & false positives

Daniel-VDM:
## Issue

Currently, the error sink just dumps any error to a ring buffer and does not remove any error report from the buffer if the tx ended up being successful. Some downstream tools (like CLI and block explorer) use the sink to check if a recent transaction failed and short circuits a failure msg if it does not. Therefore, it is important that the error sink does not duplicate error msgs and removes error messages for txs that made it through.

Moreover, I exposed the TxPool stats RPC.

SS of RPC:

Screen Shot 2020-05-05 at 3 19 46 AM

## Test

It is sufficient to test this on localnet as you can cause errors as needed. This was done & a unit test was added.

### Unit Test Coverage

Before:

```
PASS
coverage: 26.5% of statements
ok  	github.com/harmony-one/harmony/core	7.030s
```

After:

```
PASS
coverage: 26.2% of statements
ok  	github.com/harmony-one/harmony/core	6.991s
```

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/internal/configs/sharding/testnet.go
+++ b/internal/configs/sharding/testnet.go
@@ -82,5 +82,5 @@ var testnetReshardingEpoch = []*big.Int{
 	params.TestnetChainConfig.StakingEpoch,
 }
 
-var testnetV0 = MustNewInstance(4, 25, 25, numeric.OneDec(), genesis.TNHarmonyAccounts, genesis.TNFoundationalAccounts, testnetReshardingEpoch, TestnetSchedule.BlocksPerEpoch())
+var testnetV0 = MustNewInstance(4, 30, 25, numeric.OneDec(), genesis.TNHarmonyAccounts, genesis.TNFoundationalAccounts, testnetReshardingEpoch, TestnetSchedule.BlocksPerEpoch())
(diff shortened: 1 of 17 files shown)


internal/configs/sharding/testnet.go
```diff
@@ -82,5 +82,5 @@ var testnetReshardingEpoch = []*big.Int{
 	params.TestnetChainConfig.StakingEpoch,
 }
 
-var testnetV0 = MustNewInstance(4, 25, 25, numeric.OneDec(), genesis.TNHarmonyAccounts, genesis.TNFoundationalAccounts, testnetReshardingEpoch, TestnetSchedule.BlocksPerEpoch())
+var testnetV0 = MustNewInstance(4, 30, 25, numeric.OneDec(), genesis.TNHarmonyAccounts, genesis.TNFoundationalAccounts, testnetReshardingEpoch, TestnetSchedule.BlocksPerEpoch())
```
rlan35, line 85: why is this changed?

gupadhyaya, line 85: why is this change?

Daniel-VDM: It's to fix all the unit tests that use the test blockchain. We need to have slots for BLS keys so that one can send staking txs in the prestaking epoch. (previously we would get not sufficient slots for BLS keys -- hence the old TODO msgs)

janet-harmony:
Jenkins, test this please

PR #2932, opened 2020-05-06
fix super-committee total-raw-staked bug

gupadhyaya:
`total-raw-staked` field in deciders of super-committee roc was actually total effective stake. this pr fixes that and also adds `total-raw-stake` for clients to directly consume. This is still per-shard total amounts and all-shard aggregation is still required.

### DIFF
--- a/consensus/quorum/one-node-staked-vote.go
+++ b/consensus/quorum/one-node-staked-vote.go
@@ -184,18 +184,20 @@ func (v *stakedVoteWeight) MarshalJSON() ([]byte, error) {
 	}
 
 	type t struct {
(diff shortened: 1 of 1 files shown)


gupadhyaya:
sample output:
```
          "count": 110,
          "external-validator-slot-count": 80,
          "hmy-voting-power": "0.680000000000000010",
          "policy": "SuperMajorityStake",
          "staked-voting-power": "0.319999999999999990",
          "total-effective-stake": "969335743872942100000000000.000000000000000000",
          "total-raw-stake": "1544903468742533906993114286.000000000000000000"
        }
```

PR #2934, opened 2020-05-06
Fix potential wrong epoch for shard state commitment

rlan35:
Previously the shardstate was always added into DB with key of epoch+1. But it is not the case when a shard chain is lagged and jumped to a new epoch that's more than one epoch ahead.

### DIFF
--- a/core/offchain.go
+++ b/core/offchain.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"bytes"
+	"github.com/harmony-one/harmony/block"
 	"math/big"
 	"sort"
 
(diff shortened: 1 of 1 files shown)


core/offchain.go
```diff
@@ -94,25 +96,24 @@ func (bc *BlockChain) CommitOffChainData(
...
-		if err == nil && newShardState.Epoch != nil && bc.chainConfig.IsStaking(newShardState.Epoch) {
-			// After staking, the epoch will be decided by the epoch in the shard state.
-			newEpoch = new(big.Int).Set(newShardState.Epoch)
-		}
 	}
 
 	// Do bookkeeping for new staking txns
 	newVals, err := bc.UpdateStakingMetaData(
-		batch, block, state, epoch, newEpoch,
+		batch, block, state, epoch, nextBlockEpoch,
```
gupadhyaya, line 116: so, this update was wrong before? as it would right write snapshot for newly created validator even before NewEpoch?

rlan35: Right, it was duplicate snapshotted, But since those validator will be snapshotted again, the wrong snapshot will be overwritten. So it wasn't causing any issue, but main potential issue is the WriteShardState part.

PR #2936, opened 2020-05-06
[dryrun] set prestaking/staking epoch

LeoHChen:

### DIFF
--- a/internal/configs/sharding/mainnet.go
+++ b/internal/configs/sharding/mainnet.go
@@ -28,6 +28,7 @@ const (
 	mainnetV1_3Epoch = 36
 	mainnetV1_4Epoch = 46
 	mainnetV1_5Epoch = 54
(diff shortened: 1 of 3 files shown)


internal/configs/sharding/mainnet.go
```diff
@@ -28,6 +28,7 @@ const (
 	mainnetV1_3Epoch = 36
 	mainnetV1_4Epoch = 46
 	mainnetV1_5Epoch = 54
+	mainnetV2Epoch   = 177
```
rlan35, line 31: Actually, this should be 178. Should be the same as staking epoch. Not pre-staking epoch.

rlan35: But since you had the same config (the voting power doesn't take effect yet). It shouldn't be the cause of the upgrade issue.

PR #2949, opened 2020-05-07
Add SenderAddress to plain Tx Type

Daniel-VDM:
## Issue

Plain tx had a cumbersome way to fetch sender address, so added `SenderAddress` to plain tx, to make the same as staking tx. Doing this meant I could remove uses of `PoolTransactionSender`, so did that as well and refactored all associated functions. 

## Test

Tested locally, sending staking and plain tx all worked and passed through the pool.

### Unit Test Coverage

Before:

```
PASS
coverage: 26.0% of statements
ok      github.com/harmony-one/harmony/core     6.913s

```

After:

```
PASS
coverage: 26.0% of statements
ok      github.com/harmony-one/harmony/core     6.927s
```

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/core/types/transaction.go
+++ b/core/types/transaction.go
@@ -415,6 +415,22 @@ func (tx *Transaction) Copy() *Transaction {
 	return &tx2
 }
 
+// SenderAddress returns the address of transaction sender
+// Note that mainnet has unprotected transactions prior to Epoch 28
+func (tx *Transaction) SenderAddress() (common.Address, error) {
+	var signer Signer
+	if !tx.Protected() {
+		signer = HomesteadSigner{}
+	} else {
+		signer = NewEIP155Signer(tx.ChainID())
+	}
+	addr, err := Sender(signer, tx)
+	if err != nil {
+		return common.Address{}, err
+	}
+	return addr, nil
+}
+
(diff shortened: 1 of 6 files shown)


core/types/transaction.go
```diff
@@ -415,6 +415,22 @@ func (tx *Transaction) Copy() *Transaction {
...
+func (tx *Transaction) SenderAddress() (common.Address, error) {
+	var signer Signer
+	if !tx.Protected() {
+		signer = HomesteadSigner{}
+	} else {
+		signer = NewEIP155Signer(tx.ChainID())
+	}
+	addr, err := Sender(signer, tx)
+	if err != nil {
+		return common.Address{}, err
```
fxfactorial, line 429: please first one becomes pointer and nil. If giving back non-nil error, then proper value should be nil

Daniel-VDM: To be consistent with staking tx, I think it should be as-is: 

https://github.com/harmony-one/harmony/blob/master/staking/types/transaction.go#L314  

It would require a lot more change on things that consume staking tx sender addr if we decide to change it the other way IMO. Not sure if worth it.

PR #2955, opened 2020-05-08
Double sign unit test

JackyWYX:
## Description

1. Rewrite and Implemented unit test cases in `staking/slash` module. Added much more strict result checks.
2. A series of copy functions as test helper are added to new modules `staking/types/staketest` and `staking/slash/slashtest`. The module will not compile in production code.

### Unit Test Coverage

Before:

```
PASS
coverage: 46.9% of statements
ok      github.com/harmony-one/harmony/staking/slash    0.246s
```

After:

```
PASS
coverage: 85.0% of statements
ok      github.com/harmony-one/harmony/staking/slash    0.162s
```

## Compatibilities

No source code logic is changed, thus there shouldn't be compatibility issues.


staking/slash/double-sign.go
```diff
@@ -58,33 +55,84 @@ type Moment struct {
...
+}
+
 // ConflictingVotes ..
 type ConflictingVotes struct {
 	FirstVote  Vote `json:"first-vote"`
 	SecondVote Vote `json:"second-vote"`
 }
 
+// Copy makes a deep copy of the ConflictingVotes structure
+func (cv ConflictingVotes) Copy() ConflictingVotes {
```
fxfactorial, line 94: Why are all these copy functions made ?

staking/slash/double-sign.go
```diff
@@ -143,17 +191,10 @@ func (r Record) String() string {
...
-	Config() *params.ChainConfig
-	ReadShardState(epoch *big.Int) (*shard.State, error)
-	CurrentBlock() *types.Block
-}
-
 // Verify checks that the slash is valid
 func Verify(
 	chain CommitteeReader,
-	state *state.DB,
+	state stateDB,
```
fxfactorial, line 197: Why is this not a pointer anymore

staking/slash/double-sign_test.go
```diff
@@ -1,20 +1,21 @@
...
-	"encoding/hex"
+	"errors"
 	"fmt"
 	"math/big"
+	"reflect"
+	"strings"
 	"testing"
 
+	"github.com/harmony-one/harmony/consensus/votepower"
+
```
fxfactorial, line 12: No space in imports

staking/slash/interface.go
```diff
@@ -0,0 +1,24 @@
...
+
+// CommitteeReader ..
+type CommitteeReader interface {
+	Config() *params.ChainConfig
+	ReadShardState(epoch *big.Int) (*shard.State, error)
+	CurrentBlock() *types.Block
+}
+
+// stateDB is the abstracted interface for *state.DB
+type stateDB interface {
```
fxfactorial, line 21: I see why no longer ptr

fxfactorial:
The testing coverage expansion is awesome ! Why all these copy methods though ? Do you have an example of where they were strictly necessary and fixed a mistake in code or are they helpers for your test code ?

staking/slash/double-sign.go
JackyWYX: *state.DB is abstracted to an interface. This will help me to write a fake state db only for testing.

staking/slash/double-sign_test.go
JackyWYX: Thanks for pointing that out.  Fixed.

staking/slash/interface.go
JackyWYX: Yep! I am using it as a simple adapter between modules that can help with decreasing the correlation between modules and specifies a clean boundary.

JackyWYX:
> The testing coverage expansion is awesome ! Why all these copy methods though ? Do you have an example of where they were strictly necessary and fixed a mistake in code or are they helpers for your test code ?

Thanks. The `Copy` methods are just helpers for testing cases and they are not called anywhere in source code. I left them there just in case that we might need them later so we don't need to write them again. Anyway, leaving them there won't harm right? 

BTW, the correctness of the `Copy` methods has been tested in the corresponding unit tests. Check the unit tests if you are interested :)

staking/slash/double-sign.go
JackyWYX: Simply to help with the result check in test cases.

fxfactorial:
I see then keep the testing copy helpers in _test code. Main repo code should only be relevant and live code for the purpose of the binary. Plus this slows down compile time and increase binary size.

staking/slash/interface.go
fxfactorial: I prefer no interface on this , the other interfaces you see are for breaking import cycle but this is an interface for convenience for you. In general we should use interfaces only when we hit an import cycle or when we have several concrete types and need to abstract over it. Otherwise it is not worth the complexity and runtime hit

fxfactorial:
I assume you then also added extra copy methods on the other PR for staking msg test ? If so , please update as well there , ie anything that is code to help your tests should be only in _test.go files

JackyWYX:
> I see then keep the testing copy helpers in _test code. Main repo code should only be relevant and live code for the purpose of the binary. Plus this slows down compile time and increase binary size.
> 
> It is a harm - it makes source code bigger , detracts from mental attention on just things that are needed to pay attention to. If code doesn’t run , then it doesn’t need to be included

I agree that it the unused code will increase the binary size. But since it will be heavily used in different modules, if I remove it from the source code, I will need to write deep copy method again and again in each module. So do you have any solution to this? I am all ears.

BTW, to my knowledge, it is a common practice to write source code for testing that is used in multiple module tests. Please check https://github.com/ethereum/go-ethereum/blob/master/core/chain_makers.go.

JackyWYX:
> I assume you then also added extra copy methods on the other PR for staking msg test ? If so , please update as well there , ie anything that is code to help your tests should be only in _test.go files

Not yet. I haven't added copy methods to source code in previous PRs. But the next few unit test will rely on these copy methods.

fxfactorial:
It is only used in your test code, hence you should only keep in test code. Admitting that it needlessly increases binary size should be argument enough.

I’m totally fine if you copy paste it across test code files, that’s totally fine. Copy paste in non test code is fine 

Citing go ethereum as common practice is not a good source, go ethereum is not the best example of good go coding

staking/slash/double-sign.go
fxfactorial: Please remove this interface. You should not change signatures just to make testing easier. Testing consideration should not affect logic in any way

JackyWYX: The interface is not just to make tests easier. Actually it's a practice of [dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle).

JackyWYX:
> It is only used in your test code, hence you should only keep in test code. Admitting that it needlessly increases binary size should be argument enough.
> 
> I’m totally fine if you copy paste it across test code files, that’s totally fine. Copy paste in in test code is fine
> 
> Citing go ethereum as common practice is not a good source, go ethereum is not the best example of good go coding
> 
> The main repo should not have to have additional burden of code just to make testing code easier for you and in which case you or most likely anyone else will not return to this testing code in any deep way for a while. But the burden of extra code is forever .
> 
> Please remove the extra code from main compile files .

Well, I can see the disagreement here. From my knowledge, unit tests are just as important as the production code. And making Copy method in production code is an effort to make tests readable and clean. Please check this outline really quick [clean code: unit tests](https://medium.com/@shley_ng/clean-code-unit-tests-c0c871219f75)

Also, I did some tests on building binary with and without copy method. It seems that copy method will increase the binary size from 28398716 bytes to 28399052 byte, changes almost neglectable.

staking/slash/double-sign.go
rlan35: I see both of your points. But since we don't have interface in other places. It's better to not change the signature here (unless we need to change it everywhere). Is it complicated to instead use a Mock state.DB for testing?

staking/slash/double-sign.go
rlan35: I agree with Edgar, to be cleaner, it's not good to add test-only code in main binary. You can implement them as a test utility instead in test code only.

staking/slash/double-sign.go
JackyWYX: Gotta. Will treat this principle as our coding standard. Thanks. 👍

staking/slash/double-sign.go
JackyWYX: Agreed. Implemented in a module that will not be included in production code. Done.

staking/slash/double-sign.go
JackyWYX: stateDB reverted to *state.DB. Work done.

staking/slash/interface.go
JackyWYX: No interface now. Done.

JackyWYX:
> It is only used in your test code, hence you should only keep in test code. Admitting that it needlessly increases binary size should be argument enough.
> 
> I’m totally fine if you copy paste it across test code files, that’s totally fine. Copy paste in in test code is fine
> 
> Citing go ethereum as common practice is not a good source, go ethereum is not the best example of good go coding
> 
> The main repo should not have to have additional burden of code just to make testing code easier for you and in which case you or most likely anyone else will not return to this testing code in any deep way for a while. But the burden of extra code is forever .
> 
> Please remove the extra code from main compile files .

Agreed. Extracted the copy methods to a different module `staking/types/test` which will not compile in production code. Thanks for pointing that out. There does exist a solution that satisfy both our needs 😆

staking/slash/double-sign.go
rlan35: These Copy() code are still here?

JackyWYX: Thanks for reminder. Just killed them all. 👍

PR #2957, opened 2020-05-08
[blockchain] singleflight on tagged rlp registry

fxfactorial:
current work in another PR has many situation of parallelism , so during race condition testing, I hit this open issue. 

> by doing a blk.ShardID() in one thread and a rlp.DecodeBytes in another thread, they both hit up 
> the global tagged RLP registry and it exploded

### DIFF
--- a/core/types/block.go
+++ b/core/types/block.go
@@ -43,6 +43,7 @@ import (
 	"github.com/harmony-one/taggedrlp"
 	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
+	"golang.org/x/sync/singleflight"
 )
 
 // Constants for block.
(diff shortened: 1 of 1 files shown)


JackyWYX:
Interesting. The problems seems to be the race condition rooted in higher function calls. Just curious is it challenging to locate where the race happens in function calls?

fxfactorial:
The root isn’t higher function calls , need to allow only 1 thread at a time to use the registry

JackyWYX:
> The root isn’t higher function calls , need to allow only 1 thread at a time to use the registry

Gotta. Thanks for you explanation :)

rlan35:
Is this tested with the same race condition setting and it's gone?

fxfactorial:
It fixed on my branch work but it’s not too hard to setup a self contained example

PR #2960, opened 2020-05-08
Enable NatPortMap in p2p host

jz0000:
## Issue

This is commit addressing issue 2959 "[Feature] enable NAT traversal for simple 1-level NAT situation with UPNP support: connecting local host validator behind a NAT to public leaders " 

## Test

Have tried this changes behind a NAT set up in an AWS VPC. And it seems the node is running properly and blockchain keeps synced to the network. 

Note: at this point we can't test every possible NATs in the world and there could be some NATs which do not support UPNP tools, but based on my impression, UPNP should be pretty common. And whenever libp2p add support to any more NATs, we get that additional support for free. 

### Unit Test Coverage

Jians-MacBook-Pro:p2p jz$ go test --cover
PASS
coverage: 19.6% of statements
ok  	github.com/harmony-one/harmony/p2p	0.013s
Jians-MacBook-Pro:p2p jz$

./test/test_before_submit.sh

=== RUN   TestGetHashFromNodeList
--- PASS: TestGetHashFromNodeList (0.00s)
=== RUN   TestHash
--- PASS: TestHash (0.00s)
PASS
ok  	github.com/harmony-one/harmony/shard	0.023s
testing: warning: no tests to run
PASS
ok  	github.com/harmony-one/harmony/staking/types	0.068s [no tests to run]
?   	github.com/harmony-one/harmony/test/chain	[no test files]
?   	github.com/harmony-one/harmony/test/crypto/bls	[no test files]
?   	github.com/harmony-one/harmony/test/p2pchat	[no test files]
?   	github.com/harmony-one/harmony/test/txgen	[no test files]
go test succeeded.
Some checks failed; see output above.

## Operational Checklist

   NO

    NO

NO, don't have any significant changes in terms of CPU/memory/disk...

### DIFF
--- a/p2p/host/hostv2/hostv2.go
+++ b/p2p/host/hostv2/hostv2.go
@@ -234,6 +234,7 @@ func New(self *p2p.Peer, priKey libp2p_crypto.PrivKey) (*HostV2, error) {
 	ctx := context.Background()
 	p2pHost, err := libp2p.New(ctx,
 		libp2p.ListenAddrs(listenAddr), libp2p.Identity(priKey),
+		libp2p.NATPortMap(),
 	)
 	if err != nil {
 		return nil, err


jz0000:
Some details about the testbed I set up in a VPC in AWS:

I have set up a virtual private cloud (VPC) on AWS with 2 subnets: the 1st one is a public subnet which can be accessed from outside; and the 2nd one is a private subnet with only internal private IP addresses. I launched two EC2 instances, one inside public subnet with a public IP and one inside private subnet with only private IP. Then I also launched a NAT instance with a public IP to connect the two subnets. Then in the private subnet I configured a route table where all the IPs destined to outside the private subnet point to this nat instance, where IP translation is happening: SIP will be changed to NAT's own IP for outgoing packets and DIP will be changed to local private IP for incoming packets. (Note that this external/internal mapping will be done automatically once we enable the NATPortMAP option in libp2p host, if any router has UPNP support on it)

And I build a native harmony node on the server inside private subnet and launched it, it is running smoothly with blocks synced correctly from the whole harmony network outside private subnet. This proves that once the NAT IP/port mapping is configured on the router or in this case a dedicated NAT instance, harmony node can run on it smoothly.

jz0000:
Note that this external/internal mapping will be done automatically once we enable the NATPortMAP option in libp2p host, if any router has UPNP support on it, and I have tested this through my own laptop, where a libp2p is able to configure IP/port mapping with the option enabled during host creation.

PR #2961, opened 2020-05-08
[P2PNetwork] Add network configuration 

SebastianJ:
## Issue

@fxfactorial needs a dedicated testing network for his P2P PR.

This PR adds support for such a new network, called p2p/p2pnet.

### Unit Test Coverage

```
$ cd internal/configs/
$ go test -cover ./...
ok  	github.com/harmony-one/harmony/internal/configs/node	(cached)	coverage: 55.2% of statements
ok  	github.com/harmony-one/harmony/internal/configs/sharding	(cached)	coverage: 19.7% of statements
?   	github.com/harmony-one/harmony/internal/configs/viper	[no test files]
```

## Operational Checklist

    **YES|NO**

    **NO**

    **YES**

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -378,6 +378,15 @@ stn|stress|stressnet)
   dns_zone=stn.hmny.io
   syncdir=stn
   ;;
+p2p|p2pnet)
+  bootnodes=(
+    /ip4/52.40.84.2/tcp/9842/p2p/QmbPVwrqWsTYXq1RxGWcxx9SWaTUCfoo1wA6wmdbduWe29
+  )
+  REL=p2p
+  network_type=p2pnet
+  dns_zone=p2p.hmny.io
+  syncdir=p2p
+  ;;
 devnet)
   bootnodes=(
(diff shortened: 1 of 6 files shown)


scripts/node.sh
```diff
@@ -378,6 +378,15 @@ stn|stress|stressnet)
   dns_zone=stn.hmny.io
   syncdir=stn
   ;;
+p2p|p2pnet)
+  bootnodes=(
+    /ip4/52.40.84.2/tcp/9842/p2p/QmbPVwrqWsTYXq1RxGWcxx9SWaTUCfoo1wA6wmdbduWe29
```
LeoHChen, line 383: you can't use the same bootnode port 9842 with STN.  Please use a new one.

SebastianJ: Ah right, I just added it as a placeholder for now, Edgar's PR doesn't even use self-hosted bootnodes.

Now thinking of it, maybe this PR shouldn't even go in to master but only his PR.

SebastianJ:
This probably doesn't have to go into master - better to just directly send a PR to Edgar's repo. Closing and sending a PR to this repo instead.

PR #2964, opened 2020-05-08
Avoid duplicate bls as harmony nodes; Fix epoch gap issue

rlan35:
1. Avoid duplicate bls keys as harmony nodes from validators.
2. Fix a corner case issue when there is a epoch gap for shard chains, which will cause the Decider to be set incorrectly when it's already in staking epoch.
3. Add a check for div by 0

### DIFF
--- a/shard/committee/assignment.go
+++ b/shard/committee/assignment.go
@@ -129,6 +129,20 @@ func prepareOrders(
 	essentials := map[common.Address]*effective.SlotOrder{}
 	totalStaked, tempZero := big.NewInt(0), numeric.ZeroDec()
 
+	// Avoid duplicate BLS keys as harmony nodes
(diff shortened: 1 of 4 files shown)


shard/committee/assignment.go
```diff
@@ -129,6 +129,20 @@ func prepareOrders(
 	essentials := map[common.Address]*effective.SlotOrder{}
 	totalStaked, tempZero := big.NewInt(0), numeric.ZeroDec()
 
+	// Avoid duplicate BLS keys as harmony nodes
+	instance := shard.Schedule.InstanceForEpoch(stakedReader.CurrentBlock().Epoch())
```
gupadhyaya, line 133: why would there be duplicates?

rlan35: In case our key is mistakenly used as validators . It may cause nil pointer crash

PR #2972, opened 2020-05-10
Fix viewID issue on prepared block during view change

rlan35:
Previously when the view change happened with a half-done consensus (prepared reached), all the validators sign commit on the block hash and number only.

With the latest addition of view ID into the commit payload, the validator needs to sign view ID too. The problem is that the validators don't have the view ID in the already prepared block, they only have the live current viewID which can be different from the block's.

This fix ensure the the view change with prepared phase needs the full prepared block in the view change msg. Thus all validators can validate the prepared block before commit and also can commit to the right view ID in the block.


block/factory/factory.go
```diff
@@ -32,7 +32,7 @@ func (f *factory) NewHeader(epoch *big.Int) *block.Header {
 	switch {
 	case f.chainConfig.IsPreStaking(epoch) || f.chainConfig.IsStaking(epoch):
 		impl = v3.NewHeader()
-	case f.chainConfig.IsCrossLink(epoch):
+	case f.chainConfig.IsCrossLink(epoch): //Question
```
fxfactorial, line 35: What question

consensus/consensus_viewchange_msg.go
```diff
@@ -35,14 +37,21 @@ func (consensus *Consensus) constructViewChangeMessage(pubKey *bls.PublicKey, pr
 	)
 	preparedMsg := consensus.FBFTLog.FindMessageByMaxViewID(preparedMsgs)
 
+	block := consensus.FBFTLog.GetBlockByHash(preparedMsg.BlockHash)
 	var msgToSign []byte
-	if preparedMsg == nil {
+
+	encodedBlock, err := rlp.EncodeToBytes(block)
+	if err != nil {
+		consensus.getLogger().Err(err).Msg("[constructViewChangeMessage] Failed encoding block")
```
fxfactorial, line 45: Please use plain global logger

block/factory/factory.go
rlan35: removed.

consensus/consensus_viewchange_msg.go
rlan35: I will leave it for your batch update (to be consistent with existing code for now)

PR #2984, opened 2020-05-12
[p2p] Protect consensus from attack 

fxfactorial:
This PR implements the following: 

0. Builds on top of prior usage of weighted semaphore but with a vastly different scale, battled hardened tuned parameters from in the field relentless spamming attacks. 
1. Raises all issues that arise in consensus to a proper error value, lack of this was tech debt and should make debugging much easier. 
2. Remove dead functions & data structures, fields.
3. Refactor bootstrap function call
4. Remove consensus message handling from consensus thread. 
5. Add topic validator on each topic
6. Fixes open race conditions, mainly with usages of time. 
7. Introduces usage of `errors.WithStack` to the code base, an easy win for debugging
8. Fixes mistaken usage of pubsub in early start of program. 
9. Currently only temp time pubsub blacklist peer if they send not p2p sized message assumptions or not consensus or node based messages. 
10. Tested with a 4 second block time 
11. Tested with a 500 millisecond block time
12. Tested with a 50 millisecond block time (!!), 3 shards maintained consensus, 1 shard broke due to syncing code bucking.

Tested under heavy spamming attacks with consensus not missing even 1 message during over 1 hour continuous 60k p2p attack.

10, 11, 12 tested in 48 node OSTN, 12 node per shard network.


consensus/checks.go
```diff
@@ -11,10 +11,6 @@ import (
 const MaxBlockNumDiff = 100
 
 func (consensus *Consensus) validatorSanityChecks(msg *msg_pb.Message) bool {
-	if msg.GetConsensus() == nil {
```
rlan35, line 14: Why remove this check ? The consensus msg can be nil. We hit that nilpointer before, that’s why we added the check.

node/node.go
```diff
@@ -374,76 +337,71 @@ func (node *Node) Start() error {
...
+						m         msg_pb.Message
+						senderKey *bls.PublicKey
+					)
+
+					if err := protobuf.Unmarshal(cnsMsg, &m); err != nil {
+						errChan <- err
+						return false
+					}
+
+					if c := m.GetConsensus(); c.GetSenderPubkey() != nil {
```
rlan35, line 365: Consensus can be nil. Also what about view change msg validation?

p2p/host.go
```diff
@@ -71,24 +59,18 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
...
 
-	const MaxSize = 2_145_728
 	options := []libp2p_pubsub.Option{
-		libp2p_pubsub.WithPeerOutboundQueueSize(64),
-		libp2p_pubsub.WithMaxMessageSize(MaxSize),
-	}
-	if len(traceFile) > 0 {
-		tracer, _ := libp2p_pubsub.NewJSONTracer(traceFile)
-		options = append(options, libp2p_pubsub.WithEventTracer(tracer))
+		libp2p_pubsub.WithValidateQueueSize(64),
```
rlan35, line 70: What are these for?

p2p/host.go
```diff
@@ -71,24 +59,18 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
...
 		libp2p.ListenAddrs(listenAddr),
 		libp2p.Identity(key),
-		// libp2p.DisableRelay(),
 		libp2p.EnableNATService(),
 		libp2p.ForceReachabilityPublic(),
 	)
 	if err != nil {
 		return nil, errors.Wrapf(err, "cannot initialize libp2p host")
 	}
-	traceFile := os.Getenv("P2P_TRACEFILE")
```
LeoHChen, line 81: why remove tracefile option?

consensus/checks.go
fxfactorial: Because it’s handled at message validation earlier

node/node.go
fxfactorial: Iirc view change is consensus as well , so this should cover it

p2p/host.go
fxfactorial: Just parameters to control validation in libp2p

consensus/consensus_v2.go
```diff
@@ -23,12 +23,16 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
 	if len(payload) == 0 {
 		return
 	}
-	msg := &msg_pb.Message{}
-	if err := protobuf.Unmarshal(payload, msg); err != nil {
+
+	var m msg_pb.Message
+
+	if err := protobuf.Unmarshal(payload, &m); err != nil {
```
LeoHChen, line 29: what's the benefit of doing this?

node/node.go
fxfactorial: Ah ya, forgot to come back to && part

consensus/checks.go
fxfactorial: And because there is way more compute available to handle this rather than force the consensus thread to waste compute

node/node.go
fxfactorial: But ya, admittedly , not sure for view change, can come back to in next PR, after have legit way to maybe spam attack that ?

node/node.go
```diff
@@ -374,76 +337,71 @@ func (node *Node) Start() error {
...
+		if nodeconfig.GroupID(topicNamed) == nodeconfig.NewGroupIDByShardID(
+			nodeconfig.ShardID(node.Consensus.ShardID),
+		) {
+
+			pubsub.RegisterTopicValidator(
+				topicNamed,
+				func(ctx context.Context, peer libp2p_peer.ID, msg *libp2p_pubsub.Message) bool {
+					sig2 := msg.GetData()
+					if len(sig2) < p2pMsgPrefixSize {
+						return true
```
LeoHChen, line 350: return false here?

node/node.go
```diff
@@ -374,76 +337,71 @@ func (node *Node) Start() error {
...
 		topicNamed := allTopics[i].Name
+
+		if nodeconfig.GroupID(topicNamed) == nodeconfig.NewGroupIDByShardID(
+			nodeconfig.ShardID(node.Consensus.ShardID),
+		) {
+
+			pubsub.RegisterTopicValidator(
+				topicNamed,
+				func(ctx context.Context, peer libp2p_peer.ID, msg *libp2p_pubsub.Message) bool {
+					sig2 := msg.GetData()
```
LeoHChen, line 348: do you need to check if msg is nil?

node/node.go
```diff
@@ -374,76 +337,71 @@ func (node *Node) Start() error {
...
+							c.GetSenderPubkey(),
+						)
+						if err != nil {
+							errChan <- err
+							return false
+						}
+						senderKey = key
+					}
+
+					if err := consensus.VerifyMessageSig(senderKey, &m); err != nil {
```
LeoHChen, line 376: any tolerance level on the sender? what if the peer had some outdated db? will you blacklist the peer forever? This has to be very careful.

node/node.go
```diff
@@ -374,76 +337,71 @@ func (node *Node) Start() error {
...
+						pubsub.BlacklistPeer(peer)
+						errChan <- err
+						return false
+					}
+
+					return true
+
+				},
+				libp2p_pubsub.WithValidatorTimeout(24),
```
LeoHChen, line 385: maybe we can define some configuration for those critical parameters.

node/node.go
```diff
@@ -452,42 +410,30 @@ func (node *Node) Start() error {
...
-				}
-
 				nextMsg, err := sub.Next(context.Background())
 				if err != nil {
 					errChan <- err
 					continue
 				}
+
+				payload := nextMsg.GetData()
+				if len(payload) < p2pMsgPrefixSize {
```
LeoHChen, line 433: print a warning message here?

p2p/host.go
fxfactorial: It’s as Adhoc and not part of any config ,

fxfactorial:
The blacklist is time bounded , not forever

node/node.go
fxfactorial: It’s a time tolerance already, time expires, you have to dig into source of it to see. 

Db would be outdated by over an epoch for this to happen, can you say a specific scenario in which it can happen

node/node.go
fxfactorial: Not necessarily, they are part of a bigger config already proper via ipfs. I will unify then, when will bring that other pr back up to speed

node/node.go
fxfactorial: No because that will invalidate the message . There are messages that are still valid to go through here , basically beacon based things

node/node.go
fxfactorial: No, this msg is never nil

consensus/consensus_v2.go
fxfactorial: More idiomatic unmarshal and less needless allocation

node/node.go
fxfactorial: It’s would a very noisy log I think, and this part is existing logic from many months ago too.

consensus/consensus_v2.go
JackyWYX: Actually I think the following way has the least allocation.

```
var m *msg_pb.Message
if err := protobuf.Unmarshal(payload, m); err != nil {...}
```

But I cannot agree more on this idiomatic way of writing unmarshal code.

BTW, did some experiment about unmarshaling with pointer or value, the object will always escape to queue. You were definitely right about it - no difference in GC. And thanks for pointing that out in last video chat :)

node/node.go
gupadhyaya: this check also need to happen for ViewChange right? as you removed the `VerifyMessageSig` check in `viewChangeSanityCheck`

fxfactorial: Sure - but viewchange is either already a consensus message or I will handle there - viewchange/consensus should not be doing those checks - they are huge waste of compute of the consensus thread

gupadhyaya: yes, need `c := m.GetViewchange()` also.

fxfactorial: thank you!

node/node.go
```diff
@@ -403,67 +403,62 @@ func (node *Node) Start() error {
...
+						if current == 0 {
+							errChan <- errNoConsensusHandlers
+							return
+						}
+						if using > 1 {
+							sampled.Info().
+								Int32("currently-using", using).
+								Msg("sampling message handling")
 						}
+
```
daywednes, line 459: ridjlasjkldjsl

rlan35:
Please add some description to the PR.

node/node.go
```diff
@@ -350,154 +313,273 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
+		utils.Logger().Info().
+			Str("topic", topicNamed).
+			Msg("enabled topic validation on consensus bound messages")
+
+		if err := pubsub.RegisterTopicValidator(
+			topicNamed,
+			func(ctx context.Context, peer libp2p_peer.ID, msg *libp2p_pubsub.Message) bool {
+				hmyMsg := msg.GetData()
+				if len(hmyMsg) < p2pMsgPrefixSize {
+					pubsub.BlacklistPeer(peer)
```
rlan35, line 454: is it possible that valid peers are blacklisted by mistake? How long the blacklisting last?

fxfactorial: In principle no - because harmony protocol respecting nodes will never do this. 

What time period you'd like for the blacklisting to last?

node/node_handler.go
```diff
@@ -55,95 +55,130 @@ func (node *Node) processSkippedMsgTypeByteValue(
...
+) error {
 
-	msgPayload, err := proto.GetMessagePayload(content)
-	if err != nil {
-		utils.Logger().Error().
-			Err(err).
-			Msg("HandleMessage get message payload failed")
-		return
-	}
+	return nil
```
rlan35, line 62: return nil?

node/node.go
rlan35: If it's mostly for spamming. 10 mins should be enough.

node/node_handler.go
fxfactorial: means success, errors now captured in values, before these APIs didn't return errors

node/node.go
fxfactorial: k, I set it to 15 already in pushed commit , 10 or 15?

node/node_handler.go
fxfactorial: but actualy ya - now I remember, I forgot, this one should be explicit error since something should have been handled explicitly but then I'd need to add returns more and I punted.

node/node.go
fxfactorial: actually commented on this - the time based one is bad to use https://github.com/libp2p/go-libp2p-pubsub/issues/338

node/node_handler.go
rlan35: I mean this func just directly returns nil. The below logic is never executed..

node/node_handler.go
```diff
@@ -55,95 +55,130 @@ func (node *Node) processSkippedMsgTypeByteValue(
...
 
-	switch msgCategory {
-	case proto.Consensus:
-		msgPayload, _ := proto.GetConsensusMessagePayload(content)
-		if node.NodeConfig.Role() == nodeconfig.ExplorerNode {
-			node.ExplorerMessageHandler(msgPayload)
-		} else {
-			node.ConsensusMessageHandler(msgPayload)
+	switch protonode.Transaction {
```
rlan35, line 64: this is switching on a constant?

node/node_handler.go
fxfactorial: Ya it’s leftover code , I just return nil to not have to comment out , it’s handled else where

node/node_handler.go
fxfactorial: Yep

fxfactorial: This is dead code , will remove

node/node_handler.go
rlan35: I also couldn't find the reference on this func "handleClientMessage"

rlan35: Seems duplicate to HandleNodeMessage

fxfactorial: It will make sense , look at node.go only first

Issue #2976, opened 2020-05-11
[firedrill] dryrun network one shard recovery


LeoHChen:
# Goal

we need to do a fire drill of one shard recovery on the dryrun network. The fire drill will start at 8 pm Monday PDT (May 11th).

We will need to test a few scenarios.

## one shard recovery with graceful shutdown
- [x] graceful shutdown of the leader in shard3
- [x] graceful shutdown validators in shard3 until the consensus is stuck

## one shard recovery with crash/kill
- [x] kill the leader on shard2
- [x] kill validators in shard2 until the consensus is stuck 

## one shard recovery with the graceful shutdown of the beacon chain
- [x] graceful shutdown of the leader in shard0
- [x] graceful shutdown validators in shard0 until the consensus is stuck

## one shard recovery with crash/kill of the beacon chain
- [x] kill the leader on shard0
- [x] kill validators in shard0 until the consensus is stuck

LeoHChen:
@Daniel-VDM and @SebastianJ  will lead the shard recovery fire drill. We can use /zoom-eng to coordinate.

LeoHChen:
![image](https://user-images.githubusercontent.com/4513927/81634194-5740ed00-93c3-11ea-9bc6-d5a110b4c391.png)

LeoHChen:
![image](https://user-images.githubusercontent.com/4513927/81634259-73448e80-93c3-11ea-84b2-a682d896bbbf.png)

LeoHChen:
![image](https://user-images.githubusercontent.com/4513927/81634284-80fa1400-93c3-11ea-85fd-6aa16e06cb6d.png)

Daniel-VDM:
Found panic on nodes:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xd5217e]
goroutine 1634 [running]:
github.com/harmony-one/harmony/consensus.(*Consensus).onViewChange(0xc000472280, 0xc00947abe0)
/home/lc/go/src/github.com/harmony-one/harmony/consensus/view_change.go:199 +0x53e
github.com/harmony-one/harmony/consensus.(*Consensus).handleMessageUpdate(0xc000472280, 0xc0096141c6, 0x19b, 0x1ba)
/home/lc/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:91 +0x254
[...]

PR #2986, opened 2020-05-12
Fix commit payload

rlan35:
The commit signature payload was changed with a forking logic based on epoch.

The original code used data from somewhere not directly from the block data, which can cause issues if those data are inconsistent.

Should always directly use the block's data for commit payload.

Also:
1. Add commit signature into block explorer commit
2. Remove consensus.epoch

Tested locally, consensus reached.

Fixing: https://github.com/harmony-one/harmony/issues/2976

### DIFF
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -206,7 +206,6 @@ func (consensus *Consensus) String() string {
 		consensus.blockNum,
 		consensus.viewID,
 		consensus.ShardID,
-		consensus.epoch,
 	)
 }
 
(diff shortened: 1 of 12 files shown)


consensus/consensus_service.go
```diff
@@ -199,7 +199,7 @@ func (consensus *Consensus) String() string {
 	}
 
 	return fmt.Sprintf(
-		"[Duty:%s Pub:%s Header:%s Num:%d View:%d Shard:%d Epoch:%d]",
```
fxfactorial, line 202: Remove this whole string method please , it has race condition and it’s not needed

rlan35: Done.

PR #2997, opened 2020-05-13
Make leader commit m1payload itself

rlan35:
Hit a view change stuck issue in https://github.com/harmony-one/harmony/issues/2976 . The issue is that leader added M1 msg itself, while no one else votes on M1 (prepared)
image

But Leader didn't put m1payload itself, which caused the newView msg validation failure on validators. `[onNewView] M1 (prepared) Type Payload Not Have Enough Length"`

Solution is to let the leader itself add m1payload when itself added M1 signatures.

Tested locally.

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -195,16 +195,29 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 			msg_pb.MessageType_PREPARED, recvMsg.BlockNum,
 		)
 		preparedMsg := consensus.FBFTLog.FindMessageByMaxViewID(preparedMsgs)
-
-		if preparedMsg != nil && consensus.FBFTLog.GetBlockByHash(
-			preparedMsg.BlockHash,
-		) != nil {
+		hasBlock := false
+		if preparedMsg != nil {
+			if preparedBlock := consensus.FBFTLog.GetBlockByHash(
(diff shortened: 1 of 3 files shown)


consensus/view_change.go
```diff
@@ -357,7 +370,7 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 	consensus.viewIDBitmap[recvMsg.ViewID].SetKey(recvMsg.SenderPubkey, true)
 	consensus.getLogger().Debug().
 		Int("have", len(consensus.viewIDSigs[recvMsg.ViewID])).
-		Int64("needed", consensus.Decider.TwoThirdsSignersCount()).
+		Int64("total", consensus.Decider.ParticipantsCount()).
```
fxfactorial, line 373: this is race conidtion , this field prefer rid it

rlan35: The whole onViewChange is locked. I will keep it for debugging purpose.

rlan35: consensus.vcLock.Lock()

fxfactorial: k, anyway later will replace the map underlying it with a concurrent trie

rlan35:
Tested in dryrun network and this corner case view change issue that cause view change to stuck is fixed!

PR #3009, opened 2020-05-13
[rpc] add viewID and epoch to RPCMarshalBlock

ivorytowerdds:
want to get the number of block numbers in each epoch for non-beacon shards, so add epoch number to `hmy_getBlockByNumber`.

### DIFF
--- a/internal/hmyapi/apiv1/types.go
+++ b/internal/hmyapi/apiv1/types.go
@@ -353,6 +353,8 @@ func newRPCStakingTransaction(tx *types2.StakingTransaction, blockHash common.Ha
 // RPCBlock represents a block that will serialize to the RPC representation of a block
 type RPCBlock struct {
 	Number           *hexutil.Big     `json:"number"`
+	ViewID           *hexutil.Big     `json:"viewID"`
+	Epoch            *hexutil.Big     `json:"epoch"`
 	Hash             common.Hash      `json:"hash"`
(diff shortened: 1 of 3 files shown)


internal/hmyapi/apiv1/types.go
```diff
@@ -353,6 +353,8 @@ func newRPCStakingTransaction(tx *types2.StakingTransaction, blockHash common.Ha
 // RPCBlock represents a block that will serialize to the RPC representation of a block
 type RPCBlock struct {
 	Number           *hexutil.Big     `json:"number"`
+	ViewID           *hexutil.Big     `json:"viewID"`
```
gupadhyaya, line 356: why do you need these here? It is not used right? only sufficient to add to RPCMarshalBlock?

ivorytowerdds: yes, it's not used in RPCMarshalBlock and other places. But seems like we record all the other fields in RPCMarshalBlock here, so I add them here too. Do we want to keep the type RPCBlock?

gupadhyaya: ok. later rpc revamp will take care of this then.

Daniel-VDM:
Jenkins, test this please

ivorytowerdds:
> Jenkins, test this please

how to restart the jenkins again?

gupadhyaya:
@ivorytowerdds do you still need this PR?

PR #3019, opened 2020-05-14
[WIP][staking] non-beacon chain relay staking txn

### LINKED ISSUE #3017: [enhancement] staking tx shard agnostic
**Describe the bug**

To make hmy shard-agnostic, if the staking tx is received by a node on a non-beacon chain shard (currently shard #1, #2 or #3), it should broadcast this transaction to shard 0. This prevents awkward situations, such as the validator not being able to use his own node to send staking transactions.

JackyWYX:
## Issue

Closes #3017. When handling staking transaction, non-beacon chain nodes will relay the message of staking transaction to beacon chain. 

Currently doing local test.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -234,9 +234,8 @@ func (node *Node) addPendingTransactions(newTxs types.Transactions) []error {
 
 // Add new staking transactions to the pending staking transaction list.
 func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTransactions) []error {
-	if node.NodeConfig.ShardID == shard.BeaconChainShardID &&
-		node.Blockchain().Config().IsPreStaking(node.Blockchain().CurrentHeader().Epoch()) {
-		poolTxs := types.PoolTransactions{}
+	if node.NodeConfig.ShardID == shard.BeaconChainShardID && node.isPreStakingEpoch() {
+		poolTxs := make(types.PoolTransactions, 0, len(newStakingTxs))
 		for _, tx := range newStakingTxs {
 			poolTxs = append(poolTxs, tx)
 		}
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -256,19 +255,28 @@ func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTra
...
-		utils.Logger().Info().Str("Hash", newStakingTx.Hash().Hex()).Msg("Broadcasting Staking Tx")
-		node.tryBroadcastStaking(newStakingTx)
 	}
+	utils.Logger().Info().Str("Hash", newStakingTx.Hash().Hex()).Msg("Broadcasting Staking Tx")
+	node.tryBroadcastStaking(newStakingTx)
 	return nil
 }
 
+func (node *Node) isPreStakingEpoch() bool {
+	return node.Blockchain().Config().IsPreStaking(node.Blockchain().CurrentHeader().Epoch())
```
Daniel-VDM, line 277: Would it be better to use Beaconchain?

node/node.go
```diff
@@ -234,9 +234,8 @@ func (node *Node) addPendingTransactions(newTxs types.Transactions) []error {
 
 // Add new staking transactions to the pending staking transaction list.
 func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTransactions) []error {
-	if node.NodeConfig.ShardID == shard.BeaconChainShardID &&
-		node.Blockchain().Config().IsPreStaking(node.Blockchain().CurrentHeader().Epoch()) {
-		poolTxs := types.PoolTransactions{}
+	if node.NodeConfig.ShardID == shard.BeaconChainShardID && node.isPreStakingEpoch() {
+		poolTxs := make(types.PoolTransactions, 0, len(newStakingTxs))
```
fxfactorial, line 238: Do not preallocate . This is wrong premature optimization

node/node.go
```diff
@@ -256,19 +255,28 @@ func (node *Node) addPendingStakingTransactions(newStakingTxs staking.StakingTra
 func (node *Node) AddPendingStakingTransaction(
 	newStakingTx *staking.StakingTransaction,
 ) error {
+	if !node.isPreStakingEpoch() {
+		utils.Logger().Info().Str("Hash", newStakingTx.Hash().Hex()).
```
fxfactorial, line 259: There already is a .hex log thing I think

node/node.go
fxfactorial: Need to use beaconchain

JackyWYX: Sorry don't quite get it. New logic for `AddPendingStakingTransaction`:

```
if is not pre staking:
    do nothing
elif node is on beacon chain:
    Add to local tx pool
    broadcast staking tx
elif not on beacon chain:
    broadcast staking tx to beacon chain
```

Is there any logic wrong in this pseudocode?

node/node.go
JackyWYX: Done. But could you tell me a bit about why it's wrong?

node/node.go
JackyWYX: This is newly added log. In previous logic, a node not on beacon chain / not in staking epoch received staking transaction will silently return a nil error. A logging only happens when the node is broadcasting the txn. So is it true that this info doesn't have enough value to be logged?

node/node.go
Daniel-VDM: epochs are only defined by the beacon chain for staking era. `node.Blockchain()` can be a shard chain and thus be wrong.

node/node.go
fxfactorial: You are trying to be smarter than compiler when it’s not necessary and preallocate lots of space bc this is txn pool which could potentially be big. Append is heavily optimized and pretty damn good on own.

JackyWYX: I am sorry I am little bit confused. According to my knowledge, append is done through double extending. In each double extending, a new sized array is allocated. So suppose txs size is `n` and it is big. If we allocate at the beginning, we allocate `n` pointers total. But if we do not use append, we allocate `2n` pointers in total, which is a waste in performace. Am I understanding this correctly?

node/node.go
fxfactorial: No you aren’t. You don’t need to make these kinds of assumptions in a garbage collected language and it’s not necessary to think this deeply over this. Focus on business logic please

node/node.go
fxfactorial: Member
No you aren’t. You don’t need to make these kinds of assumptions in a garbage collected language and it’s not necessary to think this deeply over this. Focus on business logic please

JackyWYX: Is there any related reading materials for recommendation for this topic?

node/node.go
JackyWYX: Removed

JackyWYX:
The feature is already implemented in go-sdk. Closing.

Issue #3022, opened 2020-05-15
[panic] nil pointer returned by GetHeaderByNumber, GetHeader wasn't checked


LeoHChen:
When I was handling the out of sync ticket, found the following panic

```
	May 15 13:07:19 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: 2020/05/15 13:07:19 http: Accept error: accept tcp [::]:9500: accept4: too many open files;
May 15 13:07:19 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: 2020/05/15 13:07:19 http: Accept error: accept tcp [::]:9500: accept4: too many open files;
May 15 13:07:19 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: 2020/05/15 13:07:19 http: Accept error: accept tcp [::]:9500: accept4: too many open files;
May 15 13:07:20 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: 2020/05/15 13:07:20 http: Accept error: accept tcp [::]:9500: accept4: too many open files;
May 15 13:08:06 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: panic: runtime error: invalid memory address or nil pointer dereference
May 15 13:08:06 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb26415]
May 15 13:08:06 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: goroutine 937 [running]:
May 15 13:08:06 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: github.com/harmony-one/harmony/core.(*BlockChain).WriteBlockWithState(0xc000bf6000, 0xc00f8d
May 15 13:08:06 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: /mnt/jenkins/workspace/harmony-release/harmony/core/blockchain.go:1140 +0x545
May 15 13:08:06 ip-172-31-33-32.eu-west-1.compute.internal node.sh[25432]: github.com/harmony-one/harmony/core.(*BlockChain).insertChain(0xc000bf6000, 0xc01f146f50, 0x
[...]

PR #3030, opened 2020-05-16
Add nil checks on all GetHeader* func.

rlan35:
Fix : https://github.com/harmony-one/harmony/issues/3022

### DIFF
--- a/internal/chain/engine.go
+++ b/internal/chain/engine.go
@@ -188,6 +188,11 @@ func (e *engineImpl) VerifySeal(chain engine.ChainReader, header *block.Header) 
 	}
 	parentHash := header.ParentHash()
 	parentHeader := chain.GetHeader(parentHash, header.Number().Uint64()-1)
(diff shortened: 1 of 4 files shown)


internal/chain/engine.go
```diff
@@ -188,6 +188,11 @@ func (e *engineImpl) VerifySeal(chain engine.ChainReader, header *block.Header)
 	}
 	parentHash := header.ParentHash()
 	parentHeader := chain.GetHeader(parentHash, header.Number().Uint64()-1)
+	if parentHeader == nil {
+		return errors.New(
+			"[VerifySeal] no parent header found",
```
fxfactorial, line 193: might as well make it actually useful if we're doing this, `errors.WithStack(errors.Errorfparent for %s at height %d, ))`

rlan35: ack. can do next PR.

PR #3032, opened 2020-05-17
Optimize crosslink verification logic and add more cache

rlan35:
mainnet shard 0 leader is choking on CPU.

Mostly due to crosslink verification and processing is expensive, and there are a lot of them (redundant ones too).

Add caches on the processing and related components.

Also make crosslink batch smaller, as most of them are just duplicated data. (because of the delay of beacon sync in shard chains)

Also avoid repeated serialization of the voter bls keys in voting power computation, which happens for every validator votes!

### DIFF
--- a/consensus/quorum/one-node-staked-vote.go
+++ b/consensus/quorum/one-node-staked-vote.go
@@ -2,6 +2,7 @@ package quorum
 
 import (
 	"encoding/json"
+	errors2 "errors"
 	"math/big"
 
 	"github.com/harmony-one/harmony/consensus/votepower"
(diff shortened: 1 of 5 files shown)


consensus/quorum/one-node-staked-vote.go
```diff
@@ -2,6 +2,7 @@ package quorum
 
 import (
 	"encoding/json"
+	errors2 "errors"
```
fxfactorial, line 5: no errors2, just find usages of `pkg/errors` , use that instead of builtin errors please

fxfactorial:
Things under consensus/* could be skipped in this PR because subsumed in compute power available via other PR (protect consensus)

rlan35:
Tested locally. crosslinks and consensus are working fine.

rlan35:
> Things under consensus/* could be skipped in this PR because subsumed in compute power available via other PR (protect consensus)

Ok, But I prefer keep it. if this change is correct, better deploy it now then wait for later fix.

PR #3057, opened 2020-05-20
Transaction api improvements

Daniel-VDM:
## Issue

Currently, we don't check if transactions have been finalized before adding it to tx pool.
Moreover, we don't broadcast failed tx which makes the error sink inconsistent across all nodes. However, to protect the network from needless tx spam in the case of an attack, one can enable the option to not broadcast failed txs. 

## Test

Tested locally by sending a bunch of invalid txs to all nodes on localnet, error sink stayed synced. 
Also tested replay of same tx, error sink didn't report and saw the warning in logs that tx was not added to pool. 

### Unit Test Coverage

Before:

```
PASS
coverage: 11.3% of statements
ok      github.com/harmony-one/harmony/node     1.515s

```

After:

```
PASS
coverage: 11.2% of statements
ok      github.com/harmony-one/harmony/node     1.565s

```

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -110,8 +110,9 @@ var (
 	revertTo       = flag.Int("revert_to", 0, "The revert will rollback all blocks until and including block number revert_to")
 	revertBeacon   = flag.Bool("revert_beacon", false, "Whether to revert beacon chain or the chain this node is assigned to")
 	// Blacklist of addresses
-	blacklistPath   = flag.String("blacklist", "./.hmy/blacklist.txt", "Path to newline delimited file of blacklisted wallet addresses")
-	webHookYamlPath = flag.String(
+	blacklistPath      = flag.String("blacklist", "./.hmy/blacklist.txt", "Path to newline delimited file of blacklisted wallet addresses")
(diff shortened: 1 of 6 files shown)


cmd/harmony/main.go
```diff
@@ -110,8 +110,9 @@ var (
 	revertTo       = flag.Int("revert_to", 0, "The revert will rollback all blocks until and including block number revert_to")
 	revertBeacon   = flag.Bool("revert_beacon", false, "Whether to revert beacon chain or the chain this node is assigned to")
 	// Blacklist of addresses
-	blacklistPath   = flag.String("blacklist", "./.hmy/blacklist.txt", "Path to newline delimited file of blacklisted wallet addresses")
-	webHookYamlPath = flag.String(
+	blacklistPath      = flag.String("blacklist", "./.hmy/blacklist.txt", "Path to newline delimited file of blacklisted wallet addresses")
+	broadcastInvalidTx = flag.Bool("broadcast_invalid_tx", true, "Broadcast invalid transactions to sync pool state (default: true)")
```
rlan35, line 114: better make default to false.

Daniel-VDM: changed

PR #3066, opened 2020-05-21
[apr] fix apr for epoch 186 so that average is correctly computed

gupadhyaya:
fix apr for epoch 186 at the time of computing average. this code will be remove upon going past 30 epochs from staking epoch.


LeoHChen:
btw, please rebase your code, your branch is out of date.

gupadhyaya:
ok... there is better logic to update the stats, hence closing this PR and will open a new one.

PR #3079, opened 2020-05-22
Amount should be more than MinSelfDelegation.

zouguangxian:


LeoHChen:

gupadhyaya:
@zouguangxian no need for this check as it is done in `if err := wrapper.SanityCheck(maxBLSKeyAllowed); err != nil`

core/staking_verifier.go
```diff
@@ -87,6 +87,9 @@ func VerifyAndCreateValidatorFromMsg(
 	if msg.Amount.Sign() == -1 {
 		return nil, errNegativeAmount
 	}
+	if msg.Amount.Cmp(msg.MinSelfDelegation) == -1 {
+		return nil, errInsufficientBalanceForStake
```
rlan35, line 91: Should use this error: ErrInvalidSelfDelegation

rlan35: And also true for @gupadhyaya 's comment.

gupadhyaya:
no longer required, hence closing.

PR #3080, opened 2020-05-22
[RPC] Add RPCs to get list of BLS public keys for committees

janet-harmony:
- Fix `isBlockGreaterThanLatest` to not error when using "latest" block
- Revert change to `hmy_getValidators` to return error when called on not beacon shard
- Fix `hmy_getAllValidatorInformationByBlockNumber`, which gets list of all validators currently on the chain & would return an "validator not found" error for all data if querying a block number that any one validator was not created by
- Add `hmy_getValidatorKeys` to get BLS pub keys for epoch committee (need for P3 reward calculation)
- Add `hmy_getBlockSignerKeys` to get BLS pub keys for specific block (need for P3 reward calcluation)


gupadhyaya:
@janet-harmony can you split into three separate PRs? Also, this https://github.com/harmony-one/harmony/pull/3081 addresses partly?

internal/hmyapi/apiv1/blockchain.go
```diff
@@ -685,10 +740,9 @@ func (s *PublicBlockChainAPI) getAllValidatorInformation(
 	}
 	for i := start; i < start+validatorsNum; i++ {
 		information, err := s.b.GetValidatorInformation(addresses[i], block)
-		if err != nil {
-			return nil, err
+		if err == nil {
+			validators[i-start] = information
```
gupadhyaya, line 744: instead of making a fixed length slice and leaving some spots empty if validator not found for particular block, make the slice literal and append to it.

internal/hmyapi/apiv1/blockchain.go
```diff
@@ -71,7 +71,7 @@ func (s *PublicBlockChainAPI) isBlockGreaterThanLatest(blockNum rpc.BlockNumber)
 	// since they are never greater than latest
 	if blockNum != rpc.PendingBlockNumber &&
 		blockNum != rpc.LatestBlockNumber &&
-		uint64(blockNum) > s.b.CurrentBlock().NumberU64() {
+		uint64(blockNum) > (s.b.CurrentBlock().NumberU64()+1) {
```
gupadhyaya, line 74: why do you need +1?

internal/hmyapi/apiv1/blockchain.go
```diff
@@ -281,6 +292,50 @@ func (s *PublicBlockChainAPI) GetBlockSigners(ctx context.Context, blockNr rpc.B
 	return result, nil
 }
 
+// GetBlockSignerKeys returns bls public keys that signed the block.
+func (s *PublicBlockChainAPI) GetBlockSignerKeys(ctx context.Context, blockNr rpc.BlockNumber) ([]string, error) {
+	if uint64(blockNr) == 0 {
```
gupadhyaya, line 297: can you refactor common code out of `GetBlockSigners` and this?

internal/hmyapi/apiv1/blockchain.go
```diff
@@ -71,7 +71,7 @@ func (s *PublicBlockChainAPI) isBlockGreaterThanLatest(blockNum rpc.BlockNumber)
 	// since they are never greater than latest
 	if blockNum != rpc.PendingBlockNumber &&
 		blockNum != rpc.LatestBlockNumber &&
-		uint64(blockNum) > s.b.CurrentBlock().NumberU64() {
+		uint64(blockNum) > (s.b.CurrentBlock().NumberU64()) {
```
gupadhyaya, line 74: no need for this change.

internal/hmyapi/apiv1/blockchain.go
```diff
@@ -217,6 +214,20 @@ func (s *PublicBlockChainAPI) GetValidators(ctx context.Context, epoch int64) (m
 	return result, nil
 }
 
+// GetValidatorKeys returns list of bls public keys in the committee for a particular epoch.
+func (s *PublicBlockChainAPI) GetValidatorKeys(ctx context.Context, epoch int64) ([]string, error) {
```
gupadhyaya, line 218: can you implement this function in the backend and api1 and api2 can directly call and return it?

internal/hmyapi/apiv1/blockchain.go
```diff
@@ -260,61 +264,78 @@ func (s *PublicBlockChainAPI) GetBlockSigners(ctx context.Context, blockNr rpc.B
...
+	}
+	return committee.Slots, mask, nil
+}
+
+// GetBlockSigners returns signers for a particular block.
+func (s *PublicBlockChainAPI) GetBlockSigners(ctx context.Context, blockNr rpc.BlockNumber) ([]string, error) {
+	if uint64(blockNr) == 0 {
+		return []string{}, nil
+	}
+	if err := s.isBlockGreaterThanLatest(blockNr); err != nil {
```
gupadhyaya, line 281: move this inside `getBlockSigners` as no behavior is changed.

internal/hmyapi/apiv2/blockchain.go
```diff
@@ -65,7 +65,7 @@ func (s *PublicBlockChainAPI) isBeaconShard() error {
 }
 
 func (s *PublicBlockChainAPI) isBlockGreaterThanLatest(blockNum uint64) error {
-	if blockNum > s.b.CurrentBlock().NumberU64() {
+	if blockNum > (s.b.CurrentBlock().NumberU64() + 1) {
```
gupadhyaya, line 68: +1 needed? I think no change is needed.

internal/hmyapi/apiv2/blockchain.go
```diff
@@ -179,6 +176,20 @@ func (s *PublicBlockChainAPI) GetValidators(ctx context.Context, epoch int64) (m
 	return result, nil
 }
 
+// GetValidatorKeys returns list of bls public keys in the committee for a particular epoch.
+func (s *PublicBlockChainAPI) GetValidatorKeys(ctx context.Context, epoch int64) ([]string, error) {
```
gupadhyaya, line 180: no need for this after implementing this in backend.

internal/hmyapi/apiv2/blockchain.go
```diff
@@ -222,61 +226,82 @@ func (s *PublicBlockChainAPI) GetBlockSigners(ctx context.Context, blockNr uint6
...
+	}
+	return committee.Slots, mask, nil
+}
+
+// GetBlockSigners returns signers for a particular block.
+func (s *PublicBlockChainAPI) GetBlockSigners(ctx context.Context, blockNr uint64) ([]string, error) {
+	if blockNr == 0 || blockNr >= uint64(s.BlockNumber()) {
+		return []string{}, nil
+	}
+	if err := s.isBlockGreaterThanLatest(blockNr); err != nil {
```
gupadhyaya, line 247: same as apiv1, can be moved inside `getBlockSigners`

internal/hmyapi/apiv2/blockchain.go
janet-harmony: Requesting current block + 1 is valid in some cases.

internal/hmyapi/apiv1/blockchain.go
janet-harmony: Moved to the API backend, will keep the check in each function, because the checks are different for V1 & V2.

internal/hmyapi/apiv2/blockchain.go
gupadhyaya: I am not comfortable allowing blockNr+1, unless all callsites of `isBlockGreaterThanLatest` have proper checks for block to be nil when extracted using `BlockByNumber`. For example, `GetDelegationsByDelegatorByBlockNumber` if the block does not exists, 
`block, err := s.b.BlockByNumber(ctx, rpc.BlockNumber(blockNum))` will not return any error instead block is nil, but you check
`if err != nil {`, which will not stop a nil block from getting past this check and crash with nil pointer somewhere down. Can you please fix all the callsites for such potential nil block situations? Also, why is allowing blockNr+1 necessary?

PR #3086, opened 2020-05-23
Add quick unlock feature at epoch 190

rlan35:
Make unlocking of undelegated token to happen right at the end of current epoch.

Based on proposal in https://talk.harmony.one/t/proposal-shortening-of-undelegation-locking-period/634

Tested locally.

Will need a rolling upgrade before the QuickUnlockEpoch.


rlan35:
Jenkins, test this please.

PR #3092, opened 2020-05-24
[consensus] revert the grace period change

LeoHChen:
use the original go routine instead of the time.AfterFunc call


LeoHChen:
not needed

PR #3111, opened 2020-05-29
Remove unnecessary block verification

rlan35:
The block is fetched from fbftLog and it was already verified before it's added to the log. So we don't need to verify it again.

This actually slows down the leader by 1-2s in beacon chain because the block verification of crosslinks is very expensive.

tested locally.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -90,6 +90,7 @@ type Consensus struct {
 	// whether to ignore viewID check
 	ignoreViewIDCheck bool
 	// global consensus mutex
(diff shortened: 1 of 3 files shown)


rlan35:
> On new view handling, `blockVerifier` is handled after adding PBFT. Will this cause problem?
> 
> `consensus/view_change.go:586`
> `consensus/view_change.go:608`

It's better to be verified before adding to log in view change, but since the multi-sig is verified first in the newView msg, it's trusting that the signatures are enough to guarantee the block is valid too. Not a big deal. But you are right, it's always better to verify before adding to log.  Added a todo.

PR #3114, opened 2020-05-29
[libp2p] validate pubsub messages

LeoHChen:
This PR is mostly a subset of the original PR of https://github.com/harmony-one/harmony/pull/2984
with a few minor fixes based on review feedback.

Authored-by: Edgar Aroutiounian <edgar.factorial@gmail.com >

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"time"
 
+	"github.com/harmony-one/bls/ffi/go/bls"
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
 	"github.com/harmony-one/harmony/block"
 	"github.com/harmony-one/harmony/consensus/quorum"
@@ -18,8 +19,12 @@ import (
 	"github.com/pkg/errors"
 )
 
+var (
+	errSenderPubKeyNotLeader = errors.New("sender pubkey doesn't match leader")
+)
+
 // HandleMessageUpdate will update the consensus state according to received message
-func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb.Message) error {
+func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb.Message, senderKey *bls.PublicKey) error {
 	// when node is in ViewChanging mode, it still accepts normal messages into FBFTLog
 	// in order to avoid possible trap forever but drop PREPARE and COMMIT
 	// which are message types specifically for a node acting as leader
@@ -36,17 +41,24 @@ func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb
 
 	switch t := msg.Type; true {
 	// Handle validator intended messages first
-	case t == msg_pb.MessageType_ANNOUNCE &&
-		intendedForValidator &&
-		consensus.validatorSanityChecks(msg):
+	case t == msg_pb.MessageType_ANNOUNCE && intendedForValidator:
+
+		if !senderKey.IsEqual(consensus.LeaderPubKey) &&
+			consensus.current.Mode() == Normal && !consensus.ignoreViewIDCheck {
+			return errSenderPubKeyNotLeader
+		}
 		consensus.onAnnounce(msg)
-	case t == msg_pb.MessageType_PREPARED &&
-		intendedForValidator &&
-		consensus.validatorSanityChecks(msg):
+	case t == msg_pb.MessageType_PREPARED && intendedForValidator:
+		if !senderKey.IsEqual(consensus.LeaderPubKey) &&
+			consensus.current.Mode() == Normal && !consensus.ignoreViewIDCheck {
+			return errSenderPubKeyNotLeader
+		}
 		consensus.onPrepared(msg)
-	case t == msg_pb.MessageType_COMMITTED &&
-		intendedForValidator &&
-		consensus.validatorSanityChecks(msg):
+	case t == msg_pb.MessageType_COMMITTED && intendedForValidator:
+		if !senderKey.IsEqual(consensus.LeaderPubKey) &&
+			consensus.current.Mode() == Normal && !consensus.ignoreViewIDCheck {
+			return errSenderPubKeyNotLeader
+		}
 		consensus.onCommitted(msg)
 	// Handle leader intended messages now
 	case t == msg_pb.MessageType_PREPARE &&
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -2,10 +2,10 @@ package consensus
 
 import (
 	"bytes"
+	"context"
 	"encoding/hex"
 	"time"
 
-	protobuf "github.com/golang/protobuf/proto"
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
 	"github.com/harmony-one/harmony/block"
 	"github.com/harmony-one/harmony/consensus/quorum"
@@ -18,45 +18,16 @@ import (
 	"github.com/pkg/errors"
 )
 
-// handlemessageupdate will update the consensus state according to received message
-func (consensus *Consensus) handleMessageUpdate(payload []byte) {
-	if len(payload) == 0 {
-		return
-	}
-	msg := &msg_pb.Message{}
-	if err := protobuf.Unmarshal(payload, msg); err != nil {
-		consensus.getLogger().Error().Err(err).Msg("Failed to unmarshal message payload.")
-		return
-	}
-
+// HandleMessageUpdate will update the consensus state according to received message
+func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb.Message) error {
 	// when node is in ViewChanging mode, it still accepts normal messages into FBFTLog
 	// in order to avoid possible trap forever but drop PREPARE and COMMIT
 	// which are message types specifically for a node acting as leader
+	// so we just ignore those messages
 	if (consensus.current.Mode() == ViewChanging) &&
 		(msg.Type == msg_pb.MessageType_PREPARE ||
 			msg.Type == msg_pb.MessageType_COMMIT) {
-		return
-	}
-
-	if msg.Type == msg_pb.MessageType_VIEWCHANGE ||
-		msg.Type == msg_pb.MessageType_NEWVIEW {
-		if msg.GetViewchange() != nil &&
-			msg.GetViewchange().ShardId != consensus.ShardID {
-			consensus.getLogger().Warn().
-				Uint32("myShardId", consensus.ShardID).
-				Uint32("receivedShardId", msg.GetViewchange().ShardId).
-				Msg("Received view change message from different shard")
-			return
-		}
-	} else {
-		if msg.GetConsensus() != nil &&
-			msg.GetConsensus().ShardId != consensus.ShardID {
-			consensus.getLogger().Warn().
-				Uint32("myShardId", consensus.ShardID).
-				Uint32("receivedShardId", msg.GetConsensus().ShardId).
-				Msg("Received consensus message from different shard")
-			return
-		}
+		return nil
 	}
 
 	intendedForValidator, intendedForLeader :=
@@ -79,20 +50,17 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
 		consensus.onCommitted(msg)
 	// Handle leader intended messages now
 	case t == msg_pb.MessageType_PREPARE &&
-		intendedForLeader &&
-		consensus.leaderSanityChecks(msg):
+		intendedForLeader:
 		consensus.onPrepare(msg)
 	case t == msg_pb.MessageType_COMMIT &&
-		intendedForLeader &&
-		consensus.leaderSanityChecks(msg):
+		intendedForLeader:
 		consensus.onCommit(msg)
-	case t == msg_pb.MessageType_VIEWCHANGE &&
-		consensus.viewChangeSanityCheck(msg):
+	case t == msg_pb.MessageType_VIEWCHANGE:
 		consensus.onViewChange(msg)
-	case t == msg_pb.MessageType_NEWVIEW &&
-		consensus.viewChangeSanityCheck(msg):
+	case t == msg_pb.MessageType_NEWVIEW:
 		consensus.onNewView(msg)
 	}
+	return nil
 }
 
 func (consensus *Consensus) finalizeCommits() {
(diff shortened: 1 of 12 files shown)


LeoHChen:
ported from https://github.com/harmony-one/harmony/pull/2984

LeoHChen:
upgrade LRTN with the engineering build of this PR.

image

LeoHChen:
Also launched regression test on LRTN. 

https://jenkins.harmony.one/job/regression_test/48/console

consensus/consensus_v2.go
```diff
@@ -18,24 +18,22 @@ import (
...
-	if len(payload) == 0 {
-		return
-	}
-	msg := &msg_pb.Message{}
-	if err := protobuf.Unmarshal(payload, msg); err != nil {
-		consensus.getLogger().Error().Err(err).Msg("Failed to unmarshal message payload.")
-		return
-	}
+var (
+	errInVCNotLeader    = errors.New("something vc not leader")
```
rlan35, line 22: These errors messages are just temp placeholders. Please update with more accurate message.

consensus/consensus_v2.go
```diff
@@ -18,24 +18,22 @@ import (
...
 	// in order to avoid possible trap forever but drop PREPARE and COMMIT
 	// which are message types specifically for a node acting as leader
 	if (consensus.current.Mode() == ViewChanging) &&
 		(msg.Type == msg_pb.MessageType_PREPARE ||
 			msg.Type == msg_pb.MessageType_COMMIT) {
-		return
+		return errors.WithStack(errInVCNotLeader)
 	}
 
 	if msg.Type == msg_pb.MessageType_VIEWCHANGE ||
```
gupadhyaya, line 39: we could simplify this if-if-else-if into simple conditions:
```
isDifferentShardConsensusMsg := msg.GetConsensus() != nil && msg.GetConsensus().ShardId != consensus.ShardID
if isDifferentShardConsensusMsg {

}
isDifferentShardViewChangeMsg := msg.GetViewchange() != nil && msg.GetViewchange().ShardId != consensus.ShardID
if isDifferentShardViewChangeMsg {

}
```

consensus/consensus_v2.go
```diff
@@ -343,7 +341,7 @@ func (consensus *Consensus) Start(
 		}
 		consensus.getLogger().Info().Time("time", time.Now()).Msg("[ConsensusMainLoop] Consensus started")
 		defer close(stoppedChan)
-		ticker := time.NewTicker(3 * time.Second)
+		ticker := time.NewTicker(time.Second)
```
gupadhyaya, line 344: I recommend leaving a comment as to why 1 sec ticker is chosen which was previously 3 seconds?

rlan35: I would keep the 3 seconds. I don't see clear good reason to shorten it now.

node/node.go
```diff
@@ -363,66 +367,277 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
+		senderPubKeyViaWire = maybeVC.GetSenderPubkey()
+	} else {
+		return nil, errors.WithStack(errNoSenderPubKey)
+	}
+
+	if len(senderPubKeyViaWire) != shard.PublicKeySizeInBytes {
+		return nil, errors.WithStack(errNotRightKeySize)
+	}
+
+	key, err := bls_cosi.BytesToBLSPublicKey(
```
rlan35, line 406: Please add a TODO here to add singleFlight cache, the deserialization of BLS key is very expensive.

p2p/host.go
```diff
@@ -52,16 +48,14 @@ type Peer struct {
...
-		BLSPubKey = p.ConsensusPubKey.SerializeToHexStr()
-	}
-	return fmt.Sprintf(
-		"BLSPubKey:%s-%s/%s[%d]", BLSPubKey,
-		net.JoinHostPort(p.IP, p.Port), p.PeerID, len(p.Addrs),
-	)
-}
+const (
+	// SetAsideForConsensus ..
+	SetAsideForConsensus = 1 << 14
```
rlan35, line 53: These concurrency number seems very high.

node/node.go
```diff
@@ -363,66 +367,277 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
+						handleEArg:     openBox,
+					}
+				default:
+					return false
 				}
+
+				return true
+
+			},
+			libp2p_pubsub.WithValidatorTimeout(24),
```
rlan35, line 558: 24 seems not reasonable. It's Nano seconds.

p2p/host.go
```diff
@@ -290,15 +259,10 @@ func (host *HostV2) ConnectHostPeer(peer Peer) error {
...
-func (host *HostV2) AllTopics() []*libp2p_pubsub.Topic {
-	host.lock.Lock()
-	defer host.lock.Unlock()
-	topics := []*libp2p_pubsub.Topic{}
-	for _, g := range host.joined {
-		topics = append(topics, g)
-	}
-	return topics
+// NamedTopic ..
+type NamedTopic struct {
```
gupadhyaya, line 263: what is the purpose of this?

node/node.go
```diff
@@ -363,66 +367,277 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
+
+	if err := consensus.VerifyMessageSig(senderKey, &m); err != nil {
+		return nil, errors.WithStack(err)
+	}
+
+	return &m, nil
+}
+
 // Start kicks off the node message handling
 func (node *Node) Start() error {
```
rlan35, line 422: Should make these method smaller by refactoring out each of the big sections. Or at least add more comment on each sections' purpose. Currently it's very hard to read.

consensus/consensus_v2.go
LeoHChen: this one actually shouldn't be an error, return nil

consensus/consensus_v2.go
LeoHChen: I actually not sure. This ticker is to check the consensus or view change timeout. I guess EA want to speed up the checking of timeout of viewchange.

node/node.go
LeoHChen: done.

p2p/host.go
LeoHChen: reduced to 1<<13

node/node.go
LeoHChen: agree, set to 10 millisecond, so timeout on 100 messages per second.

p2p/host.go
LeoHChen: added comment.

node/node.go
LeoHChen: added

node/node.go
LeoHChen: validation should be very fast, only check a few byte. 100 nanoseconds should be enough.

node/node.go
LeoHChen: will add more functions

LeoHChen:
```
{"level":"debug","port":"9000","ip":"3.95.194.86","cost":"894ns","caller":"/home/leochen/go/src/github.com/harmony-one/harmony/node/node.go:577","time":"2020-05-30T09:27:40.505183648Z","message":"[cost:topic_validator]"}
{"level":"debug","port":"9000","ip":"3.95.194.86","cost":"1.253µs","caller":"/home/leochen/go/src/github.com/harmony-one/harmony/node/node.go:577","time":"2020-05-30T09:27:40.509709415Z","message":"[cost:topic_validator]"}
{"level":"debug","port":"9000","ip":"3.95.194.86","cost":"975ns","caller":"/home/leochen/go/src/github.com/harmony-one/harmony/node/node.go:577","time":"2020-05-30T09:27:40.511406363Z","message":"[cost:topic_validator]"}
{"level":"debug","port":"9000","ip":"3.95.194.86","cost":"1.392µs","caller":"/home/leochen/go/src/github.com/harmony-one/harmony/node/node.go:577","time":"2020-05-30T09
```

Got some data from LRTN, the cost of validation is no more than a few microsecond, so I set the timeout to be 50 microsecond is more than enough for validation.

node/node.go
```diff
@@ -363,66 +367,313 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
+
+	key, err := bls_cosi.BytesToBLSPublicKey(
+		senderPubKeyViaWire,
+	)
+	if err != nil {
+		return nil, nil, errors.WithStack(err)
+	}
+	senderKey = key
+
+	if !node.Consensus.IsValidatorInCommittee(senderKey) {
```
JackyWYX, line 427: Is there a data race issue here?

node/node_handler.go
```diff
@@ -506,26 +477,37 @@ func (node *Node) PostConsensusProcessing(
...
-		if numPeersNow >= node.Consensus.MinPeers {
-			utils.Logger().Info().Msg("[bootstrap] StartConsensus")
-			node.startConsensus <- struct{}{}
-			return
+// BootstrapConsensus is the a goroutine to check number of peers and start the consensus
+func (node *Node) BootstrapConsensus() error {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	defer cancel()
+	min := node.Consensus.MinPeers
+	needFriends := make(chan struct{})
```
JackyWYX, line 485: Name `needFriends` is confusing. Maybe we can find a better name for this?

consensus/consensus_v2.go
```diff
@@ -67,32 +43,32 @@ func (consensus *Consensus) handleMessageUpdate(payload []byte) {
...
-		consensus.leaderSanityChecks(msg):
+		consensus.leaderSanityChecks(msg, senderKey):
 		consensus.onPrepare(msg)
 	case t == msg_pb.MessageType_COMMIT &&
 		intendedForLeader &&
-		consensus.leaderSanityChecks(msg):
+		consensus.leaderSanityChecks(msg, senderKey):
 		consensus.onCommit(msg)
-	case t == msg_pb.MessageType_VIEWCHANGE &&
-		consensus.viewChangeSanityCheck(msg):
```
gupadhyaya, line 90: I think you still need `VerifyMessageSig` for view change messages right?

LeoHChen:
> Confusing: Is it true that this PR cannot prevent pubsub consensus spamming?
> 
> Other than that, two minor issues uncovered. Please check it out.

This PR is designed to prevent p2p attack

consensus/consensus_v2.go
LeoHChen: right, added back.

node/node.go
LeoHChen: why?  this is read-only access

node/node_handler.go
LeoHChen: renamed to enoughMinPeers

node/node.go
JackyWYX: The decider can be changed: consensus/consensus_service.go:393

JackyWYX: The Decider object can be changed: consensus/consensus_service.go:393

JackyWYX:
> > Confusing: Is it true that this PR cannot prevent pubsub consensus spamming?
> > Other than that, two minor issues uncovered. Please check it out.
> 
> This PR is designed to prevent p2p attack

As far as I see,

1. In consensus validation, the signature field is never checked before deliver to consensus module. It is only checked in consensus module. (consensus/consensus_v2.go:46)
2. The validation in libp2p validation (node/node.go:531) is only to check whether the bls public key is in the committee. This public key is very easy to be obtained by attacker. And does not screen out 
3. There is no validation on `Node` protocol thus attacker can easily spam attack on this.

So an attacker can:
1. Easily spam attack on consensus layer with a public BLS key in committee. The message will pass validation and be further passed to consensus module and check signature for every message. Which is CPU consuming (consensus/consensus_v2.go:46).
2. Spam with Node message. For example, flooding the network with transactions with random content and invalid signature. This could easily drain the memory resource on validators in transaction pool.

I would recommend the following fix:
1. Add a middleware filtering logic in validation (node/node.go:531) to
    1. Limit the request per minute to limit access based on peerId.
    2. Filter out the blacklisted peerId.
2. Do all the sanity check logic in validation, including consensus sanity check, transaction signature checks. If fails the check, add the peerId to blacklist with timeout in middleware. 
3. All nodes has the responsibility to check the validation of message before sending it to pubsub.

node/node.go
rlan35: I don't see the todo?

node/node.go
```diff
@@ -363,66 +367,313 @@ func (node *Node) AddPendingReceipts(receipts *types.CXReceiptsProof) {
...
-	ctx := context.Background()
+
+	var allTopics []u
+
+	utils.Logger().Debug().
+		Interface("topics-ended-up-with", groups).
+		Uint32("shard-id", node.Consensus.ShardID).
+		Msg("starting with these topics")
+
+		// init the pubsub topics
```
rlan35, line 471: nit: indent.

p2p/host.go
```diff
@@ -70,24 +64,29 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
...
-	const MaxSize = 2_145_728
 	options := []libp2p_pubsub.Option{
-		libp2p_pubsub.WithPeerOutboundQueueSize(64),
-		libp2p_pubsub.WithMaxMessageSize(MaxSize),
-	}
-	if len(traceFile) > 0 {
-		tracer, _ := libp2p_pubsub.NewJSONTracer(traceFile)
-		options = append(options, libp2p_pubsub.WithEventTracer(tracer))
+		// WithValidateQueueSize sets the buffer of validate queue. Defaults to 32. When queue is full, validation is throttled and new messages are dropped.
+		libp2p_pubsub.WithValidateQueueSize(64),
```
rlan35, line 81: Shall we make this bigger? The validation has sig verification and costly operations. I suspect many msg may be dropped if it's leader. Also, if there a log when msg gets dropped due to validation throttle?

p2p/host.go
```diff
@@ -70,24 +64,29 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
...
+		// WithPeerOutboundQueueSize is an option to set the buffer size for outbound messages to a peer. We start dropping messages to a peer if the outbound queue if full.
+		libp2p_pubsub.WithPeerOutboundQueueSize(32),
```
rlan35, line 83: Same for this actually, this hotspot node (like leader) may experience msg drops.

p2p/host.go
LeoHChen: We don't do sign verification on the validation anymore.  Based on my benchmarking, sig verification will triple the CPU usage on a regular node, thus it will cause a lot of latency during attack.  So I reverted it back to original logic to do late sig verification.

As to the log, I think the p2p trace will have it, but we didn't enough the p2p trace by default though.

p2p/host.go
LeoHChen: this is outbound queue size, even leader doesn't need to have a large queue for outbound unless leader has to send a lot of messages, IMHO. Leader will receive a lot of messages, but not outbound though.  We can double it for now.  But I'll dig deeper into the metrics of p2p to be able to find some more data, so that we can tune those number later though.

p2p/host.go
rlan35: The signature verification is done anyway if attacker send a lot of msg when a fake sender public key that's in the committee. But for attacker who don't use fake key in the committee, the delayed sig verification will help.

PR #3132, opened 2020-06-05
Fix multi-key related issues when keys are not in committee

rlan35:
## Issue
Previously, the multi-key feature assumes all the keys are in committee for the leader. But if it's not true, the leader node will fail to propose new blocks or fail to count votes correctly.

## Test

tested in dryrun network with slot changes.

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -211,9 +211,12 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 			consensus.getLogger().Debug().Msg("[onViewChange] add my M1 type messaage")
 			msgToSign := append(preparedMsg.BlockHash[:], preparedMsg.Payload...)
 			for i, key := range consensus.PubKey.PublicKey {
+				if err := consensus.bhpBitmap[recvMsg.ViewID].SetKey(key, true); err != nil {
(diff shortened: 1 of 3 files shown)


rlan35:
> even https://github.com/harmony-one/harmony/pull/3132/files#diff-79bde88350477bb6890cee14f9a102e7L239 needs check right?

added. good catch.

consensus/view_change.go
```diff
@@ -411,6 +420,12 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 			commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
 				block.Epoch(), block.Hash(), block.NumberU64(), block.Header().ViewID().Uint64())
 			for i, key := range consensus.PubKey.PublicKey {
+				if err := consensus.commitBitmap.SetKey(key, true); err != nil {
+					consensus.getLogger().Debug().
+						Msg("[OnViewChange] New Leader commit bitmap set failed")
```
gupadhyaya, line 425: may be we should use consistent Msg and log level throughout?

rlan35: done.

PR #3133, opened 2020-06-06
[p2p] add remote tracer

LeoHChen:

### DIFF
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 	"fmt"
 	"net"
+	"os"
 	"runtime"
 	"strings"
 	"sync"
@@ -75,7 +76,6 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
 	if err != nil {
 		return nil, errors.Wrapf(err, "cannot initialize libp2p host")
 	}
-
 	options := []libp2p_pubsub.Option{
 		// WithValidateQueueSize sets the buffer of validate queue. Defaults to 32. When queue is full, validation is throttled and new messages are dropped.
 		libp2p_pubsub.WithValidateQueueSize(64),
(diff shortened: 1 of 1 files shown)


rlan35:
what's the functionality of remote tracer? please add some description.

LeoHChen:
Tested on STN and I can collect trace from remote server.

p2p/host.go
```diff
@@ -87,6 +87,19 @@ func NewHost(self *Peer, key libp2p_crypto.PrivKey) (Host, error) {
 		libp2p_pubsub.WithValidateThrottle(MaxMessageHandlers),
 	}
 
+	traceFile := os.Getenv("P2P_TRACEFILE")
+	if len(traceFile) > 0 {
+		pi, err := libp2p_peer.AddrInfoFromP2pAddr(ma.StringCast(traceFile))
+		if err == nil {
+			tracer, _ := libp2p_pubsub.NewRemoteTracer(ctx, p2pHost, *pi)
```
JackyWYX, line 94: It seems that the remote tracer will write all `TraceEvents` to the stream to peer `pi`. But it seems that the libp2p stream currently is not handled in our code. So is this something that we expect? 

```
~/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.3.1/tracer.go:238
```

Do you think that a JSON tracer will be better for dumping the trace event to a local file?

```
~/go/pkg/mod/github.com/libp2p/go-libp2p-pubsub@v0.3.1/tracer.go:89
```

JackyWYX:
> Tested on STN and I can collect trace from remote server.

Interesting! Actually I am surprised - I thought that libp2p stream is not dealt with in our codebase. So where did you collect these trace output on remote server?

p2p/host.go
LeoHChen: we used to have local tracer actually.  We can support both though. a remote tracer is easier to collect all logs for us to analysis.   I'll add the support of both remote and local tracer then.

LeoHChen:
> > Tested on STN and I can collect trace from remote server.
> 
> Interesting! Actually I am surprised - I thought that libp2p stream is not dealt with in our codebase. So where did you collect these trace output on remote server?

I setup a remote traced, and peers send trace to the remote server I setup.

LeoHChen:
> what's the functionality of remote tracer? please add some description.

remote tracer is a central server that can collect the tracelog from peers that knows the remote server's multiaddress.  I setup a simple remote trace log server using https://github.com/libp2p/go-libp2p-pubsub-tracer.

LeoHChen:
Just FYI, @JackyWYX , @rlan35 , here is some stats I generated using the trace I collected on one peer in STN.

```
=== Trace Summary ===
Peers: 1
Topics: hmy/stressnet/0.0.1/node/shard/1
Published Messages: 1350
Delivered Messages: 1350
Duplicate Messages: 0
Rejected Messages: 0
Sent RPCs: 0
Dropped RPCs: 0
=== Trace Summary ===
Peers: 1
Topics: hmy/stressnet/0.0.1/node/beacon
Published Messages: 19
Delivered Messages: 19
Duplicate Messages: 0
Rejected Messages: 0
Sent RPCs: 0
Dropped RPCs: 0
```

Issue #3139, opened 2020-06-08
nil pointer explosion 


fxfactorial:
bec7d80;a5666b1abd77224fbc83ed423e4ce9b9ecb0b40e7b4dfaad0017638d90d010398bda2393642ff4fe1dd6c243e7252f10; -> shard 0
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xb25f7c]

goroutine 1 [running]:
github.com/harmony-one/harmony/core/types.(*Block).Root(...)
	/mnt/jenkins/workspace/harmony-release/harmony/core/types/block.go:505
github.com/harmony-one/harmony/core.(*BlockChain).repair(0xc0003f6000, 0xc000742c78, 0x23, 0xc0004c80a0)
	/mnt/jenkins/workspace/harmony-release/harmony/core/blockchain.go:538 +0x22c
github.com/harmony-one/harmony/core.(*BlockChain).loadLastState(0xc0003f6000, 0x15ea5c0, 0x0)
	/mnt/jenkins/workspace/harmony-release/harmony/core/blockchain.go:329 +0x11b8
github.com/harmony-one/harmony/core.NewBlockChain(0x1aff3a0, 0xc0000b6790, 0xc000137ae0, 0xc00042e8c0, 0x1b0a240, 0x2853e60, 0x0, 0x0, 0x0, 0x0, ...)
	/mnt/jenkins/workspace/harmony-release/harmony/core/blockchain.go:246 +0x82b
github.com/harmony-one/harmony/internal/shardchain.(*CollectionImpl).ShardChain(0xc00042e910, 0x0, 0x0, 0x0, 0x0)
	/mnt/jenkins/workspace/harmony-release/harmony/internal/shardchain/shardchains.go:95 +0x351
github.com/harmony-one/harmony/node.(*Node).Blockchain(0xc000160fc0, 0xc00042e910)
	/mnt/jenkins/workspace/harmony-release/harmony/node/node.go:164 +0x4a
github.com/harmony-one/harmony/node.New(0x1b0a2c0, 0xc00019dd60, 0xc000160d80, 0x1ad5d40, 0xc00035fcd0, 0x0, 0x0, 0x0)
	/mnt/jenkins/workspace/harmony-release/harmony/node/node.go:474 +0x487
main.setupConsensusAndNode(0xc00034a000, 0x0)
[...]

PR #3143, opened 2020-06-08
[core] fail repair on missing block

gupadhyaya:
Fixes: https://github.com/harmony-one/harmony/issues/3139
* Applying patch from go-ethereum: https://github.com/ethereum/go-ethereum/blob/4b2ff1457ac28fb2894485194e0e344e84c2bcd7/core/blockchain.go#L626 for taking care of nil pointer when rewinding in repair. 
* Move validator remove call to rewind success branch, it was dead code before.


LeoHChen:
Thanks for the PR and the cherrypick of the fix. The reported crash is also on line 538, where (*head) is nil. Please add a check there as well.

gupadhyaya:
@LeoHChen the fix will make sure that `(*head)` is never `nil` at 538. No need for additional check.

PR #3152, opened 2020-06-12
Remove mutex in message sender

LeoHChen:
atomic.Value is faster than the Mutex we used in messageSender.

### DIFF
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -2,6 +2,7 @@ package consensus
 
 import (
 	"math/big"
+	"sync/atomic"
 
 	"github.com/ethereum/go-ethereum/common"
 	protobuf "github.com/golang/protobuf/proto"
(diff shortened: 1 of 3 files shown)


LeoHChen:
> According to my knowledge, `atomic.Value` is only used for structure. Do you think using `atomic.StoreUint64` and `atomic.StoreBool` would be better?
> 
> If the naming without atomic keyword seems confusing, do you think it is a better practice to use a wrapped atomic module? ([uber go style guide](https://github.com/uber-go/guide/blob/master/style.md#use-gouberorgatomic)).

sounds good, we can use explicit func to store/load value. A wrapper module is a bit overkill now.

JackyWYX:
> > According to my knowledge, `atomic.Value` is only used for structure. Do you think using `atomic.StoreUint64` and `atomic.StoreBool` would be better?
> > If the naming without atomic keyword seems confusing, do you think it is a better practice to use a wrapped atomic module? ([uber go style guide](https://github.com/uber-go/guide/blob/master/style.md#use-gouberorgatomic)).
> 
> sounds good, we can use explicit func to store/load value. A wrapper module is a bit overkill now.

Cannot agree more. Explicit call of load have much better performance

consensus/consensus_service.go
```diff
@@ -273,9 +274,7 @@ func (consensus *Consensus) checkViewID(msg *FBFTMessage) error {
 
 // SetBlockNum sets the blockNum in consensus object, called at node bootstrap
 func (consensus *Consensus) SetBlockNum(blockNum uint64) {
```
JackyWYX, line 276: Do you think there is a race condition at `consensus/consensus_v2.go:263`.

LeoHChen: Agree. there could be race condition in all the reading.  Our original code wasn't protecting all the read operation at all.  I'd do it later.  This PR is getting too big now.

LeoHChen: updated with atomic.AddUint64

JackyWYX:
More read protection can be done later. Look good to me.

PR #3157, opened 2020-06-12
[p2p] add counters of valid/invalid messages

LeoHChen:

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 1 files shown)


LeoHChen:
> Is this the PR to be merged to production or just for debugging? If it is to be merged to be master, it will introduce a lot of race conditions. Would be better to fix?
> 
> One invalid message cause is that, when the node is starting up, it does not have the latest `stateShard` thus it may fail in `validateShardBoundMessage`. This invalid message cannot be easily fixed since it need introducing some flow control between `beaconSync` and `pubsub` handling (Only start pubsub handling when having latest stateShard).

good insight.  Actually, this could be the reason why we have failed last rolling upgrade with high latency in shard3.  I'll use atomic to keep track of it, and make it to production to monitor the situation.

PR #3161, opened 2020-06-14
Increase interval of peer finding

LeoHChen:


api/service/networkinfo/service.go
```diff
@@ -220,8 +225,13 @@ func (s *Service) DoService() {
...
 				return
 			}
+			if peerInterval < maxFindPeerInterval {
+				peerInterval *= 2
+				intervalTick.Stop()
+				intervalTick = time.NewTicker(time.Duration(peerInterval) * time.Second)
+			}
 
-			s.findPeers(ctx)
+			go s.findPeers(ctx)
```
JackyWYX, line 234: Why to add a goroutine here?

LeoHChen: don't want findPeers block this function. s.PeerInfo is a separate channel returned by libp2p.

JackyWYX: I see. So is it true that `findPeers` is never returned and the number of goroutine of findPeers is always increasing? (for channel will never end until the channel is closed)

Issue #3150, opened 2020-06-10
[libp2p] increased latency on after p2p validation change on mainnet


LeoHChen:
**Describe the bug**
After the rolling upgrade of v2.1.4, we have noticed shard 1, 3 have high consensus latency, up to 16s. shard 0 and 2 have relatively stable latency < 9s.

I was checking the leader of shard3 and noticed high latency on consensus.

```json
{"level":"info","port":"9000","ip":"13.52.219.208","myBlock":3568760,"myViewID":3568881,"phase":0,"mode":"Normal","blockNum":3568759,"epochNum":199,"ViewId":3568880,"blockHash":"0x95fc19ed368e3b55396948cb63aa3af2409a36c1bab1ae4bf9828c359aaf2232","index":0,"numTxns":0,"numStakingTxns":0,"caller":"/mnt/jenkins/workspace/harmony-release/harmony/consensus/consensus_v2.go:154","time":"2020-06-09T23:34:40.167128806Z","message":"HOORAY!!!!!!! CONSENSUS REACHED!!!!!!!"}
{"level":"info","port":"9000","ip":"13.52.219.208","myBlock":3568760,"myViewID":3568881,"phase":2,"mode":"Normal","caller":"/mnt/jenkins/workspace/harmony-release/harmony/consensus/leader.go:273","time":"2020-06-09T23:34:43.055966433Z","message":"[OnCommit] Starting Grace Period"}
{"level":"info","port":"9000","ip":"13.52.219.208","myBlock":3568760,"myViewID":3568881,"phase":2,"mode":"Normal","validatorPubKey":"cf159cbb7e222c15c42bbaed725510a707ebb891630de2dfcd2b0081aac80ce45ab433f6d2708af538327921f4841b07","MsgViewID":3568881,"MsgBlockNum":3568760,"numReceivedSoFar":165,"caller":"/mnt/jenkins/workspace/harmony-release/harmony/consensus/leader.go:276","time":"2020-06-09T23:34:48.167248787Z","message":"[OnCommit] Commit Grace Period Ended"}
[...]

PR #3162, opened 2020-06-14
[p2p] increase timeout of validator

LeoHChen:
* restore the ticker to the original 3 seconds
* try to address https://github.com/harmony-one/harmony/issues/3150

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -321,7 +321,7 @@ func (consensus *Consensus) Start(
 		}
 		consensus.getLogger().Info().Time("time", time.Now()).Msg("[ConsensusMainLoop] Consensus started")
 		defer close(stoppedChan)
-		ticker := time.NewTicker(time.Second)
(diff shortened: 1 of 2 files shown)


LeoHChen:
@JackyWYX , please review this one.  I've done a lot of tests over the weekend and found the evidence on mainnet that the validation may take longer time to finish.  So we need to increase the timeout. That's probably one of the reasons of the failed rolling upgrade last week.

rlan35:
Why is this related? The ticker is just for periodic loop for checking timeouts. How is it relate to the validation?

LeoHChen:
> Why is this related? The ticker is just for periodic loop for checking timeouts. How is it relate to the validation?

ticker is not related.  I restored it to the original 3 sec.  The main change is the validation timeout.

rlan35:
> > Why is this related? The ticker is just for periodic loop for checking timeouts. How is it relate to the validation?
> 
> ticker is not related. I restored it to the original 3 sec. The main change is the validation timeout.

I see. cool.

PR #3164, opened 2020-06-15
Sampledlog for more p2p errors

LeoHChen:
The sampledlogger helped me identify more p2p errors on mainnet so that I can tune the parameters and prepare for the blacklist feature, and adjust the timeout value for context.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -543,7 +543,7 @@ func (node *Node) Start() error {
 			// this is the validation function called to quickly validate every p2p message
 			func(ctx context.Context, peer libp2p_peer.ID, msg *libp2p_pubsub.Message) bool {
 				entryTime := time.Now()
-				defer utils.SampledLogger().Info().Str("cost", time.Now().Sub(entryTime).String()).Msg("[cost:topic_validator]")
+				defer utils.SampledLogger().Debug().Str("cost", time.Now().Sub(entryTime).String()).Msg("[cost:topic_validator]")
 
 				hmyMsg := msg.GetData()
 
@@ -552,7 +552,7 @@ func (node *Node) Start() error {
 					errChan <- withError{
 						errors.WithStack(errors.Wrapf(
 							errMsgHadNoHMYPayLoadAssumption, "on topic %s", topicNamed,
-						)), nil,
+						)), msg.GetFrom(),
 					}
 					return false
 				}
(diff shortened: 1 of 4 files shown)


node/node.go
```diff
@@ -437,7 +437,7 @@ func (node *Node) validateShardBoundMessage(
 
 	if !node.Consensus.IsValidatorInCommittee(senderKey) {
 		atomic.AddUint32(&node.NumInvalidMessages, 1)
-		return nil, nil, errors.WithStack(shard.ErrValidNotInCommittee)
+		return &m, nil, errors.WithStack(shard.ErrValidNotInCommittee)
```
JackyWYX, line 440: Why are we returning message?

node/node.go
```diff
@@ -597,11 +606,25 @@ func (node *Node) Start() error {
...
+					if errors.Is(ctx.Err(), context.DeadlineExceeded) {
+						utils.Logger().Warn().
+							Str("topic", topicNamed).Msg("[context] exceeded validation deadline")
+					}
+					errChan <- withError{errors.WithStack(ctx.Err()), msg}
+				default:
+					return true
+				}
 
+				return true
```
JackyWYX, line 627: Shall return false if timeout right?

node/node.go
```diff
@@ -552,7 +559,7 @@ func (node *Node) Start() error {
 					errChan <- withError{
```
JackyWYX, line 559: Shall we make errChan a buffered channel? Else this step can be blocked and the sample log behavior might effect message handling/

p2p/host.go
```diff
@@ -56,6 +56,8 @@ const (
 	SetAsideOtherwise = 1 << 12
 	// MaxMessageHandlers ..
 	MaxMessageHandlers = SetAsideForConsensus + SetAsideOtherwise
+	// MaxMessageSize is the 256Kb
+	MaxMessageSize = 1 << 18
```
rlan35, line 60: Our block can be as big as around 2MB.

node/node.go
LeoHChen: The message can be used later to debug and check why this node is not in the current committee.  It can be helpful when we determine if we need to blacklist the message sender.

node/node.go
LeoHChen: make sense.  I need to test this scenario though.

p2p/host.go
LeoHChen: Good catch, then we need to adjust it.  The current default max is actually 1Mb.

node/node.go
LeoHChen: ok, will make it 100.

PR #3167, opened 2020-06-16
Hard code 106 as the cap of bls keys per validator

rlan35:
## Issue

The previous cap of bls key is a dynamic number based on total open slots. It was 1/3 of open slots, so 320/3 = 106 in mainnet. This change hard code that limit and not let it float. This way, we will keep a relatively low cap after we keep adding new open slots.

## Test

Tested locally

## Operational Checklist

    **NO**

    **NO**

Yes

We will have announcement on this.

No


JackyWYX:
The code looks good to me.

PR #3168, opened 2020-06-16
Use singleflight to cache blskey deserialization

LeoHChen:
From the data/stats I collected on mainnet node, we can see that with singleflight enabled, 10 hours later, there are about 8-10% less CPU time used in cgocall, which is used for blskey deserialization.

### DIFF
--- a/crypto/bls/bls.go
+++ b/crypto/bls/bls.go
@@ -1,10 +1,14 @@
 package bls
 
 import (
-	"fmt"
-
 	"github.com/harmony-one/bls/ffi/go/bls"
 	"github.com/pkg/errors"
+
+	"golang.org/x/sync/singleflight"
+)
+
+var (
+	BLSPubKeyCache singleflight.Group
 )
 
 func init() {
@@ -18,14 +22,30 @@ func RandPrivateKey() *bls.SecretKey {
 	return &sec
 }
 
+var (
+	errEmptyByte = errors.New("BytesToBLSPublicKey: byte is empty")
+	errCastError = errors.New("BytesToBLSPublicKey: cast type error")
+)
+
 // BytesToBLSPublicKey converts bytes into bls.PublicKey pointer.
 func BytesToBLSPublicKey(bytes []byte) (*bls.PublicKey, error) {
 	if len(bytes) == 0 {
-		return nil, fmt.Errorf("[BytesToBLSPublicKey] bytes is empty")
+		return nil, errEmptyByte
 	}
-	pubKey := &bls.PublicKey{}
-	err := pubKey.Deserialize(bytes)
-	return pubKey, err
+	kkey := string(bytes)
+	key, err, _ := BLSPubKeyCache.Do(kkey, func() (interface{}, error) {
+		pubKey := &bls.PublicKey{}
+		err := pubKey.Deserialize(bytes)
+		return pubKey, err
+	})
+	if err != nil {
+		return nil, err
+	}
+	if pubkey, ok := key.(*bls.PublicKey); ok {
+		return pubkey, nil
+	} else {
+		return nil, errCastError
+	}
 }
 
 // AggregateSig aggregates all the BLS signature into a single multi-signature.


LeoHChen:
Before singleflight, pprof graph
[before-singleflight-pprof.104.199.209.190.pdf](https://github.com/harmony-one/harmony/files/4789302/before-singleflight-pprof.104.199.209.190.pdf)

LeoHChen:
Right after singleflight is enabled, pprof graph
[right-after-singleflight-pprof.104.199.209.190.pdf](https://github.com/harmony-one/harmony/files/4789305/right-after-singleflight-pprof.104.199.209.190.pdf)

image

LeoHChen:
10 hours later after singleflight enabled, pprof chart,  most of the keys are already cached.

[10-hours-later-singleflight-pprof.104.199.209.190.pdf](https://github.com/harmony-one/harmony/files/4789311/10-hours-later-singleflight-pprof.104.199.209.190.pdf)


image

LeoHChen:
The data is collected on a mainnet shard0 node, 104.199.209.190

JackyWYX:
As far as I can see, 12s sleep time is not a good pick for 8s block finalize, thus it is likely that the almost 1/4 of the cache will be a miss if each BLS key is sent twice in each block. Would you please try the following settings and wait 60s to see the performance?

1. Extend 12s sleep time to 30s.
2. Replace singleflight with LRU cache of size 400 (Cache size can be greater after slot increase).

node/node.go
```diff
@@ -426,22 +430,38 @@ func (node *Node) validateShardBoundMessage(
...
-	)
+	kkey := string(senderPubKeyViaWire)
+	key, err, _ := node.blsPubKeyCache.Do(kkey, func() (interface{}, error) {
+		k, e := bls_cosi.BytesToBLSPublicKey(senderPubKeyViaWire)
+
+		if e != nil {
+			return nil, e
+		}
+
+		if !node.Consensus.IsValidatorInCommittee(k) {
```
gupadhyaya, line 441: why do you need to cache this part? if you only cache the deserialization, then you don't even need to timer, as blskey deserialization will never change.

rlan35: This part shouldn't be cached, because the IsValidatorInCommittee result may change across epoch and if it's cached for 12 hours it will be returning wrong results. One solution is to clear the cache whenever there is a new epoch transition.

LeoHChen: good point

LeoHChen: moved it out

node/node.go
```diff
@@ -426,18 +432,29 @@ func (node *Node) validateShardBoundMessage(
 		return nil, nil, errors.WithStack(errNotRightKeySize)
 	}
 
-	key, err := bls_cosi.BytesToBLSPublicKey(
-		senderPubKeyViaWire,
-	)
+	kkey := string(senderPubKeyViaWire)
+	key, err, _ := node.blsPubKeyCache.Do(kkey, func() (interface{}, error) {
+		k, e := bls_cosi.BytesToBLSPublicKey(senderPubKeyViaWire)
```
gupadhyaya, line 437: how about moving the caching inside `BytesToBLSPublicKey`? then all the fbft message parsing can also take advantage. e.g., `ParseFBFTMessage` may speedup too.

LeoHChen: good suggestion, let me try it.

LeoHChen:
Before singleflight, the cgocall is 87.01%
image


With singleflight, the cgocall dropped to 79.26%
image

gupadhyaya:
@LeoHChen fix travis? otherwise looks good.

crypto/bls/bls.go
```diff
@@ -18,14 +22,30 @@ func RandPrivateKey() *bls.SecretKey {
...
+)
+
 // BytesToBLSPublicKey converts bytes into bls.PublicKey pointer.
 func BytesToBLSPublicKey(bytes []byte) (*bls.PublicKey, error) {
 	if len(bytes) == 0 {
-		return nil, fmt.Errorf("[BytesToBLSPublicKey] bytes is empty")
+		return nil, errEmptyByte
+	}
+	kkey := string(bytes)
+	key, err, _ := BLSPubKeyCache.Do(kkey, func() (interface{}, error) {
```
rlan35, line 36: If now we are just caching a simple data results. Better use lru cache with size such as 1000. It's much simpler in logic and probably also in performance. SingleFlight is more for complicated function call caching.

LeoHChen: make sense.

PR #3169, opened 2020-06-17
return error info if shard does not match

songgeng87:
add error info if shard does not match and make sure the rpc return it

### DIFF
--- a/internal/hmyapi/apiv1/util.go
+++ b/internal/hmyapi/apiv1/util.go
@@ -36,7 +36,6 @@ func SubmitTransaction(
 	ctx context.Context, b Backend, tx *types.Transaction,
 ) (common.Hash, error) {
 	if err := b.SendTx(ctx, tx); err != nil {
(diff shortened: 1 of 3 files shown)


node/node.go
```diff
@@ -303,8 +303,9 @@ func (node *Node) AddPendingTransaction(newTx *types.Transaction) error {
 			node.tryBroadcast(newTx)
 		}
 		return err
+	}else{
```
rlan35, line 306: go fmt ./...

Daniel-VDM:
jenkins, test this please

internal/hmyapi/apiv1/util.go
```diff
@@ -38,7 +38,7 @@ func SubmitTransaction(
 	if err := b.SendTx(ctx, tx); err != nil {
 		// legacy behavior is to never return error and always return tx hash
```
Daniel-VDM, line 39: clean-up this comment please

PR #3173, opened 2020-06-18
Reduce number of calls to consensus messages

LeoHChen:
Previous release v2.1.5 hit an issue that leaders weren't able to process all the commit messages sent by validators, even the consensus was reached.

The issue was the added validation layer as it's gone after we revert leader to v2.1.3 release.

When the pubsub validation function is added to p2p layer, every p2p message will spawn a goroutine (https://github.com/harmony-one/harmony/blob/23d1890619ba11a8f159c95dcbdb99724e7336b4/node/node.go#L390) to do validation. After the validation, every valid p2p message will spawn a new goroutine (https://github.com/harmony-one/harmony/blob/23d1890619ba11a8f159c95dcbdb99724e7336b4/consensus/consensus_v2.go#L29) to process each message.
However, due to the redundancy of p2p messages, not all messages are intended for leaders/validators, but the massive number of p2p messages will spawn massive number of goroutines that may exceed the number of goroutine limited by the system/configuration.  Thus, we've seen the drop of message processing on leaders when leaders won't be able to collect 100% of the commit messages from all validators.

Before the addition of the validation layer, as we only spawn one goroutine per p2p message, the number of goroutine didn't explode. With the validation layer, the number almost doubled. I added some logs in the code and collected the stats in mainnet to verify it. So, with this fix, we ignore most of the redundant messages that are not intended for consensus message handlers, to avoid the explosion of the goroutine.

I've tested it on the devnet and canary nodes on mainnet. We can see a clear drop of the CPU load after the update. Network traffic also dropped, but that maybe because of the restart of the node program.

We still need to the fix on the leaders though.

image

image

image

image

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -163,6 +163,7 @@ type Node struct {
 	BroadcastInvalidTx bool
 
 	// metrics of p2p messages
+	NumTotalMessages   uint32
 	NumValidMessages   uint32
 	NumInvalidMessages uint32
 }
@@ -394,6 +395,7 @@ func (node *Node) validateShardBoundMessage(
 		m         msg_pb.Message
 		senderKey *bls.PublicKey
 	)
+	atomic.AddUint32(&node.NumTotalMessages, 1)
 	entryTime := time.Now()
 	defer utils.SampledLogger().Debug().Str("cost", time.Now().Sub(entryTime).String()).Msg("[cost:validate_shard_bound_message]")
 
@@ -867,9 +869,14 @@ func New(
 		for {
 			select {
 			case <-ticker.C:
-				utils.Logger().Info().Uint32("ValidMessage", node.NumValidMessages).Uint32("InvalidMessage", node.NumInvalidMessages).Msg("MsgValidator")
+				utils.Logger().Info().
+					Uint32("TotalMessage", node.NumTotalMessages).
+					Uint32("ValidMessage", node.NumValidMessages).
+					Uint32("InvalidMessage", node.NumInvalidMessages).
+					Msg("MsgValidator")
 				atomic.StoreUint32(&node.NumInvalidMessages, 0)
 				atomic.StoreUint32(&node.NumValidMessages, 0)
+				atomic.StoreUint32(&node.NumTotalMessages, 0)
 			}
 		}
 	}()
--- a/node/node.go
+++ b/node/node.go
@@ -536,11 +536,6 @@ func (node *Node) Start() error {
 		senderPubKey   *bls.PublicKey
 	}
 
-	type p2pSenderInfo struct {
-		source    libp2p_peer.ID
-		forwarder libp2p_peer.ID
-	}
-
 	isThisNodeAnExplorerNode := node.NodeConfig.Role() == nodeconfig.ExplorerNode
 
 	for i := range allTopics {
@@ -571,7 +566,7 @@ func (node *Node) Start() error {
 					errChan <- withError{
 						errors.WithStack(errors.Wrapf(
 							errMsgHadNoHMYPayLoadAssumption, "on topic %s", topicNamed,
-						)), p2pSenderInfo{msg.GetFrom(), msg.ReceivedFrom},
+						)), msg.GetFrom(),
 					}
 					return false
 				}
@@ -586,7 +581,7 @@ func (node *Node) Start() error {
 					if !isConsensusBound {
 						errChan <- withError{
 							errors.WithStack(errConsensusMessageOnUnexpectedTopic),
-							p2pSenderInfo{msg.GetFrom(), msg.ReceivedFrom},
+							msg.GetFrom(),
 						}
 						return false
 					}
@@ -598,7 +593,8 @@ func (node *Node) Start() error {
 
 					if err != nil {
 						errChan <- withError{err,
-							p2pSenderInfo{msg.GetFrom(), msg.ReceivedFrom}}
+							msg.GetFrom(),
+						}
 						return false
 					}
 
--- a/node/node.go
+++ b/node/node.go
@@ -563,12 +563,7 @@ func (node *Node) Start() error {
 
 				// first to validate the size of the p2p message
 				if len(hmyMsg) < p2pMsgPrefixSize {
-					errChan <- withError{
-						errors.WithStack(errors.Wrapf(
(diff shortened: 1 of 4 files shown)


LeoHChen:
Before the fix, some stats I collected.  We can see the high number of consensus messages being processed by each goroutine.
```
zerolog-validator-34.87.212.7-9000-2020-06-18T13-33-37.328.log:{"level":"info","port":"9000","ip":"34.87.212.7","ConsensusMsg":3380,"PrepareMsg":0,"CommitMsg":0,"caller":"/home/leochen/go/src/github.com/harmony-one/harmony/consensus/consensus.go:223","time":"2020-06-18T11:23:33.2998771Z","message":"ConsensusMessageHandler"}
zerolog-validator-34.87.212.7-9000-2020-06-18T13-33-37.328.log:{"level":"info","port":"9000","ip":"34.87.212.7","ConsensusMsg":4201,"PrepareMsg":0,"CommitMsg":0,"caller":"/home/leochen/go/src/github.com/harmony-one/harmony/consensus/consensus.go:223","time":"2020-06-18T11:24:33.301261613Z","message":"ConsensusMessageHandler"}
```

LeoHChen:
After the fix, the number of goroutine dropped to 1/20 the total messages.

```
zerolog-validator-34.87.212.7-9000-2020-06-18T21-32-30.370.log:{"level":"info","port":"9000","ip":"34.87.212.7","ConsensusMsg":22,"PrepareMsg":0,"CommitMsg":0,"AnnounceMsg":8,"PreparedMsg":7,"CommittedMsg":7,"caller":"/home/leochen/go/src/github.com/harmony-one/harmony/consensus/consensus.go:233","time":"2020-06-18T21:30:30.968306871Z","message":"ConsensusMessageHandler"}
```

node/node.go
```diff
@@ -590,6 +602,10 @@ func (node *Node) Start() error {
 						return false
 					}
 
+					if ignore {
+						return true
```
JackyWYX, line 606: Is it true that ignored message will trigger error `errConvertToValidMessage`?

LeoHChen: you are right, should ignore it.

node/node.go
```diff
@@ -389,59 +389,71 @@ var (
...
 
 	if err := protobuf.Unmarshal(payload, &m); err != nil {
 		atomic.AddUint32(&node.NumInvalidMessages, 1)
-		return nil, nil, errors.WithStack(err)
+		return nil, nil, true, errors.WithStack(err)
 	}
+	if node.Consensus.IsLeader() {
+		switch m.Type {
+		case msg_pb.MessageType_ANNOUNCE, msg_pb.MessageType_PREPARED, msg_pb.MessageType_COMMITTED:
+			return nil, nil, true, nil
```
rlan35, line 407: We don't need the boolean.  can just return a new error here "Wrong message type" or "Unacceptable message type"

node/node.go
```diff
@@ -590,6 +602,11 @@ func (node *Node) Start() error {
 						return false
 					}
 
+					// ignore the further processing of the p2p messages as it is not intended for this node
+					if ignore {
```
rlan35, line 606: can check the error message here to do the ignoring, rather than using a extra boolean.

node/node.go
LeoHChen: It's not a real error though. Those are valid messages. that's why I want to be explicit to ignore it.

rlan35:
The ignored messages by leader is basically the messages sent from itself. First of all, I thought the p2p layer already filtered out self msgs. Second, even if it's not filtered, those messages are just a small amount compared to the messages sent from validators. I can see this PR is mitigating the problem a little. but not sure it will help much.

LeoHChen:
> The ignored messages by leader is basically the messages sent from itself. First of all, I thought the p2p layer already filtered out self msgs. Second, even if it's not filtered, those messages are just a small amount compared to the messages sent from validators. I can see this PR is mitigating the problem a little. but not sure it will help much.
> 
> it makes the validators to ignore lots of other validators's messages, but will it cause the problem that the validator's messages are not able to propagate across the network to reach leader, because the node in the middle ignored them?

Very good questions. 

For leaders, it will ignore all the redundant messages from itself. It is very true that it may still won't be able to handle all the incoming messages on-time.  

This PR will help validator a lot for sure atm. For leaders, I don't have any test data on mainnet yet. Will have to try it out.

The ignored messages are just ignored by the consensus messages handler, but not invalidated by the p2p layer.  That's why when it returns `true`, the p2p layer will still propagate the messages to other nodes, such make sure the messages can be delivered properly.

I actually tested to return `false` on the ignored messages, the outgoing traffic will drop to very little, as that's invalid messages p2p messages won't be propagated to the peers.  So, in this case ,it is safe to ignore those unrelated messages.  You can see from the chart.


image

node/node.go
LeoHChen: as above, trying to be explicit and more readable, that's why a separate variable is used.

JackyWYX:
> For leaders, it will ignore all the redundant messages from itself. It is very true that it may still won't be able to handle all the incoming messages on-time.

What about build an LRU cache with message hash as key to avoid processing on the same consensus message repeatedly?

LeoHChen:
> > For leaders, it will ignore all the redundant messages from itself. It is very true that it may still won't be able to handle all the incoming messages on-time.
> 
> What about build an LRU cache with message hash as key to avoid processing on the same consensus message repeatedly?

It is basically the singleflight pattern to process the message once at consensus message handler level.  Let me think through and test it out. I may use a later PR to address that.

LeoHChen:
> What about build an LRU cache with message hash as key to avoid processing on the same consensus message repeatedly?

I also don't think p2p layer will pass the duplicated message to the validator and message handler, as it should be handled by libp2p already.  I collected some trace from mainnet node, using tracer, and it shows me the number of messages. You can see the duplicate messages are 5x more than the delivered messages.  I did a calculation on the messages handled in current canary node and the number are expected, which means not duplicated messages.
```
=== Trace Summary ===
Peers: 1
Topics: mytopic
Published Messages: 4824
Delivered Messages: 141336
Duplicate Messages: 672758
Rejected Messages: 36132
Sent RPCs: 660649
Dropped RPCs: 870
```

PR #3176, opened 2020-06-20
Use lru to cache blspubkey

LeoHChen:
Inline validator is used to solve the missing signature issue of leader.

### DIFF
--- a/crypto/bls/bls.go
+++ b/crypto/bls/bls.go
@@ -1,10 +1,18 @@
 package bls
 
 import (
-	"fmt"
-
 	"github.com/harmony-one/bls/ffi/go/bls"
+	lru "github.com/hashicorp/golang-lru"
 	"github.com/pkg/errors"
+)
+
+const (
+	blsPubKeyCacheSize = 1024
+)
+
+var (
+	// BLSPubKeyCache is the Cache of the Deserialized BLS PubKey
+	BLSPubKeyCache, _ = lru.New(blsPubKeyCacheSize)
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


node/node.go
```diff
@@ -650,12 +652,10 @@ func (node *Node) Start() error {
...
 
 		go func() {
-
-			for msg := range msgChan {
-				// should not take more than 3 seconds to process one message
-				ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
-				msg := msg
-
+			// should not take more than 5 seconds to process one message
+			ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
rlan35, line 656: Does it work to have a single context object to use for all messages?

LeoHChen: fixed.

LeoHChen:
image

crypto/bls/bls.go
```diff
@@ -18,14 +26,28 @@ func RandPrivateKey() *bls.SecretKey {
...
 	}
 	pubKey := &bls.PublicKey{}
 	err := pubKey.Deserialize(bytes)
-	return pubKey, err
+
+	if err != nil {
+		return nil, err
+	}
+
+	BLSPubKeyCache.Add(kkey, pubKey)
```
JackyWYX, line 49: Can we use value instead of pointer for value caching?

LeoHChen: done

crypto/bls/bls.go
```diff
@@ -18,14 +26,33 @@ func RandPrivateKey() *bls.SecretKey {
...
+			return &pk, nil
+		} else {
+			return nil, errPubKeyCast
+		}
 	}
 	pubKey := &bls.PublicKey{}
 	err := pubKey.Deserialize(bytes)
-	return pubKey, err
+
+	if err != nil {
```
JackyWYX, line 50: This is also open to attack. Can we also cache the error?

JackyWYX: Ummm, this seems cannot be fully mitigated until we implement the blacklist feature. So seems good now.

PR #3177, opened 2020-06-20
[p2p] use inline validator

LeoHChen:
[validation] increase the validation queue size

[validation] increase throttle


LeoHChen:
Running on shard3 leader, >75% reached 96% of signatures collected.

rlan35:
does the leader always collect 100% sig before? I assume yes, and we have to keep 100%, otherwise, validator uptime will be bad.

LeoHChen:
still can't collect 100% signature from all validators.

PR #3181, opened 2020-06-24
Avoid validator sending msgs if the key is not in committee

rlan35:
Validator shouldn't send out msges if the key is not in committee.


LeoHChen:
not needed as we have moved from `master` to `main` branch.

PR #3186, opened 2020-06-24
Add cache to validator membership check

rlan35:
tested locally

### DIFF
--- a/consensus/quorum/quorum.go
+++ b/consensus/quorum/quorum.go
@@ -3,6 +3,7 @@ package quorum
 import (
 	"fmt"
 	"math/big"
+	"sync"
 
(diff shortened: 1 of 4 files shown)


consensus/quorum/quorum.go
```diff
@@ -148,6 +148,7 @@ type cIdentities struct {
 	// Public keys of the committee including leader and validators
 	publicKeys     []*bls.PublicKey
 	publicKeysByte []shard.BLSPublicKey
+	keyIndexMap    map[shard.BLSPublicKey]int
```
LeoHChen, line 151: The map is not concurrent safe.

rlan35: fixed.

PR #3189, opened 2020-06-25
P2p validation on main

LeoHChen:


LeoHChen:
image

image

The CPU usage on a validator will be doubled as we do pubkey checking on every p2p messages even though validators may ignore them.  This is the tradeoff to reduce the propagation of the invalid messages into the network. We can see clearly that the outgoing network traffic is lower than incoming traffic now as invalid ones are filtered out.

node/node_handler.go
```diff
@@ -56,101 +55,75 @@ func (node *Node) processSkippedMsgTypeByteValue(
...
-		} else {
-			node.ConsensusMessageHandler(msgPayload)
+// HandleNodeMessage parses the message and dispatch the actions.
+func (node *Node) HandleNodeMessage(
+	ctx context.Context,
+	payload []byte,
+) error {
+
+	// Prevent OOB crash
+	if len(payload) < (proto.MessageTypeBytes + proto.MessageTypeBytes) {
```
JackyWYX, line 70: Should it be MessageTypeBytes + MessageCatagoryBytes?

LeoHChen: fixed in new PR https://github.com/harmony-one/harmony/pull/3212

consensus/checks.go
```diff
@@ -1,99 +1,73 @@
...
-			"Failed to verify sender's signature",
-		)
-		return false
+	msgHash := hash.Keccak256(messageBytes)
+	if !msgSig.VerifyHash(signerPubKey, msgHash[:]) {
+		return errors.New("failed to verify the signature")
 	}
-
-	return true
+	message.Signature = signature
```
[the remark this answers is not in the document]
LeoHChen: Not sure. that's the original code.

LeoHChen:
New PR rebased based on latest main branch. https://github.com/harmony-one/harmony/pull/3212

PR #3191, opened 2020-06-25
[rpc] change default transaction limit to 1000

gupadhyaya:
When no pageSize is passed from clients like explorer dashboard, it only lists 100 transactions on account page. By the specification, the default should be 1000.


gupadhyaya:
Screen Shot 2020-06-25 at 11 19 31 AM
Screen Shot 2020-06-25 at 11 20 20 AM

gupadhyaya:
@denniswon may be we should totally get rid of defaultSize and use `hashes.length` instead?

PR #3193, opened 2020-06-26
[consensus] replace infoMutex with atomic bool

LeoHChen:

### DIFF
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -160,9 +160,11 @@ func (consensus *Consensus) ResetState() {
 
 // ToggleConsensusCheck flip the flag of whether ignore viewID check during consensus process
 func (consensus *Consensus) ToggleConsensusCheck() {
-	consensus.infoMutex.Lock()
-	defer consensus.infoMutex.Unlock()
-	consensus.ignoreViewIDCheck = !consensus.ignoreViewIDCheck
+	if consensus.ignoreViewIDCheck.IsSet() {
+		consensus.ignoreViewIDCheck.UnSet()
+	} else {
+		consensus.ignoreViewIDCheck.Set()
+	}
 }
 
 // IsValidatorInCommittee returns whether the given validator BLS address is part of my committee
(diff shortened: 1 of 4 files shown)


consensus/consensus_service.go
```diff
@@ -160,9 +160,11 @@ func (consensus *Consensus) ResetState() {
 
 // ToggleConsensusCheck flip the flag of whether ignore viewID check during consensus process
 func (consensus *Consensus) ToggleConsensusCheck() {
-	consensus.infoMutex.Lock()
-	defer consensus.infoMutex.Unlock()
-	consensus.ignoreViewIDCheck = !consensus.ignoreViewIDCheck
+	if consensus.ignoreViewIDCheck.IsSet() {
```
JackyWYX, line 163: There is a potential read write race condition here. Please use atomic.Bool.Toggle. abool seems not supporting toggle implementation.

JackyWYX: Toggle implementation (from go-uber):

```
func (b *Bool) Toggle() bool {
	for {
		old := b.Load()
		if b.CAS(old, !old) {
			return old
		}
	}
}
```

LeoHChen: there is a fork with toggle support. forked it to our repo.

LeoHChen:
@JackyWYX , can you review and approve this one?

JackyWYX:
Where did you find the toggle implementation, there seems still to be a race condition here.

When you call `(*int32)(ab)`, it will create a new int32 and it has nothing to do with the original abool. If you use 

```
ab = ab.Toggle() 
``` 

in multiple threads, there could be read write race condition.

JackyWYX:
Boolean variable is incrementing for each `Toggle` command, thus `IsSet` is broken. Please fix `IsSet` function.

Also I would recommend not to return value for the `Toggle` method, since there might be potential race condition if you call

```
ab = ab.Toggle()
```

LeoHChen:
> Where did you find the toggle implementation, there seems still to be a race condition here.
> 
> When you call `(*int32)(ab)`, it will create a new int32 and it has nothing to do with the original abool. If you use
> 
> ```
> ab = ab.Toggle() 
> ```
> 
> in multiple threads, there could be read write race condition.

fair, I found the implementation in the original repo in a PR.

JackyWYX:
> > Where did you find the toggle implementation, there seems still to be a race condition here.
> > When you call `(*int32)(ab)`, it will create a new int32 and it has nothing to do with the original abool. If you use
> > ```
> > ab = ab.Toggle() 
> > ```
> > 
> > 
> > in multiple threads, there could be read write race condition.
> 
> fair, I found the implementation in the original repo in a PR.

The deleted reply has some mistakes. Please refer to my last reply.

PR #3201, opened 2020-07-01
Refactor quorum calculation and signature submission process

rlan35:
IsQuorumAchieved() is a pretty costly function as it loops through all votes and sum up the voting powers. It is called twice for every consensus messages, which can be a huge amount of CPU drain on the leader.

This PR refactor the whole leader consensus checking process to avoid redundant work on quorum checking and also BLS key serialization/deserialization.

Tested locally.

This PR is fully compatible in terms of onchain logic.

### DIFF
--- a/consensus/quorum/one-node-one-vote.go
+++ b/consensus/quorum/one-node-one-vote.go
@@ -4,6 +4,10 @@ import (
 	"encoding/json"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/harmony-one/bls/ffi/go/bls"
+	"github.com/harmony-one/harmony/consensus/votepower"
+
 	bls_cosi "github.com/harmony-one/harmony/crypto/bls"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/numeric"
@@ -19,6 +23,15 @@ type uniformVoteWeight struct {
 // Policy ..
 func (v *uniformVoteWeight) Policy() Policy {
 	return SuperMajorityVote
+}
+
+// AddNewVote ..
+func (v *uniformVoteWeight) AddNewVote(
+	p Phase, pubKeyBytes shard.BLSPublicKey,
+	sig *bls.Sign, headerHash common.Hash,
+	height, viewID uint64) (*votepower.Ballot, error) {
+
+	return v.SubmitVote(p, pubKeyBytes, sig, headerHash, height, viewID)
 }
 
 // IsQuorumAchieved ..
--- a/consensus/quorum/one-node-staked-vote.go
+++ b/consensus/quorum/one-node-staked-vote.go
@@ -2,13 +2,17 @@ package quorum
 
 import (
 	"encoding/json"
-	errors2 "errors"
 	"math/big"
+
+	"github.com/harmony-one/harmony/internal/utils"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/harmony-one/bls/ffi/go/bls"
+	"github.com/pkg/errors"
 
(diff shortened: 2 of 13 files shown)


consensus/quorum/one-node-one-vote.go
```diff
@@ -21,6 +25,15 @@ func (v *uniformVoteWeight) Policy() Policy {
 	return SuperMajorityVote
 }
 
+// AddNewVote ..
+func (v *uniformVoteWeight) AddNewVote(
+	p Phase, pubKeyBytes shard.BLSPublicKey,
```
JackyWYX, line 30: Name here can be a little frustrating. Naming with `Bytes` looks like a byte slice to me since built-in module `bytes` is the module of byte slice. Can we just name it as `pubKey`?

multibls/multibls.go
```diff
@@ -13,7 +15,8 @@ type PrivateKey struct {
 
 // PublicKey stores the bls public keys that belongs to the node
 type PublicKey struct {
-	PublicKey []*bls.PublicKey
+	PublicKey      []*bls.PublicKey
+	PublicKeyBytes []shard.BLSPublicKey
```
JackyWYX, line 19: I see why we are using Bytes in naming here. Can we rename PublicKey as `libPubKey` or something else?

consensus/view_change.go
```diff
@@ -432,7 +433,7 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 				priKey := consensus.priKey.PrivateKey[i]
 				if _, err := consensus.Decider.SubmitVote(
```
gupadhyaya, line 434: This should be `AddNewVote`? how is `SubmitVote` still works?

consensus/quorum/quorum.go
```diff
@@ -79,7 +79,7 @@ type ParticipantTracker interface {
 type SignatoryTracker interface {
 	ParticipantTracker
 	SubmitVote(
```
gupadhyaya, line 81: why you need both `SubmitVote` and `AddNewVote`?

consensus/view_change.go
gupadhyaya: I see... never mind.

gupadhyaya:
how about optimizing `computeTotalPowerByMask` as well? as `IsQuorumAchievedByMask` is also extensively called?

gupadhyaya:
so, I have to change the view change code to also use AddNewVote instead of SubmitVote and also fix the quorum to use vote instead of mask. Right?

consensus/quorum/one-node-staked-vote.go
```diff
@@ -24,49 +28,88 @@ type TallyResult struct {
...
+			return v.voteTally.Commit
+		case ViewChange:
+			return v.voteTally.ViewChange
+		default:
+			// Should not happen
+			return nil
+		}
+	}()
+
+	tallyQuorum.tally = tallyQuorum.tally.Add(additionalVotePower)
```
gupadhyaya, line 84: do you need to continue adding `additionalVotePower` and recomputing `quorumAchieved` if it is already met? No need right?

rlan35:
> let's optimize IsQuorumAchievedByMask as well? and skip checking quorumAchieved if it is already.

IsQuorumAchievedByMask can not be optimized because it's a one time checking. There is no cumulative process of adding votes one by one for validators

consensus/quorum/one-node-staked-vote.go
rlan35: additionalVotePower accumulation is still needed, but the quorumCheck can be avoided. I will do it in next PR.

multibls/multibls.go
rlan35: Will do more refactoring in next PR.

PR #3202, opened 2020-07-02
[node] replace node.State with node.IsInSync atomic bool

LeoHChen:
Tested on STN with rolling upgrade.

### DIFF
--- a/node/node_syncing.go
+++ b/node/node_syncing.go
@@ -243,23 +243,17 @@ func (node *Node) doSync(bc *core.BlockChain, worker *worker.Worker, willJoinCon
 	}
 	// TODO: treat fake maximum height
 	if node.stateSync.IsOutOfSync(bc) {
-		node.stateMutex.Lock()
-		node.State = NodeNotInSync
-		node.stateMutex.Unlock()
+		node.IsInSync.UnSet()
 		if willJoinConsensus {
 			node.Consensus.BlocksNotSynchronized()
 		}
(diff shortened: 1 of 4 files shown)


node/node_syncing.go
```diff
@@ -243,23 +243,17 @@ func (node *Node) doSync(bc *core.BlockChain, worker *worker.Worker, willJoinCon
...
+		node.IsInSync.UnSet()
 		if willJoinConsensus {
 			node.Consensus.BlocksNotSynchronized()
 		}
 		node.stateSync.SyncLoop(bc, worker, false, node.Consensus)
 		if willJoinConsensus {
-			node.stateMutex.Lock()
-			node.State = NodeReadyForConsensus
-			node.stateMutex.Unlock()
+			node.IsInSync.Set()
```
JackyWYX, line 252: Is it true that validator nodes are all `IsInSync == true` no matter it is synced or not?

LeoHChen: ReadyForConsensus is not checked at all in the code,  Basically, the code logic only cares about sync'ed or not.  I can remove this set though.  I actually tested it and found no issue for node to join back consensus even though this is removed. Kept here just for backward, I assume the original logic after `SyncLoop`, it will assume the node is in sync, at least give it a try to join the consensus.

JackyWYX: OK. I see. Thanks

LeoHChen:
Can you approve this one? 

It's tested on testnet and mainnet canary nodes already.

PR #3203, opened 2020-07-02
Refactor bls public key into a wrapper

rlan35:
Bind the C bls public key object with the serialized shard.BLSPublicKey bytes to avoid many unnecessary conversion between the two.

Tested locally.

### DIFF
--- a/consensus/double_sign.go
+++ b/consensus/double_sign.go
@@ -13,11 +13,11 @@ import (
 func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 	if consensus.couldThisBeADoubleSigner(recvMsg) {
 		if alreadyCastBallot := consensus.Decider.ReadBallot(
-			quorum.Commit, recvMsg.SenderPubkeyBytes,
+			quorum.Commit, recvMsg.SenderPubkey.Bytes,
 		); alreadyCastBallot != nil {
 			firstPubKey := bls.PublicKey{}
 			alreadyCastBallot.SignerPubKey.ToLibBLSPublicKey(&firstPubKey)
-			if recvMsg.SenderPubkey.IsEqual(&firstPubKey) {
+			if recvMsg.SenderPubkey.Object.IsEqual(&firstPubKey) {
 				for _, blk := range consensus.FBFTLog.GetBlocksByNumber(recvMsg.BlockNum) {
 					firstSignedBlock := blk.Header()
 					areHeightsEqual := firstSignedBlock.Number().Uint64() == recvMsg.BlockNum
@@ -45,7 +45,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 								Msg("could not read shard state")
 							return true
 						}
-						offender := shard.FromLibBLSPublicKeyUnsafe(recvMsg.SenderPubkey)
+						offender := shard.FromLibBLSPublicKeyUnsafe(recvMsg.SenderPubkey.Object)
 						if offender == nil {
 							consensus.getLogger().Error().
 								Str("msg", recvMsg.String()).
@@ -69,7 +69,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 							return true
 						}
 
-						leaderShardKey := shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
+						leaderShardKey := shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey.Object)
 						if leaderShardKey == nil {
 							consensus.getLogger().Error().
 								Str("msg", recvMsg.String()).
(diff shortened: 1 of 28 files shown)


shard/shard_state.go
```diff
@@ -37,8 +37,14 @@ type State struct {
 	Shards []Committee `json:"shards"`
 }
 
-// BLSPublicKey defines the bls public key
-// TODO(audit): wrap c bls key object with the raw bytes
+// BLSPublicKeyWrapper defines the bls public key in both serialized and
+// deserialized form.
+type BLSPublicKeyWrapper struct {
```
LeoHChen, line 42: The wrapper should probably be in a common package or bls package. Don't think it belongs to `shard` package.

rlan35: Agree. Will take care of it in next PR.

rlan35:
> This is definitely the right way to reduce unnecessary serialization/unserialization. Can we run pprof report on a leader node in testnet to identify the reduction of the cpu load with this change.

Any existing instruction on pprof?

node/node_handler.go
```diff
@@ -397,17 +399,22 @@ func (node *Node) VerifyNewBlock(newBlock *types.Block) error {
 
 func (node *Node) numSignaturesIncludedInBlock(block *types.Block) uint32 {
 	count := uint32(0)
-	pubkeys := node.Consensus.Decider.Participants()
-	mask, err := internal_bls.NewMask(pubkeys, nil)
+	members := node.Consensus.Decider.Participants()
```
gupadhyaya, line 402: from participants to mask could be cached per epoch. may be in the next PR.

consensus/view_change.go
```diff
@@ -132,14 +134,14 @@ func (consensus *Consensus) startViewChange(viewID uint64) {
...
 		Msg("[startViewChange]")
 
-	for i, key := range consensus.PubKey.PublicKey {
-		if !consensus.IsValidatorInCommittee(key) {
+	for i, key := range consensus.PubKey {
+		if !consensus.IsValidatorInCommittee(key.Bytes) {
 			continue
 		}
-		msgToSend := consensus.constructViewChangeMessage(key, consensus.priKey.PrivateKey[i])
+		msgToSend := consensus.constructViewChangeMessage(key.Object, consensus.priKey.PrivateKey[i])
```
gupadhyaya, line 144: can you pass `key.Bytes` instead of `key.Object` as internally it is serializing again?

node/node_handler.go
rlan35: yes. it's a todo.

consensus/view_change.go
rlan35: it's done in the next PR which refactor the private key.

consensus/fbft_log.go
```diff
@@ -326,17 +327,23 @@ func (consensus *Consensus) ParseNewViewMessage(msg *msg_pb.Message) (*FBFTMessa
 		utils.Logger().Warn().Err(err).Msg("ParseViewChangeMessage failed to parse senderpubkey")
 		return nil, err
 	}
-	FBFTMsg.SenderPubkey = pubKey
-	copy(FBFTMsg.SenderPubkeyBytes[:], vcMsg.SenderPubkey[:])
 
+	FBFTMsg.SenderPubkey = &shard.BLSPublicKeyWrapper{Object: pubKey}
+	copy(FBFTMsg.SenderPubkey.Bytes[:], vcMsg.SenderPubkey[:])
+
+	members := consensus.Decider.Participants()
```
gupadhyaya, line 334: can we also cache participants to mask per epoch? can be done in the next PR.

consensus/double_sign.go
```diff
@@ -45,7 +45,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 								Msg("could not read shard state")
 							return true
 						}
-						offender := shard.FromLibBLSPublicKeyUnsafe(recvMsg.SenderPubkey)
+						offender := shard.FromLibBLSPublicKeyUnsafe(recvMsg.SenderPubkey.Object)
```
gupadhyaya, line 48: leader is going to call `checkDoubleSign` on every commit, should we use the already available bytes in the SenderPubkey? Is unsafe call really required here?

consensus/double_sign.go
```diff
@@ -69,7 +69,7 @@ func (consensus *Consensus) checkDoubleSign(recvMsg *FBFTMessage) bool {
 							return true
 						}
 
-						leaderShardKey := shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey)
+						leaderShardKey := shard.FromLibBLSPublicKeyUnsafe(consensus.LeaderPubKey.Object)
```
gupadhyaya, line 72: same as above, can we use the Bytes directly?

consensus/double_sign.go
rlan35: Yup, should be optimized. Done.

consensus/double_sign.go
rlan35: done.

consensus/fbft_log.go
rlan35: yea, it will be done in next PR.

PR #3215, opened 2020-07-14
[Explorer] Add explorer node in sync check

janet-harmony:
[explorer port]/node-sync
Returns `true` if in sync, `false` if not
Status code 200 if in sync, 418 if not in sync, 500 if error

Example CURL request:
```
root@245a9d7580dc:~/go/src/github.com/harmony-one/harmony# curl -I http://localhost:5099/node-sync
HTTP/1.1 200 OK
Content-Type: application/json
Date: Tue, 14 Jul 2020 07:52:23 GMT
Content-Length: 5
```

Also minor typo fixes & export fixes found

### DIFF
--- a/api/service/explorer/service.go
+++ b/api/service/explorer/service.go
@@ -11,9 +11,10 @@ import (
 
 	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/gorilla/mux"
-	"github.com/harmony-one/abool"
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
+	"github.com/harmony-one/harmony/api/service/syncing"
 	"github.com/harmony-one/harmony/consensus/reward"
+	"github.com/harmony-one/harmony/core"
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/numeric"
 	"github.com/harmony-one/harmony/p2p"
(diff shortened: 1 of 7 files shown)


node/service_setup.go
```diff
@@ -45,7 +45,7 @@ func (node *Node) setupForExplorerNode() {
 	)
 	// Register explorer service.
 	node.serviceManager.RegisterService(
-		service.SupportExplorer, explorer.New(&node.SelfPeer),
+		service.SupportExplorer, explorer.New(&node.SelfPeer, node.IsInSync),
```
rlan35, line 48: this doesn't work well because IsInSync will only be updated every 60seconds

```
func (node *Node) DoSyncing(bc *core.BlockChain, worker *worker.Worker, willJoinConsensus bool) {
	ticker := time.NewTicker(time.Duration(SyncFrequency) * time.Second)
	for {
		select {
		case <-ticker.C:
			node.doSync(bc, worker, willJoinConsensus)
		case <-node.Consensus.BlockNumLowChan:
			node.doSync(bc, worker, willJoinConsensus)
		}
	}
}```

The best approach is still do the checking freshly when the RPC is called. Otherwise, the RPC can return stale state of the sync (meaning it's not sync'ed, but IsInSync == true because it's stale)

File: api/service/explorer/service.go [hunk 2]
```diff
@@ -41,11 +42,14 @@ type Service struct {
...
 	messageChan chan *msg_pb.Message
+	stateSync   *syncing.StateSync
+	blockchain  *core.BlockChain
 }
 
 // New returns explorer service.
-func New(selfPeer *p2p.Peer) *Service {
-	return &Service{IP: selfPeer.IP, Port: selfPeer.Port}
+func New(selfPeer *p2p.Peer, ss *syncing.StateSync, bc *core.BlockChain) *Service {
+	fmt.Printf("%+v\n", ss)
```
rlan35, line 51: remove test code plz.

api/service/syncing/syncing.go
```diff
@@ -787,6 +787,9 @@ func (ss *StateSync) IsSameBlockchainHeight(bc *core.BlockChain) (uint64, bool)
 
 // IsOutOfSync checks whether the node is out of sync from other peers
 func (ss *StateSync) IsOutOfSync(bc *core.BlockChain) bool {
+	if ss.syncConfig == nil {
```
rlan35, line 790: why did you add this check?

node/node_syncing.go
```diff
@@ -293,6 +289,11 @@ func (node *Node) SupportSyncing() {
 		go node.SendNewBlockToUnsync()
 	}
 
+	if node.stateSync == nil {
```
rlan35, line 292: why move this here?

PR #3216, opened 2020-07-14
[RPC] Fix block "miner" field

janet-harmony:
Fetch leader address from the API Backend Leader LRU cache.
If not found in cache, then search through committee slots until found.

Refactor to the `types.go` file to have all type definitions before functions

### DIFF
--- a/internal/hmyapi/apiv2/types.go
+++ b/internal/hmyapi/apiv2/types.go
@@ -39,7 +39,7 @@ type RPCBlock struct {
 	TransactionsRoot common.Hash      `json:"transactionsRoot"`
 	ReceiptsRoot     common.Hash      `json:"receiptsRoot"`
 	Transactions     []interface{}    `json:"transactions"`
-	StakingTxs       []interface{}    `json:"stakingTxs`
+	StakingTxs       []interface{}    `json:"stakingTxs"`
 	Uncles           []common.Hash    `json:"uncles"`
 	TotalDifficulty  *big.Int         `json:"totalDifficulty"`
 	Signers          []string         `json:"signers"`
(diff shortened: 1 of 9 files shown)


internal/hmyapi/apiv2/types.go
```diff
@@ -528,21 +540,16 @@ func newRPCStakingTransactionFromBlockIndex(b *types.Block, index uint64) *RPCSt
...
-	CirculatingSupply numeric.Dec `json:"circulating-supply"`
-	EpochLastBlock    uint64      `json:"epoch-last-block"`
-	TotalStaking      *big.Int    `json:"total-staking"`
-	MedianRawStake    numeric.Dec `json:"median-raw-stake"`
-}
+// getRPCLeader returns either the leader one address (before staking)
+// or the first 20 bytes of the leader bls public key (after staking)
+func getRPCLeader(h *block.Header, isStaking bool) string {
+	if isStaking {
+		return strings.ToLower(h.Coinbase().Hex())
```
rlan35, line 547: Let's derive the correct ONE1 address and report that instead of this random hex that people don't recognize.

you can refer to this code for the derivation:

```
for _, member := range committee.Slots {
		if isStaking {
			// After staking the coinbase address will be the address of bls public key
			if utils.GetAddressFromBLSPubKeyBytes(member.BLSPublicKey[:]) == header.Coinbase() {
				if err := member.BLSPublicKey.ToLibBLSPublicKey(committerKey); err != nil {
					return nil, err
				}
				return &bls.PublicKeyWrapper{Object: committerKey, Bytes: member.BLSPublicKey}, nil
			}```

PR #3217, opened 2020-07-14
Remove expensive bls key serialization code and use the cached version

rlan35:
Remove the usage of ToLibBLSPublicKey and always use BytesToBLSPublicKey which is the cached version.

Tested locally.

### DIFF
--- a/internal/hmyapi/apiv1/blockchain.go
+++ b/internal/hmyapi/apiv1/blockchain.go
@@ -6,11 +6,13 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/harmony-one/harmony/crypto/bls"
+
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/common/math"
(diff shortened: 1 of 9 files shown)


internal/hmyapi/apiv1/blockchain.go
```diff
@@ -261,8 +263,10 @@ func (s *PublicBlockChainAPI) GetBlockSigners(ctx context.Context, blockNr rpc.B
 		if err != nil {
 			return nil, err
 		}
-		blsPublicKey := new(bls.PublicKey)
-		validator.BLSPublicKey.ToLibBLSPublicKey(blsPublicKey)
+		blsPublicKey := new(bls_core.PublicKey)
```
gupadhyaya, line 266: these allocations are not necessary right?

rlan35: it is needed if you want to inline the error check. It shouldn't have performance difference as anyway a new pointer needs to be allocated. So I will leave as is.

rlan35:
This PR reduced block sync time for beacon chain from around 0.5s to around 0.25s per block

rlan35:
CPU usage before this PR:
image

After this PR:
image

Issue #3110, opened 2020-05-29
remove the node.sh wrapper


LeoHChen:
## Summary

node.sh wrapper was originally introduced to wrap around harmony binary because we don't have a static binary build when we launched the network.  However, since we have added statically linked binary and enabled it by default, it is time to remove the wrapper.

This is the meta ticket to keep track of all related tasks about this big project.

## Current Design

node.sh is a wrapper of harmony binary and do a lot of additional work like,

* download binary and verify using signature/md5sum
* set parameters for different networks, like testnet, stress net, etc
* environment setting using sysctl
* auto-update binary and restart

## Problems

node.sh becomes a monster wrapper and it is hard for new validators to use it. Also, it becomes error prune to use.

## Proposal

As suggested in this ticket, https://github.com/harmony-one/harmony/issues/3108

And I agree with the points made in the ticket.

* remove node.sh wrapper
* use config file
* No auto-update, once the network is stable enough.  This is a security risk with auto-update.

SebastianJ:
From the conversation on Discord / here's how I personally see we can do this:

1. Start off by rewriting the CLI args parsing to use Cobra
2. Change node.sh to interact with the new Cobra binary - we need to see that it'll continue to run properly for normal starts & stops + also auto-updates
3. Port the auto-updating section of node.sh to a separate script (which would be something optional for node runners to use)
4. Phase out auto-updating from node.sh - run multiple tests to see that everything works as expected
5. After multiple iterations and testing - finally, drop node.sh altogether

There's so much stuff tied into using node.sh since it has been around for so long (e.g. deployments etc) so we 1000% have to take a gradual approach for this.

I don't agree on the idea of introducing a json config file - that adds another dependency and potential issue for validators (they might end up messing up the formatting etc). Smart refactoring of the CLI args + binary should be enough.

JackyWYX:
> From the conversation on Discord / here's how I personally see we can do this:
> 
> 1. Start off by rewriting the CLI args parsing to use Cobra
> 2. Change node.sh to interact with the new Cobra binary - we need to see that it'll continue to run properly for normal starts & stops + also auto-updates
> 3. Port the auto-updating section of node.sh to a separate script (which would be something optional for node runners to use)
> 4. Phase out auto-updating from node.sh - run multiple tests to see that everything works as expected
> 5. After multiple iterations and testing finally drop node.sh altogether
> 
[...]

PR #3219, opened 2020-07-15
[cmd] Migrate BLS load logic to harmony binary.

JackyWYX:
## Overview

This is first of the PRs to address #3110. In this PR, the following is done:

1. Modularized BLS loading logic to blsgen package with a well-defined interface (see internal/blsgen/loader.go)
2. Rewrite the BLS loading logic previously both in node.sh and harmony binary. Hope the new code is lot much easier to follow.
3. Designed and updated main module with new flags used for BLS load. The new harmony command line is 100% compatible with old node.sh.
4. Updated node.sh file. Compared with old node.sh, removed `-M` flag and bls key auto detection (now default to .hmy/blskeys folder). 
5. Updated build_go_executable.sh script to enable it to compile main module instead of only main file.

## Some more elaborations on user interfaces

This PR has updated some user interfaces, including flags used in command line tool, and some updates in node.sh. Here we will elaborate a little bit.

### 1. Backward compatibility

A lot of efforts has been put to ensure the backward compatibility, so it is not required to update node.sh if upgraded harmony to a newer version:

1. The new binary is compatible with old node.sh.
2. The new node.sh has almost the same behavior as the old node.sh (If you do not change the command to start node.sh, it is definitely ok.) Changes about node.sh will be discussed later.

So users can be unaware about this PR change. You can see test process with details in test gist document.

### 2. Updates to node.sh

Following changes has been made for node.sh

1. There is no more -M flag for multi bls bool. 
    * multibls is enabled as default, which will load keys in `.hmy/blskeys` folder.
    * If user input a -M flag, the program will tell user that -M is deprecated, but will not throw an error.
2. Now multiple keys can be parsed to `-k` flag. E.g. `-k mybls1.key,mybls2.key`
    * If user try to run node.sh with `-k mybls.key` flag, `mybls.key` file will be the bls key source.
3. Persist bls passphrase after prompt is disabled for security concern.
    * User can always prepare a `mybls.pass` file in the directory of the key file. However, persist after prompt is not supported in node.sh.
    * Persist after prompt is only supported in binary start up. If needed, just adding a new flag in node.sh can persist passphrase after success unlock.
4. For systemd programs that do not want to use interactive prompt for passphrase, can run with a newly added `-C` flag to disable interactive passphrase prompt.
5. Key file extension is crucial after this change. Need to make sure that key file end with `.key` if want to unlock with passphrase.
    * Extension `.key` will use passphrase decryption.
    * Extension `.bls` will use kms decryption.

### 3. Updates to harmony binary cli interface

1. Added flag `--save-passphrase` to save passphrase after prompt. If a pass file already exist, will prompt to ask whether overwrite.
2. Changed meaning of flag `--blspass`, specifies the source of the BLS passphrase. Now it has options: `default, no-prompt, prompt, file:$CONFIG_FILE, none`
    * default: first try to decrypt key with existing .pass file. If not exist or passphrase is wrong, prompt user to input passphrase.
    * no-prompt: Only decrypt key with existing .pass file.
    * prompt: Do not use file, use prompt only.
    * file:$PASS_FILE: Use a specific pass file for all keys.
    * none: Disable passphrase decryption. 
3. Added new flag `--save-passphrase`. If set, the passphrase will be persisted to disk after successfully unlock the key from user prompt.
4. Added new flag `--aws-config-source`, specify the aws config and credentials for kms encrypted keys. Options: default, prompt, file:$CONFIG_FILE, none)
    * default: Use default aws shared session config.
    * prompt: Use prompt for needed input fields.
    * file: Use a json file for aws session config.
    * none: Disable KMS decryption

### 1. Unit Test Coverage

Test coverage for `internal/blsgen`

Before:

```
coverage: 34.1% of statements
```

After:

```
coverage: 67.8% of statements
```

### Test/Run Logs

[Test log](https://gist.github.com/JackyWYX/b3c878bec750372170de8b86ac48a581)

Three overall pictures tested:
1. New binary with old node.sh.
2. New binary with new node.sh.
3. New added flags that's not included in node.sh.

## Operational Checklist

    **NO**

    **NO**

   **YES**

    We would encourage operators to update node.sh and not enforceable. After the update, -M flag is recommended to be removed from command line. (If you don't change, it can still work)

  **NO**

## TODO

Some updates on the node.sh documentation need to be updated. (Remove all -M flags)


JackyWYX:
@LeoHChen @rlan35 Would you please review this when you have some time?

scripts/node.sh
```diff
@@ -317,6 +318,8 @@ do
    R) TRACEFILE="${OPTARG}";;
    l) broadcast_invalid_tx=false;;
    L) log_level="${OPTARG}";;
+
+   M) msg "WARNING: deprecated flag -M: Multi BLS is always enabled, and it's safe to remove this flag.";;
```
LeoHChen, line 322: let's not remove this flag, as our internal nodes are using -M by default.

JackyWYX: I definitely agree with you. So -M flag is not removed but only deprecated. The node.sh will just run as it supposed to.

JackyWYX:
> We need to update our systemd service when we deploy this change to add `-C` option to avoid manual input of passphrase of the blskey. This also needs to broadcast to all validators. Please prepare a proposal in talk.harmony.one and ask for input from validators.

1. Since the new binary is 100% compatible with old node.sh (see my tests), so it is not required to update node.sh at this stage. 
2. I definitely agree with the -C flag with systemd, which is my initial intention to build this flag. But if updating systemd command is expensive, we can just leave it as it is for now - the node should behave exactly the same as wrong pass file. I can double check later of the wrong pass file behavior in systemd to confirm this. If that's true, it will be recommended, but not required to add -C flag.
3. The proposal in talk.harmony is really a good suggestion. I will do tomorrow.

JackyWYX:
> I definitely agree with running the -C flag with systemd, which is my initial intention to build this flag. But if updating systemd command is expensive, we can just leave it as it is for now - the node should behave exactly the same as wrong pass file. I can double check the behavior later to confirm this.

Just ran the the node.sh through systemd, Result as expected:

```
Jul 18 21:42:52 ip-172-31-12-180 node.sh: node.sh: ############### Running Harmony Process ###############
Jul 18 21:42:52 ip-172-31-12-180 node.sh: open .hmy/blskeys/3a471dea9e743b5a4ec777d192ce42c9f15564e16d2a7cb201506365b89ad15a62d5fbe885ec29e22ce98dfb8b268319.pass: no such file or directory
Jul 18 21:42:52 ip-172-31-12-180 node.sh: Enter passphrase for the BLS key file .hmy/blskeys/3a471dea9e743b5a4ec777d192ce42c9f15564e16d2a7cb201506365b89ad15a62d5fbe885ec29e22ce98dfb8b268319.key:unable to read from prompt: inappropriate ioctl for device
Jul 18 21:42:52 ip-172-31-12-180 node.sh: ERROR when loading bls key: 0 bls keys loaded
Jul 18 21:42:52 ip-172-31-12-180 node.sh: node.sh: node process finished with status 100
Jul 18 21:42:52 ip-172-31-12-180 node.sh: node.sh: exiting
```

We can see there is no ioctl provided thus an immediate failure is expected. So no operation on systemd command update. Hooray! :)

LeoHChen:
> > We need to update our systemd service when we deploy this change to add `-C` option to avoid manual input of passphrase of the blskey. This also needs to broadcast to all validators. Please prepare a proposal in talk.harmony.one and ask for input from validators.
> 
> 1. Since the new binary is 100% compatible with old node.sh (see my tests), so it is recommended but not required to update node.sh at this stage. Actually all changes in this PR is a "Soft-fork".
> 2. I definitely agree with running the -C flag with systemd, which is my initial intention to build this flag. But if updating systemd command is expensive, we can just leave it as it is for now - the node should behave exactly the same as wrong pass file. I can double check the behavior later to confirm this. If that's true, it will be only recommended, but not required to add -C flag.
> 3. The proposal in talk.harmony is really a good suggestion. I will make the talk tomorrow.



> > I definitely agree with running the -C flag with systemd, which is my initial intention to build this flag. But if updating systemd command is expensive, we can just leave it as it is for now - the node should behave exactly the same as wrong pass file. I can double check the behavior later to confirm this.
> 
> Just ran the the node.sh through systemd, Result as expected:
> 
> ```
> Jul 18 21:42:52 ip-172-31-12-180 node.sh: node.sh: ############### Running Harmony Process ###############
> Jul 18 21:42:52 ip-172-31-12-180 node.sh: open .hmy/blskeys/3a471dea9e743b5a4ec777d192ce42c9f15564e16d2a7cb201506365b89ad15a62d5fbe885ec29e22ce98dfb8b268319.pass: no such file or directory
> Jul 18 21:42:52 ip-172-31-12-180 node.sh: Enter passphrase for the BLS key file .hmy/blskeys/3a471dea9e743b5a4ec777d192ce42c9f15564e16d2a7cb201506365b89ad15a62d5fbe885ec29e22ce98dfb8b268319.key:unable to read from prompt: inappropriate ioctl for device
> Jul 18 21:42:52 ip-172-31-12-180 node.sh: ERROR when loading bls key: 0 bls keys loaded
> Jul 18 21:42:52 ip-172-31-12-180 node.sh: node.sh: node process finished with status 100
> Jul 18 21:42:52 ip-172-31-12-180 node.sh: node.sh: exiting
> ```
> 
> We can see there is no ioctl provided thus an immediate failure is expected. So no operation on systemd command update. Hooray! :)

So, as long as the pass file is provided, there should be no errors, right?  If that's the case, we don't need to update the systemd service file.  Please test it on our testnet or STN.

JackyWYX:
> So, as long as the pass file is provided, there should be no errors, right? If that's the case, we don't need to update the systemd service file. Please test it on our testnet or STN.

Yes, exactly. Do you mean updating the whole network nodes or just a single node? Where can I find the handbook for this?

JackyWYX:
Do as this handbook to deploy on testnet. 

First deploy a single ip `40.69.201.193`

JackyWYX:
Then try to upgrade shard 1.

Step 1: Rolling upgrade validators

All nodes except leader node is updated. All goes well.

Step 2: Upgrade leader

Hooray observed. Not even a view change. 

Shard 1 upgrade finished.



Step 2: Force update leader

JackyWYX:
Next step, upgrade shard 0.

Successfully updated the whole network. No interruption observed. Looks good.

internal/blsgen/kms.go
```diff
@@ -0,0 +1,279 @@
+package blsgen
```
rlan35, line 1: is currently anyone using kms?

JackyWYX: None I believe.

PR #3224, opened 2020-07-15
[node.sh] Remove self ip probe in node.sh

JackyWYX:
## Issue

This is another PR for #3110. Remove the self IP probe logic, and use 127.0.0.1 as default is fine. Libp2p will auto detect IP to listen.

## Test

Tested with branch: https://github.com/JackyWYX/harmony/tree/nodesh_remove_ip_test, in which added some print message when handling pubsub message. 

1. When starting with new node.sh (without specifying ip), pubsub message is correctly listened from print message, meaning the p2p network is joined.
2. The test is done both on local macbook and AWS instance.

## Operational Checklist

    **NO**

    **NO**

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -36,34 +36,6 @@ random() {
    echo $(( b + rand ))
 }
 
-# https://www.linuxjournal.com/content/validating-ip-address-bash-script
-function valid_ip()
-{
-    local  ip=$1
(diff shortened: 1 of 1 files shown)


JackyWYX:
The IP is still used for public endpoint, and we shall not remove it at this stage. Will be changed in later PRs.

Issue #3225, opened 2020-07-15
[optimization] duplicated commit message signature verification consumed quite a lot of CPU on leader


LeoHChen:
## Summary

Seemingly duplicated commit message signature verification consumed lots of CPU.

## Current Design

When the leader is handling the consensus messages, the leader always do a senderKeySanityChecks call for every message,
https://github.com/harmony-one/harmony/blob/14e017defc18d712a854f70104e173e6eea9c058/consensus/checks.go#L41

However, during the process of commit/prepare message, the leader will check the signature of the commit messages again, 
https://github.com/harmony-one/harmony/blob/14e017defc18d712a854f70104e173e6eea9c058/consensus/leader.go#L192

## Problems

Duplicated deserialized and verification consumed a lot of leader's CPU, which leads to slow consensus and can't collect 100% of signatures from all validators on time.

## Proposal

Deep dive and reduce duplicated signature verification.

rlan35:
The two signatures are serving different purpose. Once is signing on the p2p message to make sure that the sender is legit. the other signature is signed on the block to make sure it's a valid consensus signature.

We can think about clever way to avoid the first type of signature checking possibly by tagging on the second type.

LeoHChen:
Thanks @rlan35  for the explanation.  So, the main purpose of the first check is to make sure the message sender is legit.

In this regard, I don't think we need to verify or even do the first signature to serve the purpose.  As our network is an open network, anyone can send any messages to the p2p network, we didn't have any protection before at p2p layer. The verification logic is already in the application layer.

Also, the current p2p messages are encrypted already at p2p layer.  If we intend to filter out invalid senders, it should be done at the peer discovery service to prevent them from connecting to the p2p network. Like using some advanced logic at boot node.

Since the signature verification is quite heavy using bls signing, it is actually becoming the weakest link if we have to verify every single message.  A valid defense in this case would be like PoW. Taking quite some effort to sign the message and minimal effort to verify it.  Right now, in our case, to verify the sign take most of the time/cpu as RJ benchmarked in https://github.com/harmony-one/harmony/pull/3211, which didn't' serve the purpose of defense.

So, I suggest we shall remove the first sanity check and only keep the second one to make sure the payload is signed properly.  This can save quite some CPU time of leader, especially during the final commit phase.

LeoHChen:
And we already have the 2nd check, which means illegal sender still can't compromise the consensus.  That's good enough.

LeoHChen:
[...]

PR #3226, opened 2020-07-16
[p2p] add a timed blocklist

LeoHChen:
blocklist peers not in the current committee for 120 seconds.

there is a risk that internal nodes may still
use the keys that are not in the current committee
if we increase the slot from 480 to 640.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -420,7 +420,7 @@ func (node *Node) validateShardBoundMessage(
 
 	if !node.Consensus.IsValidatorInCommittee(senderKey) {
 		atomic.AddUint32(&node.NumSlotMessages, 1)
-		return nil, nil, true, errors.WithStack(shard.ErrValidNotInCommittee)
+		return nil, nil, true, shard.ErrValidNotInCommittee
 	}
 
 	// ignore mesage not intended for validator
@@ -561,7 +561,15 @@ func (node *Node) Start() error {
 					)
 
 					if err != nil {
-						errChan <- withError{err, msg.GetFrom()}
+						// block the peer if it is not in the current committee
+						// FIXME: this is for testing only, will need to enable it after we increase the slot to 640
+						// as the internal nodes may contain invalid keys and we don't want to block internal nodes
+						if err == shard.ErrValidNotInCommittee {
+							node.host.PubSub().BlacklistPeer(msg.GetFrom())
+							utils.Logger().Info().Str("peer", msg.GetFrom().String()).Msg("BlacklistPeer")
+						} else {
+							errChan <- withError{err, msg.GetFrom()}
+						}
 						return libp2p_pubsub.ValidationReject
 					}
 
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -9,6 +9,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/harmony-one/bls/ffi/go/bls"
 	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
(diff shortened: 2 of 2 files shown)


node/node.go
```diff
@@ -561,7 +561,15 @@ func (node *Node) Start() error {
 					)
 
 					if err != nil {
-						errChan <- withError{err, msg.GetFrom()}
+						// block the peer if it is not in the current committee
+						// FIXME: this is for testing only, will need to enable it after we increase the slot to 640
+						// as the internal nodes may contain invalid keys and we don't want to block internal nodes
+						if err == shard.ErrValidNotInCommittee {
```
JackyWYX, line 567: Why are we only blocking this specific error? Maybe we should also block invalid signature and more?

JackyWYX:
I see the risk of committee change which might even cause system break in that specific epoch change. But there is also risks for any epoch change since it could happen that new validators is not acknowledged by other validators. I would recommend also to test the epoch change scenario as well.

JackyWYX:
Also some comments on previous PRs:

1. node/node.go:555: If node is not consensus bound, sould return `ValidationAccept`?
2. node/node.go:541: If node is delivering empty message like spamming, should we return `ValidationReject` instead?

LeoHChen:
> Also some comments on previous PRs:
> 
> 1. node/node.go:555: If node is not consensus bound, sould return `ValidationAccept`?
> 2. node/node.go:541: If node is delivering empty message like spamming, should we return `ValidationReject` instead?
> 3. Shall we verify signature in p2p message validation before entering consensus?

1) this is not a risk as we already switch case the messageCategory before the validation. https://github.com/harmony-one/harmony/blob/14e017defc18d712a854f70104e173e6eea9c058/node/node.go#L547

2) agree, we shall probably reject empty message as well

3) verify the signature of p2p message is very expensive, also, as I raised the question of the duplicated signature verification in https://github.com/harmony-one/harmony/issues/3225,  I think we shall not do signature verification at p2p layer.

LeoHChen:
> I see the risk of committee change which might even cause system break in that specific epoch change. But there is also risks for any epoch change since it could happen that new validators is not acknowledged by other validators. I would recommend also to test the epoch change scenario as well.

definitely will test on the epoch change scenario. For epoch change, there may be a risk, when the leader already moved to a new epoch, and validators still in the previous epoch, and keep sending messages signed with old keys, thus may be banned in the network for some time.  However, unless all nodes are doing the same kind of blskey manipulation, this shouldn't break the network.

Also, the block is timed, we can change it like 1 minute or so to shorter the block time.

node/node.go
LeoHChen: Yes, we can also block other errors.  This is the initial intention as we have seen many error messages like this. Ofc, to protect the network, we shall also block on other errors.

JackyWYX:
> verify the signature of p2p message is very expensive, also, as I raised the question of the duplicated signature verification in #3225, I think we shall not do signature verification at p2p layer.

I see your point. But I wonder whether we have done any spamming where attackers send consensus message with public key in committee but invalid signature?

rlan35:
> > I see the risk of committee change which might even cause system break in that specific epoch change. But there is also risks for any epoch change since it could happen that new validators is not acknowledged by other validators. I would recommend also to test the epoch change scenario as well.
> 
> definitely will test on the epoch change scenario. For epoch change, there may be a risk, when the leader already moved to a new epoch, and validators still in the previous epoch, and keep sending messages signed with old keys, thus may be banned in the network for some time. However, unless all nodes are doing the same kind of blskey manipulation, this shouldn't break the network.
> 
> Also, the block is timed, we can change it like 1 minute or so to shorter the block time.

I will definitely cause trouble during epoch transition. We should change the logic to accept both new epoch and last epoch's keys. More optimized way is to accept last epoch's key within a certain range of time after entering new epoch. But I think simply accept new and last epoch should be fine.

PR #3227, opened 2020-07-16
[p2p] higher weight to process node messages

LeoHChen:
Total number of weight is 8192+8192=16384

weight of consensus message is 1
weight of node message / tx is 64

This is to prevent massive number of tx messages from getting the node OOM.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -63,6 +63,10 @@ const (
 	GlobalRxWorkers = 32
 	// MsgChanBuffer is the buffer of consensus message handlers.
 	MsgChanBuffer = 1024
+	// WeightConsensus is the weight of semaphore for consensus messages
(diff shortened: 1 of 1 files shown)


JackyWYX:
I would recommend to use two different `sem` to process `consensus` and `node` message, thus consensus protocol will not be effected by E.g. transaction spamming.

rlan35:
> I would recommend to use two different `sem` to process `consensus` and `node` message, thus consensus protocol will not be effected by E.g. transaction spamming.

Agree, if we give node message 64, then certainly people can spam with node message and block consensus message from being processed.

LeoHChen:
replaced by https://github.com/harmony-one/harmony/pull/3232

PR #3228, opened 2020-07-16
[consensus] remove duplicated signature verification

LeoHChen:
Check the following issue on the detailed explanation of why this commit makes sense.
https://github.com/harmony-one/harmony/issues/3225

The onPrepare and onCommit will do signature verification of the message payload.
So, there is no need to do duplicated sanity check when handling Prepare/Commit messages.

We reserve the signature verification on other messages, as they are not duplicated.
This PR can reduce the CPU load of leader during consensus.

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -204,11 +204,6 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 	consensus.mutex.Lock()
 	defer consensus.mutex.Unlock()
 
-	// Check for potential double signing
-	if consensus.checkDoubleSign(recvMsg) {
-		return
-	}
-
 	validatorPubKey, commitSig, commitBitmap :=
 		recvMsg.SenderPubkey, recvMsg.Payload, consensus.commitBitmap
 	logger := consensus.getLogger().With().
@@ -241,6 +236,11 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 
 	if !sign.VerifyHash(recvMsg.SenderPubkey.Object, commitPayload) {
 		logger.Error().Msg("[OnCommit] Cannot verify commit message")
+		return
+	}
+
+	// Check for potential double signing
+	if consensus.checkDoubleSign(recvMsg) {
 		return
 	}
 
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -79,14 +79,8 @@ func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb
 
 	// Handle leader intended messages now
 	case t == msg_pb.MessageType_PREPARE && intendedForLeader:
-		if !consensus.senderKeySanityChecks(msg, senderKey) {
-			return errVerifyMessageSignature
-		}
 		consensus.onPrepare(msg)
 	case t == msg_pb.MessageType_COMMIT && intendedForLeader:
-		if !consensus.senderKeySanityChecks(msg, senderKey) {
-			return errVerifyMessageSignature
-		}
 		consensus.onCommit(msg)
 
 		// Handle view change messages


gupadhyaya:
big assumption, but intention looks correct. however, I think a malicious validator can fool the leader in `checkDoubleSign` validation. here is how: a malicious validator will modify the commit message to make it look like double sign and send it on behalf of some other validator. since we didn't check that commit message was actually from the claimed sender (note that this part comes later, `checkDoubleSign` is before that), now the leader will falsely mark a genuine validator as double signing validator, hence banning him. 

may be moving the `checkDoubleSign` after the verifyHash could fix this issue. but needs a careful look.

LeoHChen:
> big assumption, but intention looks correct. however, I think a malicious validator can fool the leader in `checkDoubleSign` validation. here is how: a malicious validator will modify the commit message to make it look like double sign and send it on behalf of some other validator. since we didn't check that commit message was actually from the claimed sender (note that this part comes later, `checkDoubleSign` is before that), now the leader will falsely mark a genuine validator as double signing validator, hence banning him.
> 
> may be moving the `checkDoubleSign` after the verifyHash could fix this issue. but needs a careful look.

Thanks for the comments and good insight, I didn't look closely on the double-sign checking though.  Let me dive deeper.

LeoHChen:
@gupadhyaya , @rlan35 , please review again.

consensus/leader.go
```diff
@@ -244,6 +239,11 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 		return
 	}
 
+	// Check for potential double signing
+	if consensus.checkDoubleSign(recvMsg) {
```
rlan35, line 243: So this won't simply work. Actually lots of logic reply on the fact the p2p message are verified and legit. If people now can manipulate the message since we don't check sig anymore, then people can attack.
image

For example, the block hash in the p2p message are used for double sign check.

rlan35:
instead of not doing any sanity check on p2p message with signatures, we should try explore p2p level message checking to make sure it's not tampered. P2p

consensus/leader.go
LeoHChen: the block signature is verified in onCommit https://github.com/harmony-one/harmony/blob/6029ba5cf9a608aad9256b212a319d201dc4512c/consensus/leader.go#L242

that VerifyHash verified the blocknum, viewed, and hash already, so the messages verified.  As I mentioned, the p2p message level verification only have 2 additional info, msg type and service type, so why do you think the p2p messages was tampered?

LeoHChen: noted the commitPayload was created from `recvMsg`, which is copied from `msg` p2p data.  It is the signature verification of the p2p message payload.

gupadhyaya:
@LeoHChen `VerifyHash` is not sufficient to guarantee that the tampered receive message will have no harm. It is still possible to make the leader think that a genuine validator has double signed by keeping the blockHash and payload intact, but tampering the blockNum and viewID (checkDoubleSign lines 25&26). I suggest carefully evaluating all fields of the received message to see the impact of tampering and making the fixes.

LeoHChen:
After discussion with @gupadhyaya , I realized that the 2nd verification is actually not needed.  We can just do some direct comparison to make sure the recvMsg has signed the right commit message, similar to the check we did in double-sign checking. Thus to avoid another signature verification, which is quite expensive.

LeoHChen:
We shall still need the first signature verification though to make sure the p2p message is from the legit validator w/o tempering.

consensus/leader.go
rlan35: You are right. The block signature verification here actually made sure the blockHash and blocknum viewid are all good and verified. Resolved.

rlan35:
I just did round of checking on the p2p message fields and how they are used. Seems for leader, this change will prevent potential problems. But for validators, there are cases that can cause issue:
image

Here the BlockNum is trusted and used to check sync status. Here we should use the BlockNum associated with the actual block (blockObj.NumberU64())

And for other parts, mostly the issue is with ViewID, since we have view change, so the sender commit to a viewID by signing on it. Before, we rely on the signature on the p2p message to commit to the viewID. So if we remove it, attacker can mess up the viewID of messages and cause confusion.

rlan35:
And for validator OnAnnounce. The message fields are directly used for serious checks that may trigger view change. e.g in onAnnounceSanityChecks.

Actually for OnAnnounce, since there is no block data to check (block data is in Prepared message). The p2p message signature is the only verification to verify it's the real leader. We should keep the sig checking for onAnnounce then.

rlan35:
> And for validator OnAnnounce. The message fields are directly used for serious checks that may trigger view change. e.g in onAnnounceSanityChecks.
> 
> Actually for OnAnnounce, since there is no block data to check (block data is in Prepared message). The p2p message signature is the only verification to verify it's the real leader. We should keep the sig checking for onAnnounce then.

Nevermind these comments. The sig checks are only removed for leader logic. Then it looks good.

rlan35:
> @LeoHChen `VerifyHash` is not sufficient to guarantee that the tampered receive message will have no harm. It is still possible to make the leader think that a genuine validator has double signed by keeping the blockHash and payload intact, but tampering the blockNum and viewID (checkDoubleSign lines 25&26). I suggest carefully evaluating all fields of the received message to see the impact of tampering and making the fixes.

blockNum and ViewID are checked in isRightBlockNumAndViewID, and BlockHash are verified with the sig, so they are all good.

PR #3229, opened 2020-07-17
Remove all wrong usage of singlr flight; also fix voting power cache key

rlan35:
Single flight is meant for duplicate func call avoidance (same-time duplicate calls). it's not cache. The func call results are not cached (https://medium.com/@vCabbage/go-avoid-duplicate-requests-with-sync-singleflight-311601b3068b)

Found the issue after seeing lots of accumulated pending goroutines in our node. They are caused by the sleep() and forget() part of singleFlight. Since it's not cache, actually the func got called many times: 
image

Besides the rpc layer, all other part of the singleFlight usage are wrong. Should instead just use lru cache.

This PR also fix the cache key on voting power cache which was hidden due to that single flight didn't do the cache work for voting power.

Tested locally.

### DIFF
--- a/internal/chain/reward.go
+++ b/internal/chain/reward.go
@@ -5,6 +5,8 @@ import (
 	"math/big"
 	"sort"
 
+	lru "github.com/hashicorp/golang-lru"
+
 	"github.com/harmony-one/harmony/numeric"
 	types2 "github.com/harmony-one/harmony/staking/types"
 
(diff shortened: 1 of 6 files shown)


internal/chain/reward.go
```diff
@@ -45,34 +46,35 @@ func ballotResultBeaconchain(
...
-			if err != nil {
-				return nil, err
-			}
-
-			// For new calc, remove old data from 3 epochs ago
-			deleteEpoch := big.NewInt(0).Sub(epoch, big.NewInt(3))
-			deleteKey := fmt.Sprintf("%s-%d", deleteEpoch.String(), subComm.ShardID)
-			votingPowerCache.Forget(deleteKey)
+	// Look up
+	key := fmt.Sprintf("%s-%d", epoch.String(), subComm.Hash())
```
JackyWYX, line 57: format `"%s-%x"`?

internal/chain/reward.go
```diff
@@ -45,34 +46,35 @@ func ballotResultBeaconchain(
...
+	fmt.Println(key)
+	if b, ok := votingPowerCache.Get(key); ok {
+		if b == nil {
+			fmt.Println("DAMN")
```
LeoHChen, line 61: log this error please

internal/chain/reward.go
```diff
@@ -45,34 +46,35 @@ func ballotResultBeaconchain(
 }
 
 var (
-	votingPowerCache   singleflight.Group
-	delegateShareCache singleflight.Group
+	votingPowerCache, _   = lru.New(16)
```
LeoHChen, line 49: why 16 here?  can you elaborate it a bit?

internal/chain/reward.go
```diff
@@ -45,34 +46,35 @@ func ballotResultBeaconchain(
...
+	votingPower, err := votepower.Compute(subComm, epoch)
 	if err != nil {
 		return nil, err
 	}
 
-	return results.(*votepower.Roster), nil
+	// Put in cache
+	votingPowerCache.Add(key, votingPower)
+	if votingPower == nil {
+		fmt.Println("DAMN2")
```
LeoHChen, line 75: log the error please

internal/chain/reward.go
```diff
@@ -81,40 +83,35 @@ func lookupDelegatorShares(
...
+	}
 
-			for i := range validatorSnapshot.Delegations {
-				delegation := validatorSnapshot.Delegations[i]
-				// NOTE percentage = <this_delegator_amount>/<total_delegation>
-				percentage := numeric.NewDecFromBigInt(delegation.Amount).Quo(totalDelegationDec)
-				result[delegation.DelegatorAddress] = percentage
-			}
+	// If not found, construct
+	result := map[common.Address]numeric.Dec{}
```
LeoHChen, line 95: will the map cause concurrency issue?

LeoHChen: do you need to protect it?

internal/chain/reward.go
rlan35: sorry, those are debug code. Forgot to remove. Removed.

internal/chain/reward.go
rlan35: done.

internal/chain/reward.go
rlan35: votingPower is the calcuation per committee (or per epoch/shard). so 16 is good enough for multiple epochs.

internal/chain/reward.go
rlan35: this map is atomically created and only read, never updated. So no concurrency issue.

rlan35:
tested on a node connected to mainnet shard 0. The goroutine usage all looks normal now.

PR #3238, opened 2020-07-18
[p2p] node message validation

LeoHChen:

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -386,8 +386,7 @@ func (node *Node) validateNodeMessage(ctx context.Context, payload []byte) (
 		switch proto_node.BlockMessageType(payload[p2pNodeMsgPrefixSize]) {
 		case proto_node.Sync:
 			// only non-beacon nodes process the beacon block sync messages
-			if node.Blockchain().ShardID() == shard.BeaconChainShardID ||
-				node.NodeConfig.Role() == nodeconfig.ExplorerNode {
+			if node.Blockchain().ShardID() == shard.BeaconChainShardID {
 				atomic.AddUint32(&node.NumInvalidNodeMsg, 1)
 				return nil, 0, errIgnoreBeaconMsg
 			}
--- a/node/node.go
+++ b/node/node.go
@@ -128,6 +128,17 @@ type Node struct {
 	NumInvalidMessages uint32
 	NumSlotMessages    uint32
 	NumIgnoredMessages uint32
+
+	// metrics of node messages
+	NumTotalNodeMsg   uint32
+	NumInvalidNodeMsg uint32
+	NumOversizedMsg   uint32
+	NumTransactionMsg uint32
+	NumStakingMsg     uint32
+	NumBlockSyncMsg   uint32
+	NumSlashMsg       uint32
+	NumReceiptMsg     uint32
+	NumCrossLinkMsg   uint32
 }
 
 // Blockchain returns the blockchain for the node's current shard.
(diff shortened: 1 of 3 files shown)


node/node.go
```diff
@@ -568,13 +646,28 @@ func (node *Node) Start() error {
...
 						consensusBound: false,
 						handleE:        node.HandleNodeMessage,
-						handleEArg:     openBox,
+						handleEArg:     validMsg,
+						actionType:     actionType,
 					}
+					return libp2p_pubsub.ValidationAccept
 				default:
+					// ignore garbled messages
+					atomic.AddUint32(&node.NumIgnoredMessages, 1)
```
JackyWYX, line 670: Shall this message be rejected since it does not have valid message type?

node/node.go
```diff
@@ -344,8 +355,72 @@ var (
...
 	errNoSenderPubKey  = errors.New("no sender public BLS key in message")
 	errWrongShardID    = errors.New("wrong shard id")
+	errInvalidNodeMsg  = errors.New("invalid node message")
+	errIgnoreBeaconMsg = errors.New("ignore beacon sync block")
 )
 
+// validateNodeMessage validate node message
+func (node *Node) validateNodeMessage(ctx context.Context, payload []byte) (
+	[]byte, proto_node.MessageType, error) {
+	atomic.AddUint32(&node.NumTotalNodeMsg, 1)
```
JackyWYX, line 365: Is it true that these metrics is only increasing? Is there any possibility that integer overflow might happen?

node/node.go
LeoHChen: good point, will do.

SebastianJ:
@LeoHChen: Stress testing report and findings reported in separate issue/report in priv repo

node/node.go
LeoHChen: The metrics will be reset every 1 minute, so no worries.

LeoHChen:
> @LeoHChen: Stress testing report and findings reported in separate issue/report in priv repo

Acknowledged.  Need to tune more parameters though.  This validation  shouldn't add additional risk though.

node/node.go
```diff
@@ -344,8 +355,72 @@ var (
...
+		atomic.AddUint32(&node.NumTransactionMsg, 1)
+	case proto_node.Staking:
+		// nothing much to validate staking message unless decode the RLP
+		atomic.AddUint32(&node.NumStakingMsg, 1)
+	case proto_node.Block:
+		switch proto_node.BlockMessageType(payload[p2pNodeMsgPrefixSize]) {
+		case proto_node.Sync:
+			// only non-beacon nodes process the beacon block sync messages
+			if node.Blockchain().ShardID() == shard.BeaconChainShardID ||
+				node.NodeConfig.Role() == nodeconfig.ExplorerNode {
```
rlan35, line 390: explorer node also need beacon sync.

node/node.go
```diff
@@ -344,8 +355,72 @@ var (
...
+			if node.NodeConfig.ShardID != shard.BeaconChainShardID {
+				atomic.AddUint32(&node.NumInvalidNodeMsg, 1)
+				return nil, 0, errIgnoreBeaconMsg
+			}
+			atomic.AddUint32(&node.NumSlashMsg, 1)
+		case proto_node.Receipt:
+			atomic.AddUint32(&node.NumReceiptMsg, 1)
+		case proto_node.CrossLink:
+			// only beacon chain node process crosslink messages
+			// TODO (lc) Does explorer node need to process crosslink message?
```
rlan35, line 406: explorer node don't need to process crosslinks.

node/node.go
LeoHChen: The original logic is here. https://github.com/harmony-one/harmony/blob/43d0696505927b7586426cc358c4d639bea2c254/node/node_handler.go#L101

explorer node on other shards doesn't participate in beacon block sync.

node/node.go
LeoHChen: okay, will skip

node/node.go
rlan35: it's likely a bug. Let's fix it here. The explorer node is using state sync to catch up on beacon chain now. With this fix, it will help explorer node to sync beacon chain better

LeoHChen: okay.

PR #3239, opened 2020-07-19
Use finer grained RWMutex for consensus logic

rlan35:
Instead of locking on on all consensus logic with a single mutex. Use more fine grained RWMutex to lock on Read and Write separately, and also make signature checking not locked to parallelize the heavy CPU process.

Tested locally.

Did more benchmarking on the code.

Compared to the version without this change. This PR makes the CPU usage lower by around **10%**.

However, on my instrumented node, I didn't see obvious improvement on the speed of processing consensus messages between code with and without this PR. It could be because the instrumented node can not exercise all operations of a real leader.

But compared to the mainnet leader node on shard 0, the instrumented node did perform about 20% better in consensus message processing.

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -36,9 +36,6 @@ type Consensus struct {
 	phase FBFTPhase
 	// current indicates what state a node is in
 	current State
-	// blockNum: the next blockNumber that FBFT is going to agree on,
-	// should be equal to the blockNumber of next block
-	blockNum uint64
 	// How long to delay sending commit messages.
 	delayCommit time.Duration
 	// Consensus rounds whose commit phase finished
@@ -64,12 +61,9 @@ type Consensus struct {
 	bhpBitmap    map[uint64]*bls_cosi.Mask
 	nilBitmap    map[uint64]*bls_cosi.Mask
 	viewIDBitmap map[uint64]*bls_cosi.Mask
-	m1Payload    []byte     // message payload for type m1 := |vcBlockHash|prepared_agg_sigs|prepared_bitmap|, new leader only need one
-	vcLock       sync.Mutex // mutex for view change
+	m1Payload    []byte // message payload for type m1 := |vcBlockHash|prepared_agg_sigs|prepared_bitmap|, new leader only need one
 	// The chain reader for the blockchain this consensus is working on
 	ChainReader *core.BlockChain
-	// map of nodeID to validator Peer object
-	validators sync.Map // key is the hex string of the blsKey, value is p2p.Peer
 	// Minimal number of peers in the shard
 	// If the number of validators is less than minPeers, the consensus won't start
 	MinPeers   int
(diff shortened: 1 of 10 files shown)


consensus/leader.go
```diff
@@ -127,66 +122,89 @@ func (consensus *Consensus) onPrepare(msg *msg_pb.Message) {
...
 	}
 
 	consensus.getLogger().Debug().
-		Int64("NumReceivedSoFar", consensus.Decider.SignersCount(quorum.Prepare)).
+		Int64("NumReceivedSoFar", signerCount).
 		Int64("PublicKeys", consensus.Decider.ParticipantsCount()).
 		Msg("[OnPrepare] Received New Prepare Signature")
+
+	//// Lock Write - Start
+	consensus.prepareMutex.Lock()
```
JackyWYX, line 177: As far as I know, this operation will introduce race condition which is not discoverable by --cover flag. Please compare the following two functions in multithreading execution:

```
type myStruct struct {
    val int
    mu sync.RWMutex
}

func (ms *myStruct) CompareAndSwap1(compareVal, targetVal int) {
    ms.mu.RLock()
    prev := ms.val
    ms.mu.RUnlock()

    if prev != compareVal {
        return
    }

    ms.mu.Lock()
    ms.val = targetVal  // At this point, ms.val can be already been changed by other goroutine
    ms.mu.Unlock()
}

// Use double lock check
func (ms *myStruct) CompareAndSwap1(compareVal, targetVal int) {
    ms.mu.RLock()
    prev := ms.val
    ms.mu.RUnlock()

    if prev == compareVal {
        ms.mu.Lock()
        if ms.val == compareVal {
            ms.val = targetVal    // value is protected. So this is fine
        }
        ms.mu.Unlock()
    }
}
```

If we really want to optimize in this code, I would recommend using the double lock checking.

consensus/consensus.go
```diff
@@ -89,9 +86,11 @@ type Consensus struct {
...
 	// IgnoreViewIDCheck determines whether to ignore viewID check
 	IgnoreViewIDCheck *abool.AtomicBool
-	// global consensus mutex
-	// TODO(optimization): Avoid mutex and use more efficient locking primitives
-	mutex sync.Mutex
+	// consensus mutex
+	prepareMutex sync.RWMutex
+	commitMutex  sync.RWMutex
+	// mutex for view change
+	vcLock sync.Mutex
```
JackyWYX, line 93: vcLock is not exclusive with prepareMutex & commitMutex, but somehow they are actually protecting the same data structure. So this will cause a race condition.

JackyWYX:
Sorry, didn't see the WIP flag before I made my comment. But I hope the comment will provide some insights anyway.

consensus/leader.go
rlan35: Good catch. Added the check of consistent block hash to make sure the read and write are operating on the same block. Note that checking on the same block is enough because the signature addition are always valid if it's the same block.

rlan35:
> Sorry, didn't see the WIP flag before I made my comment. But I hope the comment will provide some insights anyway.

yea, you are right, the concurrency between view change and consensus are not strictly protected. But it's semi-protected by the ViewChanging mode which will let the node only process one type of messages (either consensus or view change) in specific mode. So far, the view change and consensus are working fine without concurrency issue. So this PR will not address this separate issue here.

consensus/consensus.go
rlan35: yea, you are right, the concurrency between view change and consensus are not strictly protected. But it's semi-protected by the ViewChanging mode which will let the node only process one type of messages (either consensus or view change) in specific mode. So far, the view change and consensus are working fine without concurrency issue. So this PR will not address this separate issue here.

LeoHChen:
Great effort to have finer lock!

Seems you have added readonly lock on IsQuorumAchieved call in all places, to simplify the code, maybe you can put the readonly lock inside the IsQuorumAchieved function itself?

JackyWYX:
Nice fix in GOMAXPROCS!

rlan35:
> Great effort to have finer lock!
> 
> Seems you have added readonly lock on IsQuorumAchieved call in all places, to simplify the code, maybe you can put the readonly lock inside the IsQuorumAchieved function itself?

The lock on on all read operations, not only IsQuorumAchieved. It will be better to keep the locks minimum. Besides, IsQuorumAchieved is basically a read on boolean operation, better not add another lock on it.

rlan35:
> Also, I see we almost call SignersCount after IsQuorumAchieved all the time, maybe we can also return the SignersCount as part of the IsQuorumAchieved.

I think it's better to keep the logic separate. IsQuorumAchieved is just reading a boolean, it doesn't touch the signer count actually.

Issue #3222, opened 2020-07-15
[panic] nil pointer in map crash


LeoHChen:
**Describe the bug**
During a spamming test on STN by @SebastianJ , we found a nil map crash during view change.

The following is the crash dump.

```
Jul 15 19:27:38 stn-node-12 harmony[60674]: panic: assignment to entry in nil map
Jul 15 19:27:38 stn-node-12 harmony[60674]: goroutine 14214463 [running]:
Jul 15 19:27:38 stn-node-12 harmony[60674]: github.com/harmony-one/harmony/consensus.(*Consensus).onViewChange(0xc0005cc480, 0xc04bf5a720)
Jul 15 19:27:38 stn-node-12 harmony[60674]:         /home/lc/go/src/github.com/harmony-one/harmony/consensus/view_change.go:356 +0x152c
Jul 15 19:27:38 stn-node-12 harmony[60674]: github.com/harmony-one/harmony/consensus.(*Consensus).HandleMessageUpdate(0xc0005cc480, 0x1bbfe40, 0xc04bf5b8c0, 0xc04bf5a>
Jul 15 19:27:38 stn-node-12 harmony[60674]:         /home/lc/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:98 +0x17f
Jul 15 19:27:38 stn-node-12 harmony[60674]: github.com/harmony-one/harmony/node.(*Node).Start.func2.1(0xc01c8a67f0, 0xc008c40500, 0x1, 0xc01c8a66f0, 0xc04bf5a720, 0x0>
Jul 15 19:27:38 stn-node-12 harmony[60674]:         /home/lc/go/src/github.com/harmony-one/harmony/node/node.go:640 +0x447
Jul 15 19:27:38 stn-node-12 harmony[60674]: created by github.com/harmony-one/harmony/node.(*Node).Start.func2
Jul 15 19:27:38 stn-node-12 harmony[60674]:         /home/lc/go/src/github.com/harmony-one/harmony/node/node.go:626 +0x1e9

```

The code is in view_change.go. We didn't check the map.

https://github.com/harmony-one/harmony/blob/14e017defc18d712a854f70104e173e6eea9c058/consensus/view_change.go#L356

PR #3242, opened 2020-07-21
fix view change race issue

rlan35:
The signature map can be reset if without the lock (https://github.com/harmony-one/harmony/issues/3222)

Also correct the condition of sanity check to avoid the problems that cause view change to stuck.

Tested locally.

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -190,7 +190,7 @@ func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
 }
 
 func (consensus *Consensus) onNewViewSanityCheck(recvMsg *FBFTMessage) bool {
-	if recvMsg.ViewID <= consensus.viewID {
+	if recvMsg.ViewID < consensus.viewID {
 		consensus.getLogger().Warn().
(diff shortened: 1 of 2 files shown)


consensus/checks.go
```diff
@@ -190,7 +190,7 @@ func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
 }
 
 func (consensus *Consensus) onNewViewSanityCheck(recvMsg *FBFTMessage) bool {
-	if recvMsg.ViewID <= consensus.viewID {
+	if recvMsg.ViewID < consensus.viewID {
```
gupadhyaya, line 193: why is it ok to allow a new view for the same view id as the last successful view id?

rlan35: This is more of a temporary workaround to solve the view change stuck issue. We do see it in mainnet and other scenarios that this check got view change stuck. Likely the consensus.viewID is updated incorrectly somewhere else. Before we have the total view change revamp. I think it's ok to make this change to make view change more robust.

gupadhyaya: sure. I will address this later. thanks for explanation.

PR #3243, opened 2020-07-22
Optimize bls Mask infra code to avoid expensive loops

rlan35:
The bls Mask implementation contains lot of expensive loops which is replaced with efficient lookup index.

This should further speed up the leader process as leader need to do lots of BLS Mask operations.

tested locally.

### DIFF
--- a/crypto/bls/mask.go
+++ b/crypto/bls/mask.go
@@ -67,6 +67,7 @@ func AggregateSig(sigs []*bls.Sign) *bls.Sign {
 type Mask struct {
 	Bitmap          []byte
 	Publics         []*bls.PublicKey
+	PublicsIndex    map[SerializedPublicKey]int
 	AggregatePublic *bls.PublicKey
 }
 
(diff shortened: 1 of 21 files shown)


crypto/bls/mask.go
```diff
@@ -67,27 +67,32 @@ func AggregateSig(sigs []*bls.Sign) *bls.Sign {
...
+	PublicsIndex    map[SerializedPublicKey]int
 	AggregatePublic *bls.PublicKey
 }
 
 // NewMask returns a new participation bitmask for cosigning where all
 // cosigners are disabled by default. If a public key is given it verifies that
 // it is present in the list of keys and sets the corresponding index in the
 // bitmask to 1 (enabled).
-func NewMask(publics []*bls.PublicKey, myKey *bls.PublicKey) (*Mask, error) {
+func NewMask(publics []PublicKeyWrapper, myKey *PublicKeyWrapper) (*Mask, error) {
```
JackyWYX, line 78: If this function is used a lot in leader node, we can do it in one iteration and no need to build two maps, right?

```
func NewMask(publics []PublicKeyWrapper, myKey *PublicKeyWrapper) (*Mask, error) {
    if myKey != nil {
        return nil, errors.New("...")
    }
    m := ...
    for i, key := range publics {
        if myKey .Bytes == public.Bytes {
            m.SetBit(i, true)
            return m, nil
        }
    }
    return nil, errors.New("...")
}

rlan35: The map needs to be populated with the public keys and indexes.

JackyWYX: I see. My bad.

PR #3245, opened 2020-07-23
[node.sh] print out running binary version

LeoHChen:
We used to check only the installed version of harmony binary.
There are cases when validator didn't restrart the harmony process
even though the binary was updated.

So, we print out both on-disk binary version and the running version.

### DIFF
--- a/scripts/node.sh
+++ b/scripts/node.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-version="v1 20200606.0"
+version="v1 20200723.0"
 
 unset -v progname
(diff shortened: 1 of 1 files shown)


LeoHChen:
Tested on one node.
```
./node.sh -V
Binary  Version: Harmony (C) 2020. harmony, version v6137-v2.1.9-61-gb9b462d9 (lc@ 2020-07-21T05:56:59+0000)
Running Version: Harmony (C) 2020. harmony, version v6137-v2.1.9-61-gb9b462d9 (lc@ 2020-07-21T05:56:59+0000)
```

PR #3250, opened 2020-07-25
Add 5 seconds block time at epoch 230 and slot increase 480->640 at epoch 231

rlan35:
Add 5 second block and reward at epoch 230 (around July 39 Wed 2020)

Add slot increase from 480 to 640 at epoch 231 (around July 31 Fri 2020)

### DIFF
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -3,6 +3,7 @@ package consensus
 import (
 	"math/big"
 	"sync/atomic"
+	"time"
 
 	"github.com/harmony-one/harmony/crypto/bls"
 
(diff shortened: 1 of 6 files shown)


consensus/consensus_service.go
```diff
@@ -346,6 +347,10 @@ func (consensus *Consensus) UpdateConsensusInformation() Mode {
 		}
 	}
 
+	if consensus.ChainReader.Config().IsFiveSeconds(nextEpoch) {
+		consensus.BlockPeriod = time.Duration(5 * time.Second)
```
JackyWYX, line 351: Shall we make 5s a const?

rlan35: This code is only temporary, after we enter that epoch, we can just remove this code and change the BlockPeriod flag to 5 seconds.

JackyWYX: Well, instead of providing the option to the user I would suggest to hard code this value in consensus. Because block time in flag is just a sanity field. However, according to protocol, it shall be enforced to be a certain value (8s or 5s). Parsing through flag would not be a great option for this "enforced" field as far as I see.

LeoHChen: we can just remove the flag then.

LeoHChen: It was there for testing purpose initially.

JackyWYX: > we can just remove the flag then.

Thanks, will add TODO for after my current cmd overhaul.

PR #3256, opened 2020-07-27
[p2p] propogate cross-shard node messages

LeoHChen:

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -385,29 +385,26 @@ func (node *Node) validateNodeMessage(ctx context.Context, payload []byte) (
 	case proto_node.Block:
 		switch proto_node.BlockMessageType(payload[p2pNodeMsgPrefixSize]) {
 		case proto_node.Sync:
+			atomic.AddUint32(&node.NumBlockSyncMsg, 1)
 			// only non-beacon nodes process the beacon block sync messages
 			if node.Blockchain().ShardID() == shard.BeaconChainShardID {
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -667,9 +664,16 @@ func (node *Node) Start() error {
...
 					)
 					if err != nil {
-						// TODO (lc): block peers sending error messages
-						errChan <- withError{err, msg.GetFrom()}
-						return libp2p_pubsub.ValidationReject
+						switch err {
+						case errIgnoreBeaconMsg:
+							// ignore the further processing of the ignored messages as it is not intended for this node
+							// but propogate the messages to other nodes
+							return libp2p_pubsub.ValidationAccept
```
JackyWYX, line 671: Is it better to be `ValidationIgnore` so that the node will forward ignored message?

LeoHChen:
Deployed to STN, and see the crosslink start processing now.

http://watchdog.hmny.io/report-stn#top-of-page

node/node.go
rlan35, line 671: should you add the result to "ValidatorData" before return?

LeoHChen: No need. The ValidatorData will be used by the real node message handler.  Since we are not processing those messages by the real handler, but just forward the message in p2p layer.  We don't need to add the result to validatorData.  Those ignored messages won't be sent to node message handlers.  It is similar to the ignored consensus message, in https://github.com/harmony-one/harmony/blob/d829d227dc843bb257a5865a1c988332b24efdcf/node/node.go#L649

Issue #3210, opened 2020-07-08
Node API Overhaul (Design & Planning)


Daniel-VDM:
# Node API Overhaul

> Last updated: Monday, October 12th, 2020.

[Original Draft](https://gist.github.com/Daniel-VDM/24930faecf8813cfc6b07fd64b600d24)

## Overview
The API layer of a Harmony node is its main interface with the outside world. Therefore, it must be **consistent**, **robust**, and **well-documented** in order to provide easy integration with developers who wish to build on Harmony. 

## Current issues with v1 & v2 of Harmony's JSON-RPC based API
* Inconsistancies of response types (hex v.s. hex string v.s. number) and names (cammelCase v.s. kebab-case v.s. snake_case).
* Adding 1 RPC call consists of 5-7 files to edit and often contains **duplicate code**.
* Silent crashes and no logs of any error during a failed procedure calls. 
* RPC's Http server occasionally fails to start on node boot.
* Documentation is not part of the dev cycle and must be manually created afterward.
* Testing is not part of the dev cycle.

## Context
For Harmony to grow and get external contributors, we must provide a professional interface for developers to hook into. Moreover, an easily consumable Node API can attract developers to take a deeper dive into Harmony's codebase and contribute to the core protocol. 

Since Harmony has completed open staking, all core functionalities for an API layer have been defined. Therefore, **now** is an opportune time to correct the issues stated above as we are in the design phase of other major core features. 

It is important to note that Harmony has numerous tools and partners that use the current JSON-RPC based API. Therefore, we must be careful to NOT break/change the existing RPC output format during the overhaul. 

## Goals and Non-Goals

**What this overhaul aims to do:**
* Create a 'sanity-check' test suite that ensures the format & functionality of the current JSON-RPC.
* Clean-up the existing [`./internal/hmyapi`](https://github.com/harmony-one/harmony/tree/1be1ddf0b529725f8be162a353f8ad0bab9c8954/internal/hmyapi) packages to reduce code duplication & dead code. 
* Create an *additional* API layer that provides consistency in all responses and retains all functionality from the v1 & v2 JSON-RPC services.
* Integrate documentation as part of the API dev cycle.
* Integrate testing as part of the API dev cycle.
* Provide error reports of call crashes in the existing & new API layer (low priority).

**What this overhaul aims NOT to do:**
* Fix race conditions of the underlying function calls.
* Break old v1 & v2 JSON-RPC functionalities.
[...]

PR #3259, opened 2020-07-28
Node API Refactor - pt2 (Stage 2.2 of Node API Overhaul)

Daniel-VDM:
# Stage 2.2 of [Node API Overhaul](https://github.com/harmony-one/harmony/issues/3210)

This PR removes (as much as reasonably possible) the code duplication that was introduced earlier this year between the 2 RPC versions. Note that this refactor aims to keep the **same** functionality (args & methods names) as we currently have. However, there are some added errors that now return to help developers (this should not affect properly written code or SDK). 

This PR exposes the RPC methods and structs so that any go project (like the go-SDK & CLI) can use this to interact with the chain more easily. 

This PR removes the `addrlock` object as it was not used. It also removes the utils function as they could be baked into the 1 RPC that used them.

This PR also adds inline documentation and some logs to help with debugging and maintenance of the RPC layer in the future.

Lastly, this PR organizes each logical group of methods (i.e: staking, transactions, blockchain, pool, etc...) into their own services for clarity. 

Note that some less than ideal choices had to be made in the implementation to not break old behavior. For instance, some methods must return an interface to account for hex strings vs integers. Structured responses must be marshaled and unmarshaled to account for the different RPC structure versions. That said, I tried to keep this at a minimum and did not change any of the node functions that fetch the appropriate data for the RPCs. 

## TODO:
* Testnet integration test with partners


rpc/blockchain.go
```diff
@@ -0,0 +1,664 @@
...
+			continue
+		}
+		if ok, err := mask.KeyEnabled(validator.BLSPublicKey); err == nil && ok {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+// GetSignedBlocks returns how many blocks a particular validator signed for last defaultBlocksPeriod (3 hours ~ 1500 blocks).
```
LeoHChen, line 367: why 1500 blocks as default?

Daniel-VDM: Think this is an arbitrary number that we have. The comment is actually wrong, its 15k blocks, which is around an epoch's worth of blocks. Therefore I'll make `defaultBlocksPeriod` the blocks per epoch from the config.

PR #3253, opened 2020-07-25
[CMD] Command overhaul

JackyWYX:
## 1. Description

This is another PR for issue [#3110]. This PR overhauls the command line interface.

1. Commands are rewritten with cobra.
2. All flags (new flags and legacy flags) are parsed into `harmonyConfig`.
3. The `harmonyConfig` can be dumped to a file with `dumpconfig` command.
4. The config file can be loaded with `--config` flag. Loaded config can override with custom flags.
5. Functionality of all existing flags are preserved in this PR (all in legacy flags) and legacy logic are not changed either. 

### Claim 1. Break change in command line interface

1. All flags should be specified with double dash instead of single dash.
2. Staking mode is on by default. If want to run in legacy mode, use `--staking=false` or `--run.legacy`
3. Though all legacy fields become hidden for `-h`, all these functionality are supported.

### Claim 2. Not compatible with previous node.sh

Because of the fact in claim 1, this PR is NOT compatible with previous node.sh.

On the other side, node.sh command is still compatible.

### Claim 3. This PR is NOT ready to release

Reasons:
1. Some legacy code need to check with Leo and RJ for whether we should continue supporting these flags - Dealing with legacy is expensive. We shall make this command line interface a clean start.
2. Some legacy flags (IP and port) are separated to different flags (rpc.ip/port, ws.ip/port, p2p.ip/port) in this PR. However these flags are just defined but not used so far.

### Claim 4. This PR is big

1. This PR contains unmerged code in https://github.com/harmony-one/harmony/pull/3219. (~2000 lines)
2. There is a natural dilemma between make it work and make it small. In order to make it work, this size is the best I can do.

### Claim 5. Foundation node supports

This PR shall not be released. We shall only release once after the mega project removing node.sh is totally done (to minimize the impact of the foundation nodes).

### 2.1 Change the ip/port usage. 

Do all port maths (+500, -3000...) at command line instead of confusing user with a single 9000 port but actually taking 4 ports (6000, 9000, 9500, 9800). Also user can choose whether to open these RPC calls separately by specifying the IP address to listen.  Currently these ports are just defined but not used.

Also, we can preserve the legacy --port or --ip field for simplicity. (Done in https://github.com/harmony-one/harmony/pull/3261)

### 2.2 Determine whether we can hard code and remove these flags:

1. bls.maxkeys
2. consensus.delay-commit
3. consensus.block-time
4. consensus.min-peers
5. txpool.broadcast-invalid-tx

[...]

PR #3261, opened 2020-07-29
[cmd] Seperate the usage of p2p port, http port and ws port

JackyWYX:
## Issue

Currently, we are using one ip and port for multiple listening addresses: P2P, DNS, HTTP, Web Socket. But this is actually confusing the user with the port magic, and at the same time, users are not able to customize these endpoints seperately. This PR aims to decouple the usage of IP and port to different service, and make it much clearer for users, while still preserve the legacy ip / port flags for backward compatibility.

 Note: DNS port is still set to 9000 since I see that some of the syncing protocol are using this number as the GRPC registration. So I will leave it for now. It shall be gone soon after implemented fastSync with distributed sync protocol. 

## Test

Test log: https://gist.github.com/JackyWYX/2554d561ccca74617d68a74a53ca64e3

## Operational Checklist

    **NO**

    **NO**

    **NO**


JackyWYX:
My original intention to build multiple PRs for command overhaul (#3253, #3261, #3274) was to incrementally add changes to the repository. However, since #3253 has not been merged for a long time, so multiple branches will only cause ambiguity. Other than that, some bugs is found so this PR is not completely self-contained. So will close this PR and reopen a new one.

PR #3262, opened 2020-07-29
add support for redelegation and bring back 7 epoch locking time

rlan35:
Add redelegation support. The redelegation works without disturbing the existing staking transaction interface, so no need to modify all the client integrations such as Trust wallet or staking dashboard. It's a two step process where the user first undelegate the tokens which will be locked, and then the new delegation will first try to use the locked token, if not enough, then use the liquid token from normal user balance.

The locked tokens will only be used for redelegation if they are already passed 1 epoch after it's first locked. For example, if 100 ONE is undelegated and locked at epoch 10, then they are only considered for redelegation if it's epoch 11 now. This is to prevent the case where malicious attackers commit attack and then immediately switch token to other validators to avoid slashing.

At the same epoch for redelegation fork, also bring back the 7 epoch undelegation locking time.

Local Integration Tests Done:
1. Delegation without using locked token.
2. Delegation using all locked tokens (redelegation).
3. Delegation using balance when the locked token are not 1 epoch old.
4. Delegation using partial locked token and partial balance.
5. Delegation using locked token from multiple undelegations.

### DIFF
--- a/internal/shardchain/dbfactory.go
+++ b/internal/shardchain/dbfactory.go
@@ -22,7 +22,7 @@ type LDBFactory struct {
 // NewChainDB returns a new LDB for the blockchain for given shard.
 func (f *LDBFactory) NewChainDB(shardID uint32) (ethdb.Database, error) {
 	dir := path.Join(f.RootDir, fmt.Sprintf("harmony_db_%d", shardID))
-	return ethdb.NewLDBDatabase(dir, 128, 1024)
+	return ethdb.NewLDBDatabase(dir, 128, 64)
 }
 
 // MemDBFactory is a memory-backed blockchain database factory.


LeoHChen:
Can we add some those test cases in nightly build/test or Jenkins job? This is a very impactful feature and we shouldn't introduce regression later.  @SebastianJ , @Daniel-VDM

core/tx_pool.go
```diff
@@ -823,7 +823,20 @@ func (pool *TxPool) validateStakingTx(tx *staking.StakingTransaction) error {
 			return errors.WithMessagef(ErrInvalidSender, "staking transaction sender is %s", b32)
 		}
 
-		_, _, err = VerifyAndDelegateFromMsg(pool.currentState, stkMsg)
+		chain, ok := pool.chain.(ChainContext)
+		if !ok {
+			return nil // for testing, chain could be testing blockchain
```
LeoHChen, line 828: what does this mean here? any info/log needed?

internal/params/config.go
```diff
@@ -25,89 +25,95 @@ var EpochTBD = big.NewInt(10000000)
...
-		S3Epoch:          big.NewInt(0),
-		ReceiptLogEpoch:  big.NewInt(0),
+		ChainID:           TestnetChainID,
+		CrossTxEpoch:      big.NewInt(0),
+		CrossLinkEpoch:    big.NewInt(2),
+		StakingEpoch:      big.NewInt(2),
+		PreStakingEpoch:   big.NewInt(1),
+		QuickUnlockEpoch:  big.NewInt(0),
+		FiveSecondsEpoch:  big.NewInt(16500),
+		RedelegationEpoch: big.NewInt(10000000),
```
LeoHChen, line 50: shall we enable it earlier on testnet ?

rlan35: I will enable it once the testnet current issues are fixed.

core/tx_pool.go
rlan35: This part is copied from CollectReward. I think it's for unittests. Added log.

PR #3267, opened 2020-07-30
[p2p] forward valid commit messages by validators

LeoHChen:


LeoHChen:
> NVM. Is it better to change to ValidationIgnore instead of ValidationReject for validator not in committee?

Same thing Ignore or Reject on application logic.  By moving down, we won't reject this kind of error on regular validator.

JackyWYX:
I know it's no difference in application level, but accepting these message will cause the p2p layer vulnerable to some kind of spamming attacks. A systematic way to fix is to add an extra PBFT round of the epoch change event. Anyway, this could be a hot fix for this certain issue.

PR #3276, opened 2020-07-31
[build] add arm_static build target

LeoHChen:


AmazingAng:
How to build in arm64?

PR #3296, opened 2020-08-12
add support for redelegation and bring back 7 epoch locking time

rlan35:
Add redelegation support. The redelegation works without disturbing the existing staking transaction interface, so no need to modify all the client integrations such as Trust wallet or staking dashboard. It's a two step process where the user first undelegate the tokens which will be locked, and then the new delegation will first try to use the locked token, if not enough, then use the liquid token from normal user balance.

The locked tokens will only be used for redelegation if they are already passed 1 epoch after it's first locked. For example, if 100 ONE is undelegated and locked at epoch 10, then they are only considered for redelegation if it's epoch 11 now. This is to prevent the case where malicious attackers commit attack and then immediately switch token to other validators to avoid slashing.

At the same epoch for redelegation fork, also bring back the 7 epoch undelegation locking time.

Local Integration Tests Done:

Delegation without using locked token.
Delegation using all locked tokens (redelegation).
Delegation using balance when the locked token are not 1 epoch old.
Delegation using partial locked token and partial balance.
Delegation using locked token from multiple undelegations.


core/tx_pool.go
```diff
@@ -823,7 +823,21 @@ func (pool *TxPool) validateStakingTx(tx *staking.StakingTransaction) error {
...
+			utils.Logger().Debug().Msg("Missing chain context in txPool")
+			return nil // for testing, chain could be testing blockchain
+		}
+		delegations, err := chain.ReadDelegationsByDelegator(stkMsg.DelegatorAddress)
+		if err != nil {
+			return err
+		}
+		pendingEpoch := pool.chain.CurrentBlock().Epoch()
+		if shard.Schedule.IsLastBlock(pool.chain.CurrentBlock().Number().Uint64()) {
+			pendingEpoch = new(big.Int).Add(pendingEpoch, big.NewInt(1))
```
gupadhyaya, line 837: why add 1? you still want to validate for the current epoch right?

JackyWYX: Either way it will work, but I would prefer Ganesha's approach to make this simple. Or is there any concerns with the transition, RJ?

core/staking_verifier.go
```diff
@@ -214,40 +214,111 @@ func VerifyAndDelegateFromMsg(
...
 		return nil, nil, errDelegationTooSmall
 	}
-	wrapper, err := stateDB.ValidatorWrapperCopy(msg.ValidatorAddress)
-	if err != nil {
-		return nil, nil, err
+
+	updatedValidatorWrappers := []*staking.ValidatorWrapper{}
+	delegateBalance := big.NewInt(0).Set(msg.Amount)
+
+	var delegateeWrapper *staking.ValidatorWrapper
```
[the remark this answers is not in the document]
rlan35: it's actually "delegatee", no typo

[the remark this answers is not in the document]
JackyWYX: Ah, see.

core/tx_pool.go
rlan35: it's the same logic as collectRewards. If currentBlock is from last epoch, then currently it's epoch+ 1

rlan35:
> Looks good to me. My only concerns is whether it will work with slashing. Maybe we can test it out in integration test. Also note Ganesha's comment.

Slashing should work as before as the logic is fairly decoupled. Yea, we will test in integration.

JackyWYX:
For port number, take a look at this? https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=9700

rlan35:
> For port number, take a look at this? https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=9700

wrong PR? Please approve btw if this PR looks good.

PR #3297, opened 2020-08-13
Node API Refactor - pt3 (Stage 3.1 of Node API Overhaul)

Daniel-VDM:
# Stage 3.1 of [Node API Overhaul](https://github.com/harmony-one/harmony/issues/3210)

This PR adds the rosetta server to the node program. The rosetta server is hosted on port 9700 by default on an explorer node. Other nodes can enable rosetta it if they wish, but by default, it is off for all validator nodes. 

This PR also outlines the structure of the `rosetta` package to follow the example provided [here](https://github.com/coinbase/rosetta-sdk-go/tree/master/examples/server). 

The server was tested using rosetta's example client [here](https://github.com/coinbase/rosetta-sdk-go/tree/master/examples/client) and passes.

In order to use rosetta's SDK, we must use go-ethereum v1.9.18. As this is not compatible with our current code, I've added a replace in the go.mod file to support go-ethereum v1.8.27 as well. I've also updated/cleaned up the relative packages needed for rosetta's SDK.


go.mod
```diff
@@ -10,18 +10,19 @@ require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
 	github.com/cespare/cp v1.1.1
+	github.com/coinbase/rosetta-sdk-go v0.3.4
 	github.com/davecgh/go-spew v1.1.1
 	github.com/davidlazar/go-crypto v0.0.0-20190912175916-7055855a373f // indirect
 	github.com/deckarep/golang-set v1.7.1
 	github.com/edsrzf/mmap-go v1.0.0 // indirect
-	github.com/ethereum/go-ethereum v1.8.27
+	github.com/ethereum/go-ethereum v1.9.18
```
rlan35, line 18: why update ethereum? any potential compatibility issue?

Daniel-VDM: Rosetta SDK depends on this version of eth. As it is not compatible with our version, I've added a replacement to our dependent version here: https://github.com/harmony-one/harmony/pull/3297/files#diff-37aff102a57d3d7b797f152915a6dc16R83

Daniel-VDM:
> 1. Since we are adding new ports, we should do our best try not to collide our port with well-known services. 10000 is binded with NDMP (https://www.snia.org/ndmp). Can we use another one?
> 2. You need to consider the backward compatibility of the harmonyConfig structure (It's dumped to the config file along with a version). My suggestions are:
>    
>    * Increase the version to v1.1.0. (Preferred to import the version module from libp2p)
>    * If user are still using old version , recommend user to upgrade to config v1.1.0 with command `./harmony dumpconfig harmony.conf`
>    * Add a unit test case to make sure the data loaded from v1.0.0 config file is expected.

1. Changed the port to 9700
2. I bumped the config version and tested running the node with the old config, it works. Also added a unit test and warning/suggestion message if loading an old config.

PR #3300, opened 2020-08-17
Add multi-sig merged messaging logic

rlan35:
For a node with multiple keys, the old logic for consensus messaging will sign the block and one message for each of the keys. This causes unnecessary network load and leave rooms for optimization since BLS signature support elegant multi-sig signing and aggregation.

This PR in general enables the aggregated multi-sig signing and messaging for nodes with multiple keys. Specifically:

1. A new field "sender_pubkey_bitmap" is added to the ConsensusRequest protobuf. When multi-key nodes send aggregated signature, they don't need to explicitly specify the signing keys in plain bytes, but rather just a simply bitmap. The leader can derive the corresponding BLS public keys from the bitmap as long as the validator shared the updated committee as the leader, which is expected for a sync'ed validator.

2. Multi-sig messaging is only available for validator's Prepare and Commit message, which is the most common messages during consensus. Leader's messages of Announce, Prepared and Committed does not fit with multi-sig because there is only one leader. View change messages are not supported with multi-sig for now as view change is not speed sensitive.

3. If the multi-sig is enabled, a node with multiple keys will sign on the block with each of the keys that's active in the committee and the signatures will be aggregated as a single multi-sig which is put in the message. The bitmap of the signing keys will also be put in the message. A node with only a single key will continue the old messaging logic where the "sender_pubkey" is populated with the plain BLS public key bytes.

4. The slashing logic is updated to take care of the multi-sig situation where keys signing on a multi-sig did the double sign. The slashing record will log the multi-sig and all the signing keys on the multi-sig. In the multi-sig keys, only the keys that double signed will be accounted in the slashing.

Tested in local environment, consensus and view change works fine with multi-sig nodes.

Tested slash logic when multi-sig having double signs:

```{"level":"info","port":"127.0.0.1","ip":"9000","rate":"1.000000000000000000","records":[{"evidence":{"epoch":2,"shard-id":1,"height":50,"view-id":50,"first-vote":{"bls-public-keys":["034510e7619874d71a18ff25639dd64efd1b250e4b77180b8fadd78323375f5db10af98fc1534b689cf378d843f10085","3107c8b3f8c96817433221bdac48f38ae3e4cef35da90a43736102cf71ce4b46539ac01c25509d16791951029c4c3a97"],"block-header-hash":"0xa0430a61e0ea09865d90267da89c03d95afa077c1c44aa134ab9fdbd5380f80e","bls-signature":"J8/PIOrhyz/PsVVDLSSWuWwGTkwToEYpF0L1kVis5VTCvMqRL7wusDfYDSX5SJATqDpCGtJgjgOr5TUzo4D6ebBTvyBXGpZuIMcUmnD9F49tsFPjlh1qXnSfgBYtrMCI"},"second-vote":{"bls-public-keys":["034510e7619874d71a18ff25639dd64efd1b250e4b77180b8fadd78323375f5db10af98fc1534b689cf378d843f10085","3107c8b3f8c96817433221bdac48f38ae3e4cef35da90a43736102cf71ce4b46539ac01c25509d16791951029c4c3a97"],"block-header-hash":"0x00430a61e0ea09865d90267da89c03d95afa077c1c44aa134ab9fdbd5380f80e","bls-signature":"3sp4wf/D/H81rQRbJ6s4IcSbbXm4Q33lwaoJ6+DP7BADcPDlOh4vBmDgPRgBof4Sael/VMFPy9TYzKgHlWyHMAKFwlAYtfvzmTh5hTqlcuxglhCfxalTScn8/zIPZLgO"}},"beneficiary":"one1m6m0ll3q7ljdqgmth2t5j7dfe6stykucpj2nr5","offender":"one1zksj3evekayy90xt4psrz8h6j2v3hla4qwz4ur"}],"applied":{"total-slashed":10000000000000000000000,"total-snitch-reward":5000000000000000000000},"caller":"/Users/rongjianlan/workspace/go/src/github.com/harmony-one/harmony/internal/chain/engine.go:423","time":"2020-08-27T22:06:24.821022-07:00","message":"slash applied successfully"}```

### CODE REVIEW

File: consensus/double_sign.go [hunk 1]
```diff
@@ -12,96 +12,111 @@ import (
...
-					areHeightsEqual := firstSignedBlock.Number().Uint64() == recvMsg.BlockNum
-					areViewIDsEqual := firstSignedBlock.ViewID().Uint64() == recvMsg.ViewID
-					areHeadersEqual := firstSignedBlock.Hash() == recvMsg.BlockHash
+		addrSet := map[common.Address]struct{}{}
+		for _, pubKey2 := range recvMsg.SenderPubkeys {
+			if alreadyCastBallot := consensus.Decider.ReadBallot(
+				quorum.Commit, pubKey2.Bytes,
+			); alreadyCastBallot != nil {
+				for _, pubKey1 := range alreadyCastBallot.SignerPubKeys {
+					firstPubKey, err := bls.BytesToBLSPublicKey(pubKey1[:])
```
gupadhyaya, line 21: can we simply compare the bytes of pubKey1 with pubKey2.Bytes and skip, instead of `BytesToBLSPublicKey` and then skipping?

consensus/quorum/quorum.go
```diff
@@ -168,10 +168,29 @@ type depInject struct {
...
+		alreadyCollected := false
+		for _, key := range ballot.SignerPubKeys {
+			if _, ok := collectedKeys[key]; ok {
+				alreadyCollected = true
+				break
+			}
+		}
+		if alreadyCollected {
+			continue
+		} else {
```
gupadhyaya, line 188: readability: `else` not needed?

consensus/validator.go
```diff
@@ -59,11 +61,38 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
...
-	for _, key := range consensus.priKey {
+	priKeys := []*bls.PrivateKeyWrapper{}
+	for i, key := range consensus.priKey {
 		if !consensus.IsValidatorInCommittee(key.Pub.Bytes) {
 			continue
 		}
-		networkMessage, err := consensus.construct(msg_pb.MessageType_PREPARE, nil, &key)
+		priKeys = append(priKeys, &consensus.priKey[i])
+		if !consensus.MultiSig {
+			networkMessage, err := consensus.construct(msg_pb.MessageType_PREPARE, nil, []*bls.PrivateKeyWrapper{&key})
```
gupadhyaya, line 71: can constructing the message and sending be refactored out to a function? lot of duplicate code... :(

consensus/view_change.go
```diff
@@ -624,14 +633,37 @@ func (consensus *Consensus) onNewView(msg *msg_pb.Message) {
...
-		for _, key := range consensus.priKey {
+
+		priKeys := []*bls.PrivateKeyWrapper{}
+		for i, key := range consensus.priKey {
 			if !consensus.IsValidatorInCommittee(key.Pub.Bytes) {
 				continue
 			}
+			priKeys = append(priKeys, &consensus.priKey[i])
+			if !consensus.MultiSig {
+				network, err := consensus.construct(
```
gupadhyaya, line 644: same thing as validator.go, if we can refactor the code here. if it is too complicated, then leave it.

consensus/double_sign.go
rlan35: Yup, definitely. done.

consensus/quorum/quorum.go
rlan35: good point. done.

consensus/validator.go
rlan35: Originally, planned to leave it as code to delete once migrated. But may leave it as a option. So refactored.

JackyWYX:
LGTM now!

rlan35:
jenkins, test this please

PR #3304, opened 2020-08-19
build official RPM package for release

LeoHChen:
This is the first attempt to build the first Harmony RPM package for official release.


LeoHChen:
TODO: sysctl reload

LeoHChen:
https://github.com/harmony-one/harmony/issues/3305

nyetwurk:
Looks good, but I'm unclear on how harmony.cfg works, and how the user can customize either the entrypoint command line or the config.

What happens when the deb is upgraded from upstream?

I didn't look closely at the deb/rpm creation yet, so I'm not sure but normally the way this works is one or more of the following :

* An `/etc/default/<env file>` which is installed on first installation, then never touched in case the user customizes it. It is generally environment variables passed to a wrapper script. This doesn't work in systemd unless systemd itself is told to use a wrapper (this is sysv style).

* A config file`/etc/<app>/<app>.conf` which includes all files in a dir `/etc/<app>/conf.d`, which has defaults installed by the upstream, and the user can add in addtional configs which are merged in with the default. They're often prefixed by number (e.g. `00-` `01-` `99-`) so the precedence is deterministic

LeoHChen:
> Looks good, but I'm unclear on how `harmony.cfg` works, and how the user can customize either the entrypoint command line or the config.
> 
> What happens when the deb is upgraded from upstream?
> 
> I didn't look closely at the deb/rpm creation yet, so I'm not sure but normally the way this works is one or more of the following :
> 
> * An `/etc/default/<envfile>` which is installed on first installation, then never touched in case the user customizes it. It is generally environment variables passed to a wrapper script. This doesn't work in systemd unless systemd itself is told to use a wrapper (this is sysv style). Systemd also allows including an env file directly in the unit file.
> * A config file`/etc/<app>/<app>.conf` which includes all files in a dir `/etc/<app>/conf.d`, which has defaults installed by the upstream, and the user can add in addtional configs which are merged in with the default. They're often prefixed by number (e.g. `00-` `01-` `99-`) so the precedence is deterministic


1) harmony.cfg is used to determine the node is an `validator` node or an `explorer` node.  We have exchange/wallet partners that need to run `explorer` node instead of regular `validator` node, so the cli parameter is different in those two types. It is a one-time configuration though.

2) we have another config file, that user can also use it to customize the node.  It was usually in home directory.  

I guess I shall unite both cases.  We don't support multiple config files. It is just one big toml file though.

Thanks a lot for the feedback.

JackyWYX:
Make `--run.archive` default for explorer node (https://github.com/harmony-one/harmony/pull/3306). Now the command line part looks good to me.

LeoHChen:
@nyetwurk, @Daniel-VDM , can you review this PR again?  I've simplified the config file to just the single configuration file on `/etc/harmony` directory by default.

Makefile
```diff
@@ -76,3 +84,40 @@ arm_static:
 	make -C $(TOP)/bls minimised_static BLS_SWAP_G=1 -j8
 	bash ./scripts/go_executable_build.sh -a arm64 -s
 	git checkout go.mod
+
+deb_init: bin/harmony
+	rm -rf $(DEBBUILD)
+	mkdir -p $(DEBBUILD)/scripts
+	mkdir -p $(DEBBUILD)/$(PKGNAME)-$(VERSION)/{etc/systemd/system,usr/local/sbin,data/harmony/.config/rclone,etc/sysctl.d,etc/harmony}
```
nyetwurk, line 91: this requires `SHELL:=bash`

nyetwurk:
I have addressed other issues with @LeoHChen in DM in discord.. will summarize here on request.

PR #3306, opened 2020-08-19
[cmd] if running explorer node, set archive to true

JackyWYX:
## Description

Command line fix for explorer node. (https://github.com/harmony-one/harmony/pull/3304) Archive mode is now default to enabled when running explorer node.

## Testing

Two test cases added for explorer archival mode.

### DIFF
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -244,6 +244,10 @@ func applyGeneralFlags(cmd *cobra.Command, config *harmonyConfig) {
 		config.General.NodeType = cli.GetStringFlagValue(cmd, legacyNodeTypeFlag)
 	}
 
+	if config.General.NodeType == nodeTypeExplorer {
+		config.General.IsArchival = true
+	}
+
 	if cli.IsFlagChanged(cmd, shardIDFlag) {
 		config.General.ShardID = cli.GetIntFlagValue(cmd, shardIDFlag)
 	} else if cli.IsFlagChanged(cmd, legacyShardIDFlag) {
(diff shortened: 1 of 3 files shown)


LeoHChen:
Can I run explorer in non-archival mode though?  Or must we run explorer in archival mode?

JackyWYX:
@LeoHChen This is the default option. You can run non-archival mode with `--run.archive=false`. I believe we are running archival mode for all our explorer node right?

LeoHChen:
> @LeoHChen This is the default option. You can run non-archival mode with `--run.archive=false`. I believe we are running archival mode for all our explorer node right?

sounds good, as long as we can run explorer in non-archival mode.  Yes, it's true we shall run explorer in archival mode by default.

cmd/harmony/flags.go
```diff
@@ -244,6 +244,10 @@ func applyGeneralFlags(cmd *cobra.Command, config *harmonyConfig) {
 		config.General.NodeType = cli.GetStringFlagValue(cmd, legacyNodeTypeFlag)
 	}
 
+	if config.General.NodeType == nodeTypeExplorer {
+		config.General.IsArchival = true
```
LeoHChen, line 248: always override it?

JackyWYX: If flag is set by user, user's setting will override this option. Please check `cmd/harmony/flags.go:263`.

Also, a unit test is implemented to show the archival mode is overriden by user's option `cmd/harmony/flags_test.go:178`

LeoHChen: thanks for the explanation.

Issue #3322, opened 2020-08-28
PubSub Refactor


JackyWYX:
## Issue

Currently, PubSub handle logic are written in function `node.Start` as a whole. This function is the core part of harmony p2p, but the code is not well designed and structured so all code logic from different modules exist in a single function. This is a blocker for adding more features in the pub sub handling logic. Thus we need to refactor this piece of code to give it a bit more extensibility for new features.

The refactor has 4 requirements:

1. Different message handling logics are decoupled in different implementation of interface `PubSubHandler`.
2. `PubSubHandler` can be added, started, stopped and removed dynamically for extensibility.
3. Do not change the code logic, so that the code should work exactly the same as the current implementation.
4. After change, the code can resist the existing attacks.

## Phases

The refactor will take 3 phases, each phase will have a PR with the change.

### Phase 1: Refactor pub sub handling logic

This is basically add feature to p2p `Host` to have the capability of handling `PubSubHandlers` to meet the requirements 1. In this step, only adding new modules and do not edit the existing code logic.

### Phase 2: Implement different handlers

Decouple the existing message handling logic into different handlers, which is to implement `consensusChainHandler`, `nodeHandler`, `beaconChainHandler`. In this phase, write new code and do not edit the existing code logic.

### Phase 3: Make them a whole

Using the code in Phase 1 and 2, replace the existing code logic.

Do all the tests:
1. Normal use cases
2. Stress use cases (attacks)

rlan35:
can we close this now?

PR #3323, opened 2020-08-28
Pubsub refactor Phase 1 of 3

JackyWYX:
## Issue

This is the phase one for [PubSub refactor](https://github.com/harmony-one/harmony/issues/3322). Implement a new module `harmony/p2p/pubsub` to provide PubSub interface. 

1. This PR refactors the current code logic we are using with libp2p PubSub. 
2. All business logics can be implemented within the `Handler` interface, and be decoupled according to handler's usage. 
3. The handlers can be dynamically registered, added, stopped and removed.

Followings are some public interfaces that will give a taste about how to use this module.

```
type PubSub interface {
	AddHandler(psh Handler) error
	StartHandler(spec HandlerSpecifier) error
	StopHandler(spec HandlerSpecifier) error
	RemoveHandler(spec HandlerSpecifier) error
	RemoveTopic(topic Topic) error
	SendMessageToTopic(ctx context.Context, topic Topic, msg []byte) error
	Start()
	Close()
}
```

Supported handlers:

```
type Handler interface {
	Topic() Topic
	Specifier() HandlerSpecifier
	ValidateMsg(ctx context.Context, peer PeerID, rawData []byte) ValidateResult
	DeliverMsg(ctx context.Context, rawData []byte, cache ValidateCache)
	Report()
}
```

### Unit Test Coverage

```
PASS
coverage: 81.4% of statements
```

Also two race tests are implemented to detect race conditions. 

* `TestTopicRunner_race`
* `TestPubSubHost_Race`

Since race tests will take 10 seconds to finish and we do not have `--race` flag in our test suite, these two tests are skipped for now. To run these two tests, simply to comment out `t.Skip` line and run test with `--race`flag.


p2p/pubsub/psHost.go
```diff
@@ -0,0 +1,366 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func (psh *pubSubHost) deleteHandlersWithTopic(topic Topic) {
+	for spec, handler := range psh.handlers {
+		if handler.Topic() == topic {
+			delete(psh.handlers, spec)
```
rlan35, line 316: These are run in single threaded code, so no race condition right?

JackyWYX: Yes. All updates on the running handlers are done in a single thread. You can see the logic in `pubSubHost.run`

p2p/pubsub/topicRunner.go
```diff
@@ -0,0 +1,227 @@
...
+
+	return
+}
+
+func (tr *topicRunner) validateMsg(ctx context.Context, peer PeerID, raw *libp2p_pubsub.Message) libp2p_pubsub.ValidationResult {
+	m := newMessage(raw)
+	handlers := tr.getHandlers()
+	vResults := make([]ValidateResult, 0, len(handlers))
+
+	for _, handler := range handlers {
```
LeoHChen, line 93: are you expecting multiple validator for one message? This can be low efficient as validator has to run on every single libp2p messages.

p2p/pubsub/metric.go
```diff
@@ -0,0 +1,86 @@
+package pubsub
```
LeoHChen, line 1: if we are designing a new metrics system, we'd better use a global metrics system that can be added for each layer.  So that we don't have to add metrics system for each layer.

LeoHChen: I am thinking to add a new metrics system using Prometheus https://prometheus.io/docs/guides/go-application/

LeoHChen:
if we are going to redesign the network layer, we should re-consider to have stream support as well as the pubsub.  So that the node can deliver specific message to certain node for streaming, while also deliver broadcast message using pubsub for consensus.

p2p/pubsub/psHost.go
```diff
@@ -0,0 +1,366 @@
...
+		}
+	}
+}
+
+// Close close the pubSubHost.
+func (psh *pubSubHost) Close() {
+	task := closeTask{closed: make(chan struct{})}
+	psh.closeTaskC <- task
+
+	<-task.closed
```
rlan35, line 119: why is this needed?

p2p/pubsub/topicRunner.go
JackyWYX: Yes, there will be node handler validating node message and consensus handler handling consensus message. So basically, messages are only validated by one handler - Another one is always return `MsgAccept`

p2p/pubsub/psHost.go
JackyWYX: This is to wait for the close task to be executed at `run`.

p2p/pubsub/metric.go
JackyWYX: Oh, yea. This can be much greater than the simple metric we implement ourselves. Let me think it over and maybe we can do in another thread.

JackyWYX:
> if we are going to redesign the network layer, we should re-consider to have stream support as well as the pubsub. So that the node can deliver specific message to certain node for streaming, while also deliver broadcast message using pubsub for consensus.

Totally agree. This is definitely what I have been thinking of. Will implement later.

rlan35 (approved):
lgtm.

p2p/pubsub/topicRunner.go
```diff
@@ -0,0 +1,227 @@
...
+	handlers := tr.getHandlers()
+
+	for _, handler := range handlers {
+		tr.deliverMessageForHandler(ctx, msg, handler)
+	}
+}
+
+func (tr *topicRunner) deliverMessageForHandler(ctx context.Context, msg *message, handler Handler) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
JackyWYX, line 130: Wrong usage of cancel. Take care of this.

PR #3324, opened 2020-08-28
Explorer node return errors

LeoHChen:
Found many repeated "Address not found" error messages on explorer node, but it didn't return error to the caller.  So, return errors to the caller.

### DIFF
--- a/node/node_explorer.go
+++ b/node/node_explorer.go
@@ -28,6 +28,9 @@ var (
 	errFailFindingValidCommit = errors.New(
 		"explorer failed finding a valid committed message",
 	)
+	errAddressNotFound = errors.New(
+		"address not found",
+	)
 )
 
 // explorerMessageHandler passes received message in node_handler to explorer service
@@ -235,7 +238,7 @@ func (node *Node) GetTransactionsCount(address, txType string) (uint64, error) {
 	bytes, err := explorer.GetStorageInstance(node.SelfPeer.IP, node.SelfPeer.Port).GetDB().Get([]byte(key), nil)
 	if err != nil {
 		utils.Logger().Error().Err(err).Str("addr", address).Msg("[Explorer] Address not found")
-		return 0, nil
+		return 0, errAddressNotFound
 	}
 	if err = rlp.DecodeBytes(bytes, &addressData); err != nil {
 		utils.Logger().Error().Err(err).Msg("[Explorer] Cannot convert address data from DB")
@@ -258,7 +261,7 @@ func (node *Node) GetStakingTransactionsCount(address, txType string) (uint64, e
 	bytes, err := explorer.GetStorageInstance(node.SelfPeer.IP, node.SelfPeer.Port).GetDB().Get([]byte(key), nil)
 	if err != nil {
 		utils.Logger().Error().Err(err).Str("addr", address).Msg("[Explorer] Address not found")
-		return 0, nil
+		return 0, errAddressNotFound
 	}
 	if err = rlp.DecodeBytes(bytes, &addressData); err != nil {
 		utils.Logger().Error().Err(err).Msg("[Explorer] Cannot convert address data from DB")
--- a/node/node_explorer.go
+++ b/node/node_explorer.go
@@ -31,6 +31,9 @@ var (
 	errAddressNotFound = errors.New(
 		"address not found",
 	)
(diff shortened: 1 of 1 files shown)


LeoHChen:
Before
```bash
curl -s --request POST 'http://127.0.0.1:9500/' --header 'Content-Type: application/json' --data-raw '{ "jsonrpc": "2.0", "method": "hmyv2_getTransactionsCount", "params": ["one10akf7c089sl2w3e9yrmfn47g6fzy002h6zw47", "ALL"], "id": 1}' | jq .
```
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": 0
}
```


===
After
```bash
curl -s --request POST 'http://127.0.0.1:9500/' --header 'Content-Type: application/json' --data-raw '{ "jsonrpc": "2.0", "method": "hmyv2_getTransactionsCount", "params": ["one10akf7c089sl2w3e9yrmfn47g6fzy002h6zw47", "ALL"], "id": 1}' | jq .
```
```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "error": {
    "code": -32000,
    "message": "address not found"
  }
}
```

LeoHChen:
> I think nil error was intended since it seems that users are querying for the nonce of account that does not exist, which is 0 as expected. I think It would better to change log level instead of changing return errors. What do you think?

GetTransactionsCount return `nonce`, it sounds weird to me.  `nonce` shouldn't be equal to the transaction count, right?
 Also, it return errors to the client as the address is not found.  It's the client's job to understand that the nonce should be 0 if the address is not found.

LeoHChen:
Also, this function is in node_explorer, which is only intended for explorer node.

JackyWYX:
> GetTransactionsCount return `nonce`, it sounds weird to me. `nonce` shouldn't be equal to the transaction count, right?
> Also, it return errors to the client as the address is not found. It's the client's job to understand that the nonce should be 0 if the address is not found.

Yes, it is weird. But the fact is that we are using this function for nonce in our JS SDK, and changing this will surely break JS SDK. We should be more cautious if we are to apply the change. Also this misuse of `getTransactionCount` as nonce should be the root cause why we cannot send multiple transactions at a time (have to wait for one transaction to confirm to send next one). I discussed with RJ once, and proposed the fix [in this doc indexed 14](https://docs.google.com/document/d/10bVwqSt90jt6T34hm2rcO-5GWSSl0c46R0hUyxiVCds/edit#). Please take a look.

rlan35:
> > GetTransactionsCount return `nonce`, it sounds weird to me. `nonce` shouldn't be equal to the transaction count, right?
> > Also, it return errors to the client as the address is not found. It's the client's job to understand that the nonce should be 0 if the address is not found.
> 
> Yes, it is weird. But the fact is that we are using this function for nonce in our JS SDK, and changing this will surely break JS SDK. We should be more cautious if we are to apply the change. Also this misuse of `getTransactionCount` as `getNonce` should be the root cause why we cannot send multiple transactions at a time (have to wait for one transaction to confirm to send next one). I discussed with RJ once, and proposed the fix [in this doc indexed 14](https://docs.google.com/document/d/10bVwqSt90jt6T34hm2rcO-5GWSSl0c46R0hUyxiVCds/edit#). Please take a look.

getTransactionCount is conforming with Eth's naming, since initially we are just following eth's sdk. https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#gettransactioncount   And transaction count is the same as nonce in evm world. https://ethereum.stackexchange.com/questions/2527/is-there-a-way-to-find-an-accounts-current-transaction-nonce

JackyWYX:
> And transaction count is the same as nonce in evm world.

Yea, but is it true that when sending transaction, nonce should better refer to nonce in txPool instead of nonce in state?

rlan35:
I think it's better to change the log level from error, It's basically not an error if the user is supplying a new address that's not present in the state. 

Logically, if a user creates a new address and query with it. The result should be 0 because there is no history/txn on the address in the blockchain state, it shouldn't be an error.

rlan35:
> > And transaction count is the same as nonce in evm world.
> 
> Yea, but is it true that when sending transaction, nonce should better refer to nonce in txPool instead of nonce in state?

We have the support. Just need to specify it in the query:

image

PR #3325, opened 2020-08-28
[rosetta] Implement account endpoint

janet-harmony:
[ #3210 ] Implements Rosetta's `/account/balance` endpoint
There is no major logic outside of existing RPC functions, so no testing file included.

### DIFF
--- a/rosetta/services/account.go
+++ b/rosetta/services/account.go
@@ -26,6 +26,7 @@ func NewAccountAPI(hmy *hmy.Harmony) server.AccountAPIServicer {
 	}
 }
 
+// AccountBalance ...
 func (s *AccountAPI) AccountBalance(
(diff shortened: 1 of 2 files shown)


rosetta/services/account.go
```diff
@@ -0,0 +1,84 @@
...
+				return nil, common.NewError(common.BlockNotFoundError, map[string]interface{}{
+					"message": "block index not found",
+				})
+			}
+		}
+	}
+
+	addr := hmyCommon.ParseAddr(req.AccountIdentifier.Address)
+	blockNum := rpc.BlockNumber(block.Header().Header.Number().Int64())
+	balance, err := s.hmy.GetBalance(ctx, addr, blockNum)
```
rlan35, line 63: Should it get balance based on the state of the requested block? Otherwise, the block and the balance may not be consistent with each other.

PR #3337, opened 2020-09-10
Add back multi-sig support and fix the multi-sig check for harmony nodes

rlan35:
For multi-sig, there was a condition to disallow multi-sig signed from multiple accounts. It is to prevent malicious people to double sign and merge with others signature to trigger slashing for other people.

This condition broke testnet since in testnet our nodes contain multiple keys from different accounts. This PR fixes the condition so Harmony nodes can sign blocks with multi-sig without problems.

Basically, the first commit is a revert of a revert, which is already reviewed. The second commit is the fix: eeb8e31a31d164812f1d6231d45b62c73cf613c4

Tested on local net. Will deploy to testnet.


JackyWYX:
Looks good to me!

JackyWYX (approved):

PR #3338, opened 2020-09-11
Rosetta Implementation - pt2 FIX2 (Stage 3.2 of Node API Overhaul)

Daniel-VDM:
# Stage 3.2 FIX2 of [Node API Overhaul](https://github.com/harmony-one/harmony/issues/3210)

This PR is related to #3312 .

TLDR: This fixes the Undelegation reward operation as well as correctly accounts for the redelegation balance changes. This also adds an info dump for the last block that was garbage collected for non-archival DBs. 

## Details

Previously, the Undelegation transaction would result in a balance changing operation (other than gas). As this is not the case due to the locking period, the balance at a given block could be different than what it is on-chain. Therefore a 'special' transaction is added to the implementation to account for the Undelegation payouts at the last block of an epoch (similar to the pre-staking era block rewards). 

Note that all nodes (archival & pruned) store the last block of an epoch, however, the Undelegations saved for the said block are **after** the Undelegations have been paid out. Since the unpaid Undelegations are needed to recompute the paid out amount for each delegator, the state of the *second to the last block* of an epoch is required. This means that the non-archival nodes/DBs will not be able to provide this data unless the state is still in memory. This is why I have added the 'oldest block identifier' on the `network` endpoint for nodes running in the non-archival mode. 

The oldest block number is computed from the biggest block number that has been garbage collected plus 1. Since the block garbage collection is done monotonically w.r.t. block number, this should always give a block number that is in memory.

Lastly, I used the transaction receipt to report how much of the redelegation took from the account's balance. As well as did some general refactoring/clean-up given the new functions & variables.

All logic was tested using localnet as well as testnet using the [rosetta-cli](https://github.com/Daniel-VDM/rosetta-cli/tree/harmony).

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -161,14 +161,14 @@ type BlockChain struct {
 	procInterrupt int32          // interrupt signaler for block processing
 	wg            sync.WaitGroup // chain processing wait group for shutting down
 
-	engine                  consensus_engine.Engine
-	processor               Processor // block processor interface
-	validator               Validator // block and state validator interface
-	vmConfig                vm.Config
-	badBlocks               *lru.Cache              // Bad block cache
-	shouldPreserve          func(*types.Block) bool // Function used to determine whether should preserve the given block.
-	pendingSlashes          slash.Records
-	lastGarbCollectedBlkNum int64
+	engine                 consensus_engine.Engine
+	processor              Processor // block processor interface
+	validator              Validator // block and state validator interface
+	vmConfig               vm.Config
+	badBlocks              *lru.Cache              // Bad block cache
+	shouldPreserve         func(*types.Block) bool // Function used to determine whether should preserve the given block.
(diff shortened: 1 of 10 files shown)


core/blockchain.go
```diff
@@ -1168,6 +1170,9 @@ func (bc *BlockChain) WriteBlockWithState(
 						bc.triegc.Push(root, number)
 						break
 					}
+					if -number > bc.maxGarbCollectedBlkNum {
```
rlan35, line 1173: why is number negative?

Daniel-VDM: b/c its a prio queue, key by negative block nums. Notice the push here: 

https://github.com/Daniel-VDM/harmony/blob/834e09bf03a3ce0cda83866ee50755d5a4118b28/core/blockchain.go#L1134

rlan35 (approved):

PR #3340, opened 2020-09-11
[consensus] enable the consensus module can spin up doSync

JackyWYX:
## Description

This PR is to enable better coordination between consensus and sync. Two problem is fixed:

1. Logic optimization. Move consensus spin up the state sync after `tryCatchup`. Remove the 2 block delay.
2. Remove an unnecessary time.Ticker in sync logic. Thus will reduce the sync speed (from 3.5s -> 1.5s in local test).

## Test

Test done locally. Test branch: https://github.com/JackyWYX/harmony/tree/sync_consensus_fix_test.

In this branch, Following changes are made:
1. Disable the consensus insertChain logic in `tryCatchup`. So all block update are done in sync module.
2. Added a lot of print logs to observe the behaviour.

Run a local node on 

Following are observed:
1. Consensus module inform sync module upon receiving onCommitted message.
2. Each sync process takes < 1.5s to complete.

### DIFF
--- a/api/service/syncing/syncing.go
+++ b/api/service/syncing/syncing.go
@@ -914,10 +914,7 @@ func (ss *StateSync) SyncLoop(bc *core.BlockChain, worker *worker.Worker, isBeac
 	if !isBeacon {
 		ss.RegisterNodeInfo()
 	}
-	// remove SyncLoopFrequency
-	ticker := time.NewTicker(SyncLoopFrequency * time.Second)
-	defer ticker.Stop()
-	for range ticker.C {
+	for {
 		otherHeight := ss.getMaxPeerHeight(isBeacon)
 		currentHeight := bc.CurrentBlock().NumberU64()
 		if currentHeight >= otherHeight {
--- a/consensus/validator.go
+++ b/consensus/validator.go
@@ -188,6 +188,10 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
 
 	// tryCatchup is also run in onCommitted(), so need to lock with commitMutex.
 	consensus.tryCatchup()
+	if recvMsg.BlockNum > consensus.blockNum {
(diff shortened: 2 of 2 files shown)


rlan35:
What bug is it? I think the logic were intended and correct. If the node is consensusBlockNumBuffer behind the latest committed msg, it will trigger state sync.

JackyWYX:
> What bug is it? I think the logic were intended and correct. If the node is consensusBlockNumBuffer behind the latest committed msg, it will trigger state sync.

Previously, consensus informing sync about low block number is unreachable. Since `onCommitted` has already returned in check `consensus.isRightBlockNumCheck`.

```
func (consensus *Consensus) isRightBlockNumCheck(recvMsg *FBFTMessage) bool {
	if recvMsg.BlockNum < consensus.blockNum {
			Uint64("MsgBlockNum", recvMsg.BlockNum).
			Msg("Wrong BlockNum Received, ignoring!")
		return false
	} else if recvMsg.BlockNum-consensus.blockNum > MaxBlockNumDiff {
			Uint64("MsgBlockNum", recvMsg.BlockNum).
			Uint64("MaxBlockNumDiff", MaxBlockNumDiff).
			Msg("Received blockNum that is MaxBlockNumDiff further from the current blockNum!")
		return false
	}
	return true
}
```

You can see MaxBlockNumDiff have value 2 and is exactly the same as the condition to trigger a state sync. So it logic is well written here but never reachable.

rlan35:
> isRightBlockNumCheck

MaxBlockNumDiff has value of 100. It's not not a bug. I am sure the commit out of sync was able to be triggered.

JackyWYX:
> MaxBlockNumDiff has value of 100. It's not not a bug. I am sure the commit out of sync was able to be triggered.

You got me. You are right, not sure why I have that number of value 2 in the first place. Thus consensus related change in this PR is to put spin StateSync logic after executing `tryCatchup` in both `onPrepared` and `onCommitted`. Do you think this is a reasonable change?

rlan35 (approved):

PR #3342, opened 2020-09-12
[viewchange] remove redundant viewID

LeoHChen:

### DIFF
--- a/consensus/consensus.go
+++ b/consensus/consensus.go
@@ -77,7 +77,6 @@ type Consensus struct {
 	// blockNum: the next blockNumber that FBFT is going to agree on,
 	// should be equal to the blockNumber of next block
 	blockNum uint64
-	viewID   uint64
 	// Blockhash - 32 byte
 	blockHash [32]byte
 	// Block to run consensus on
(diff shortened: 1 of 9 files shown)


consensus/consensus.go
```diff
@@ -77,7 +77,6 @@ type Consensus struct {
 	// blockNum: the next blockNumber that FBFT is going to agree on,
 	// should be equal to the blockNumber of next block
 	blockNum uint64
-	viewID   uint64
```
rlan35, line 80: This view ID is not redundant. This is the viewID for the last successful consenus. The current.viewID is the moving viewID during view change. They can not be merged together, at least based on the current logic. Things will break if we don't keep track of them separately in current code.

LeoHChen: wow! that's a precious hidden knowledge. I can see why it is very confusing.

rlan35: yup. The whole view change code logic is not clean. Very error-prone.

LeoHChen:
this PR is replaced by https://github.com/harmony-one/harmony/pull/3347

PR #3345, opened 2020-09-15
Sync consensus fix

JackyWYX:
## Issue

Try to fix the consensus / sync coordination issue. Basically, this PR is:
1. Refactor consensus tryCatchup logic.
2. Add consensus last mile block in the end of state sync.

## Test

Tested locally with customized code. 

Also run a node on both testnet and mainnet, seems good so far.


JackyWYX:
There is some known issue in this PR. Will open another one once fixed.

PR #3351, opened 2020-09-17
[viewchange] reset view ID per epoch

LeoHChen:
This PR reset the current view ID and view changing ID at the beginning of each epoch to the block number.
ViewID itself is not saved on-chain, and only used an in-memory data.
So when a node is restarted, it didn't know what's the latest view ID and can only set it to the block height.
On the mainnet, as view change had happened many times, the view ID is far from the latest block number.
During a view change process, when the node is restarted, it fall far behind the latest view ID and took a while to catch up with latest view ID. This can cause a problem for a shard to recovery during view change when many nodes are restarted and they can't agree on the viewID.

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -178,6 +178,19 @@ func (consensus *Consensus) finalizeCommits() {
 		Int("numStakingTxns", len(block.StakingTransactions())).
 		Msg("HOORAY!!!!!!! CONSENSUS REACHED!!!!!!!")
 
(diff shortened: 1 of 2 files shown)


LeoHChen:
after discussion with RJ, found it is not needed for this PR as the view ID will be set to the last successful viewID during the blockchain init stage. https://github.com/harmony-one/harmony/blob/3adab4496c0327e232608ac427885d2b14d4b336/cmd/harmony/main.go#L602

PR #3352, opened 2020-09-17
[consensus][sync] Better coordination between state sync and consensus module.

JackyWYX:
## Issue

This PR is rebased on top of https://github.com/harmony-one/harmony/pull/3378. The PR solves the state sync and consensus coordination problem. Which is basically do the following change:

1. Sync module add last mile block in consensus cache at the end of state sync.
2. Consensus spin up state sync logic in both on prepared and on committed.
3. When calling onPrepared, the block will be saved to FBFT log for future use even if it fails verification at the moment. The verification will be called again in state sync.
4. Rewrite FBFTLog since it's not thread safe + used mapset is not stable. Replace with map.
5. Failure in `IsQuorumAchievedByMask` will also spin up state sync.
6. Some code refactors and some logic optimization.

## Test

Did on local test using [test branch](https://github.com/JackyWYX/harmony/tree/sync_consensus_fix_test2) on testnet.

In this test branch, `onPrepared` message is ignored at 10x+1 block and `onCommitted` message is ignored at 10x block. And unselected validator will mock submit the COMMIT message. Please take a look at code if you are interested. Following is an example console output:

```
[onCommitted] ---------------- 1779159 1779159
	isInSync: true
	signing rate [178/209]=85%
[onPrepared] ----------------- 1779160 1779160
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779160 1779160
	SKIPPED
	signing rate [179/210]=85%
[onPrepared] ----------------- 1779161 1779160
	SKIPPED
[onCommitted] ---------------- 1779161 1779160
	isInSync: false
	spin sync
	signing rate [179/211]=84%
[onPrepared] ----------------- 1779162 1779160
	inSync: false
	spin sync
[onCommitted] ---------------- 1779162 1779162
	isInSync: true
	signing rate [179/212]=84%
[onPrepared] ----------------- 1779163 1779163
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779163 1779163
	isInSync: true
	signing rate [180/213]=84%
[onPrepared] ----------------- 1779164 1779164
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779164 1779164
	isInSync: true
	signing rate [181/214]=84%
[onPrepared] ----------------- 1779165 1779165
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779165 1779165
	isInSync: true
	signing rate [182/215]=84%
[onPrepared] ----------------- 1779166 1779166
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779166 1779166
	isInSync: true
	signing rate [183/216]=84%
[onPrepared] ----------------- 1779167 1779167
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779167 1779167
	isInSync: true
	signing rate [184/217]=84%
[onPrepared] ----------------- 1779168 1779168
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779168 1779168
	isInSync: true
	signing rate [185/218]=84%
[onPrepared] ----------------- 1779169 1779169
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779169 1779169
	isInSync: true
	signing rate [186/219]=84%
[onPrepared] ----------------- 1779170 1779170
	inSync: true
	BINGO - SIGNED
[onCommitted] ---------------- 1779170 1779170
	SKIPPED
	signing rate [187/220]=85%
[onPrepared] ----------------- 1779171 1779170
	SKIPPED
[onCommitted] ---------------- 1779171 1779170
	isInSync: false
	spin sync
	signing rate [187/221]=84%
```

Following is the observation:

1. 2 out of 10 blocks are abnormal, either lacks onPrepared or lacks onCommitted.
2. State sync for small block difference can be finished within 1 block time.
3. Sync can catch up and the validator is able to sign all left 8 normal blocks. Obtaining ~85% signing signature, which is quite good.


consensus/checks.go
```diff
@@ -141,24 +144,6 @@ func (consensus *Consensus) onPreparedSanityChecks(
 			Msg("[OnPrepared] BlockHash not match")
 		return false
 	}
-	if consensus.current.Mode() == Normal {
```
rlan35, line 144: why is this removed?

consensus/fbft_log.go
```diff
@@ -65,127 +61,160 @@ func (m *FBFTMessage) String() string {
...
+	return hash.FromRLPNew256(m)
+}
+
+// FBFTLog represents the log stored by a node during FBFT process
+type FBFTLog struct {
+	blocks         map[common.Hash]*types.Block // store blocks received in FBFT
+	messages       map[common.Hash]*FBFTMessage // store messages received in FBFT
+	verifiedBlocks map[common.Hash]struct{}     // store block hashes for blocks that has already been verified
+
+	lock sync.Mutex
```
rlan35, line 75: how about use separate locks to avoid conflict?

consensus/checks.go
JackyWYX: It is not removed but instead the logic is rewritten. Previously, if the node is in syncing or listening mode, the block will be always added to FBFT log without block verification. This might cause consensus memory have some malicious block. Thus, the block verification logic is also applied to all blocks in FBFTLogs and added `verifiedBlocks` in FBFTLog. The block verification logics are:

1. tryCatchup (consensus/consensus_v2.go:495)
2. `LastMileBlockIterator.Next` (consensus/consensus_v2.go:446)
3.  `onPrepared` (consensus/validator.go:204)

consensus/validator.go
```diff
@@ -140,13 +142,11 @@ func (consensus *Consensus) onPrepared(msg *msg_pb.Message) {
 		consensus.getLogger().Error().Err(err).Msg("ReadSignatureBitmapPayload failed!")
 		return
 	}
-
 	if !consensus.Decider.IsQuorumAchievedByMask(mask) {
-		consensus.getLogger().Warn().
-			Msgf("[OnPrepared] Quorum Not achieved")
+		consensus.getLogger().Warn().Msgf("[OnPrepared] Quorum Not achieved.")
+		consensus.spinUpStateSync()
```
rlan35, line 147: not sure this logic is correct. If the quorum check failure is caused by node out of sync, they it should detect that by checking block number somewhere else.

consensus/validator.go
```diff
@@ -301,16 +313,19 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 	if !consensus.isRightBlockNumCheck(recvMsg) {
 		return
 	}
+	if len(recvMsg.SenderPubkeys) != 1 {
```
rlan35, line 316: why not directly use isSendByLeader()

consensus/fbft_log.go
JackyWYX: Do you mean one lock for each of the map? I thought this is not the performance bottle neck so I didn't optimize this. Do you think this might be the potential bottleneck?

consensus/validator.go
JackyWYX: This is the temporary hack before we add the leader public key check. Considering the attack scenario, we can only spin up state sync if 

1. block number is left behind. 
2. Message is sent by leader. 
3. Message passed all validation (requires latest block in state and consensus cache).

Currently, the second part is missing and state sync is spinned up after all check pass. Only if we have leader public key verified (in the next PR), we can spin up state sync in condition 1+2 or 2+3.

consensus/validator.go
JackyWYX: Since currently there is no guarantee that the leader message is signed by leader key (on Testnet). The consensus will not work if we do check the leader public key. This will be resolved in future PR.

consensus/validator.go
JackyWYX: The leader check is outside `onCommitted` and `onPrepared`. Added spin up sync logic right after low block number is detected.

consensus/fbft_log.go
JackyWYX: Different sync mutex on different fields is done. Please check the latest code.

consensus/validator.go
JackyWYX: The function key check is already outside of `onPrepared` and `onCommitted`. Thus we don't need to write this logic again. This piece of code is removed.

consensus/fbft_log.go
```diff
@@ -60,127 +56,179 @@ func (m *FBFTMessage) String() string {
...
+		BlockHash: m.BlockHash,
+		Payload:   m.Payload,
+	}
+	if m.SenderPubkey != nil {
+		hashData.SenderKey = m.SenderPubkey.Bytes
+	}
+	if m.LeaderPubkey != nil {
+		hashData.LeaderKey = m.LeaderPubkey.Bytes
+	}
+	return hash.FromRLPNew256(hashData)
```
rlan35, line 79: the rlp hash could be a performance bottleneck. better directly use some of the basic data fields such as blockHash+msgType+sender as key.

JackyWYX: Thanks for feedback. Updated with ID function to specify a unique FBFT message.

consensus/validator.go
```diff
@@ -316,3 +316,14 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
...
 }
+
+func (consensus *Consensus) spinUpStateSync() {
+	select {
+	case consensus.BlockNumLowChan <- struct{}{}:
+		consensus.current.SetMode(Syncing)
+		for _, v := range consensus.consensusTimeout {
+			v.Stop()
+		}
+	default:
```
rlan35, line 327: what if the reader are not reading. Then this part will block forever right? potential goroutine leak. The original code has a timeout

consensus/validator.go
```diff
@@ -287,22 +297,12 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 	consensus.aggregatedCommitSig = aggSig
 	consensus.commitBitmap = mask
 
-	if recvMsg.BlockNum > consensus.blockNum && recvMsg.BlockNum-consensus.blockNum > consensusBlockNumBuffer {
+	consensus.tryCatchup()
+	if recvMsg.BlockNum > consensus.blockNum {
```
rlan35, line 301: do we still need this part when we already have the "spinUpStateSync" above?

consensus/validator.go
JackyWYX: If `consesus.BlockNumLowChan` is occupied, the select will go to default, and the function will return immediately. Also, the function is not called with a new goroutine. So it shall not be not a problem as far as I see.

consensus/validator.go
```diff
@@ -247,16 +255,18 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 	if !consensus.isRightBlockNumCheck(recvMsg) {
 		return
 	}
+	if recvMsg.BlockNum > consensus.blockNum {
```
rlan35, line 258: the previous condition has "recvMsg.BlockNum-consensus.blockNum > consensusBlockNumBuffer" to avoid entering into sync mode prematurely. Now the buffer is gone, is the logic taking care of this case? "spinUpStateSync" won't disturb normal consensus logic?

consensus/validator.go
rlan35: ok, right.

consensus/validator.go
JackyWYX: Removing buffer is to enable better sync consensus coordination, which spin sync immediately,  resulting in the benchmark illustrated in the PR description. For the previous condition, when will it happen when "entering into sync mode prematurely"? Could you specify?

Yes, `spinUpStateSync` does not interrupt message handling logic. The message will be added to FBFT log if passed aggregated signature verification.

consensus/validator.go
JackyWYX: I think yes. If tryCatchup does not help to get consensus block number to the latest (In most cases, this could be the error of missing state in block verification), committed message is not handled correctly. So I don't want it to interfere with the consensus time out stuff. Thus I return immediately.

RLlane2126: Approve

RLlane2126 (approved):

consensus/validator.go
rlan35: prematurely meaning it doesn't really miss the block, just that some messages are delivered out of order. Say committed_10 is delivered before committed_9, then it will trigger the sync, when it could self recover when the committed_9 is delivered. But as long as the sync won't interfere with consensus, it is fine to trigger sync anytime. Previously, I think when the node is in sync mode, it may interfere with consensus.

rlan35 (approved):

rlan35:
looks good.

consensus/validator.go
JackyWYX: Thanks for clarification. Yes, Sync will only used when deciding whether or not to sign a block during message handling. So shall be good here :)

PR #3353, opened 2020-09-18
[rpc] expose consensus internal data via rpc

LeoHChen:
This is used to debug consensus/viewchange state.

### DIFF
--- a/hmy/hmy.go
+++ b/hmy/hmy.go
@@ -93,6 +93,8 @@ type NodeAPI interface {
 	ListTopic() []string
 	ListBlockedPeer() []peer.ID
 
+	GetConsensusInternal() commonRPC.ConsensusInternal
+
 	// debug API
 	GetConsensusMode() string
 	GetConsensusPhase() string
@@ -176,6 +178,8 @@ func (hmy *Harmony) GetNodeMetadata() commonRPC.NodeMetadata {
 	c := commonRPC.C{}
 	c.TotalKnownPeers, c.Connected, c.NotConnected = hmy.NodeAPI.PeerConnectivity()
 
+	consensusInternal := hmy.NodeAPI.GetConsensusInternal()
+
 	return commonRPC.NodeMetadata{
 		BLSPublicKey:   blsKeys,
 		Version:        nodeconfig.GetVersion(),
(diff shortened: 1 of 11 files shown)


LeoHChen:
image


Example of watchdog output on STN.

node/api.go
```diff
@@ -126,3 +127,21 @@ func (node *Node) GetConsensusViewChangingID() uint64 {
...
+
+// GetConsensusInternal returns consensus internal data
+func (node *Node) GetConsensusInternal() rpc_common.ConsensusInternal {
+	return rpc_common.ConsensusInternal{
+		ViewID:        node.GetConsensusCurViewID(),
+		VCID:          node.GetConsensusViewChangingID(),
+		Mode:          node.GetConsensusMode(),
+		Phase:         node.GetConsensusPhase(),
+		BlockNum:      node.GetConsensusBlockNum(),
+		InSync:        !node.IsOutOfSync(node.Blockchain()),
```
Daniel-VDM, line 144: This locks the syncing service, prob best to not put this in the node metadata as its called quite frequently.

hmy/hmy.go
```diff
@@ -190,6 +194,7 @@ func (hmy *Harmony) GetNodeMetadata() commonRPC.NodeMetadata {
 		Archival:       cfg.GetArchival(),
 		NodeBootTime:   hmy.NodeAPI.GetNodeBootTime(),
 		PeerID:         nodeconfig.GetPeerID(),
+		Consensus:      consensusInternal,
```
Daniel-VDM, line 197: let's not put this in metadata, reason is in a different comment.

rpc/common/types.go
```diff
@@ -37,6 +37,17 @@ type C struct {
...
 
+// ConsensusInternal captures consensus internal data
+type ConsensusInternal struct {
+	ViewID        uint64 `json:"viewid"`
+	VCID          uint64 `json:"vcid"`
+	Mode          string `json:"mode"`
+	Phase         string `json:"phase"`
+	BlockNum      uint64 `json:"blocknum"`
+	InSync        bool   `json:"insync"`
+	ConsensusTime int64  `json:"finality"`
```
Daniel-VDM, line 48: `vcid` is not clear IMO. Maybe also stick to [kebab case](https://wiki.c2.com/?KebabCase) for the unmarshalled names if we decide to keep it in node metadata. Or use snake case.

node/api.go
Daniel-VDM: Suggestion is to put it as a separate RPC.

Daniel-VDM (changes requested):
Some concerns about putting it in node metadata & some naming nits

node/api.go
LeoHChen: got it, will remove it.

hmy/hmy.go
LeoHChen: I'll remove the IsSync variable then.

Daniel-VDM (approved):

PR #3354, opened 2020-09-18
Fast viewchange converge

LeoHChen:


rlan35 (changes requested):
Discussed in person. We will probably choose the synchronous timestamp based viewID setting protocol.

PR #3355, opened 2020-09-19
Make block proposal async and receiving last commit sig through channel

rlan35:
Following the idea of pipelining: https://www.notion.so/harmonyone/3s-Finality-5a324362edd647f692c91c9c31c6d6ea

This PR prepares for the async block proposal logic to make the block proposal able to start first and wait for the last commit sigs's arrival to fully complete the process.

Note this pr is just adding the async part for the block proposal to wait on the commitSigs. This PR is not implementing the real early block proposal (pipelining) part yet. So the block proposal is still after last consensus fully done.


consensus/engine/consensus_engine.go
```diff
@@ -115,6 +115,6 @@ type Engine interface {
 		state *state.DB, txs []*types.Transaction,
 		receipts []*types.Receipt, outcxs []*types.CXReceipt,
 		incxs []*types.CXReceiptsProof, stks staking.StakingTransactions,
-		doubleSigners slash.Records,
+		doubleSigners slash.Records, sigsReady chan bool,
```
JackyWYX, line 118: I am not a big fan of this hack approach since there are huge risks to add multi-threading logic in the state machine which is designed to be single threaded which returns immediately. 

1. Now function `InsertChain` holds the blockchain global mutex now will probably get blocked until sigsReady timeout. Do you think it will cause other goroutines running `InsertChain` intentions to be failed unexpected. causing the locking scenario of blockchain to be extensively complicated. I would suggest not to do this "hack" coding, instead I would prefer do it in a dummy way that only finalize after collected enough commit signature.

2. Is there any code that provides a constraint on leader node cannot enter prepare phase until last block is committed?

node/node_newblock.go
```diff
@@ -47,13 +47,20 @@ func (node *Node) WaitForConsensusReadyV2(readySignal chan struct{}, stopChan ch
 						Uint64("blockNum", node.Blockchain().CurrentBlock().NumberU64()+1).
 						Msg("PROPOSING NEW BLOCK ------------------------------------------------")
 
+					// Prepare last commit signatures
+					commitSigs := make(chan []byte)
+					sigs, err := node.Consensus.BlockCommitSigs(node.Blockchain().CurrentBlock().NumberU64())
```
JackyWYX, line 52: I am a little confused here: The sigs results are computed simultaneously. What's the intention of adding channel logic here? Oh, is it true that the pipeline feature is not implemented in this PR?

JackyWYX (commented):
After all I think we would better to do Finalize in a "dummy" way unless there are very good reasons to support this. Do you have any unspoken ideas in your mind?

consensus/consensus.go
```diff
@@ -86,6 +86,9 @@ type Consensus struct {
 	OnConsensusDone func(*types.Block) error
 	// The verifier func passed from Node object
 	BlockVerifier BlockVerifierFunc
+	// Block Proposer proposes new block at the specific block number
+	BlockProposer func(blockNum uint64) (*types.Block, error)
```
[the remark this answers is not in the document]
rlan35: yup. removed.

node/node_newblock.go
rlan35: Sorry, didn't clarify in the description. This pr is just adding the async part for the block proposal to wait on the commitSigs. This PR is not implementing the real early block proposal (pipelining) part yet. just want to make things separate.

JackyWYX: OK, sure, NP

consensus/engine/consensus_engine.go
rlan35: 1. Yea, now all calls on Finalize needs to be async. That's a tradeoff of adding this feature. Any other idea to do this in a better way? Note we don't want to block/wait and do the Finalize() as a whole afterwards, because finalize contains the block reward logic for shard chains, which is expensive and could be computed earlier without the commit signatures of last beacon block.

2. We need the commit signatures to compute block reward for the new block.

rlan35:
> After all I think we would better to do Finalize in a "dummy" way unless there are very good reasons to support this. Do you have any unspoken ideas in your mind?

what do you mean in dummy way? take a look at the comment about block reward computation logic in Finalize().

JackyWYX:
My "dummy" way would be simply insert block after collected commit sigs. Compared to current hack approach, it is true that it might increase the delay of committing blocks after received enough commit sigs, but it will also keep the code itself robust against future changes. What this PR is doing is actually a very radical change (though small change in code) to the usage patterns for function `InsertChain`.

I am curious about following questions:
1. How much time does it take to process a block before accumulating rewards with commit sigs?
2. How will this time effect the block period? Does this duration directly add to the block period? Can we design a pipeline (a design pattern) to mitigate this impact?
3. What about divide current `Processor.Process` to two different functions? The intermediate state can be cached to a temporary memory cache and is ready to grab and use once the commit sigs are arrived?

rlan35:
> My "dummy" way would be simply insert block after collected commit sigs. Compared to current hack approach, it is true that it might increase the delay of committing blocks after received enough commit sigs, but it will also keep the code itself robust against future changes. What this PR is doing is actually a very radical change (though small change in code) to the usage patterns for function `InsertChain`.
> 
> I am curious about following questions:
> 
> 1. How much time does it take to process a block before accumulating rewards with commit sigs?

The block reward logic for crosslinks takes around 0.2+ seconds.

> 2. How will this time effect the block period? Does this duration directly add to the block period? Can we design a pipeline (a design pattern) to mitigate this impact?

This is the pipelining we are trying to do to reduce block time. Take a look https://docs.google.com/presentation/d/1vNegkz7sgNjb0OANiR0pwdxI3NkVa-ykHaQ9ne02B08/edit#slide=id.ga1bfef749f_0_335
> 3. What about divide current `Processor.Process` to two different functions? The intermediate state can be cached to a temporary memory cache and is ready to grab and use once the commit sigs are arrived?

It's the same problem we talked about. Then the called of Process needs to know the details about how to call them in a correct way and how to assemble the results correctly. The complexity is there anyway, because we are making things async. Don't think there is easy way to remove the complexity while adding async here.

JackyWYX:
> This is the pipelining we are trying to do to reduce block time. Take a look https://docs.google.com/presentation/d/1vNegkz7sgNjb0OANiR0pwdxI3NkVa-ykHaQ9ne02B08/edit#slide=id.ga1bfef749f_0_335

I see. I thought the consensus pipeline is a little different from what is in the slide. What do you think is the problem in the pipeline diagram in my mind?

Screen Shot 2020-10-20 at 3 57 11 PM

Also, there is another word "pipeline" refers to a design pattern "pipeline". I would encourage you to take a look at Ethereum ChainIndexer implementation (core/chain_indexer.go:197, core/chain_indexer.go:304) might gives some coding insights?

rlan35:
> > This is the pipelining we are trying to do to reduce block time. Take a look https://docs.google.com/presentation/d/1vNegkz7sgNjb0OANiR0pwdxI3NkVa-ykHaQ9ne02B08/edit#slide=id.ga1bfef749f_0_335
> 
> I see. I thought the consensus pipeline is a little different from what is in the slide. What do you think is the problem in the pipeline diagram in my mind?
> 
> Screen Shot 2020-10-20 at 4 05 32 PM
> 
> Also, there is another word "pipeline" refers to a design pattern "pipeline". I would encourage you to take a look at Ethereum ChainIndexer implementation (core/chain_indexer.go:197, core/chain_indexer.go:304) might gives some coding insights?

The new block can not be proposed before 66% of commit, because that's when the block is theoretically confirmed and finalized (the txns and state are confirmed). before that, the newly proposed block can not guarantee correctness.

for the chain_indexer code, it seems the logic is much simpler since it's only updating some fields of the header. In our case, we need to preserve the order of the state processing, which is critical in the correctness of state computation. I would rather not change the current state processing logic which could introduce more issues.

JackyWYX:
> The new block can not be proposed before 66% of commit, because that's when the block is theoretically confirmed and finalized (the txns and state are confirmed). before that, the newly proposed block can not guarantee correctness.

Thanks for clarification. Now I understand the scenario better. Thanks.

>  I would rather not change the current state processing logic which could introduce more issues.

I also agree this. It's definitely not single pipeline pattern alone can address all consensus logic.

> It's the same problem we talked about. With this, the caller of Process needs to know the details about how to call them in a correct way and how to assemble the results correctly. The complexity is there anyway, because we are making things async. Don't think there is easy way to remove the complexity while adding async here.

I am still a little obsessed about `InsertChain` holding the blockchain mutex until sigs channel times out. I would still suggest to make InsertChain single threaded, but I could also live with the hack solution since it takes the least effort to accomplish the target.

consensus/engine/consensus_engine.go
```diff
@@ -110,11 +110,12 @@ type Engine interface {
 	// and assembles the final block.
 	// Note: The block header and state database might be updated to reflect any
 	// consensus rules that happen at finalization (e.g. block rewards).
+	// sigsReady signal indicates whether the commit sigs are populated in the header object.
```
JackyWYX, line 113: Do you think adding a note in comment saying that the `Finalize` is blocked would be better?

JackyWYX (approved):

node/worker/worker.go
```diff
@@ -492,15 +485,39 @@ func (w *Worker) FinalizeNewBlock(
 	}
 	state := w.current.state.Copy()
 	copyHeader := types.CopyHeader(w.current.header)
```
gupadhyaya, line 487: is there a side-effect to not copying the sig and bitmap to `w.current.header`? earlier, everything was written to `w.current.header` and then deep-copied.

node/node_newblock.go
```diff
@@ -47,13 +47,22 @@ func (node *Node) WaitForConsensusReadyV2(readySignal chan struct{}, stopChan ch
...
+
 					node.Consensus.StartFinalityCount()
-					newBlock, err := node.proposeNewBlock()
 					if err != nil {
-						utils.Logger().Err(err).Msg("!!!!!!!!!Failed Proposing New Block!!!!!!!!!")
+						utils.Logger().Error().Err(err).Msg("[proposeNewBlock] Cannot get commit signatures from last block")
```
gupadhyaya, line 56: minor `proposeNewBlock` => `ProposeNewBlock`

node/node_newblock.go
```diff
@@ -225,17 +235,22 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
 
-	// Prepare last commit signatures
-	sig, mask, err := node.Consensus.BlockCommitSig(header.Number().Uint64() - 1)
+	finalizedBlock, err := node.Worker.FinalizeNewBlock(
+		commitSigs, node.Consensus.GetCurBlockViewID(),
+		coinbase, crossLinksToPropose, shardState,
+	)
 	if err != nil {
-		utils.Logger().Error().Err(err).Msg("[proposeNewBlock] Cannot get commit signatures from last block")
+		utils.Logger().Error().Err(err).Msg("[proposeNewBlock] Failed finalizing the new block")
```
gupadhyaya, line 243: minor `proposeNewBlock` => `ProposeNewBlock`

node/node_newblock.go
```diff
@@ -225,17 +235,22 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
 		return nil, err
 	}
+	utils.Logger().Info().Msg("[proposeNewBlock] verifying the new block header")
```
gupadhyaya, line 246: minor `proposeNewBlock` => `ProposeNewBlock`. what is the purpose of this info msg?

node/node_newblock.go
```diff
@@ -225,17 +235,22 @@ func (node *Node) proposeNewBlock() (*types.Block, error) {
...
+	err = node.Blockchain().Validator().ValidateHeader(finalizedBlock, true)
 
-	return node.Worker.FinalizeNewBlock(
-		sig, mask, node.Consensus.GetCurBlockViewID(),
-		coinbase, crossLinksToPropose, shardState,
-	)
+	if err != nil {
+		utils.Logger().Error().Err(err).Msg("[proposeNewBlock] Failed verifying the new block header")
```
gupadhyaya, line 250: minor `proposeNewBlock` => `ProposeNewBlock`

node/node_newblock.go
rlan35: done.

node/node_newblock.go
rlan35: done.

node/worker/worker.go
rlan35: it is fine since the commit sigs are empty is the header here. Will be populated later.

consensus/engine/consensus_engine.go
rlan35: Done.

gupadhyaya (approved):

node/node_newblock.go
rlan35: Done.

node/node_newblock.go
rlan35: Done.

PR #3357, opened 2020-09-20
[viewchange] fix getNextLeader

LeoHChen:
use a separate NextLeaderPubKey variable to keep track of the pubkey of the next leader

do not change current pubkey unless leader change succeeded

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -120,8 +120,12 @@ func (consensus *Consensus) switchPhase(desired FBFTPhase, override bool) {
 }
 
 // GetNextLeaderKey uniquely determine who is the leader for given viewID
-func (consensus *Consensus) GetNextLeaderKey() *bls.PublicKeyWrapper {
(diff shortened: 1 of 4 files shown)


consensus/view_change.go
```diff
@@ -163,13 +167,13 @@ func (consensus *Consensus) startViewChange(viewID uint64) {
 	consensus.consensusTimeout[timeoutBootstrap].Stop()
 	consensus.current.SetMode(ViewChanging)
 	consensus.SetViewChangingID(viewID)
-	consensus.LeaderPubKey = consensus.GetNextLeaderKey()
+	consensus.NextLeaderPubKey = consensus.GetNextLeaderKey(viewID)
```
rlan35, line 170: This NextLeaderPubKey field is only used in here. Don't need to be a struct variable. it can just be a local variable.

LeoHChen: Fixed, can you review again.

rlan35 (approved):

PR #3372, opened 2020-09-27
make receipt log sorted

rlan35:
Non-deterministic ordering of the map will trigger invalid receipt root for some of the unlucky machines with a different ordering than the leader.


JackyWYX (approved):
lgtm. This is a hard fork which requires extensive caution and testing.

rlan35:
> lgtm. This is a hard fork which requires extensive caution and testing.

I've sync'ed the node to latest block without problem.

PR #3373, opened 2020-09-29
viewchange revamp

LeoHChen:


consensus/view_change_construct.go
```diff
@@ -0,0 +1,415 @@
...
+					vc.m1Payload = append(preparedMsg.BlockHash[:], preparedMsg.Payload...)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// IsM1PayloadEmpty returns true if m1Payload is not set
+func (vc *ViewChange) IsM1PayloadEmpty() bool {
```
JackyWYX, line 393: I see this function is used in both locked and unlocked version. Is is better to write a private version like: 
```
func (vc *ViewChange) IsM1PayloadEmpty() bool {
        vc.vcLock.RLock()
        defer vc.vcLock.RUnlock()
	return vc.isM1PayloadEmpty()
}

func (vc *ViewChange) isM1PayloadEmpty() bool {
	return len(vc.m1Payload) == 0
}
```

And private version can be used in ViewChange public methods.

consensus/consensus.go
```diff
@@ -85,15 +70,15 @@ type Consensus struct {
 	IgnoreViewIDCheck *abool.AtomicBool
 	// consensus mutex
 	mutex sync.Mutex
-	// mutex for view change
-	vcLock sync.Mutex
+	// ViewChange struct
+	VC *ViewChange
```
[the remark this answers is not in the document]
LeoHChen: noted. let me check then.

consensus/view_change.go
```diff
@@ -98,30 +90,83 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+// The next view ID is calculated based on the difference of validator's timestamp
+// and the block's timestamp. So that it can be deterministic to return the next view ID
+// only based on the blockchain block and the validator's current timestamp.
+// The next view ID is the single factor used to determine
+// the next leader, so it is mod the number of nodes per shard.
+// It returns the next viewID and duration of the view change
+// The view change duration is a fixed duration now to avoid stuck into offline nodes during
+// the view change.
+// viewID is only used as the fallback mechansim to determine the nextViewID
+func (consensus *Consensus) GetNextViewID(viewID uint64) (uint64, time.Duration) {
```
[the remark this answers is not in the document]
LeoHChen: makes sense.  Let me double check then.

consensus/view_change_construct.go
LeoHChen: let me check, better protected in rlock.

consensus/view_change_construct.go
```diff
@@ -0,0 +1,415 @@
...
+	M2
+	// ERR is the error condition when parsing the message type
+	ERR
+
+	// ValidPayloadLength is the valid length for viewchange payload
+	ValidPayloadLength = 32 + bls.BLSSignatureSizeInBytes
+)
+
+// ViewChange encapsulate all the view change related data structure and functions
+type ViewChange struct {
```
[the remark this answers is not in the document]
LeoHChen: let me double check.

core/types/block.go
```diff
@@ -223,6 +223,9 @@ type Block struct {
 	commitSigAndBitmap []byte
 }
 
+// BlockVerifier is a function used to verify the block
+type BlockVerifier func(*Block) error
```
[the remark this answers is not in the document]
LeoHChen: sounds good.

consensus/view_change.go
```diff
@@ -98,30 +90,83 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+// the next leader based on the gap of the viewID of the view change and the last
+// know view id of the block.
+func (consensus *Consensus) GetNextLeaderKey(viewID uint64) *bls.PublicKeyWrapper {
+	gap := 1
+	if viewID > consensus.GetCurBlockViewID() {
+		gap = int(viewID - consensus.GetCurBlockViewID())
+	}
+	var lastLeaderPubKey *bls.PublicKeyWrapper
+	var err error
+	if consensus.ChainReader == nil {
```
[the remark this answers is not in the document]
LeoHChen: this tricky though.  it may happen during bootstrap stage.

consensus/checks.go
```diff
@@ -174,18 +176,28 @@ func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
 	}
 	if consensus.blockNum < recvMsg.BlockNum {
 		consensus.getLogger().Warn().
-			Msg("[onViewChange] New Leader Has Lower Blocknum")
+			Msg("[onViewChangeSanityCheck] MsgBlockNum is different from my BlockNumber")
 		return false
 	}
 	if consensus.IsViewChangingMode() &&
-		consensus.GetViewChangingID() > recvMsg.ViewID {
+		consensus.GetCurBlockViewID() > recvMsg.ViewID {
```
rlan35, line 183: So this is intentionally changed so we are more lenient on the viewID received?

consensus/config.go
```diff
@@ -4,14 +4,15 @@ import "time"
...
+	// default timeout configuration is shorten to 30 seconds as the consensus is 5s
+	viewChangeTimeout = 30
+	// The duration of viewChangeTimeout for each view change
+	viewChangeDuration time.Duration = viewChangeTimeout * time.Second
 
 	// timeout duration for announce/prepare/commit
-	phaseDuration     time.Duration = 60 * time.Second
-	bootstrapDuration time.Duration = 600 * time.Second
+	// shorten the duration from 60 to 30 seconds as the consensus is 5s
+	phaseDuration     time.Duration = 30 * time.Second
```
rlan35, line 14: Let's keep the timeout for consensus 60s? so the retries may save the consensus before we have to enter view change.

consensus/view_change.go
```diff
@@ -98,30 +90,18 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+	gap := 1
+	if viewID > consensus.GetCurBlockViewID() {
+		gap = int(viewID - consensus.GetCurBlockViewID())
+	}
+	consensus.getLogger().Info().
+		Str("leaderPubKey", consensus.LeaderPubKey.Bytes.Hex()).
+		Uint64("newViewID", viewID).
+		Uint64("myCurBlockViewID", consensus.GetCurBlockViewID()).
+		Msg("[GetNextLeaderKey] got leaderPubKey from coinbase")
+	wasFound, next := consensus.Decider.NextAfter(consensus.LeaderPubKey, gap)
```
rlan35, line 104: consensus.LeaderPubKey is updated for every view change. This logic change will mess up the leader change. Let's just keep the same logic as a pure refactor?

rlan35, line 104: consensus.LeaderPubKey is updated for every view change. This logic change will mess up the leader change. Let's just keep the same logic as a pure refactor?

consensus/view_change.go
```diff
@@ -214,228 +191,72 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-			}
-
-			// Verify the multi-sig for prepare phase
-			if !aggSig.VerifyHash(mask.AggregatePublic, blockHash[:]) {
-				consensus.getLogger().Warn().
-					Hex("blockHash", blockHash).
-					Msg("[onViewChange] failed to verify multi signature for m1 prepared payload")
-				return
+	if msgType == M1 {
+		preparedMsgs := consensus.FBFTLog.GetMessagesByTypeSeq(
```
rlan35, line 219: For the viewChange's M1 msg. The block is parsed from the message. Not from local log, the FBFTLog is for self commit of viewChange (should be in InitPayload). Can you double check?

consensus/checks.go
LeoHChen: The true lower bound is curBlockViewID, not the viewchangingID.

consensus/config.go
LeoHChen: we've shorten the consensus time from 8 to 5. Don't you think we can shorten the phaseDuration? The retry is faster as well, right now.

LeoHChen: I also want to speed up the converge process with shorter phaseDuration.

consensus/view_change.go
LeoHChen: okay, I'll keep the old one for this PR.  But I need this change for the next PR though.

consensus/view_change.go
LeoHChen: will double check.

consensus/checks.go
rlan35: cool. just double check that this change is intentional.

consensus/config.go
rlan35: The view change protocol is like the last resort. before that, it's better to rely on the retry logic because eventually the system's problems should only happen on the network level and retry should be able to get things through most of the time. We don't want to prematurely enter view change too quickly to make things more complicated. But it's reasonable to shorten the timeout a bit. Maybe 45 seconds?

consensus/view_change.go
rlan35: Yea, I know it's for new logic, let's keep the new logic separate.

consensus/view_change.go
LeoHChen: The original logic also add a new prepared message to FBFTLog if the validator can't find it locally.  So this validator has prepared message, but it received view change with M1.

image

LeoHChen:
Use new PR for review. https://github.com/harmony-one/harmony/pull/3379

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+				for i, key := range privKeys {
+					if err := vc.bhpBitmap[viewID].SetKey(key.Pub.Bytes, true); err != nil {
+						vc.getLogger().Warn().Int("index", i).Msg("[InitPayload] bhpBitmap setkey failed")
+						continue
+					}
+					vc.bhpSigs[viewID][key.Pub.Bytes.Hex()] = key.Pri.SignHash(msgToSign)
+				}
+				// if m1Payload is empty, we just add one
+				if vc.isM1PayloadEmpty() {
+					vc.m1Payload = append(preparedMsg.BlockHash[:], preparedMsg.Payload...)
```
rlan35, line 384: "consensus.m1Payload = append(recvMsg.Payload[:0:0], recvMsg.Payload...)" is this logic changed here?

consensus/view_change.go
rlan35: The original logic only add the m1 payload as fbft log if the payload is never seen before. Did you keep that logic?

consensus/view_change_construct.go
LeoHChen: en, will add it back to set m1Payload to the recvMsg.Paylaod it can't be initialized from local prepared message.

consensus/view_change.go
LeoHChen: discussed in the new PR.

PR #3375, opened 2020-09-30
[core] Return EVM errors from ApplyMessage for external use

janet-harmony:
Part 1 of [EVM Error changes](https://gist.github.com/janet-harmony/980f6cacc3e398a0a116b17fa79c6fd6)

**Reverted changes to RPC response to keep backwards compatibility**
Example error response from failed `require` statement :
```
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": "0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000b56616c75652066616c7365000000000000000000000000000000000000000000"
}
```

### DIFF
--- a/core/state_transition.go
+++ b/core/state_transition.go
@@ -163,7 +163,7 @@ func NewStateTransition(evm *vm.EVM, msg Message, gp *GasPool, bc ChainContext) 
 // the gas used (which includes gas refunds) and an error if it failed. An error always
 // indicates a core error meaning that the message would always fail for that particular
 // state and would never be accepted within a block.
-func ApplyMessage(evm *vm.EVM, msg Message, gp *GasPool) (*ExecutionResult, error) {
+func ApplyMessage(evm *vm.EVM, msg Message, gp *GasPool) (ExecutionResult, error) {
 	return NewStateTransition(evm, msg, gp, nil).TransitionDb()
 }
 
@@ -224,9 +224,9 @@ func (st *StateTransition) preCheck() error {
 // TransitionDb will transition the state by applying the current message and
 // returning the result including the used gas. It returns an error if failed.
 // An error indicates a consensus issue.
-func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
+func (st *StateTransition) TransitionDb() (ExecutionResult, error) {
 	if err := st.preCheck(); err != nil {
-		return nil, err
+		return ExecutionResult{}, err
 	}
 	msg := st.msg
 	sender := vm.AccountRef(msg.From())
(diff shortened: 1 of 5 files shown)


JackyWYX (changes requested):
Looks good to me.

JackyWYX (approved):

LeoHChen:
will it cause any compatibility issue during upgrade process?

janet-harmony:
> will it cause any compatibility issue during upgrade process?

No, shouldn't. The changes are only internal to each node & does not affect messages in the network.

core/state_transition.go
```diff
@@ -218,9 +224,9 @@ func (st *StateTransition) preCheck() error {
 // TransitionDb will transition the state by applying the current message and
 // returning the result including the used gas. It returns an error if failed.
 // An error indicates a consensus issue.
-func (st *StateTransition) TransitionDb() (ret []byte, usedGas uint64, failed bool, err error) {
-	if err = st.preCheck(); err != nil {
-		return
+func (st *StateTransition) TransitionDb() (*ExecutionResult, error) {
```
rlan35, line 227: can you make the return the struct, instead of pointer? To loose the client's burden of checking nil and avoid potential crash.

rlan35:
> Part 1 of [EVM Error changes](https://gist.github.com/janet-harmony/980f6cacc3e398a0a116b17fa79c6fd6)
> 
> **Reverted changes to RPC response to keep backwards compatibility**
> Example error response from failed `require` statement :
> 
> ```
> {
>   "jsonrpc": "2.0",
>   "id": 1,
>   "result": "0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000b56616c75652066616c7365000000000000000000000000000000000000000000"
> }
> ```

Did you check the the vmErr contains the require() statement string?   I remember I always got something like "transaction reverted" for the vmErr without detailed data in `utils.Logger().Info().Err(vmErr).Msg("VM returned with error")`

janet-harmony:
The VMErr field is `transaction reverted`, but the return data from the EVM is the error message in the `require(false, "error")` statement.
So the current implementation of `hmy_call` already supports returning error messages.

rlan35:
> The VMErr field is `transaction reverted`, but the return data from the EVM is the error message in the `require(false, "error")` statement.
> So the current implementation of `hmy_call` already supports returning error messages.
> 
> Changed to a struct, but the consumer should always check the error anyways to be safe, which is the only place the ExecutionResult returns as nil.

If we can, we should choose to systematically avoid potential mistakes which could cause crashes. Sounds good on the return data.

rlan35 (approved):

PR #3378, opened 2020-10-01
[core] node stuck fix

JackyWYX:
## Issue

Previously, it could happen that the node can fall out of sync when leveldb get an error `too many open files`. Sometimes the error will cause node stuck because the non-atomic operation in `WriteBlockWithState`. This PR aims to fix this error. 

1. Modify `WriteBlockWithState` logic to make db write operations atomic.
2. Add error handling to all rawdb related operations, which allows node to continue running if `too many open files` error happens.
3. Add fdlimit module to raise nofile limit for the process, which is effective for macos.

## Test

Tested locally on both MacOs and Linux machine. After running for about 1 days, there is no more too many open files error.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/harmony-one/bls/ffi/go/bls"
 	"github.com/harmony-one/harmony/api/service/syncing"
+	"github.com/harmony-one/harmony/common/fdlimit"
 	"github.com/harmony-one/harmony/consensus"
 	"github.com/harmony-one/harmony/consensus/quorum"
 	"github.com/harmony-one/harmony/core"
@@ -109,10 +110,13 @@ func runHarmonyNode(cmd *cobra.Command, args []string) {
 		os.Exit(0)
 	}
 
-	prepareRootCmd(cmd)
+	if err := prepareRootCmd(cmd); err != nil {
+		fmt.Fprint(os.Stderr, err)
+		os.Exit(128)
+	}
 	cfg, err := getHarmonyConfig(cmd)
 	if err != nil {
-		fmt.Println(err)
+		fmt.Fprint(os.Stderr, err)
 		cmd.Help()
 		os.Exit(128)
 	}
@@ -122,7 +126,7 @@ func runHarmonyNode(cmd *cobra.Command, args []string) {
 	setupNodeAndRun(cfg)
 }
 
-func prepareRootCmd(cmd *cobra.Command) {
+func prepareRootCmd(cmd *cobra.Command) error {
 	// HACK Force usage of go implementation rather than the C based one. Do the right way, see the
 	// notes one line 66,67 of https://golang.org/src/net/net.go that say can make the decision at
 	// build time.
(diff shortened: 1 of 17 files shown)


cmd/harmony/main.go
```diff
@@ -131,6 +135,20 @@ func prepareRootCmd(cmd *cobra.Command) {
 	runtime.GOMAXPROCS(runtime.NumCPU())
 	// Set up randomization seed.
 	rand.Seed(int64(time.Now().Nanosecond()))
+	// Raise fd limits
+	return raiseFdLimits()
+}
+
+func raiseFdLimits() error {
```
rlan35, line 142: does this require the user to run ./harmony as root?

rlan35: how is .Maximum() determined?

JackyWYX: No, not necessarily.

JackyWYX: By system configuration cap and a hard cap defined in the module. This piece of code is copied from ethereum code.

core/blockchain.go
```diff
@@ -1048,11 +1088,9 @@ func (bc *BlockChain) InsertReceiptChain(blockChain types.Blocks, receiptChain [
 			batch.Reset()
 		}
 	}
-	if batch.ValueSize() > 0 {
```
rlan35, line 1051: why remove the > 0 condition?

JackyWYX: This is to make `InsertReceiptChain` atomic. Else there will be some unwritten data at the end of `InsertReceiptChain` (though the function is not used in production code now)

JackyWYX: Ummm, there is no strong reason to do this change. Let me switch it back.

core/blockchain.go
```diff
@@ -2944,3 +3010,14 @@ func (bc *BlockChain) SuperCommitteeForNextEpoch(
 	}
 	return nextCommittee, err
 }
+
+var (
+	leveldbErrSpec         = "leveldb"
+	tooManyOpenFilesErrStr = "Too many open files"
+)
+
+func isUnrecoverableErr(err error) bool {
```
rlan35, line 3019: how do you define unrecoverable errors? Please add a comment. The conditions is not clear since the reader don't know the scope of all the errors and their full messages.

core/headerchain.go
```diff
@@ -493,7 +511,7 @@ func (hc *HeaderChain) SetHead(head uint64, delFn DeleteCallback) {
 	}
 	hc.currentHeaderHash = hc.CurrentHeader().Hash()
 
-	rawdb.WriteHeadHeaderHash(hc.chainDb, hc.currentHeaderHash)
```
rlan35, line 496: why move this above? the hc.currentHeaderHash was updated before the write.

JackyWYX: Since the cache field shall be updated only after the successful db write. Else this is not an atomic operation, and will cause data discrepancy in memory cache and storage.

rlan35: I see. cool.

core/blockchain.go
JackyWYX: Added more comments. Please tell me whether the comment gives more context about this error. Thanks.

rlan35: thanks. looks good.

rlan35 (approved):

PR #3379, opened 2020-10-02
View Change Revamp 1

LeoHChen:
This PR addressed review comments from previous one. https://github.com/harmony-one/harmony/pull/3373


consensus/config.go
```diff
@@ -4,14 +4,15 @@ import "time"
...
+	// default timeout configuration is shorten to 30 seconds as the consensus is 5s
+	viewChangeTimeout = 30
+	// The duration of viewChangeTimeout for each view change
+	viewChangeDuration time.Duration = viewChangeTimeout * time.Second
 
 	// timeout duration for announce/prepare/commit
-	phaseDuration     time.Duration = 60 * time.Second
-	bootstrapDuration time.Duration = 600 * time.Second
+	// shorten the duration from 60 to 30 seconds as the consensus is 5s
+	phaseDuration     time.Duration = 30 * time.Second
```
rlan35, line 14: following on the last PR's comment. How about 45 seconds? We don't want to jump into view change prematurely if retry can fix the transient network problems (we expect most of the problems are transient network issues eventually when we made our protocol code stable)

consensus/quorum/quorum.go
```diff
@@ -187,14 +187,15 @@ func (s *cIdentities) IndexOf(pubKey bls.SerializedPublicKey) int {
 	return -1
 }
 
-func (s *cIdentities) NextAfter(pubKey *bls.PublicKeyWrapper) (bool, *bls.PublicKeyWrapper) {
+// NextAfter return the next pubkey, next can be negative number
+func (s *cIdentities) NextAfter(pubKey *bls.PublicKeyWrapper, next int) (bool, *bls.PublicKeyWrapper) {
```
rlan35, line 191: should rename this func now the logic is different. It's not "next" anymore. It's NthAfter

consensus/view_change.go
```diff
@@ -183,21 +151,26 @@ func (consensus *Consensus) startViewChange(viewID uint64) {
...
 		)
 	}
-
-	consensus.consensusTimeout[timeoutViewChange].SetDuration(duration)
-	consensus.consensusTimeout[timeoutViewChange].Start()
 }
 
+// onViewChange is called when the view change message is received.
 func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
+	consensus.getLogger().Debug().Msg("[onViewChange] Received ViewChange Message")
```
rlan35, line 158: Let's make all the view change related log Info()? view change doesn't happen all the time, but when it happens, it's critical have the log for troubleshooting.

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+	myPubKey string,
+	privKeys multibls.PrivateKeys,
+) error {
+	vc.vcLock.Lock()
+	defer vc.vcLock.Unlock()
+
+	// add my own M2 type message signature and bitmap
+	_, ok := vc.nilSigs[viewID][myPubKey]
+
+	if !ok {
```
rlan35, line 335: The logic should be as the original condition `if !(ok1 || ok2) {`  . meaning: when I never added either m1 or m2 myself, I proceed to check whether I have prepared block myself, if yes, I add to m1, otherwise, I add to m2. Honest node can only commit to one choice, either m1 or m2. They are mutual exclusive. Let's simply keep the original logic (copy and paste is better to avoid mistakes)

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+		if err := vc.blockVerifier(preparedBlock); err != nil {
+			return nil, errors.New("[VerifyNewViewMsg] Prepared block verification failed")
+		}
+		return preparedBlock, nil
+	}
+	return nil, nil
+}
+
+// VerifyViewChangeMsg returns the type of the view change message after verification
+func (vc *viewChange) VerifyViewChangeMsg(recvMsg *FBFTMessage, members multibls.PublicKeys) (ViewChangeMsgType, *types.Block, error) {
```
rlan35, line 242: This function is more like "ProcessViewChangeMsg" then just verify, because it's modifying the state (not a pure function)

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+			Str("validatorPubKey", senderKeyStr).
+			Msg("[VerifyViewChangeMsg] Add M1 (prepared) type message")
+		vc.bhpSigs[recvMsg.ViewID][senderKeyStr] = recvMsg.ViewchangeSig
+		vc.bhpBitmap[recvMsg.ViewID].SetKey(senderKey.Bytes, true) // Set the bitmap indicating that this validator signed.
+
+		vc.getLogger().Info().Uint64("viewID", recvMsg.ViewID).
+			Str("validatorPubKey", senderKeyStr).
+			Msg("[VerifyViewChangeMsg] Add M3 (ViewID) type message")
+
+		vc.viewIDSigs[recvMsg.ViewID][senderKeyStr] = recvMsg.ViewidSig
```
rlan35, line 289: Missing the viewID sig checking.

rlan35: Also better to keep the old logic flow where 
if m1 {
 do m1 thing
} else {
 do m2 thing
}

// at last
do m3 (viewID) thing

consensus/view_change.go
```diff
@@ -214,228 +188,73 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-					Hex("blockHash", blockHash).
-					Msg("[onViewChange] failed to verify multi signature for m1 prepared payload")
-				return
+	if msgType == M1 {
+		preparedMsgs := consensus.FBFTLog.GetMessagesByTypeSeq(
+			msg_pb.MessageType_PREPARED, recvMsg.BlockNum,
+		)
+		// search for local prepared msg and add one if not found
+		preparedMsg := consensus.FBFTLog.FindMessageByViewID(preparedMsgs, recvMsg.ViewID)
+		if preparedMsg == nil {
```
rlan35, line 221: This should be the original condition to add new prepared block:
`if len(consensus.m1Payload) == 0 || !bytes.Equal(consensus.m1Payload, recvMsg.Payload) {`

new logic is different now.

consensus/view_change.go
```diff
@@ -214,228 +188,73 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-				return
-			}
-			blockHash := recvMsg.Payload[:32]
-			aggSig, mask, err := consensus.ReadSignatureBitmapPayload(recvMsg.Payload, 32)
-			if err != nil {
-				consensus.getLogger().Error().Err(err).Msg("[onViewChange] M1 RecvMsg Payload Read Error")
-				return
-			}
-
-			if !consensus.Decider.IsQuorumAchievedByMask(mask) {
```
rlan35, line 351: Seems the quorum check on the prepared msg is lost in the new code. Please double check.

consensus/view_change.go
rlan35: Please double check the logic about adding a m1/prepared block to local fbft log.

consensus/view_change.go
```diff
@@ -468,22 +287,18 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-					consensus.getLogger().Warn().Msg("submit vote on viewchange commit failed")
+					consensus.getLogger().Warn().Err(err).Msg("submit vote on viewchange commit failed")
 					return
 				}
 
 			}
 		}
 
-		consensus.SetViewChangingID(recvMsg.ViewID)
+		consensus.SetViewIDs(recvMsg.ViewID)
```
rlan35, line 297: Should not update block view ID, because no new block is confirmed. Only view change reached quorum, doesn't mean the rest will succeed and a new consensus can be reached yet.

rlan35, line 297: can you add a todo here? // TODO: reconsider whether we should update CurBlockViewID yet because there is no new block confirmed yet and problem could still happen before that. It could cause problem if some nodes update their curBlockViewID locally while others don't, since view change now are based on time and CurBlockViewID.

consensus/view_change.go
```diff
@@ -494,105 +309,65 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
 		consensus.consensusTimeout[timeoutConsensus].Start()
-		consensus.getLogger().Info().Str("myKey", newLeaderKey.Bytes.Hex()).Msg("[onViewChange] I am the New Leader")
+		consensus.getLogger().Info().
+			Str("myKey", newLeaderKey.Bytes.Hex()).
+			Msg("[onViewChange] I am the New Leader")
 	}
 }
 
-// TODO: move to consensus_leader.go later
+// onViewChange is called when validators received newView message from the new leader
```
rlan35, line 326: onNewView is called.

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+	senderKeyStr := senderKey.Bytes.Hex()
+
+	// first time received the new view message
+	_, ok := vc.newViewMsg[recvMsg.ViewID]
+	if !ok {
+		newViewMap := map[string]uint64{}
+		vc.newViewMsg[recvMsg.ViewID] = newViewMap
+	}
+
+	_, ok = vc.newViewMsg[recvMsg.ViewID][senderKeyStr]
```
rlan35, line 196: Don't think there is a need to add this newViewMsg map and check whether new view is already received. As long as the newView message is verified and valid, we should always response to it. Cases, like the leader failed to deliver newView message to everyone and need retry logic, in this case, better let everyone to always response to the same newView message until consensus can be reached.

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+	// bhpSigs: blockHashPreparedSigs is the signature on m1 type message
+	bhpSigs map[uint64]map[string]*bls_core.Sign
+	// nilSigs: there is no prepared message when view change,
+	// it's signature on m2 type (i.e. nil) messages
+	nilSigs      map[uint64]map[string]*bls_core.Sign
+	viewIDSigs   map[uint64]map[string]*bls_core.Sign
+	bhpBitmap    map[uint64]*bls_cosi.Mask
+	nilBitmap    map[uint64]*bls_cosi.Mask
+	viewIDBitmap map[uint64]*bls_cosi.Mask
+	viewIDCount  map[uint64]map[string]int
```
rlan35, line 54: viewIDCount remove if not used.

consensus/view_change_construct.go
```diff
@@ -0,0 +1,422 @@
...
+					vc.m1Payload = append(preparedMsg.BlockHash[:], preparedMsg.Payload...)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// isM1PayloadEmpty returns true if m1Payload is not set
+func (vc *viewChange) isM1PayloadEmpty() bool {
```
rlan35, line 393: why have two version of this, one private and one public? I see the private one are called somewhere else which could still cause race. Might just have one IsM1PayloadEmpty()

consensus/config.go
LeoHChen: I can adjust it to 45 seconds.  One other thing is the retry frequency, right now, we do retry every 10 seconds if consensus is not reached.  I think we shall shorten the time of retry as well since the consensus is shortened to 5 seconds.   I'll change this to 45 seconds, and retry interval to 7 seconds.

consensus/quorum/quorum.go
LeoHChen: sure.

consensus/view_change.go
LeoHChen: make senses.

consensus/view_change_construct.go
LeoHChen: this is a change requested by Jacky.  The private version is not locked as it was called in a locked context.  The public version is locked.

consensus/view_change_construct.go
LeoHChen: not needed, will remove.

consensus/view_change_construct.go
LeoHChen: this situation won't happen as the validator received a valid newview message, it will enter normal mode, not in viewchange anymore. So, it won't process the newview message. Otherwise, it's a security risk.  

The original function locked the entire function, so there is not reentry of the processing logic.  Now, since I removed the monolithic lock, I simply don't want to process the same message again, just like how we handle viewchange messages.

consensus/view_change.go
LeoHChen: sure, will change.

consensus/view_change.go
LeoHChen: This makes sense, actually, the curblockviewID should be updated along with the commit of each block.  Let me add a TODO and dig deeper.

consensus/view_change_construct.go
LeoHChen: will rename

consensus/view_change_construct.go
LeoHChen: okay,

consensus/view_change_construct.go
LeoHChen: m3 is already processed at end end, so the new code processed m1 and return, if not m1, then do m2 and return.  The same logic. ViewID sign checking is moved to consensus/check.go, onViewChangeSanityCheck function.

consensus/view_change.go
LeoHChen: I've commented on the old PR.  The original code, the m1payload is always set to payload found on the local prepared message at first. And it only sets to payload from recvMsg when it is empty which means it can't find the local prepared message.  So the new logic is the same, only add a local block when it can't find the prepared message from the local fbft log.

The validator should always trust the local message than the received message. that's my understanding.  We can further discuss it.

consensus/view_change.go
LeoHChen: Quorum check is in line 241 of the new file. Always use m3 to check quorum, as we always add m3 sig and bitmap.  Does it make sense?  The original code checked quorum on m1payload, which is not necessary, IMO.

consensus/view_change_construct.go
rlan35: I think we can err on the safer side and respond to the same message as long as it's valid (meaning I am still in view change, and block number and all other checks pass). Of course, we don't need to respond to a newView message if we already enter normal mode. But before we successfully finish a view change, better respond to every messages for better robustness. Otherwise, think about cases, where half of the nodes responds but others don't. Then if the leader is restarted, it will lose the collected votes from the first half and when it retries, those first half won't respond anymore and this view change won't succeed anymore on retries.

consensus/view_change.go
rlan35: It's not about the quorum on m1 view change message. This is about the quorum on the prepared signatures in m1. We definitely need to check the prepared signature reached quorum on the block before trusting it.

consensus/view_change.go
rlan35: ok, got it. thanks.

consensus/checks.go
```diff
@@ -174,18 +176,28 @@ func (consensus *Consensus) onViewChangeSanityCheck(recvMsg *FBFTMessage) bool {
...
+		consensus.GetCurBlockViewID() > recvMsg.ViewID {
 		consensus.getLogger().Warn().
-			Msg("[onViewChange] ViewChanging ID Is Low")
+			Msg("[onViewChangeSanityCheck] ViewChanging ID Is Low")
 		return false
 	}
 	if recvMsg.ViewID-consensus.GetViewChangingID() > MaxViewIDDiff {
 		consensus.getLogger().Debug().
-			Msg("Received viewID that is MaxViewIDDiff (100) further from the current viewID!")
+			Msg("[onViewChangeSanityCheck] Received viewID that is MaxViewIDDiff (100) further from the current viewID!")
```
rlan35, line 190: (249)

consensus/fbft_log.go
```diff
@@ -239,6 +244,19 @@ func (log *FBFTLog) FindMessageByMaxViewID(msgs []*FBFTMessage) *FBFTMessage {
 	return msgs[maxIdx]
 }
 
+// FindMessageByViewID returns the mesasge with matching view ID
+func (log *FBFTLog) FindMessageByViewID(msgs []*FBFTMessage, viewID uint64) *FBFTMessage {
```
rlan35, line 248: if not used, remove please.

rlan35 (approved):

consensus/view_change.go
```diff
@@ -214,228 +200,47 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
 		} else {
-			consensus.getLogger().Info().
-				Str("From", consensus.phase.String()).
-				Str("To", FBFTCommit.String()).
-				Msg("[OnViewChange] Switching phase")
-			consensus.switchPhase(FBFTCommit, true)
-			copy(consensus.blockHash[:], consensus.m1Payload[:32])
-			aggSig, mask, err := consensus.ReadSignatureBitmapPayload(consensus.m1Payload, 32)
+			consensus.switchPhase("onViewChange", FBFTCommit)
+			payload := consensus.vc.GetM1Payload()
```
rlan35, line 241: this has potential race condition.

consensus/view_change.go
```diff
@@ -214,228 +200,47 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-		Int("have", len(consensus.viewIDSigs[recvMsg.ViewID])).
-		Int64("total", consensus.Decider.ParticipantsCount()).
-		Msg("[onViewChange]")
 
 	// received enough view change messages, change state to normal consensus
-	if consensus.Decider.IsQuorumAchievedByMask(consensus.viewIDBitmap[recvMsg.ViewID]) {
+	if consensus.Decider.IsQuorumAchievedByMask(consensus.vc.GetViewIDBitmap(recvMsg.ViewID)) {
+		consensus.getLogger().Info().Msg("[onViewChange] View Change Message Quorum Reached")
 		consensus.current.SetMode(Normal)
 		consensus.LeaderPubKey = newLeaderKey
```
rlan35, line 230: These state changes have high risk for race conditions. As when multiple viewChangeMsg are received that all makes IsQuorumAchievedByMask() == true, then these states will be updated concurrently. And this is very likely to happen since those messages can arrive very closely.

rlan35: I think it's better to keep the old lock logic to avoid troubles. The view change protocol is not performance concerned after all, stability of the nodes is much more important for view change.

LeoHChen: why is there an issue if multiple view change messages arrived if they are all valid? we can check the mode during the quorum check as well.

consensus/view_change.go
LeoHChen: we can lock the entire quorum reach block though.  what's the potential race condition here?

PR #3380, opened 2020-10-06
Rosetta Implementation - pt4 (Stage 3.4 of Node API Overhaul)

Daniel-VDM:
# Stage 3.4 of [Node API Overhaul](https://github.com/harmony-one/harmony/issues/3210)

TLDR: This PR implements Rosetta's [Construction API](https://www.rosetta-api.org/docs/ConstructionApi.html) as well as correct some small bugs in the Node API found during the integration test of the Construction API. The integration test was done using the [rosetta-cli](https://www.rosetta-api.org/docs/rosetta_cli.html) using [this](https://github.com/Daniel-VDM/rosetta-cli/blob/f8a6b805540334d490f6094018d239afc2232ae6/examples/configuration/harmony.json) configuration. Lastly, some refactoring of the block transaction formatter was done to make it more general so that the Construction API could consume said formatter. 

## Constriction API Details

For reference, this is the flow of the Construction API:
```
                               Caller (i.e. Coinbase)                + Construction API Implementation
                              +---------------------------------------------------------------------------------+
                                                                     |
                               Derive Address   +----------------------------> /construction/derive
                               from Public Key                       |
                                                                     |
                             X                                       |
                             X Create Metadata Request +---------------------> /construction/preprocess
                             X (array of operations)                 |                    +
    Get metadata needed      X                                       |                    |
    to construct transaction X            +-----------------------------------------------+
                             X            v                          |
                             X Fetch Online Metadata +-----------------------> /construction/metadata (online)
                             X                                       
                                                                     |
                             X                                       |
                             X Construct Payloads to Sign +------------------> /construction/payloads
                             X (array of operations)                 |                   +
                             X                                       |                   |
 Create unsigned transaction X          +------------------------------------------------+
                             X          v                            |
                             X Parse Unsigned Transaction +------------------> /construction/parse
                             X to Confirm Correctness                |
                             X                                       |
                                                                     |
                             X                                       |
                             X Sign Payload(s) +-----------------------------> /construction/combine
                             X (using caller's own detached signer)  |                 +
                             X                                       |                 |
   Create signed transaction X         +-----------------------------------------------+
                             X         v                             |
                             X Parse Signed Transaction +--------------------> /construction/parse
                             X to Confirm Correctness                |
                             X                                       |
                                                                     |
                             X                                       |
                             X Get hash of signed transaction +--------------> /construction/hash
Broadcast Signed Transaction X to monitor status                     |
                             X                                       |
                             X Submit Transaction +--------------------------> /construction/submit (online)
                             X                                       |
                                                                     +
``` 

**All endpoints do NOT have access to the current state of the chain/pool unless specified that it is 'online'**.

### [`/construction/derive`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionderive) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction.go#L46-L64)

An [Address Identifier](https://www.rosetta-api.org/docs/models/AccountIdentifier.html) can be derived from a *compressed* secp256k1 public key. The key's bytes given to this endpoint are assumed to be said compressed key.

### [`/construction/preprocess`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionpreprocess) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_check.go#L47-L93)

One can determine if a set of operations creates a valid Harmony Transaction by using this endpoint. The underlying function(s) to check the operation(s) can be found [here](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/operation_components.go). Once the operations are validated, all metadata options (and curred data) are packaged into [this](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_check.go#L19-L23) struct and returned to the client to be forwarded (unchanged) on to the `/construction/metadata` endpoint. 

### [`/construction/metadata`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionmetadata) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_check.go#L123-L202)

This endpoint is considered to be 'online'. Given the metadata option struct from `/construction/preprocess`, it fetches the correct nonce & gas fee estimation and returns it in [this](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_check.go#L96-L101) struct to the client to be forwarded (unchanged) on to the `/construction/payloads` endpoint.

### [`/construction/payloads`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionpayloads) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L76-L150)

This endpoint actually constructs the Harmony transaction given the [metadata](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_check.go#L96-L101) from `/construction/metadata` and [OperationComponents](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/operation_components.go#L19-L25) from the given operations. The returned payload to be signed by the client is the 'signer' (EIP115 Signer) hash of a Harmony transaction (i.e: [this](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/core/types/transaction_signing.go#L58)). Note that the returned harmony transaction object is wrapped in [this](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L25-L30) struct.

### [`/construction/parse`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionparse) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_parse.go#L14-L28)

One can check that an unsigned or signed ([wrapped](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L25-L30)) transaction results in the intended rosetta operations using this endpoint. Note that parsing uses the same formatting functions as in the `/block` endpoint of the Node API. **Logic for unpacking and checking the wrapped transaction can be found [here](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L33-L73)**.

### [`/construction/combine`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructioncombine) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L174-L267)

One can get a signed ([wrapped](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L25-L30)) transaction object by giving the signed payload to this endpoint (along with the unsigned wrapped transaction).

### [`/construction/hash`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionhash) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_submit.go#L15-L33)

One can get the hash of a *signed* ([wrapped](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L25-L30)) transaction using this endpoint.

### [`/construction/submit`](https://www.rosetta-api.org/docs/ConstructionApi.html#constructionsubmit) | [Implementation](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_submit.go#L36-L91)
This endpoint is considered to be 'online'. One can submit a *signed* ([wrapped](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/construction_create.go#L25-L30)) transaction to the pool using this endpoint.

## Misc Details

The transaction formatting needed to be changed to not error on unsigned transactions. Hence the introduction of the [DefaultSenderAddress](https://github.com/Daniel-VDM/harmony/blob/5bb3d5f6dcd158556cc591c28494d9606009a088/rosetta/services/block.go#L793).

Moreover, all error messages now also return the line number, file, and node version to help with debugging.

## Testing

Lots of effort was put into writing unit tests. Moreover, test scenarios were written for the rosetta-cli to test the construction API. One can run these scenarios by downloading and installing the [rosetta-cli](https://github.com/Daniel-VDM/rosetta-cli/tree/testing), and running the tests with the following command once **localnet** is started:
```
./rosetta-cli check:construction --configuration-file ./examples/configuration/harmony.json
```
> The scenario file is [here](https://github.com/Daniel-VDM/rosetta-cli/blob/testing/examples/configuration/harmony.json)

Here is a screenshot of the successful tests:

image

Note that this also checks a large part of the Node API as it uses it to ensure that transactions are finalized.

## TODO

Once merged, all functionalities of the constriction endpoint will be tested on testnet.


rosetta/services/network.go
```diff
@@ -55,7 +57,16 @@ func (s *NetworkAPI) NetworkStatus(
 	}
 
 	// Fetch relevant headers, syncing status, & peers
-	currentHeader, err := s.hmy.HeaderByNumber(ctx, rpc.LatestBlockNumber)
+	currBlock := s.hmy.CurrentBlock()
+	var currentHeader *block.Header
+	var err error
+	if currBlock.Number().Cmp(big.NewInt(0)) == 1 && !s.hmy.IsStakingEpoch(currBlock.Epoch()) {
```
LeoHChen, line 63: block 0? can you add a bit comment here?

rosetta/services/operation_components.go
```diff
@@ -0,0 +1,232 @@
...
+			"message": "nil operations",
+		})
+	}
+	if len(operations) > maxNumOfConstructionOps || len(operations) == 0 {
+		return nil, common.NewError(common.InvalidTransactionConstructionError, map[string]interface{}{
+			"message": fmt.Sprintf("invalid number of operations, must <= %v & > 0", maxNumOfConstructionOps),
+		})
+	}
+
+	if len(operations) == 2 {
```
LeoHChen, line 53: == maxNumOfConstructionOps ?

rosetta/services/operation_components.go
```diff
@@ -0,0 +1,232 @@
...
+		return getContractCreationOperationComponents(operations[0])
+	default:
+		return nil, common.NewError(common.InvalidTransactionConstructionError, map[string]interface{}{
+			"message": fmt.Sprintf("%v is unsupported or invalid operation type", operations[0].Type),
+		})
+	}
+}
+
+// getTransferOperationComponents ..
+func getTransferOperationComponents(
```
LeoHChen, line 69: add a comment here that this function assume operations has length 2

rosetta/services/transaction_construction.go
```diff
@@ -0,0 +1,143 @@
...
+	if metadata.Transaction.FromShardID != nil && *metadata.Transaction.FromShardID != sourceShardID {
+		return nil, common.NewError(common.InvalidTransactionConstructionError, map[string]interface{}{
+			"message": fmt.Sprintf("intended source shard %v != tx metadata source shard %v",
+				sourceShardID, *metadata.Transaction.FromShardID),
+		})
+	}
+
+	var tx hmyTypes.PoolTransaction
+	switch components.Type {
+	case common.CrossShardTransferOperation:
```
LeoHChen, line 33: why does crossshardtransfer different from plaintransfer?

LeoHChen (commented):
did a quick review, in general looks good to me with lots of good unit tests.

rosetta/services/block.go
```diff
@@ -765,51 +788,59 @@ func formatUndelegationPayoutTransaction(
 
 }
 
+var (
+	// DefaultSenderAddress ..
+	DefaultSenderAddress = ethcommon.HexToAddress("0xEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEE")
```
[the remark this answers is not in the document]
Daniel-VDM: Its a valid address and I see it often being used in contract code.

The b32 addr for this is: one1amhwamhwamhwamhwamhwamhwamhwamhwzdcv76

rosetta/services/block.go
```diff
@@ -1240,6 +1269,15 @@ func findLogsWithTopic(
 	return logs
 }
 
+// formatNegativeValue ..
+func formatNegativeValue(num *big.Int) string {
```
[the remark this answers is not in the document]
Daniel-VDM: This function is specific to rosetta. In some cases (all be it rare), a -0 may be valid. Also, blocks.go will be split into separate files (kinda like how construction is) in a separate PR so the organization will make more sense then. This PR is already big as is.

rosetta/services/construction.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/harmony-one/harmony/common/denominations"
+	hmyTypes "github.com/harmony-one/harmony/core/types"
+	"github.com/harmony-one/harmony/hmy"
+	"github.com/harmony-one/harmony/rosetta/common"
+	stakingTypes "github.com/harmony-one/harmony/staking/types"
+)
+
+const (
+	// DefaultGasPrice ..
+	DefaultGasPrice = denominations.Nano
```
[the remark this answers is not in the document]
Daniel-VDM: My thought here was that a consumer can import our rosetta package and have access to what the implementation uses as the default gas price. This is needed since a user can provide the `GasPriceMultiplier` when constructing a transaction, and this number is multiplied by the `DefaultGasPrice` for the final gas price of the transaction.

rosetta/services/construction.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+var (
+	// DefaultGasLimit ..
+	DefaultGasLimit = big.NewInt(24000)
```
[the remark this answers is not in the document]
Daniel-VDM: This can actually be replaced by `params.TxGas`

rosetta/services/construction_check.go
```diff
@@ -0,0 +1,233 @@
...
+				"message": errors.WithMessage(err, "invalid tx data format").Error(),
+			})
+		}
+	}
+
+	var estGasUsed uint64
+	if !isStakingOperation(options.OperationType) {
+		if options.OperationType == common.ContractCreationOperation {
+			estGasUsed, err = rpc.EstimateGas(ctx, s.hmy, rpc.CallArgs{Data: &data}, nil)
+			estGasUsed *= 2 // HACK to account for imperfect contract creation estimation
```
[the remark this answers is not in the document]
Daniel-VDM: Sure ill add

rosetta/services/construction_create.go
```diff
@@ -0,0 +1,267 @@
...
+	"github.com/pkg/errors"
+
+	hmyTypes "github.com/harmony-one/harmony/core/types"
+	"github.com/harmony-one/harmony/rosetta/common"
+	stakingTypes "github.com/harmony-one/harmony/staking/types"
+)
+
+const (
+	// SignedPayloadLength ..
+	SignedPayloadLength = 65
```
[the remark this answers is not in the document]
Daniel-VDM: it is the required length of the ECDSA payload, ill clarify in the comment.

rosetta/services/construction_create.go
```diff
@@ -0,0 +1,267 @@
...
+			"message": "require metadata",
+		})
+	}
+	metadata := &ConstructMetadata{}
+	if err := metadata.UnmarshalFromInterface(request.Metadata); err != nil {
+		return nil, common.NewError(common.InvalidTransactionConstructionError, map[string]interface{}{
+			"message": errors.WithMessage(err, "invalid metadata").Error(),
+		})
+	}
+	if request.PublicKeys == nil || len(request.PublicKeys) != 1 {
```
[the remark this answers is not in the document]
Daniel-VDM: right for got to change this

rosetta/services/construction_create.go
```diff
@@ -0,0 +1,267 @@
...
+	ctx context.Context, request *types.ConstructionCombineRequest,
+) (*types.ConstructionCombineResponse, *types.Error) {
+	if err := assertValidNetworkIdentifier(request.NetworkIdentifier, s.hmy.ShardID); err != nil {
+		return nil, err
+	}
+	wrappedTransaction, tx, rosettaError := unpackWrappedTransactionFromString(request.UnsignedTransaction)
+	if rosettaError != nil {
+		return nil, rosettaError
+	}
+	if request.Signatures == nil || len(request.Signatures) != 1 {
```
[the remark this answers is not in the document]
Daniel-VDM: forgot to check for this

rosetta/services/network.go
Daniel-VDM: The comment directly below it should be clear enough.
```
// all blocks in the era before staking epoch requires the next block to get the block reward transactions
```

The block 0 check is an edge case check so that we don't attempt to look for block -1.

rosetta/services/operation_components.go
Daniel-VDM: `maxNumberOfConstructionOps` can change in the future. I'll just make a const for it. This should also take care of the comment for getTransferOperationComponents

rosetta/services/operation_components.go
Daniel-VDM: ref above

rosetta/services/transaction_construction.go
Daniel-VDM: The assumed metadata provided is different. Cross shard has to provide from & to shard in the metadata whereas plain transactions do not. More rosetta implementation do not require metadata for transfers, thus the decision to do it this way.

rosetta/common/errors.go
```diff
@@ -63,18 +65,61 @@ var (
...
+	InvalidTransactionConstructionError = types.Error{
+		Code:      9,
+		Message:   "invalid transaction construction",
+		Retriable: false,
+	}
 )
 
+// chopPath returns the source filename after the last slash.
+// Inspired by https://github.com/jimlawless/whereami
+func chopPath(original string) string {
```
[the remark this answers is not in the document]
Daniel-VDM: moved into internal

rosetta/common/errors.go
```diff
@@ -63,18 +65,61 @@ var (
...
+	if i == -1 {
+		return original
+	}
+	return original[i+1:]
+}
+
+// traceMessage return a string containing the file name, function name
+// and the line number of a specified entry on the call stack.
+// Inspired by https://github.com/jimlawless/whereami
+func traceMessage(depthList ...int) string {
```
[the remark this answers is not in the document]
Daniel-VDM: moved into internal

rosetta/services/block.go
```diff
@@ -585,27 +586,33 @@ type blockSignerInfo struct {
 
 // TransactionMetadata ..
 type TransactionMetadata struct {
+	// CrossShardIdentifier is the transaction identifier on the from/source shard
 	CrossShardIdentifier *types.TransactionIdentifier `json:"cross_shard_transaction_identifier,omitempty"`
-	ToShardID            uint32                       `json:"to_shard,omitempty"`
-	FromShardID          uint32                       `json:"from_shard,omitempty"`
-	Data                 string                       `json:"data,omitempty"`
+	ToShardID            *uint32                      `json:"to_shard,omitempty"`
```
rlan35, line 591: why change to pointers?

rosetta/services/block.go
```diff
@@ -1104,19 +1112,25 @@ func newTransferOperations(
...
 ) ([]*types.Operation, *types.Error) {
-	// Sender address should have errored prior to call this function
-	senderAddress, _ := tx.SenderAddress()
-	receiverAddress := *tx.To()
 	senderAccountID, rosettaError := newAccountIdentifier(senderAddress)
 	if rosettaError != nil {
 		return nil, rosettaError
 	}
-	receiverAccountID, rosettaError := newAccountIdentifier(receiverAddress)
+	receiverAccountID, rosettaError := newAccountIdentifier(*tx.To())
```
rlan35, line 1121: any risk of nil pointer?

rosetta/services/operation_components.go
```diff
@@ -0,0 +1,235 @@
...
+			"message": "nil operation",
+		})
+	}
+	amount, err := types.AmountValue(operation.Amount)
+	if err != nil {
+		return nil, common.NewError(common.InvalidTransactionConstructionError, map[string]interface{}{
+			"message": err.Error(),
+		})
+	}
+	if amount.Sign() == 1 {
```
rlan35, line 213: for contract creation, you could still send token along to the contract address right? By making the contract constructor payable.

rlan35:
any risk that these new codes will cause issues for mainnet validators? Seems very separated.

rosetta/services/operation_components.go
Daniel-VDM: correct, but then the amount should be **negative** for the rosetta spec as it would deduct funds from the contract deployer.

rosetta/services/block.go
Daniel-VDM: Don't think so since cross-shard transactions need a destination address and some verification is done to the tx prior to calling this internal function. That said, better safe than sorry, so I'll review and add some checks for all uses of `*tx.To()`.

rosetta/services/block.go
Daniel-VDM: The shards are optional for the metadata and I don't want to unmarshal it to 0 if it's not present since that is a valid shard ID. 

https://play.golang.org/p/9yNLRRwdFw7

Daniel-VDM:
> any risk that these new codes will cause issues for mainnet validators? Seems very separated.

Rosetta is only enabled by default on the explorer node. Validators have to manually enable it, therefore there shouldn't be any issues.

LeoHChen (approved):
LGTM in general, thanks for the work.

rlan35 (approved):

PR #3385, opened 2020-10-13
Rosetta Dockerfile (Stage 4 of Node API Overhaul)

Daniel-VDM:
# Stage 4 of [Node API Overhaul](https://github.com/harmony-one/harmony/issues/3210)

This PR adds the Docker file and relevant README to deploy a rosetta enabled Harmony node. 

I had to make the pool's `transaction.rlp` file dependent on the `datadir` param to ensure that the path of that file can be determined by said param. This should not break any existing node deployment, as the default directory is the same as the `harmony_db_*` directory. In the worst case, a user would have to re-submit their transaction (if the `datadir` was used prior).

## TODO
* Add uploading to dockerhub to the release pipeline.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -904,6 +904,7 @@ func New(
 		node.BeaconBlockChannel = make(chan *types.Block)
 		txPoolConfig := core.DefaultTxPoolConfig
 		txPoolConfig.Blacklist = blacklist
+		txPoolConfig.Journal = fmt.Sprintf("%v/%v", node.NodeConfig.DBDir, txPoolConfig.Journal)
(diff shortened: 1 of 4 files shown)


LeoHChen (approved):
LGTM in general.  If we want to support wider usage of this docker file, we can include some rclone information. Also, user can run node in non-archival mode. Those info could be helpful.  

it would be better to also include some more info about how to check Harmony runtime log.

Daniel-VDM:
> LGTM in general. If we want to support wider usage of this docker file, we can include some rclone information. Also, user can run node in non-archival mode. Those info could be helpful.
> 
> it would be better to also include some more info about how to check Harmony runtime log.

Sure I've clarified & shown some examples use cases in the README for running in non-archival, viewing logs, and using a rcloned DB.

PR #3386, opened 2020-10-13
[viewchange] time-based synchronuous view change

LeoHChen:
For view change, the view change ID is the key to determine which node
will be next leader. In our original algorithm, the view ID always increased one step
at a time. The view change period increased exponetially based on the square of
the gap between current view changing ID and previous known block view change ID.

However, it is very slow to converge the view change if multiple nodes are offline and
view change can't be reached. Especially, during the shard down event, multiple nodes
are offline and other nodes have advanced their current view changing ID.

The new time-baed synchronuous view change algorithm uses the local timestamp and
the timestamp of the block to calculate the expected view changing ID. In this case,
offline nodes can immediately catch up with the latest view changing ID as long as
the offline nodes have the latest sync'ed block and relatively acturate local
clock. This algorithm will converge the view change faster in one or two view change
duration.

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -92,7 +92,7 @@ func (consensus *Consensus) onAnnounceSanityChecks(recvMsg *FBFTMessage) bool {
 					"[OnAnnounce] Already in ViewChanging mode, conflicing announce, doing noop",
 				)
 			} else {
-				consensus.startViewChange(consensus.GetCurBlockViewID() + 1)
+				consensus.startViewChange(consensus.GetCurBlockViewID())
 			}
 		}
 		consensus.getLogger().Debug().
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -92,7 +92,7 @@ func (consensus *Consensus) onAnnounceSanityChecks(recvMsg *FBFTMessage) bool {
 					"[OnAnnounce] Already in ViewChanging mode, conflicing announce, doing noop",
 				)
 			} else {
-				consensus.startViewChange(consensus.GetCurBlockViewID())
+				consensus.startViewChange()
 			}
 		}
 		consensus.getLogger().Debug().
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -81,7 +81,7 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 	consensus.Decider.UpdateParticipants(pubKeys)
 	utils.Logger().Info().Msg("My Committee updated")
 	for i := range pubKeys {
-		utils.Logger().Debug().
+		utils.Logger().Info().
 			Int("index", i).
 			Str("BLSPubKey", pubKeys[i].Bytes.Hex()).
 			Msg("Member")
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -95,7 +95,6 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 	} else {
 		utils.Logger().Error().
 			Msg("[UpdatePublicKeys] Participants is empty")
-		consensus.pubKeyLock.Unlock()
 	}
 	consensus.pubKeyLock.Unlock()
 	// reset states after update public keys
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -92,6 +92,10 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 		consensus.LeaderPubKey = &allKeys[0]
 		utils.Logger().Info().
 			Str("info", consensus.LeaderPubKey.Bytes.Hex()).Msg("My Leader")
+	} else {
+		utils.Logger().Error().
+			Msg("[UpdatePublicKeys] Participants is empty")
+		consensus.pubKeyLock.Unlock()
 	}
 	consensus.pubKeyLock.Unlock()
 	// reset states after update public keys
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -265,12 +265,12 @@ func (consensus *Consensus) Start(
 					if k != timeoutViewChange {
 						consensus.getLogger().Warn().Msg("[ConsensusMainLoop] Ops Consensus Timeout!!!")
 						viewID := consensus.GetCurBlockViewID()
-						consensus.startViewChange(viewID + 1)
+						consensus.startViewChange(viewID)
 						break
 					} else {
 						consensus.getLogger().Warn().Msg("[ConsensusMainLoop] Ops View Change Timeout!!!")
 						viewID := consensus.GetViewChangingID()
-						consensus.startViewChange(viewID + 1)
+						consensus.startViewChange(viewID)
 						break
 					}
 				}
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -264,13 +264,12 @@ func (consensus *Consensus) Start(
 					}
 					if k != timeoutViewChange {
 						consensus.getLogger().Warn().Msg("[ConsensusMainLoop] Ops Consensus Timeout!!!")
-						viewID := consensus.GetCurBlockViewID()
-						consensus.startViewChange(viewID)
+						consensus.SetViewChangingID(consensus.GetCurBlockViewID())
+						consensus.startViewChange()
 						break
 					} else {
 						consensus.getLogger().Warn().Msg("[ConsensusMainLoop] Ops View Change Timeout!!!")
-						viewID := consensus.GetViewChangingID()
-						consensus.startViewChange(viewID)
+						consensus.startViewChange()
 						break
 					}
 				}
--- a/consensus/quorum/quorum.go
+++ b/consensus/quorum/quorum.go
@@ -10,6 +10,7 @@ import (
 	bls_core "github.com/harmony-one/bls/ffi/go/bls"
 	"github.com/harmony-one/harmony/consensus/votepower"
 	bls_cosi "github.com/harmony-one/harmony/crypto/bls"
+	shardingconfig "github.com/harmony-one/harmony/internal/configs/sharding"
 	"github.com/harmony-one/harmony/multibls"
 	"github.com/harmony-one/harmony/numeric"
 	"github.com/harmony-one/harmony/shard"
@@ -73,6 +74,7 @@ type ParticipantTracker interface {
 	IndexOf(bls.SerializedPublicKey) int
 	ParticipantsCount() int64
 	NthNext(*bls.PublicKeyWrapper, int) (bool, *bls.PublicKeyWrapper)
+	NthNextHmy(shardingconfig.Instance, *bls.PublicKeyWrapper, int) (bool, *bls.PublicKeyWrapper)
 	UpdateParticipants(pubKeys []bls.PublicKeyWrapper)
 }
 
@@ -214,7 +216,29 @@ func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bl
 	if idx != -1 {
 		found = true
 	}
-	idx = (idx + next) % int(s.ParticipantsCount())
+	numNodes := int(s.ParticipantsCount())
+	// sanity check to avoid out of bound access
+	if numNodes <= 0 || numNodes > len(s.publicKeys) {
+		numNodes = len(s.publicKeys)
+	}
+	idx = (idx + next) % numNodes
+	return found, &s.publicKeys[idx]
+}
+
+// NthNextHmy return the Nth next pubkey of Harmony nodes, next can be negative number
+func (s *cIdentities) NthNextHmy(instance shardingconfig.Instance, pubKey *bls.PublicKeyWrapper, next int) (bool, *bls.PublicKeyWrapper) {
+	found := false
+
+	idx := s.IndexOf(pubKey.Bytes)
+	if idx != -1 {
+		found = true
+	}
+	numNodes := instance.NumHarmonyOperatedNodesPerShard()
+	// sanity check to avoid out of bound access
+	if numNodes <= 0 || numNodes > len(s.publicKeys) {
+		numNodes = len(s.publicKeys)
+	}
+	idx = (idx + next) % numNodes
 	return found, &s.publicKeys[idx]
 }
 
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -95,16 +95,15 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
 
 // fallbackNextViewID return the next view ID and duration when there is an exception
 // to calculate the time-based viewId
-func (consensus *Consensus) fallbackNextViewID(viewID uint64) (uint64, time.Duration) {
-	consensus.getLogger().Error().
-		Uint64("viewID", viewID).
-		Msg("[fallbackNextViewID] use legacy viewID algorithm")
-
-	if viewID < consensus.current.GetCurBlockViewID() {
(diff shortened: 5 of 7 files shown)


consensus/view_change.go
```diff
@@ -90,20 +90,111 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+// The next view ID is calculated based on the difference of validator's timestamp
+// and the block's timestamp. So that it can be deterministic to return the next view ID
+// only based on the blockchain block and the validator's current timestamp.
+// The next view ID is the single factor used to determine
+// the next leader, so it is mod the number of nodes per shard.
+// It returns the next viewID and duration of the view change
+// The view change duration is a fixed duration now to avoid stuck into offline nodes during
+// the view change.
+// viewID is only used as the fallback mechansim to determine the nextViewID
+func (consensus *Consensus) GetNextViewID(viewID uint64) (uint64, time.Duration) {
```
JackyWYX, line 117: Why these methods are public?

JackyWYX (commented):
Do you think we should check ntp clock when initializing harmony node? If the ntp time is off too much, return an error to prompt the user to fix the system clock?

LeoHChen:
> Do you think we should check ntp clock when initializing harmony node? If the ntp time is off too much, return an error to prompt the user to fix the system clock?

yes, will add it during node initialization.

consensus/view_change.go
LeoHChen: true, not necessary. will update.

LeoHChen:
> > Do you think we should check ntp clock when initializing harmony node? If the ntp time is off too much, return an error to prompt the user to fix the system clock?
> 
> yes, will add it during node initialization in a new PR though.

https://github.com/harmony-one/harmony/pull/3394

consensus/view_change.go
```diff
@@ -90,20 +90,111 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+	blockTimestamp := curHeader.Time().Int64()
+	curTimestamp := time.Now().Unix()
+
+	// timestamp messed up in current validator node
+	if curTimestamp < blockTimestamp {
+		return consensus.fallbackNextViewID(viewID)
+	}
+	totalNode := consensus.Decider.ParticipantsCount()
+	// diff is at least 1, and it won't exceeded the totalNode
+	diff := uint64(((curTimestamp - blockTimestamp) / viewChangeTimeout) % int64(totalNode))
```
rlan35, line 135: the blockHeader timestamp is in seconds, while the viewChangeTimeout is in nanoseconds. Need to be in the same unit.

consensus/view_change.go
```diff
@@ -90,20 +90,111 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+	if diff == 0 {
```
rlan35, line 136: the case of 0 is normal too. I can be that curtimestamp < blockTimestamp + viewChangeTimeout, or the view change goes round the whole list and comes back to the first leader again. So don't need to do diff = 1 here.

consensus/view_change.go
```diff
@@ -90,20 +90,111 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+		if curHeader == nil {
+			consensus.getLogger().Error().Msg("[getNextLeaderKey] Failed to get current header from blockchain")
+			lastLeaderPubKey = consensus.LeaderPubKey
+		} else {
+			// this is the truth of the leader based on blockchain blocks
+			lastLeaderPubKey, err = consensus.getLeaderPubKeyFromCoinbase(curHeader)
+			if err != nil || lastLeaderPubKey == nil {
+				consensus.getLogger().Error().Err(err).
+					Msg("[getNextLeaderKey] Unable to get leaderPubKey from coinbase. Set it to consensus.LeaderPubKey")
+				lastLeaderPubKey = consensus.LeaderPubKey
```
rlan35, line 178: consensus.LeaderPubKey should take priority than consensus.getLeaderPubKeyFromCoinbase(curHeader), because it's the real current leader. If consensus.LeaderPubKey != nil, then we should just use it. If it's nil, we can fall back on the leader of the last confirmed blocks (which could be the same as consensus.LeaderPubKey, but it could be not, ie. previous leader)

consensus/checks.go
```diff
@@ -90,7 +90,7 @@ func (consensus *Consensus) onAnnounceSanityChecks(recvMsg *FBFTMessage) bool {
 					"[OnAnnounce] Already in ViewChanging mode, conflicing announce, doing noop",
 				)
 			} else {
-				consensus.startViewChange(consensus.GetCurBlockViewID() + 1)
+				consensus.startViewChange(consensus.GetCurBlockViewID())
```
rlan35, line 93: We can remove the input param for startViewChange now. As the view change only depends on last block's viewID and the time difference. In the first 45 seconds after last block is confirmed, it's the current expected leader's job to propose. If it didn't do the job, then view change happens, it's the second 45 seconds and the next leader should be selected. 

So when the viewChange is triggered, based on the time difference and last block, we can already decide the new leaders. No need to refer to consensus.GetCurBlockViewID() anymore.

And with that, we can remove the ViewChangingID actually.

consensus/view_change.go
LeoHChen: make sense. will remove this additional check.

consensus/view_change.go
LeoHChen: why consensus.LeaderPubKey is higher priority than the leader from leader from block? I thought the leader from the block is the truth.  Also, I used the leader from the block is to make sure we can always get to the same next leader.

consensus/view_change.go
LeoHChen: let me double check.

consensus/checks.go
LeoHChen: I kept it for fallback.  Let me double check.

consensus/view_change.go
rlan35: because the leader could change automatically, in cases like epoch change, the leader is reset to the first one. Once we have resharding, the leader will change every N blocks automatically. So the last block's leader may not be the leader for the new block (current in consensus).

consensus/view_change.go
LeoHChen: viewChangeTimeout is a scalar 45 defined in consensus/config.go

both curTimestamp and blockTimestamp are in unit of second.  So no issue here.

consensus/view_change.go
LeoHChen: I see your point. However, the algorithm we used is to determine who is the next leader. It is an deterministic algorithm to use the known last leader to determine who is the next leader.  The leaderpubkey is also an in memory variable, which is not deterministic. So, we should still use the true last known leader to determine the next leader.

In case leader changed automatically, the leader's key will be written to blockchain after a block is confirmed.  Noted, automatically leader change should be independent of the view change.

consensus/view_change.go
rlan35: I see. cool

consensus/view_change.go
rlan35: The current GetNextLeaderKey() depends on the consensus.LeaderPubKey, which was determined and set by other logics such as the epoch change or view change. In this case, we have to rely on the in-memory data to determine the next leaders and consensus.LeaderPubKey should take priority.

Think about this case, the leader is latest confirmed block is A. A should produce next block, but it didn't. View change started, and after a few rounds, successfully done on leader D. Now, every nodes set their consensus.LeaderPubKey = D. But D could fail to produce the block too. And new view change needs to happen. Note here, there is no new block confirmed, so the latest confirmed block still have leader A. Then if view change start again, the whole process will happen again, and D could fail everything again.

So bottomline is, we have to rely on in-memory state in this case. And it's ok, as long as most of the nodes don't crash at the same time repeatedly

consensus/checks.go
LeoHChen: in the getNextViewID function, the viewID is still the fallback mechanism when the chain has not blocks, especially during bootstrap procedure when there is a view change needed.  So, I think we may still need to keep the viewID.  Otherwise, there is no fallback viewID then.

consensus/view_change.go
LeoHChen: If D is failed to produce block, a new view change will start, the view ID will still be based on A, but noted, the time has moved, so D won't be the leader anymore.  The next leader, assuming E, will be the new leader.

LeoHChen:
I've tested the PR on the STN, and the view change at the epoch change worked as expected, comparing to the previous version.

consensus/consensus_service.go
```diff
@@ -90,6 +90,11 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 		consensus.LeaderPubKey = &allKeys[0]
 		utils.Logger().Info().
 			Str("info", consensus.LeaderPubKey.Bytes.Hex()).Msg("My Leader")
+	} else {
+		utils.Logger().Error().
+			Msg("[UpdatePublicKeys] Participants is empty")
+		consensus.pubKeyLock.Unlock()
+		return 0
```
rlan35, line 97: why return early? If there is no participants. It doesn't mean it's an error. We don't prevent empty pubKeys anyway.

consensus/consensus_v2.go
```diff
@@ -368,12 +368,12 @@ func (consensus *Consensus) Start(
 					if k != timeoutViewChange {
 						consensus.getLogger().Warn().Msg("[ConsensusMainLoop] Ops Consensus Timeout!!!")
 						viewID := consensus.GetCurBlockViewID()
```
rlan35, line 370: Can directly use consensus.GetCurBlockViewID(). can you help remove viewID local var?

consensus/quorum/quorum.go
```diff
@@ -199,6 +201,18 @@ func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bl
...
 
+// NthNextHmy return the Nth next pubkey of Harmony nodes, next can be negative number
+func (s *cIdentities) NthNextHmy(instance shardingconfig.Instance, pubKey *bls.PublicKeyWrapper, next int) (bool, *bls.PublicKeyWrapper) {
+	found := false
+
+	idx := s.IndexOf(pubKey.Bytes)
+	if idx != -1 {
+		found = true
+	}
+	idx = (idx + next) % instance.NumHarmonyOperatedNodesPerShard()
```
rlan35, line 212: add a check to make sure idx is within index bound?

core/blockchain.go
```diff
@@ -293,6 +293,16 @@ func IsEpochBlock(block *types.Block) bool {
 	return shard.Schedule.IsLastBlock(block.NumberU64() - 1)
 }
 
+// IsEpochBlockByNumber returns whether this block is the first block of an epoch.
+// by checking the block number
+func IsEpochBlockByNumber(blockNum uint64) bool {
```
rlan35, line 298: Daniel add a utility for this purpose. IsLastBlockInEpoch(). that's the correct way to check. Because only beacon chain follow fixed epoch schedule. Other chains follow beacon chain's epoch change timing.

consensus/quorum/quorum.go
LeoHChen: it doesn't matter, right?  It mod on the number of harmony node.

consensus/consensus_v2.go
LeoHChen: sure.

consensus/consensus_service.go
LeoHChen: okay.

core/blockchain.go
LeoHChen: let me check.

consensus/view_change.go
```diff
@@ -124,12 +230,13 @@ func (consensus *Consensus) startViewChange(viewID uint64) {
 	consensus.consensusTimeout[timeoutConsensus].Stop()
 	consensus.consensusTimeout[timeoutBootstrap].Stop()
 	consensus.current.SetMode(ViewChanging)
-	consensus.SetViewChangingID(viewID)
```
rlan35, line 127: startViewChange(viewID uint64);  can remove the param for this func as the viewID is now basically internal var of consensus.

consensus/consensus_v2.go
rlan35: can remove the viewID param for consensus.startViewChange() actually as the value used in inside consensus.

consensus/view_change.go
```diff
@@ -90,20 +93,123 @@ func (pm *State) GetViewChangeDuraion() time.Duration {
...
+		gap = int(viewID - consensus.GetCurBlockViewID())
+	}
+	var lastLeaderPubKey *bls.PublicKeyWrapper
+	var err error
+	epoch := big.NewInt(0)
+	if consensus.ChainReader == nil {
+		consensus.getLogger().Error().Msg("[getNextLeaderKey] ChainReader is nil. Use consensus.LeaderPubKey")
+		lastLeaderPubKey = consensus.LeaderPubKey
+	} else {
+		curHeader := consensus.ChainReader.CurrentHeader()
```
rlan35, line 169: If we can't get currentHeader, it's an error case which we should just return error since we don't know which epoch anymore. Epoch = 0 won't work.

LeoHChen: I think it may be the bootstrap phase, and view change happened.  It is not applicable to current mainnet though.

consensus/quorum/quorum.go
rlan35: if NumHarmonyOperatedNodesPerShard returns something off. We don't want crash.

LeoHChen: okay

consensus/consensus_v2.go
LeoHChen: done.

consensus/view_change.go
LeoHChen: done

consensus/view_change.go
LeoHChen: how to calculate the right epoch then if the header is not available? that's why I use epoch 0 as the fallback value. in case of this error case happened.

core/blockchain.go
LeoHChen: changed.

consensus/consensus_service.go
```diff
@@ -92,6 +92,10 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 		consensus.LeaderPubKey = &allKeys[0]
 		utils.Logger().Info().
 			Str("info", consensus.LeaderPubKey.Bytes.Hex()).Msg("My Leader")
+	} else {
+		utils.Logger().Error().
+			Msg("[UpdatePublicKeys] Participants is empty")
+		consensus.pubKeyLock.Unlock()
```
rlan35, line 98: remove this line. double unlocked.

consensus/consensus_v2.go
```diff
@@ -264,13 +264,12 @@ func (consensus *Consensus) Start(
 					}
 					if k != timeoutViewChange {
 						consensus.getLogger().Warn().Msg("[ConsensusMainLoop] Ops Consensus Timeout!!!")
-						viewID := consensus.GetCurBlockViewID()
-						consensus.startViewChange(viewID + 1)
+						consensus.SetViewChangingID(consensus.GetCurBlockViewID())
```
rlan35, line 267: why set SetViewChangingID here? It will be set inside startViewChange() right?

consensus/consensus_service.go
LeoHChen: oh, forget to remove this one.

consensus/consensus_v2.go
LeoHChen: This is for fallback mode, in case of get view id failed from blockchain.  Yes it will be set after successfully getnextviewid withing startviewchange func.  The fallback mode will use the current viewchanging id to get the next one, which is current + 1, that's the old algorithm.  that's why it is the fallback mode.

rlan35 (approved):

PR #3388, opened 2020-10-14
View Change Revamp 2

LeoHChen:
This set of PR further improves the stability of the view change process.

The enhancement includes:

* retry sending view change messages, in case the new leader started view change process slower than the other validators that may miss a few view change messages. This commit can reduce another round of view change process.
* ignore viewchange/newview message if the validator is not in viewchanging mode
* RLock to protect the reading of m1Payload
* use consistent data to construct the newview message
* use sendCommitMessages functions to reduce code duplication onViewChange
* further protection of consensus data change during view change process using consensus.mutex

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -43,6 +43,13 @@ func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb
 		return nil
 	}
 
+	// when node is not in ViewChanging mode, ignore viewchange / newview messages
+	if !consensus.IsViewChangingMode() &&
+		(msg.Type == msg_pb.MessageType_VIEWCHANGE ||
+			msg.Type == msg_pb.MessageType_NEWVIEW) {
+		return nil
+	}
+
 	intendedForValidator, intendedForLeader :=
 		!consensus.IsLeader(),
 		consensus.IsLeader()
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -43,13 +43,6 @@ func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb
 		return nil
 	}
 
-	// when node is not in ViewChanging mode, ignore viewchange / newview messages
-	if !consensus.IsViewChangingMode() &&
-		(msg.Type == msg_pb.MessageType_VIEWCHANGE ||
-			msg.Type == msg_pb.MessageType_NEWVIEW) {
-		return nil
-	}
-
 	intendedForValidator, intendedForLeader :=
 		!consensus.IsLeader(),
 		consensus.IsLeader()
--- a/consensus/validator.go
+++ b/consensus/validator.go
@@ -89,6 +89,41 @@ func (consensus *Consensus) prepare() {
 	consensus.switchPhase("Announce", FBFTPrepare)
 }
 
+// sendCommitMessages send out commit messages to leader
+func (consensus *Consensus) sendCommitMessages(blockObj *types.Block) {
+	// Sign commit signature on the received block
+	commitPayload := signature.ConstructCommitPayload(consensus.ChainReader,
+		blockObj.Epoch(), blockObj.Hash(), blockObj.NumberU64(), blockObj.Header().ViewID().Uint64())
+	groupID := []nodeconfig.GroupID{
+		nodeconfig.NewGroupIDByShardID(nodeconfig.ShardID(consensus.ShardID)),
+	}
+	for _, key := range consensus.priKey {
+		if !consensus.IsValidatorInCommittee(key.Pub.Bytes) {
+			continue
+		}
+
+		networkMessage, _ := consensus.construct(
+			msg_pb.MessageType_COMMIT,
+			commitPayload,
+			&key,
+		)
+
+		if consensus.current.Mode() != Listening {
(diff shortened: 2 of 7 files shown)


rlan35:
description about the PR please.

consensus/view_change_construct.go
```diff
@@ -439,6 +440,8 @@ func (vc *viewChange) InitPayload(
 
 // isM1PayloadEmpty returns true if m1Payload is not set
 func (vc *viewChange) isM1PayloadEmpty() bool {
+	vc.vcLock.RLock()
```
rlan35, line 443: deadlock with IsM1PayloadEmpty

consensus/view_change.go
```diff
@@ -157,14 +157,55 @@ func (consensus *Consensus) startViewChange(viewID uint64) {
...
-		)
+		); err != nil {
+			consensus.getLogger().Err(err).
+				Msg("could not send out the ViewChange message")
+		}
 	}
 }
 
+// stopViewChange stops the current view change
+func (consensus *Consensus) stopViewChange(viewID uint64, newLeaderPriKey *bls.PrivateKeyWrapper) error {
```
rlan35, line 174: Let's use "StartNewView()"?. stopViewChange could mean preemptively stopping something.

consensus/consensus_v2.go
```diff
@@ -43,6 +43,13 @@ func (consensus *Consensus) HandleMessageUpdate(ctx context.Context, msg *msg_pb
 		return nil
 	}
 
+	// when node is not in ViewChanging mode, ignore viewchange / newview messages
+	if !consensus.IsViewChangingMode() &&
```
rlan35, line 47: there are already logic to filter messages in validateShardBoundMessage. 
```	// when node is in ViewChanging mode, it still accepts normal messages into FBFTLog
	// in order to avoid possible trap forever but drop PREPARE and COMMIT
	// which are message types specifically for a node acting as leader
	// so we just ignore those messages
	if node.Consensus.IsViewChangingMode() {
		switch m.Type {
		case msg_pb.MessageType_PREPARE, msg_pb.MessageType_COMMIT:
			return nil, nil, true, nil
		}
	} else {
		// ignore newview message if the node is not in viewchanging mode
		switch m.Type {
		case msg_pb.MessageType_NEWVIEW:
			return nil, nil, true, nil
		}
	}```
No need to add filter here. (the new logic is different, which we shouldn't change)

File: consensus/validator.go [hunk 4]
```diff
@@ -89,6 +89,41 @@ func (consensus *Consensus) prepare() {
...
+		blockObj.Epoch(), blockObj.Hash(), blockObj.NumberU64(), blockObj.Header().ViewID().Uint64())
+	groupID := []nodeconfig.GroupID{
+		nodeconfig.NewGroupIDByShardID(nodeconfig.ShardID(consensus.ShardID)),
+	}
+	for _, key := range consensus.priKey {
+		if !consensus.IsValidatorInCommittee(key.Pub.Bytes) {
+			continue
+		}
+
+		networkMessage, _ := consensus.construct(
```
rlan35, line 105: shouldn't ignore the err. If err != nil, networkMessage will be nil.

consensus/view_change.go
LeoHChen: sure

consensus/consensus_v2.go
LeoHChen: good point. let me check. I'll ignore viewchange message in the validation process.

consensus/validator.go
LeoHChen: will check.

consensus/view_change.go
```diff
@@ -224,88 +265,34 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-					key.Pri.SignHash(commitPayload),
-					common.BytesToHash(consensus.blockHash[:]),
-					block.NumberU64(),
-					block.Header().ViewID().Uint64(),
-				); err != nil {
-					consensus.getLogger().Warn().Err(err).Msg("submit vote on viewchange commit failed")
-					return
-				}
+			consensus.ResetState()
+			consensus.LeaderPubKey = newLeaderKey
```
rlan35, line 277: why not put 
consensus.ResetState()
consensus.LeaderPubKey = newLeaderKey

inside startNewView?   they need to be lock protected too.

consensus/view_change.go
```diff
@@ -224,88 +265,34 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
...
-		consensus.consensusTimeout[timeoutViewChange].Stop()
-		consensus.consensusTimeout[timeoutConsensus].Start()
-		consensus.getLogger().Info().
-			Str("myKey", newLeaderKey.Bytes.Hex()).
-			Msg("[onViewChange] I am the New Leader")
+		if err := consensus.startNewView(recvMsg.ViewID, newLeaderPriKey); err != nil {
+			consensus.getLogger().Error().Err(err).Msg("[onViewChange] startNewView failed")
+			return
+		}
+		consensus.ResetState()
```
rlan35, line 294: same here.

consensus/view_change.go
LeoHChen: make sense.

consensus/view_change.go
LeoHChen: will change.

rlan35 (approved):

PR #3389, opened 2020-10-14
[RPC] Port EVM tracer

janet-harmony:
* Ports `api_tracer.go` from Geth to the Harmony object
  - This allows these underlying functions to be used by the Explorer service in the future
* Adds `debug` namespace to the RPC layer for new EVM tracing RPC
* Some `debug` RPCs are currently disabled & will return "RPC not available yet" errors
* Available RPCs: `debug_traceCall` & `debug_traceTransactions`

Example for `debug_traceTransaction`: [here](https://gist.github.com/janet-harmony/707b9d40b9abd3ec3eed6669cc156d67).


rlan35:
Just want to confirm. The end goal of this effort is to have the ability to display something like this. 
image

janet-harmony:
This will provide the EVM execution traces necessary to generate the internal transaction data, but something else will have to parse the traces in order to generate the readable format that Etherscan has.

rpc/contract.go
```diff
@@ -116,44 +113,8 @@ func doCall(
 		return core.ExecutionResult{}, err
 	}
 
-	// Set sender address or use a default if none specified
-	var addr common.Address
```
rlan35, line 120: why are these codes removed?

hmy/tracer.go
```diff
@@ -0,0 +1,743 @@
+// Copyright 2017 The go-ethereum Authors
```
rlan35, line 1: is this file modified meaningfully or it's basically a simple port?

rpc/tracer.go
```diff
@@ -0,0 +1,177 @@
+// Copyright 2017 The go-ethereum Authors
```
rlan35, line 1: is this file modified meaningfully or it's basically a simple port?

rlan35 (approved):

PR #3390, opened 2020-10-15
Rosetta Implementation Cleanup (Stage 3 of Node API Overhaul)

Daniel-VDM:
# Stage 3 Cleanup of [Node API Overhaul](https://github.com/harmony-one/harmony/issues/3210)

This PR clean's up the Rosetta implementation for the initial release. A large part of this PR is just moving functions/methods/tests around and putting useful util functions deeper into the codebase. 

I've also fixed a bug in the pre-staking era block reward calculation. Essentially the calculation needs to be processed in the order of the committee slots to handle some floating-point precision rounding. I've noted that this order is important in the fix. This error was caught using the rosetta-CLI (thx god for that tool 😄 ).  

Lastly, I've renamed all of the transfer operations to native transfer operations where applicable. This is in anticipation of integrating HRC20 tokens to rosetta's data & construction API at some point down the line.


core/types/receipt.go
```diff
@@ -218,3 +219,19 @@ func (r Receipts) ToShardID(i int) uint32 {
 func (r Receipts) MaxToShardID() uint32 {
 	return 0
 }
+
+// FindLogsWithTopic returns all the logs that contain the given receipt
+func FindLogsWithTopic(
+	receipt *Receipt, targetTopic ethcommon.Hash,
+) []*Log {
+	logs := []*Log{}
+	for _, log := range receipt.Logs {
```
LeoHChen, line 228: go through all the logs ,will it have any performance concern on mainnet?

hmy/blockchain.go
```diff
@@ -73,6 +73,51 @@ func (hmy *Harmony) GetBlockSigners(
...
+	// Mask is the bitmap Mask for the block.
+	Mask      *bls.Mask
+	BlockHash common.Hash
+}
+
+// GetDetailedBlockSignerInfo fetches the block signer information for any non-genesis block
+func (hmy *Harmony) GetDetailedBlockSignerInfo(
+	ctx context.Context, blk *types.Block,
+) (*DetailedBlockSignerInfo, error) {
+	slotList, mask, err := hmy.GetBlockSigners(
```
LeoHChen, line 94: will committee change per block, or just change per epoch?  will you consider to cache the data?

hmy/staking.go
```diff
@@ -133,6 +133,14 @@ func (hmy *Harmony) IsPreStakingEpoch(epoch *big.Int) bool {
 	return hmy.BlockChain.Config().IsPreStaking(epoch)
 }
 
+// IsCommitteeSelectionBlock checks if the given block is the committee selection block
+func (hmy *Harmony) IsCommitteeSelectionBlock(blk *types.Block) bool {
+	isBeaconChain := blk.ShardID() == shard.BeaconChainShardID
+	isNewEpoch := len(blk.Header().ShardState()) > 0
```
LeoHChen, line 139: is there any util function to check isnewepoch?  this is a very common operation, maybe we can add a function is not alrready.

LeoHChen (approved):
LGTM in general. Good you catch the issue using tool.

core/types/receipt.go
Daniel-VDM: Possibly, but this is probably the best we can do since the logs are stored in an array.

hmy/blockchain.go
Daniel-VDM: It is cached...

The committee is fetched from shard state, which is cached.

https://github.com/harmony-one/harmony/blob/6ce8aeac3d448d0bc13e5f1286ce97856de42f1f/core/blockchain.go#L1800-L1816

https://github.com/harmony-one/harmony/blob/6ce8aeac3d448d0bc13e5f1286ce97856de42f1f/hmy/staking.go#L178-L190

https://github.com/harmony-one/harmony/blob/6ce8aeac3d448d0bc13e5f1286ce97856de42f1f/hmy/blockchain.go#L36-L53

hmy/staking.go
Daniel-VDM: Sure, the last commit of this PR refactored this part.

node/node_explorer.go
```diff
@@ -123,7 +123,7 @@ func (node *Node) explorerMessageHandler(ctx context.Context, msg *msg_pb.Messag
 func (node *Node) AddNewBlockForExplorer(block *types.Block) {
 	utils.Logger().Info().Uint64("blockHeight", block.NumberU64()).Msg("[Explorer] Adding new block for explorer node")
 	if _, err := node.Blockchain().InsertChain([]*types.Block{block}, true); err == nil {
-		if len(block.Header().ShardState()) > 0 {
+		if block.Header().IsLastBlockInEpoch() {
```
LeoHChen, line 126: you already have block.IsLastBlockInEpoch

node/worker/worker.go
```diff
@@ -317,7 +317,7 @@ func (w *Worker) GetNewEpoch() *big.Int {
 		// have an epoch and it will decide the next epoch for following blocks
 		epoch = new(big.Int).Set(shardState.Epoch)
 	} else {
-		if len(parent.Header().ShardState()) > 0 && parent.NumberU64() != 0 {
+		if parent.Header().IsLastBlockInEpoch() && parent.NumberU64() != 0 {
```
LeoHChen, line 320: will you just use parent.IslastBlockInEpoch?

LeoHChen (approved):
two minor function calls.

core/offchain.go
```diff
@@ -39,7 +39,6 @@ func (bc *BlockChain) CommitOffChainData(
 	isStaking := bc.chainConfig.IsStaking(block.Epoch())
 	isPreStaking := bc.chainConfig.IsPreStaking(block.Epoch())
 	header := block.Header()
-	isNewEpoch := len(header.ShardState()) > 0
```
rlan35, line 42: better keep the isNewEpoch variable so not repeated call IsLastBlockInEpoch(). Also less code change.

internal/chain/engine.go
```diff
@@ -312,6 +310,14 @@ func payoutUndelegations(
 	return nil
 }
 
+// IsCommitteeSelectionBlock checks if the given header is for the committee selection block
+// which can only occur on beacon chain and in pre-staking era
```
rlan35, line 314: saying "after staking is enabled (epoch > pre-staking epoch)" is less confusing then "in pre-staking era."

rlan35 (approved):

node/node_explorer.go
Daniel-VDM: yep forgot to change this

core/offchain.go
Daniel-VDM: Sure changed back.

internal/chain/engine.go
Daniel-VDM: agreed, I changed the comment

PR #3394, opened 2020-10-17
[ntp] check local time against ntp server

LeoHChen:
* add sys.ntp flag
* exit the node program if local time is 30 seconds off the NTP time

### DIFF
--- a/cmd/harmony/config.go
+++ b/cmd/harmony/config.go
@@ -27,6 +27,7 @@ type harmonyConfig struct {
 	TxPool    txPoolConfig
 	Pprof     pprofConfig
 	Log       logConfig
+	Sys       sysConfig
 	Consensus *consensusConfig `toml:",omitempty"`
 	Devnet    *devnetConfig    `toml:",omitempty"`
 	Revert    *revertConfig    `toml:",omitempty"`
@@ -96,6 +97,10 @@ type logConfig struct {
 type logContext struct {
 	IP   string
 	Port int
+}
+
+type sysConfig struct {
+	NtpServer string
 }
 
 type httpConfig struct {
--- a/cmd/harmony/config.go
+++ b/cmd/harmony/config.go
@@ -27,7 +27,7 @@ type harmonyConfig struct {
 	TxPool    txPoolConfig
 	Pprof     pprofConfig
 	Log       logConfig
-	Sys       sysConfig
+	Sys       *sysConfig       `toml:",omitempty"`
 	Consensus *consensusConfig `toml:",omitempty"`
 	Devnet    *devnetConfig    `toml:",omitempty"`
 	Revert    *revertConfig    `toml:",omitempty"`
--- a/cmd/harmony/default.go
+++ b/cmd/harmony/default.go
@@ -62,6 +62,9 @@ var defaultConfig = harmonyConfig{
 		RotateSize: 100,
 		Verbosity:  3,
 	},
+	Sys: sysConfig{
+		NtpServer: "1.pool.ntp.org",
+	},
 }
 
 var defaultDevnetConfig = devnetConfig{
@@ -79,6 +82,10 @@ var defaultRevertConfig = revertConfig{
 var defaultLogContext = logContext{
 	IP:   "127.0.0.1",
 	Port: 9000,
+}
+
+var defaultSysConfig = sysConfig{
+	NtpServer: "1.pool.ntp.org",
 }
 
 var defaultConsensusConfig = consensusConfig{
@@ -117,6 +124,11 @@ func getDefaultLogContextCopy() logContext {
 
 func getDefaultConsensusConfigCopy() consensusConfig {
 	config := defaultConsensusConfig
+	return config
+}
+
+func getDefaultSysConfigCopy() sysConfig {
+	config := defaultSysConfig
 	return config
 }
 
--- a/cmd/harmony/default.go
+++ b/cmd/harmony/default.go
@@ -62,9 +62,10 @@ var defaultConfig = harmonyConfig{
 		RotateSize: 100,
 		Verbosity:  3,
 	},
-	Sys: sysConfig{
-		NtpServer: "1.pool.ntp.org",
-	},
+}
+
+var defaultSysConfig = sysConfig{
+	NtpServer: "1.pool.ntp.org",
 }
 
 var defaultDevnetConfig = devnetConfig{
@@ -100,6 +101,11 @@ func getDefaultHmyConfigCopy(nt nodeconfig.NetworkType) harmonyConfig {
 		devnet := getDefaultDevnetConfigCopy()
 		config.Devnet = &devnet
 	}
+	return config
+}
+
+func getDefaultSysConfigCopy() sysConfig {
+	config := defaultSysConfig
 	return config
 }
 
--- a/cmd/harmony/default.go
+++ b/cmd/harmony/default.go
@@ -84,10 +84,6 @@ var defaultLogContext = logContext{
 	Port: 9000,
 }
 
-var defaultSysConfig = sysConfig{
-	NtpServer: "1.pool.ntp.org",
-}
-
 var defaultConsensusConfig = consensusConfig{
 	MinPeers: 6,
 }
@@ -124,11 +120,6 @@ func getDefaultLogContextCopy() logContext {
 
 func getDefaultConsensusConfigCopy() consensusConfig {
 	config := defaultConsensusConfig
-	return config
-}
-
-func getDefaultSysConfigCopy() sysConfig {
-	config := defaultSysConfig
 	return config
 }
 
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -908,13 +908,12 @@ var (
 )
 
 func applySysFlags(cmd *cobra.Command, config *harmonyConfig) {
-	if cli.HasFlagsChanged(cmd, sysFlags) {
-		cfg := getDefaultSysConfigCopy()
-		config.Sys = cfg
-	}
-
 	if cli.IsFlagChanged(cmd, sysNtpServerFlag) {
 		config.Sys.NtpServer = cli.GetStringFlagValue(cmd, sysNtpServerFlag)
+	}
+
+	if config.Sys.NtpServer == "" {
+		config.Sys.NtpServer = defaultConfig.Sys.NtpServer
 	}
 }
 
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -903,18 +903,19 @@ var (
 	sysNtpServerFlag = cli.StringFlag{
 		Name:     "sys.ntp",
 		Usage:    "the ntp server",
-		DefValue: defaultConfig.Sys.NtpServer,
+		DefValue: defaultSysConfig.NtpServer,
 		Hidden:   true,
 	}
 )
 
 func applySysFlags(cmd *cobra.Command, config *harmonyConfig) {
-	if cli.IsFlagChanged(cmd, sysNtpServerFlag) {
-		config.Sys.NtpServer = cli.GetStringFlagValue(cmd, sysNtpServerFlag)
+	if cli.HasFlagsChanged(cmd, sysFlags) || config.Sys == nil {
+		cfg := getDefaultSysConfigCopy()
+		config.Sys = &cfg
 	}
 
-	if config.Sys.NtpServer == "" {
-		config.Sys.NtpServer = defaultConfig.Sys.NtpServer
+	if cli.IsFlagChanged(cmd, sysNtpServerFlag) {
+		config.Sys.NtpServer = cli.GetStringFlagValue(cmd, sysNtpServerFlag)
 	}
 }
 
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -904,6 +904,7 @@ var (
 		Name:     "sys.ntp",
 		Usage:    "the ntp server",
 		DefValue: defaultConfig.Sys.NtpServer,
+		Hidden:   true,
 	}
 )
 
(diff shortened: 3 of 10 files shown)


cmd/harmony/default.go
```diff
@@ -120,6 +127,11 @@ func getDefaultConsensusConfigCopy() consensusConfig {
 	return config
 }
 
+func getDefaultSysConfigCopy() sysConfig {
```
JackyWYX, line 130: I was using these `getDefaultConsensusConfigCopy` functions to get default value for optional config fields (fields are pointers in harmonyConfig). I think currently `sysConfig` is a mandatory field in harmony config, so maybe we don't need this function?

cmd/harmony/flags.go
```diff
@@ -894,6 +899,25 @@ func applyLogFlags(cmd *cobra.Command, config *harmonyConfig) {
...
+
+func applySysFlags(cmd *cobra.Command, config *harmonyConfig) {
+	if cli.HasFlagsChanged(cmd, sysFlags) {
+		cfg := getDefaultSysConfigCopy()
+		config.Sys = cfg
+	}
+
+	if cli.IsFlagChanged(cmd, sysNtpServerFlag) {
+		config.Sys.NtpServer = cli.GetStringFlagValue(cmd, sysNtpServerFlag)
+	}
```
JackyWYX, line 918: config loaded by harmony.conf is also parsed here, so it is possible that sys flags are not changed and ntp field is empty. Probably add:

```
if config.Sys.NtpServer == "" {
    config.Sys.NtpServer = defaultConfig.Sys.NtpServer
}
```

Do you think so?

cmd/harmony/default.go
```diff
@@ -81,6 +84,10 @@ var defaultLogContext = logContext{
 	Port: 9000,
 }
 
+var defaultSysConfig = sysConfig{
+	NtpServer: "1.pool.ntp.org",
```
JackyWYX, line 88: Probably we can remove this variable since we may not need `getDefaultSysConfigCopy` (Check the other comment). Also, the value of NtpServer is defined twice in this package, which is not very good.

common/ntp/ntp.go
```diff
@@ -0,0 +1,35 @@
...
+	beevik_ntp "github.com/beevik/ntp"
+)
+
+const (
+	// tolerance range of local clock with NTP time server
+	toleranceRange = time.Duration(50 * time.Millisecond)
+)
+
+// IsLocalTimeAccurate returns whether the local clock acturate or not
+func IsLocalTimeAccurate(ntpServer string) bool {
```
JackyWYX, line 15: For most of the cases, returning an error instead of boolean would be more preferable since it gives more context about what error is about. For this part maybe returning error like 

```
errors.Wrapf("system clock is off by %v", diff)
```

will be more resourceful.

common/ntp/ntp.go
```diff
@@ -0,0 +1,35 @@
...
+	response, err := beevik_ntp.Query(ntpServer)
```
JackyWYX, line 16: Probably we shall do a sanity check on the ntpServer configuration here. So that if a user has wrong ntpServer config, he will not get a non-sense error saying system clock is off.

cmd/harmony/main.go
```diff
@@ -531,6 +538,8 @@ func createGlobalConfig(hc harmonyConfig) (*nodeconfig.ConfigType, error) {
 		nodeConfig.WebHooks.Hooks = config
 	}
 
+	nodeConfig.NtpServer = hc.Sys.NtpServer
```
JackyWYX, line 541: Since ntp check is done only once, do you think it would be better to put local system check in function `runHarmonyNode` (cmd/harmony/main.go:125) instead of adding more complexity to already exploded module `nodeConfig`?

common/ntp/ntp.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"time"
+
+	beevik_ntp "github.com/beevik/ntp"
+)
+
+const (
+	// tolerance range of local clock with NTP time server
+	toleranceRange = time.Duration(50 * time.Millisecond)
```
JackyWYX, line 11: How did we come with this tolerance? I am actually not sure whether this 50 milliseconds works in general, but Ethereum has this clock drift threshold with only 10 seconds. (go-ethereum/p2p/discover/v4_udp.go:59)

JackyWYX (commented):
After all, I wonder whether do we need user to config the ntp endpoint? For Ethereum, they have a lot of configs but only have ntpPool hard coded (go-ethereum/p2p/discover/ntp.go:32). I guess the thoughts behind this is that ntp is a well-known service and they don't want to give options that's not very helpful from user's perspective. What's your idea?

rlan35:
Don't think we should hard-rely on a centralized service for time, especially should not exit based on the result. What is the centralized service is hacked and give a malicious results? Then the whole network will exit. We could let the know the check the time and print a warning to the node runner if the local time is suspected to be way off (maybe 2 seconds, since we have 45 seconds window). but definitely shouldn't trust the centralized service and do any significant actions on the node based on it.

common/ntp/ntp.go
rlan35: yea, we could tolerant seconds level too, as our view change window is 45 seconds.

JackyWYX:
> Don't think we should hard-rely on a centralized service for time, especially should not exit based on the result. 

I think we can rely on these services but not relying too hard on it. 50ms would be a super strong dependency that goes too far. Instead, loosing constraint and using it as a sanity check would be better from my point of view.

> It could happen that the centralized service is hacked and give a malicious results. Then the whole network will exit.

The check logic happens only once at bootstrap, only effecting new nodes and will not effect nodes currently running. This probably is not a problem.

rlan35:
> > Don't think we should hard-rely on a centralized service for time, especially should not exit based on the result.
> 
> I think we can rely on these services but not relying too hard on it. 50ms would be a super strong dependency that goes too far. Instead, loosing constraint and using it as a sanity check would be better from my point of view.
> 
> > It could happen that the centralized service is hacked and give a malicious results. Then the whole network will exit.
> 
> The check logic happens only once at bootstrap, only effecting new nodes and will not effect nodes currently running. This probably is not a problem.

Still, if a malicious attack on the centralized server could prevent node from start running, it's also potential attack vector. If the service is hacked, then no new node could start running (or crashed node can't restart anymore) and it's dangerous for the network.

We could follow what ethereum is doing with the time check, 10seconds threshold and only warning to users.

```
if drift < -driftThreshold || drift > driftThreshold {
		log.Warn(fmt.Sprintf("System clock seems off by %v, which can prevent network connectivity", drift))
		log.Warn("Please enable network time synchronisation in system settings.")
	} else {
		log.Debug("NTP sanity check done", "drift", drift)
	}
```

LeoHChen:
> > > Don't think we should hard-rely on a centralized service for time, especially should not exit based on the result.
> > 
> > 
> > I think we can rely on these services but not relying too hard on it. 50ms would be a super strong dependency that goes too far. Instead, loosing constraint and using it as a sanity check would be better from my point of view.
> > > It could happen that the centralized service is hacked and give a malicious results. Then the whole network will exit.
> > 
> > 
> > The check logic happens only once at bootstrap, only effecting new nodes and will not effect nodes currently running. This probably is not a problem.
> 
> Still, if a malicious attack on the centralized server could prevent node from start running, it's also potential attack vector to the network. If that happens, then no new node could start running (or crashed node can't restart anymore) and it's dangerous for the network.
> 
> We could simply follow what ethereum is doing with the time check, 10seconds threshold and only warning to users.
> 
> ```
>         if drift < -driftThreshold || drift > driftThreshold {
> 		log.Warn(fmt.Sprintf("System clock seems off by %v, which can prevent network connectivity", drift))
> 		log.Warn("Please enable network time synchronisation in system settings.")
> 	} else {
> 		log.Debug("NTP sanity check done", "drift", drift)
> 	}
> ```

Our view change will rely on the relative accuracy of the time, if the local time is way off, it may not be able to join in the view change.  We can definitely relax the threshold, 50ms is a strong dependency I agree.  We can maybe use 5~10 seconds. Ethereum doesn't really rely on local time, so it is not a dependency.

We can have two levels of threshold maybe.  If the drift is within 5~10 sec, we can give warning. If more than 30 sec, we can error out.

common/ntp/ntp.go
LeoHChen: sure. we can have two levels of tolerance levels.

cmd/harmony/main.go
LeoHChen: ntpServer may still be used later in get current time, so we need it in a configuration variable. Any suggestion on where to put it?

common/ntp/ntp.go
LeoHChen: let me see if we can get more error info output.

cmd/harmony/main.go
JackyWYX: I think after initial check,  we can use `time.Now()` as local current time. So why do we need ntpServer when after initialization?

LeoHChen:
> After all, I wonder whether do we need user to config the ntp endpoint? For Ethereum, they have a lot of configs but only have ntpPool hard coded (go-ethereum/p2p/discover/ntp.go:32). I guess the thoughts behind this is that ntp is a well-known service and they don't want to give options that's not very helpful from user's perspective. What's your idea?

how about hide the option?

JackyWYX:
> > After all, I wonder whether do we need user to config the ntp endpoint? For Ethereum, they have a lot of configs but only have ntpPool hard coded (go-ethereum/p2p/discover/ntp.go:32). I guess the thoughts behind this is that ntp is a well-known service and they don't want to give options that's not very helpful from user's perspective. What's your idea?
> 
> how about hide the option?

So what is the situation do you think that a user (or us) probably want to use a custom ntp server other than `1.pool.ntp.org` in production? If not, I would suggest totally not to expose this option to users or us, neither hidden or not.

cmd/harmony/flags.go
LeoHChen: removed

cmd/harmony/main.go
LeoHChen: I still want to use ntp time when view change started, and default to time.Now(). The local maybe off, but within our range during initialization.  It may get off gradually later, since view change will be rely on time, I may still try to get ntp time during view change process.

LeoHChen:
> > > After all, I wonder whether do we need user to config the ntp endpoint? For Ethereum, they have a lot of configs but only have ntpPool hard coded (go-ethereum/p2p/discover/ntp.go:32). I guess the thoughts behind this is that ntp is a well-known service and they don't want to give options that's not very helpful from user's perspective. What's your idea?
> > 
> > 
> > how about hide the option?
> 
> So what is the situation do you think that a user (or us) probably want to use a custom ntp server other than `1.pool.ntp.org` in production? If not, I would suggest totally not to expose this option to users or us, neither hidden or not.

I agree, it may not be common to use a vendor ntp server even though it is possible. but what if the 1.pool.ntp.org is not available, and we can still give user an easy option to switch to other ntp server pool though.  It's an external dependency, I think it is still better to make it configurable.

cmd/harmony/config.go
```diff
@@ -27,6 +27,7 @@ type harmonyConfig struct {
 	TxPool    txPoolConfig
 	Pprof     pprofConfig
 	Log       logConfig
+	Sys       sysConfig
```
JackyWYX, line 30: If we are to use make sysConfig hidden, simply make a pointer and add toml tag here.

JackyWYX:
> I agree, it may not be common to use a vendor ntp server even though it is possible. but what if the 1.pool.ntp.org is not available, and we can still give user an easy option to switch to other ntp server pool though. It's an external dependency, I think it is still better to make it configurable.

This reasoning justifies the optional system flags. Please make it an optional config in `harmonyConfig`. Thanks.

cmd/harmony/default.go
LeoHChen: removed

common/ntp/ntp.go
LeoHChen: added error

LeoHChen:
> 
> This reasoning justifies the optional system flags. Please make it an optional config in `harmonyConfig`. Thanks.

How to make it optional?  any example?  I didn't find existing examples of optional flag.

JackyWYX (approved):

cmd/harmony/main.go
rlan35: We shouldn't blindly trust a centralized service to set the local node's time. There will be attack vector from that. Ethereum just bring up a warning after checking with ntp, but ntp is not blindly used to set local time I think. It's the validator's responsibility to set their local time correctly. We can add warning about time drift. But eventually, how to set the local time shouldn't be automatically controlled by a centralized service.

common/ntp/ntp.go
```diff
@@ -0,0 +1,46 @@
...
+// CheckLocalTimeAccurate returns whether the local clock accurate or not
+func CheckLocalTimeAccurate(ntpServer string) (bool, error) {
+	response, err := beevik_ntp.Query(ntpServer)
+	// failed to query ntp time
+	if err != nil {
+		return false, err
+	}
+	// drift too much
+	if response.ClockOffset > toleranceRangeError {
+		return false, errDriftTooMuch
```
rlan35, line 30: the return value for this func is a bit messy. error should only be used to indicate something wrong happened in the normal operations of the func, causing failure to do the work and return a value. The errDriftTooMuch and errDriftInRange are actually just different return values, not errors. It would be better to have a enum for the return value: <Normal, DriftTooMuch, DriftInRange>, rather than piggy back on error for value return.

LeoHChen: I felt that's a bit over-engineering for such a simple func. we used error to provide additional info on why the function failed.

Issue #3399, opened 2020-10-21
[regression] view change during epoch change is broken


LeoHChen:
**Describe the bug**
I was testing the new synchronous view change algorithm, especially for view change at the first block of the epoch change. 
Then I found the view change doesn't work during the epoch change.

**Expected behavior**

The expected behavior is to always return to the validator index 0 after epoch change.
However, the latest code on `main` branch doesn't exhibit this behavior.
The leader remains the same one before the epoch change.

**To Reproduce**

This problem happened on the latest `main` branch, but I think I have pinpointed to this PR (https://github.com/harmony-one/harmony/pull/3352), especially the `commitBlock` function.

I've tested the code on (https://github.com/harmony-one/harmony/pull/3388), the view change exhibited the expected behavior.

To reproduce the behavior on `stn` shard0, after updated the code to a certain build.
Refer to this gist on `stn` testing, https://gist.github.com/LeoHChen/f8eb8593092c65a12cb1c78f4569fba2

- stop the current leader, normally the node index 0 on shard 0
- watch the leader change to node index 2, and get hooray
- resume the node index 0, node index 2 is still the leader, node index 0 becomes a normal validator
- wait for epoch change, after epoch change, node index 2 is still the leader. The expected behavior is node index 0 should become the new leader after epoch change.

**Logs**

I've captured some logs of the node index 2 after epoch change, on two different build in this gist,
https://gist.github.com/LeoHChen/34dc4c30968e70396a67a151bc5dd24f

JackyWYX:
Addressed in #3400. Closing

PR #3400, opened 2020-10-22
[consensus] fix leader change during epoch change

JackyWYX:
## Issue 

#3399 Sync consensus fix branch breaks the leader change during epoch change. The reason is that the leader public key was overwritten by fbft message sender key. After this fix, the leader will be automatically switched to index 0 bls key after epoch change.

## Test

Tested on stn shard 1. The epoch change can restore the leader key.

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -521,6 +521,10 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
 	atomic.AddUint64(&consensus.blockNum, 1)
 	consensus.SetCurBlockViewID(committedMsg.ViewID + 1)
 	consensus.LeaderPubKey = committedMsg.SenderPubkeys[0]
+	// Update consensus keys at last so the change of leader status doesn't mess up normal flow
+	if blk.IsLastBlockInEpoch() {
+		consensus.SetMode(consensus.UpdateConsensusInformation())
+	}
 	consensus.ResetState()
 	return nil
 }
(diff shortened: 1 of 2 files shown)


consensus/consensus_v2.go
```diff
@@ -521,6 +521,10 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
 	atomic.AddUint64(&consensus.blockNum, 1)
 	consensus.SetCurBlockViewID(committedMsg.ViewID + 1)
 	consensus.LeaderPubKey = committedMsg.SenderPubkeys[0]
```
rlan35, line 523: I think the better solution is to move 521-523 above OnConsensusdone(), rather than move that code from OnConsensusDone to here.

rlan35: nvm. this works too. approved. These code needs a bit more restructure anyway in the pipelining code.

rlan35 (approved):

consensus/consensus_v2.go
JackyWYX: It's possible that `OnConsensusDone` returns an error. If we move `OnConsensusDone` (which calls `InsertChain`),  the node might be blocked in consensus module since the blockNum has been incremented while the block is not added in blockchain. 

Also, `UpdateConsensusInformation` would better not to be exposed to public. Please let me know your further thinking.

JackyWYX: Yea, please let me know if there are any restructure ideas. I would be super happy to be engaged in the discussion.

PR #3407, opened 2020-10-25
Viewchange stability

LeoHChen:
This PR handles a few more corner cases like empty NewView message.

Also, we shall stop sending retry viewchange messages once NewView message sent or received.

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -303,6 +303,8 @@ func (consensus *Consensus) startNewView(viewID uint64, newLeaderPriKey *bls.Pri
 		Hex("M1Payload", consensus.vc.GetM1Payload()).
 		Msg("[startNewView] Sent NewView Messge")
 
+	consensus.msgSender.StopRetry(msg_pb.MessageType_VIEWCHANGE)
+
 	consensus.current.SetMode(Normal)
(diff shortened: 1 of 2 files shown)


consensus/view_change.go
```diff
@@ -278,9 +278,17 @@ func (consensus *Consensus) startNewView(viewID uint64, newLeaderPriKey *bls.Pri
 	consensus.mutex.Lock()
 	defer consensus.mutex.Unlock()
 
+	if !consensus.IsViewChangingMode() {
```
JackyWYX, line 281: Can we move this piece of code to the start of `onViewChange(consensus/view_change.go:317)`?

LeoHChen: That won't help.  The new leader may still in the viewchanging mode, when it received two view change messages at the same time.  However, one is processed and ready to start a new view, another one is still being processed and just came a bit late.  the OnViewChange is not protected by mutex.  However, startNewView should only be called once. That's why the check at the top of this function, since this function is protected by mutex. That's guaranteed to be called only once per viewchange.

JackyWYX: Fair. Thanks for clarification

JackyWYX (approved):

PR #3411, opened 2020-10-26
Update viewchangingid; Keep block hash in the case of m1 view change

rlan35:
Fix the issue that view changing id isn't updated as new block committed.

Also should keep the block hash in new view as the leader still need to finalize the block hash if it's a M1 view change.


LeoHChen (approved):
tested on STN and worked as expected, one concern is the copy is not protected by mutex.

LeoHChen:
This is right fix on the race condition!

PR #3414, opened 2020-10-28
Update to v1.4.6 of Rosetta SDK

Daniel-VDM:
This PR also sets the release version for the `rosetta` pkg.

### DIFF
--- a/rosetta/common/config.go
+++ b/rosetta/common/config.go
@@ -12,7 +12,7 @@ import (
 
 const (
 	// RosettaVersion tied back to the version of the rosetta go-sdk
-	RosettaVersion = "0.4.4" // TODO (dm): set variable via build flags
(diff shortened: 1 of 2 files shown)


rosetta/common/config.go
```diff
@@ -12,7 +12,7 @@ import (
 
 const (
 	// RosettaVersion tied back to the version of the rosetta go-sdk
-	RosettaVersion = "0.4.4" // TODO (dm): set variable via build flags
+	RosettaVersion = "1.4.6" // TODO (dm): set variable via build flags
```
LeoHChen, line 15: 1.4.6 or 0.4.6?

Daniel-VDM: 1.x.x since it's now a release. I previously set it to 0.x.x since we were still working on it.

I've made it match https://www.rosetta-api.org/versions.html.

LeoHChen (approved):

PR #3415, opened 2020-10-28
Allow validator to update block commit sigs if it has more sigs

rlan35:
Some refactoring +

Add the logic so when validator receives new committed message with more signatures, it can update it's local database with that.

### DIFF
--- a/consensus/validator.go
+++ b/consensus/validator.go
@@ -227,6 +227,11 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 		consensus.getLogger().Warn().Msg("[OnCommitted] unable to parse msg")
 		return
 	}
+	consensus.getLogger().Info().
+		Uint64("MsgBlockNum", recvMsg.BlockNum).
+		Uint64("MsgViewID", recvMsg.ViewID).
+		Msg("[OnCommitted] Received committed message")
+
 	// NOTE let it handle its own logs
 	if !consensus.isRightBlockNumCheck(recvMsg) {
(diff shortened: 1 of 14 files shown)


consensus/validator.go
```diff
@@ -273,6 +278,21 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 	consensus.aggregatedCommitSig = aggSig
 	consensus.commitBitmap = mask
 
+	// If we already have a committed signature received before, check whether the new one
```
JackyWYX, line 281: I am a little confused about these logic:

1. Is it true that leader can possibly send out multiple committed message? And why multiple committed is required for pipelining?
2. If the user shutdown the node before receiving the full commit sig, is it true that it could possibly cause database corruption in `commitSig`?

rlan35: 1. yea, leader will send out the 67% committed message first. And later could possibly also send out the 100% committed if the next consensus doesn't happen in time. Because it better pass the 100% committed sig to the next leader.

2. No, the db operation is atomic, it's just updating the commitSig for the block. Nothing else. The original logic of updating the block with the commit sig atomically is still there. This is just to potentially update 67% committed with 100% committed sigs

JackyWYX: I see. Thanks for clarification.

JackyWYX (approved):

sophoah:
@rlan35 are the reward distributed based on the signature received the first time or based of the updated number of sig afterward ?

PR #3417, opened 2020-10-29
Rosetta Contract Operations Fix

Daniel-VDM:
This PR fixes a couple of contract related things for Rosetta:

* Contract creations are now represented as a pair of operations, where the receiver's account is the contract account. This is needed to express contract instantiation that takes native ONE tokens (other than gas) from the deployer account.
* All transactions involving smart contract calls will have the 'main' contract address reported
* All constructed transactions using the construction API involving contract code will run through an EVM evaluation for the client to handle (depending on the evm result/error).

In general, this PR makes developing contracts with Rosetta a bit easier.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -615,7 +615,7 @@ func setupConsensusAndNode(hc harmonyConfig, nodeConfig *nodeconfig.ConfigType) 
 	}
 
 	// TODO: refactor the creation of blockchain out of node.New()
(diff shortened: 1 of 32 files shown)


LeoHChen (approved):
It's our own operation, right?  How to capture this error in our regression test of travis? Maybe we shall add a contract creation test as well.

rlan35 (approved):

PR #3419, opened 2020-10-31
Fix view change stuck issue

rlan35:
View change still got stuck when manually abort consensus on commit phase every 7 block, with following code added in leader.go:

```
 func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
+       utils.Logger().Info().Msgf("ViewChanging %d %d", consensus.GetCurBlockViewID(), consensus.GetViewChangingID())
+       if consensus.GetCurBlockViewID()%7 == 0 {
+               return
+       }
        recvMsg, err := consensus.ParseFBFTMessage(msg)
        if err != nil {
                consensus.getLogger().Debug().Err(err).Msg("[OnCommit] Parse pbft message failed")
```

There are two bugs fixed in this PR:

1. When selfCommit, we need to use AddNewVote, instead of SubmitVote. SubmitVote is a lower level logic which doesn't take care of voting power calculation. Fixed and also refactored to make submitVote private.

2. The view changing id changes in cycle, like 7, 8, 9, 10, 7, 8, 9, 10, 7.....  rather than monotonically increase. Actually View ID should never decrease.

Also made view changing id to grow with block view id to make things cleaner and more consistent. Basically, view change id always >= block view id.

PS: also find a corner case leaderPubKey mistake. See the todo added. The mitigation is to make block sync happen less frequent. Because block sync will make the node reject all consensus messages, which is not good for stability.

### DIFF
--- a/api/service/syncing/syncing.go
+++ b/api/service/syncing/syncing.go
@@ -32,7 +32,7 @@ const (
 	TimesToFail                     = 5 // downloadBlocks service retry limit
 	RegistrationNumber              = 3
 	SyncingPortDifference           = 3000
-	inSyncThreshold                 = 0    // when peerBlockHeight - myBlockHeight <= inSyncThreshold, it's ready to join consensus
+	inSyncThreshold                 = 1    // when peerBlockHeight - myBlockHeight <= inSyncThreshold, it's ready to join consensus
 	SyncLoopBatchSize        uint32 = 1000 // maximum size for one query of block hashes
 	verifyHeaderBatchSize    uint64 = 100  // block chain header verification batch size
 	SyncLoopFrequency               = 1    // unit in second
--- a/consensus/quorum/quorum.go
+++ b/consensus/quorum/quorum.go
@@ -81,6 +81,7 @@ type ParticipantTracker interface {
 // SignatoryTracker ..
 type SignatoryTracker interface {
 	ParticipantTracker
+	// This func shouldn't be called directly from outside of quorum. Use AddNewVote instead.
 	submitVote(
 		p Phase, pubkeys []bls.SerializedPublicKey,
 		sig *bls_core.Sign, headerHash common.Hash,
@@ -118,6 +119,7 @@ type Decider interface {
 	DependencyInjectionWriter
 	SetVoters(subCommittee *shard.Committee, epoch *big.Int) (*TallyResult, error)
 	Policy() Policy
+	// Add new vote will add the signature in the memory and increase the cumulative voting power
 	AddNewVote(
 		p Phase, pubkeys []*bls_cosi.PublicKeyWrapper,
 		sig *bls_core.Sign, headerHash common.Hash,
--- a/consensus/quorum/quorum.go
+++ b/consensus/quorum/quorum.go
@@ -81,7 +81,7 @@ type ParticipantTracker interface {
 // SignatoryTracker ..
 type SignatoryTracker interface {
 	ParticipantTracker
-	SubmitVote(
+	submitVote(
 		p Phase, pubkeys []bls.SerializedPublicKey,
 		sig *bls_core.Sign, headerHash common.Hash,
 		height, viewID uint64,
--- a/consensus/quorum/quorum.go
+++ b/consensus/quorum/quorum.go
@@ -273,7 +273,7 @@ func (s *cIdentities) SignersCount(p Phase) int64 {
 	}
 }
 
-func (s *cIdentities) SubmitVote(
+func (s *cIdentities) submitVote(
 	p Phase, pubkeys []bls.SerializedPublicKey,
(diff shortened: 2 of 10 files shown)


api/service/syncing/syncing.go
```diff
@@ -32,7 +32,7 @@ const (
 	TimesToFail                     = 5 // downloadBlocks service retry limit
 	RegistrationNumber              = 3
 	SyncingPortDifference           = 3000
-	inSyncThreshold                 = 0    // when peerBlockHeight - myBlockHeight <= inSyncThreshold, it's ready to join consensus
+	inSyncThreshold                 = 1    // when peerBlockHeight - myBlockHeight <= inSyncThreshold, it's ready to join consensus
```
LeoHChen, line 35: why this change?

consensus/quorum/quorum.go
```diff
@@ -273,7 +273,7 @@ func (s *cIdentities) SignersCount(p Phase) int64 {
 	}
 }
 
-func (s *cIdentities) SubmitVote(
+func (s *cIdentities) submitVote(
```
LeoHChen, line 276: need a better description on how to use this function, can you add some comments.

consensus/view_change.go
```diff
@@ -131,9 +131,8 @@ func (consensus *Consensus) getNextViewID() (uint64, time.Duration) {
 	if curTimestamp <= blockTimestamp {
 		return consensus.fallbackNextViewID()
 	}
-	totalNode := consensus.Decider.ParticipantsCount()
 	// diff is at least 1, and it won't exceed the totalNode
-	diff := uint64(((curTimestamp - blockTimestamp) / viewChangeTimeout) % int64(totalNode))
+	diff := uint64((curTimestamp - blockTimestamp) / viewChangeTimeout)
```
LeoHChen, line 135: this is good, we already take care of nextLeader rotation with mod.

LeoHChen (approved):
some minor comments, LGTM.  Can you add some comments in AddNewVote function, so that we won't misuse it later.

api/service/syncing/syncing.go
rlan35: see the todo added. because block sync will make the validator fail to accept consensus message. and in the rare case when the sync check happens in a bad timing, the validator could enter sync mode prematurely and miss consensus messages.

consensus/quorum/quorum.go
rlan35: Done.

consensus/quorum/quorum.go
```diff
@@ -81,7 +81,8 @@ type ParticipantTracker interface {
 // SignatoryTracker ..
 type SignatoryTracker interface {
 	ParticipantTracker
-	SubmitVote(
+	// This func shouldn't be called directly from outside of quorum. Use AddNewVote instead.
+	submitVote(
```
JackyWYX, line 85: Consider removing from interface?

consensus/validator.go
```diff
@@ -272,6 +270,11 @@ func (consensus *Consensus) onCommitted(msg *msg_pb.Message) {
 
 	consensus.FBFTLog.AddMessage(recvMsg)
 
+	if recvMsg.BlockNum > consensus.blockNum {
```
JackyWYX, line 273: If a prepared message is missing, onCommitted will never spin state sync since the it is already returned in 

```
	blockObj := consensus.FBFTLog.GetBlockByHash(recvMsg.BlockHash)
	if blockObj == nil {
			Uint64("blockNum", recvMsg.BlockNum).
			Uint64("viewID", recvMsg.ViewID).
			Str("blockHash", recvMsg.BlockHash.Hex()).
			Msg("[OnCommitted] Failed finding a matching block for committed message")
		return
	}
```

Is this what we expect?

rlan35: yes, we want to spin up the sync when it's very certain. Don't want to turn to sync mode prematurely. It's risky.

JackyWYX: I mean this sync code is unreachable since the block from prepared message is missing.

consensus/quorum/quorum.go
rlan35: it's used as typing for different type of structs. So not easy to remove.

rlan35:
jenkins, test this please.

consensus/validator.go
rlan35: Ok, then we probably don't need to have this sync code in oncommitted. What do you think?

JackyWYX: OK, it's actually a trade off between catch up performance and how sync mode effects the rest of validator consensus logic. I definitely agree with reducing complexity of the consensus logic for now, and refactoring the message handling logic for sync mode can be a good todo in the future.

JackyWYX (approved):

PR #3423, opened 2020-11-01
[p2p] reduce min peer for non mainnet bootstrap

JackyWYX:
## Issue

For new network bootstrap, it could happen that the peers in p2pnetwork is smaller than lower threshold. This PR is to increase the min consensus peer for non-mainnet node during bootstrap.

E.g. During the bootnode migration, the dht data in boot nodes are removed and caused very limited peers when bootstrap.

```
"p2p-connectivity": {
      "connected": 2,
      "not-connected": 1,
      "total-known-peers": 3
},
```

## Test

Tested locally. After the change, the node will not be terminated because of not enough peers.

### DIFF
--- a/cmd/harmony/config_test.go
+++ b/cmd/harmony/config_test.go
@@ -130,7 +130,7 @@ func TestPersistConfig(t *testing.T) {
 		},
 		{
 			config: makeTestConfig("mainnet", func(cfg *harmonyConfig) {
-				consensus := getDefaultConsensusConfigCopy()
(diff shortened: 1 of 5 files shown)


rlan35 (approved):

LeoHChen:
minpeer is a configurable parameter anyway

JackyWYX:
> minpeer is a configurable parameter anyway

OK. Yea, this is a rare case maybe we don't need this change. Also this PR conflicts with https://github.com/harmony-one/harmony/pull/3438. Closing

PR #3438, opened 2020-11-06
add aggregateSig flag

rlan35:
add flag to control aggregate signature turn-on

tested on local net

### DIFF
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -92,6 +92,7 @@ var (
 	// consensusValidFlags are flags that are effective
 	consensusValidFlags = []cli.Flag{
 		consensusMinPeersFlag,
+		consensusAggregateSigFlag,
 		legacyConsensusMinPeersFlag,
 	}
 
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -772,7 +772,7 @@ var (
 )
 
 func applyConsensusFlags(cmd *cobra.Command, config *harmonyConfig) {
-	if cli.HasFlagsChanged(cmd, consensusValidFlags) {
+	if config.Consensus == nil && cli.HasFlagsChanged(cmd, consensusValidFlags) {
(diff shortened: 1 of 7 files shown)


consensus/consensus_service.go
```diff
@@ -313,13 +311,6 @@ func (consensus *Consensus) UpdateConsensusInformation() Mode {
 
 	consensus.BlockPeriod = 5 * time.Second
 
-	// Enable aggregate sig at epoch 1000 for mainnet, at epoch 53000 for testnet, and always for other nets.
```
LeoHChen, line 316: so we don't need to enable the aggregate sig all together in an epoch?

rlan35: both way works. in a specific epoch. then we have to wait for that epoch and be on hold. Or we just update the first few nodes with aggregate-sig = false (the potential leaders, leader will accept aggregate signatures once the binary is updated, regardless of the flag, the flag is only controlling whether validator will send aggregated signature). and then update the rest with aggregate-sig = true. With this flag. we could have more control on the behavior if needed anyway. 

Which way do you prefer?

cmd/harmony/flags.go
```diff
@@ -776,6 +782,10 @@ func applyConsensusFlags(cmd *cobra.Command, config *harmonyConfig) {
 	} else if cli.IsFlagChanged(cmd, legacyConsensusMinPeersFlag) {
 		config.Consensus.MinPeers = cli.GetIntFlagValue(cmd, legacyConsensusMinPeersFlag)
 	}
+
```
JackyWYX, line 785: It was my fault. In applyConsensusFlags, it would better to be 

```
if config.Consensus == nil && cli.HasFlagsChanged(cmd, consensusValidFlags) {
	cfg := getDefaultConsensusConfigCopy()
	config.Consensus = &cfg
}
```

Would you please update this change for me?

rlan35: done.

JackyWYX (approved):

PR #3442, opened 2020-11-08
Avoid reset state that's resetting committed content.

rlan35:
make precommit's committed message construction in the same thread as the main consensus thread, so that no other thread can clear the state in between.


api/service/syncing/syncing.go
```diff
@@ -1025,7 +1025,6 @@ func (ss *StateSync) SyncLoop(bc *core.BlockChain, worker *worker.Worker, isBeac
 		if err := ss.addConsensusLastMile(bc, consensus); err != nil {
 			utils.Logger().Error().Err(err).Msg("[SYNC] Add consensus last mile")
 		}
-		consensus.SetMode(consensus.UpdateConsensusInformation())
```
rlan35, line 1028: double check that removing this unlocked update is ok.

JackyWYX: We can test it!

JackyWYX (approved):

PR #3448, opened 2020-11-09
Viewchange fix out of sync

LeoHChen:
This PR fixed the gap calculation in getNextLeader.

It also increase the period when calculating the nextView ID to avoid view ID out of sync when nodes start view change at different time.  If the view ID period is the same as the view change timeout, the nodes will always have different view ID.  I have found this issue during another stress test on STN.

Tested on STN, shutdown 2/10 nodes and can still recover consensus after view change.


consensus/view_change.go
```diff
@@ -351,7 +353,7 @@ func (consensus *Consensus) onViewChange(msg *msg_pb.Message) {
 		return
 	}
 
-	if consensus.Decider.IsQuorumAchieved(quorum.ViewChange) {
```
rlan35, line 354: ok, this basically doesn't work before. Good catch.

rlan35 (approved):

PR #3451, opened 2020-11-09
[viewchange] shorten the viewchange timeout to 29s

LeoHChen:

### DIFF
--- a/consensus/config.go
+++ b/consensus/config.go
@@ -4,12 +4,14 @@ import "time"
 
 // timeout constant
 const (
-	// default timeout configuration is shorten to 45 seconds as the consensus is 5s
-	viewChangeTimeout = 45
-	// viewChangeSlot means every 90 seconds, the view change ID will be advanced.
-	// so that the nodes init view change process within the 90 seconds range will
+	// default timeout configuration is shorten to 27 seconds as the consensus is 5s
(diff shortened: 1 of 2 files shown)


consensus/config.go
```diff
@@ -4,12 +4,12 @@ import "time"
...
-	viewChangeTimeout = 45
-	// viewChangeSlot means every 90 seconds, the view change ID will be advanced.
-	// so that the nodes init view change process within the 90 seconds range will
+	// default timeout configuration is shorten to 29 seconds as the consensus is 5s
+	viewChangeTimeout = 29
+	// viewChangeSlot means every 47 seconds, the view change ID will be advanced.
+	// so that the nodes init view change process within the 47 seconds range will
 	// be have the same view change ID
-	viewChangeSlot = 90
+	viewChangeSlot = 47
```
rlan35, line 12: how about 45? just to make it a more formal number for public to remember.    Then ViewChangeTime can be 27 then.

rlan35 (approved):

consensus/config.go
LeoHChen: sure, then it's still the original 45s.

rlan35: yea, only changing the viewChangeTime to be off with the slot timing.

PR #3453, opened 2020-11-10
be more lenient on grace period

rlan35:
if there are more time left for the block. Wait more than 1s up until block time is reached. Just to be more lenient on the slow nodes if it allows.

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -305,7 +305,7 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 
 		go func(viewID uint64) {
 			waitTime := 1000 * time.Millisecond
-			maxWaitTime := time.Until(consensus.NextBlockDue)
+			maxWaitTime := time.Until(consensus.NextBlockDue) - 100*time.Millisecond
 			if maxWaitTime > waitTime {
 				waitTime = maxWaitTime
 			}
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -303,9 +303,15 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 			consensus.preCommitAndPropose(blockObj)
 		}
 
-		consensus.getLogger().Info().Msg("[OnCommit] Starting Grace Period")
 		go func(viewID uint64) {
-			time.Sleep(1000 * time.Millisecond)
+			waitTime := 1000 * time.Millisecond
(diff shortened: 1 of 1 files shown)


consensus/leader.go
```diff
@@ -303,9 +303,15 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 			consensus.preCommitAndPropose(blockObj)
 		}
 
-		consensus.getLogger().Info().Msg("[OnCommit] Starting Grace Period")
 		go func(viewID uint64) {
-			time.Sleep(1000 * time.Millisecond)
+			waitTime := 1000 * time.Millisecond
+			maxWaitTime := time.Until(consensus.NextBlockDue)
+			if maxWaitTime > waitTime {
```
JackyWYX, line 309: From the variable name I read the condition is supposed to be `waitTime > maxWaitTime`?

rlan35: not really, we want the waitTime to be Max(1s, <rest of the block time>)

consensus/leader.go
```diff
@@ -303,9 +303,15 @@ func (consensus *Consensus) onCommit(msg *msg_pb.Message) {
 			consensus.preCommitAndPropose(blockObj)
 		}
 
-		consensus.getLogger().Info().Msg("[OnCommit] Starting Grace Period")
 		go func(viewID uint64) {
-			time.Sleep(1000 * time.Millisecond)
+			waitTime := 1000 * time.Millisecond
+			maxWaitTime := time.Until(consensus.NextBlockDue)
```
LeoHChen, line 308: shouldn't you just recalculate the maxWaitTime based on the confirmed block timestamp? The consensus.NextBlockDue was updated during the block propose time, will it messed up with async block proposal? As it is the only variable, so there is chance that the nextblockdue was updated to the next next blockdue time.

rlan35:
```				// Sleep to wait for the full block time
				consensus.getLogger().Info().Msg("[ConsensusMainLoop] Waiting for Block Time")
				**<-time.After(time.Until(consensus.NextBlockDue))**
				consensus.StartFinalityCount()

				// Update time due for next block
				consensus.NextBlockDue = time.Now().Add(consensus.BlockPeriod)```
it's updated after the last blockTime is passed. So no worry of timing issues.

LeoHChen (approved):

PR #3454, opened 2020-11-11
[prometheus] initial support of prometheus metrics

LeoHChen:
New prometheus end point of 9900 to expose metrics.

/metrics
/goroutinez


api/service/prometheus/service.go
```diff
@@ -0,0 +1,122 @@
...
+	Handler func(http.ResponseWriter, *http.Request)
+}
+
+var (
+	svc = &Service{}
+)
+
+// NewService sets up a new instance for a given address host:port.
+// An empty host will match with any IP so an address like ":19000" is perfectly acceptable.
+func NewService(config nodeconfig.PrometheusServerConfig, additionalHandlers ...Handler) {
```
JackyWYX, line 39: I would suggest to define PrometheusServerConfig in this module instead of in nodeconfig.

api/service/prometheus/service.go
```diff
@@ -0,0 +1,122 @@
...
+	if err := pprof.Lookup("goroutine").WriteTo(w, 2); err != nil {
+		getLogger().Error().Err(err).Msg("Failed to write pprof goroutines")
+	}
+}
+
+// Start the prometheus service.
+func (s *Service) Start() {
+	go func() {
+		// See if the port is already used.
+		conn, err := net.DialTimeout("tcp", s.server.Addr, time.Second)
```
JackyWYX, line 83: 1. Why we do need to check availability before start a server? If the port is already in use, is it true that the error can be returned from ListenAndServe?
2. Consider refactor this into a new function called `checkServerAddrReachable`? This piece of code looks like it is doing a client job at the first sight.

api/service/prometheus/service.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Status checks for any service failure conditions.
+func (s *Service) Status() error {
+	if s.failStatus != nil {
+		return s.failStatus
+	}
+	return nil
+}
+
+func getLogger() *zerolog.Logger {
```
JackyWYX, line 117: I would suggest to use zerolog without pointer (for new code). If you look at the zerolog implementation, all receivers are structs instead of pointers. And structs are preferable to pointers.

api/service/prometheus/service.go
```diff
@@ -0,0 +1,122 @@
...
+	logger := utils.Logger().With().
```
JackyWYX, line 118: Each time you call a getLogger, a new logger struct is created and destructed after use. Do you think it would be better to add logger to struct `Service` instead of this fashion?

cmd/harmony/main.go
```diff
@@ -329,6 +329,13 @@ func setupNodeAndRun(hc harmonyConfig) {
 		HTTPPort:    hc.HTTP.RosettaPort,
 	}
 
+	// Pares Prometheus config
+	nodeConfig.PrometheusServer = nodeconfig.PrometheusServerConfig{
+		HTTPEnabled: hc.HTTP.PrometheusEnabled,
+		HTTPIp:      hc.HTTP.IP,
```
JackyWYX, line 335: Shall we create a different IP field for prometheus service? E.g, user want to open prometheus to 0.0.0.0, but RPC to local 127.0.0.1?

Can we also create the IP flag for rosetta?

api/service/prometheus/service.go
LeoHChen: we use node config for all the configuration, I think that's easier to see and change.

api/service/prometheus/service.go
LeoHChen: I'll look into it.  Any sample usage?

api/service/prometheus/service.go
LeoHChen: agree. that's better.

cmd/harmony/main.go
LeoHChen: make sense.  For rosetta, @Daniel-VDM can add a new variable for the IP.

api/service/prometheus/service.go
LeoHChen: removed to make it simpler

JackyWYX (approved):

PR #3458, opened 2020-11-14
[Rosetta] Make all side effects one transaction

Daniel-VDM:
## Issue

Currently, all 'special' transactions are done on a per-account basis. For example, each validator in the pre-staking era has its own 'special' transaction to indicate its block reward. Some preliminary feedback of our rosetta implementation indicates that this is excess clutter and bloats some blocks with seemingly 'fake' transactions.

## Proposed Solution

This PR consolidates all 'special' operations (renamed as 'side effects') into 1 transaction (at most) per block. This side effect transaction's identifier is `<block_hash>_side_effect`. These transactions can have multiple side effects operations and can affect many different accounts. Going forward, any on-chain side effects should be reflected as an operation in this transaction **only**.

## Docs

Pull request for the doc changes on postman can be found [here](https://harmony.postman.co/collections/8725027-aa1fe329-5052-408c-882f-b1b91aa7a540/pull-requests/dc55f925-d84c-4e34-8cc1-969b610fef57?view=overview&workspace=4f2f1b50-78d3-43c5-8070-5c563fd22e3b) (perms required).


rosetta/services/block.go
```diff
@@ -48,31 +48,34 @@ func (s *BlockAPI) Block(
...
-	if err != nil {
-		return nil, common.NewError(common.CatchAllError, map[string]interface{}{
-			"message": err.Error(),
-		})
-	}
-	prevBlockID = &types.BlockIdentifier{
-		Index: prevBlock.Number().Int64(),
-		Hash:  prevBlock.Hash().String(),
+
+	if blk.NumberU64() == 0 {
```
LeoHChen, line 56: please add a comment here on the special handling.

LeoHChen (approved):
LGTM in general. just a minor comment.

rosetta/services/block.go
Daniel-VDM: IMO comment here would not serve any benefit. The genesis block handling is clear by the nature of the code.

rosetta/services/block_side_effect.go
```diff
@@ -0,0 +1,168 @@
...
+func (s *BlockAPI) getSideEffectTransaction(
+	ctx context.Context, blk *hmytypes.Block,
+) (*types.Transaction, *types.Error) {
+	if !s.containsSideEffectTransaction(ctx, blk) {
+		return nil, common.NewError(common.TransactionNotFoundError, map[string]interface{}{
+			"message": "no side effect transaction found for given block",
+		})
+	}
+
+	var startingOpIndex *int64
```
rlan35, line 78: for normal transactions, is the opIndex same as transaction index in the block?

Daniel-VDM: This operation index corresponds to the operation index of a single transaction according to the rosetta spec. 
Here is [operation documentation](https://www.rosetta-api.org/docs/models/Operation.html).

rlan35 (approved):

PR #3463, opened 2020-11-17
use the correct block hash for newview

rlan35:
use the blockhash from m1payload as it is the canonical one.

Also optimistically continue to propose new block even block commit failed on leader itself.


consensus/consensus_v2.go
```diff
@@ -157,7 +157,6 @@ func (consensus *Consensus) finalCommit() {
 		consensus.getLogger().Err(err).
 			Uint64("beforeCatchupBlockNum", beforeCatchupNum).
 			Msg("[finalCommit] Leader failed to commit the confirmed block")
-		return
```
LeoHChen, line 160: what's the reason to remove the return here?

LeoHChen (commented):
the getPreparedBlock change makes sense, but please mix it with the optimistic block proposal change though.

consensus/consensus_v2.go
rlan35: just in case something weird happened in the leader logic, it should still try propose new blocks. As long as validators still verify the blocks and messages. It doesn't hurt to try move forward optimisitically.

LeoHChen (approved):

LeoHChen:
Please squash and merge this PR. No need to show three commits in the history. Thanks.

PR #3466, opened 2020-11-18
Prometheus PushGateway Support

LeoHChen:
This PR adds Prometheus Push Gateway support to the Harmony node.

All harmony metrics will be pushed to https://gateway.harmony.one.
This is enabled by default for mainnet nodes.

The metrics are still retrievable via local query, `http://localhost:9900/metrics`
Every 1 minute, the metrics will be pushed to the gateway server.

- [x] enable only on mainnet by default
- [x] differentiate harmony node and external nodes
- [x] support local metrics in the meantime

Caveat/Todo:

- [ ] how to protect the gateway server from being attacked
- [ ] add customized push handler to send p2p messages to the network
- [x] add blskey to the metrics


LeoHChen:
https://gateway.harmony.one/#

rlan35 (approved):

rlan35:
looks good. Is this PR tested in some way?

LeoHChen:
> looks good. Is this PR tested in some way?

yes, tested on STN and a local mainnet node.

api/service/prometheus/service.go
```diff
@@ -29,32 +51,83 @@ type Handler struct {
...
+
+	return fmt.Sprintf("%s/%s%d", config.network, node, config.shard)
+}
+
 // NewService sets up a new instance for a given address host:port.
 // An empty host will match with any IP so an address like ":19000" is perfectly acceptable.
-func NewService(config nodeconfig.PrometheusServerConfig, additionalHandlers ...Handler) {
-	if !config.HTTPEnabled {
+func NewService(additionalHandlers ...Handler) {
+	if !config.enabled {
```
JackyWYX, line 81: Do you think enabled would better be decided in upper function calls?

api/service/prometheus/service.go
```diff
@@ -29,32 +51,83 @@ type Handler struct {
...
 
-	utils.Logger().Debug().Int("port", config.HTTPPort).
-		Str("ip", config.HTTPIp).
+	utils.Logger().Debug().Int("port", config.port).
+		Str("ip", config.ip).
 		Msg("Starting Prometheus server")
-	endpoint := fmt.Sprintf("%s:%d", config.HTTPIp, config.HTTPPort)
+	endpoint := fmt.Sprintf("%s:%d", config.ip, config.port)
 	svc.server = &http.Server{Addr: endpoint, Handler: mux}
+	if config.enablePush {
```
JackyWYX, line 109: Do you think it would be better to move this part of function in `svc.Start`?

api/service/prometheus/service.go
```diff
@@ -99,3 +172,43 @@ func (s *Service) Status() error {
 	}
 	return nil
 }
+
+// SetConfig initialize the prometheus config
+func SetConfig(
```
JackyWYX, line 177: Too many arguments. For common practice, we shall not let number of args larger than 4. It would be better to make Config exported and initialized in New.

consensus/metrics.go
```diff
@@ -83,3 +90,15 @@ func (consensus *Consensus) UpdateLeaderMetrics(numCommits float64, blockNum flo
 	consensusCounterVec.With(prometheus.Labels{"consensus": "num_commits"}).Add(numCommits)
 	consensusGaugeVec.With(prometheus.Labels{"consensus": "num_commits"}).Set(numCommits)
 }
+
+func initMetrics() {
+	onceMetrics.Do(func() {
+		prom.PromRegistry().MustRegister(
```
JackyWYX, line 96: Is it possible that this function will probably panic? It is ok if we exit node program if this problem happens?

api/service/prometheus/service.go
```diff
@@ -29,32 +51,83 @@ type Handler struct {
 }
 
 var (
-	svc = &Service{}
+	registryOnce sync.Once
+	svc          = &Service{}
+	config       Config
```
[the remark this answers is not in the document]
LeoHChen: I can pass in the cfg variable.

api/service/prometheus/service.go
```diff
@@ -8,17 +8,39 @@ import (
...
-	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
 	"github.com/harmony-one/harmony/internal/utils"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/client_golang/prometheus/push"
 )
 
+// Config is the config for the prometheus service
+type Config struct {
+	enabled    bool
```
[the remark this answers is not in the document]
LeoHChen: sure

api/service/prometheus/service.go
LeoHChen: not really, better to keep all the config in one place though.

api/service/prometheus/service.go
LeoHChen: sure, I can move.

api/service/prometheus/service.go
LeoHChen: okay, will export all config variable and pass in the variable then.

consensus/metrics.go
LeoHChen: it will panic when duplicated metrics are registered.  In our case, since we have static metrics created here and not dynamic metrics created, so the risk is very low. We should be able to detect this kind of panic in tests though.

JackyWYX (approved):

PR #3467, opened 2020-11-20
change lock scope for validator to avoid premature sync check

rlan35:
Enlarge the lock scope to include the out of sync check, so when committed messages takes too long to process (due to block commit), new messages such as prepared in next round won't trigger "out of sync" check. (tested locally, it works well)

Also update some of the log levels for better debugging.


JackyWYX (approved):
Yes, this PR will surely mitigate the lock issue. However, have we tested the performance on mainnet? I am a little worried that this change will cause much performance change.

rlan35:
This change only affect the validators. And the newly included check are very fast. It's not much concern on performance.

PR #3468, opened 2020-11-21
Add out of order message processing logic

rlan35:
In the case where committed is received before prepared, the current logic will fail and trigger sync'ing. 

Add a logic to handle out of order messages when committed is received before prepared. When the committed is received, it will be cached as a unverified message. Later when prepared is received, it will try lookup whether there is any unverified committed msg and if yes, will proceed and process it.

Tested locally, works perfect. Before this PR, there are lots of "low consensus" messages. With this PR, there is not a single out of sync cases.

Also refactor fbft message processing for easy handling of the unverified committed messages.


consensus/consensus_service.go
```diff
@@ -621,13 +615,11 @@ func (consensus *Consensus) NumSignaturesIncludedInBlock(block *types.Block) uin
 
 // getLogger returns logger for consensus contexts added
 func (consensus *Consensus) getLogger() *zerolog.Logger {
-	logOnce.Do(func() {
```
rlan35, line 624: revert this change because the data inside the logger will be stale if only init once. It's printing garbage data which confuses debugging.

JackyWYX (approved):
LGTM!

PR #3470, opened 2020-11-25
[viewchange] use consensus mutex, remove vc lock

LeoHChen:

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -280,9 +280,6 @@ func (consensus *Consensus) startViewChange() {
 
 // startNewView stops the current view change
 func (consensus *Consensus) startNewView(viewID uint64, newLeaderPriKey *bls.PrivateKeyWrapper, reset bool) error {
-	consensus.mutex.Lock()
-	defer consensus.mutex.Unlock()
-
 	if !consensus.IsViewChangingMode() {
 		return errors.New("not in view changing mode anymore")
 	}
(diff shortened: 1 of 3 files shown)


consensus/view_change.go
```diff
@@ -368,6 +365,9 @@ func (consensus *Consensus) onViewChange(recvMsg *FBFTMessage) {
 	// already checked the length of SenderPubkeys in onViewChangeSanityCheck
 	senderKey := recvMsg.SenderPubkeys[0]
 
+	consensus.mutex.Lock()
```
rlan35, line 368: suggest moving these locks at the top of the func so that the logs about receiving the messages can be printed out in order of processing them. Otherwise, those logs are printed together at first and later on the messages are processed separately after acquiring locks. Basically not much performance concern to include the above in the lock. Consensus does that too.

consensus/view_change.go
```diff
@@ -280,9 +280,6 @@ func (consensus *Consensus) startViewChange() {
 
 // startNewView stops the current view change
 func (consensus *Consensus) startNewView(viewID uint64, newLeaderPriKey *bls.PrivateKeyWrapper, reset bool) error {
-	consensus.mutex.Lock()
```
rlan35, line 283: Deadlock inside selfCommit(). Please double check all cases.

consensus/view_change.go
LeoHChen: moved.

consensus/view_change.go
LeoHChen: fixed.

rlan35 (approved):

PR #3471, opened 2020-11-25
Separate archival configuration for beacon chain and shard chain

LeoHChen:
We used to use one `run.archive` flag to represent the archival mode for both beacon chain and shard chain.

This PR separates the archival configuration for beacon chain and shard chain.
By default, the beacon chain is not in archival mode, which is not necessary even for explorer node on shard chain.

This change can save a lot of disk space on the explorer node for shard chain.

### DIFF
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -15,6 +15,7 @@ var (
 		noStakingFlag,
 		shardIDFlag,
 		isArchiveFlag,
+		isBeaconArchiveFlag,
 		isOfflineFlag,
 		dataDirFlag,
 
@@ -197,6 +198,11 @@ var (
 		Usage:    "run node in archive mode",
 		DefValue: defaultConfig.General.IsArchival,
 	}
+	isBeaconArchiveFlag = cli.BoolFlag{
+		Name:     "run.beacon-archive",
+		Usage:    "beacon chain also in archival mode",
+		DefValue: defaultConfig.General.IsBeaconArchival,
+	}
 	isOfflineFlag = cli.BoolFlag{
 		Name:     "run.offline",
 		Usage:    "run node in offline mode",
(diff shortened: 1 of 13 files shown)


cmd/harmony/flags.go
```diff
@@ -197,6 +198,11 @@ var (
 		Usage:    "run node in archive mode",
 		DefValue: defaultConfig.General.IsArchival,
 	}
+	isBeaconArchiveFlag = cli.BoolFlag{
+		Name:     "run.beacon-archive",
+		Usage:    "beacon chain also in archival mode",
```
JackyWYX, line 203: Suggest edit: "run beacon chain in archive mode"

cmd/harmony/flags.go
```diff
@@ -197,6 +198,11 @@ var (
 		Usage:    "run node in archive mode",
```
JackyWYX, line 198: Suggest edit: "run shard chain in archive mode",

JackyWYX (approved):

node/node.go
```diff
@@ -870,7 +870,7 @@ func New(
 	consensusObj *consensus.Consensus,
 	chainDBFactory shardchain.DBFactory,
 	blacklist map[common.Address]struct{},
-	isArchival bool,
```
rlan35, line 873: suggest rename isArchival (usually a bool) to archivalModes or archivalSettings etc.

internal/configs/node/config.go
```diff
@@ -218,8 +219,19 @@ func SetNetworkType(networkType NetworkType) {
...
+func (conf *ConfigType) SetArchival(bcArchival, archival bool) {
+	if conf.isArchival == nil {
+		conf.isArchival = make(map[uint32]bool)
+	}
+	conf.isArchival[shard.BeaconChainShardID] = bcArchival
+	conf.isArchival[conf.ShardID] = archival
+}
+
+// IsArchival return the isArchival map
+func (conf *ConfigType) IsArchival() map[uint32]bool {
```
rlan35, line 233: suggest rename isArchival (usually a bool) to ArchivalModes or ArchivalSettings etc.

internal/shardchain/shardchains.go
```diff
@@ -35,7 +35,7 @@ type CollectionImpl struct {
 	engine       engine.Engine
 	mtx          sync.Mutex
 	pool         map[uint32]*core.BlockChain
-	disableCache bool
```
rlan35, line 38: init this in NewCollection() to avoid nil map in any cases

PR #3472, opened 2020-12-01
Change blocks per epoch for testnet.

rlan35:
Update 3s finality to 2s finality.

Add epoch transition logic to turn blocks per epoch from 38 to 8192.

Enable the above at epoch 73000 for testnet.


JackyWYX (approved):
LGTM

PR #3480, opened 2020-12-07
node.sh v3.0

LeoHChen:
This PR re-write the node.sh.

The new node.sh greatly simplified the node.sh.
The code is mainly borrowed from prysm.sh (eth2 client).

The new node.sh is only used to

* download latest harmony binary or specified version
* download from github release page
* use harmony public key to do verify the pgp signature of the binary

`539910B02B5BAC637F6615C9799ACE3A9B83DF60` is the fingerprint of the harmony key.

To sign the binary, using the following command,
```
gpg -u 539910B02B5BAC637F6615C9799ACE3A9B83DF60 --output harmony-amd64.sig --detach-sig harmony-amd64
```

To verify the key, using the following command,
```
gpg -u 539910B02B5BAC637F6615C9799ACE3A9B83DF60 --verify harmony-amd64.sig harmony-amd64
```


.hmy/pgp_keys.asc
```diff
@@ -0,0 +1,29 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
JackyWYX, line 1: Shall we move this file to other repository?

scripts/node.sh
```diff
@@ -1,838 +1,210 @@
...
-   mkdir -p "${outdir}"
-   if _curl_check_exist $url; then
-      curl --progress-bar -Sf $url -o "${outdir}/$filename" || return $?
-      return 0
-   else
-      msg "failed to find/download $url"
-      return 1
-   fi
-}
+case "$1" in
```
JackyWYX, line 179: Is it reasonable to have node.sh in another repository?

JackyWYX (approved):

.hmy/pgp_keys.asc
LeoHChen: this is the public key, better reside in the main repo for ppl to trust, IMO.

scripts/node.sh
LeoHChen: better stay within the main repo IMO.

LeoHChen:
@yelllowsin , @sophoah , can you prepare some document about the usage of the new node.sh?  For validators that are still using the old node.sh, this is a breaking change.

PR #3482, opened 2020-12-09
[btcrelay] init version of btcrelay

LeoHChen:
* prototype of btcrelay
* build support
* add LICENSE file


Daniel-VDM:
I don’t see why this needs to be in the main repo. 

IMHO should be in it’s own repo or in the harmony-one/onebtc repo. 

According to the interlay spec, a staked relayer is suppose to run their own btc node and report/verify block headers to/from the relayer contract and possibly also interact with the relayer’s governance mechanism. If this is what this code is for (with is unclear by the PR message btw) then it def needs to be somewhere else. 

If the code is just for some tool to fetch btc block info then this should also not be in the main repo and prob be made as a script.

Daniel-VDM (changes requested):
Ref comment above

gupadhyaya:
@LeoHChen why this needs to be here? let's have all modules and tools under: https://github.com/harmony-one/onebtc

LeoHChen:
okay, I'll move it to onebtc repo. The staked relayer will run the btc node, and the relay tool is a client of the btc node, it needs to fetch block info from btc node, and feed the data to the relayer contract.

PR #3483, opened 2020-12-10
[node] graceful shutdown leader

LeoHChen:
when leader is doing consensus, wait until it finish the consensus
before shutdown the leader.


cmd/harmony/main.go
```diff
@@ -296,16 +296,26 @@ func setupNodeAndRun(hc harmonyConfig) {
...
 				const msg = "Got %s signal. Gracefully shutting down...\n"
-				utils.Logger().Printf(msg, sig)
-				fmt.Printf(msg, sig)
+				fmt.Fprintf(os.Stderr, msg, sig)
+				if node.Consensus.Mode() == consensus.Normal {
+					phase := node.Consensus.GetConsensusPhase()
+					utils.Logger().Warn().Str("phase", phase).Msg("[shutdown] commit phase has to wait")
+					maxWait := time.Now().Add(time.Second * 2) // wait up to 2 seconds in commit phase
+					for time.Now().Before(maxWait) &&
+						node.Consensus.GetConsensusPhase() == "Commit" {
```
rlan35, line 310: this asynchronous approach can be unreliable for bad timing. Better add synchronous signal to stop new block proposal too (otherwise, new consensus could also be started when the node is killed after the delay).
image

rlan35: This way, can first send signal to stop the new block proposal and then wait for more than 2 seconds (like 5s) for the current consensus to fully finish.

LeoHChen: good way

rlan35 (approved):

PR #3485, opened 2020-12-11
Fix inconsistent block proposal and verification

rlan35:
It happened that when a leader is restarted, two part of the code triggered new block proposal and started announce phase, which messed up the consensus states (block content and block hash). This caused the second consensus to fail due to malformed prepared message (containing the second proposed block in last round):

```{"level":"warn","myBlock":1497743,"myViewID":1507399,"phase":"Prepare","mode":"Normal","MsgBlockNum":1497743,"blockNum":1497742,"caller":"/home/lc/go/src/github.com/harmony-one/harmony/consensus/checks.go:129","time":"2020-12-10T05:32:48.149366601Z","message":"[OnPrepared] BlockNum not match"}```

Then view change started with this bad prepared message, but it succeeded because verifyBlock() method doesn't take care of whether the block is at the tip of not. The second proposed block is a valid block itself, it's just it's a old block which should be ignored.

Then view change succeeded with M1 message and all validators signed commit message on this old block. Then the committed message got sent to all validators who override the commit sig for the canonical block, causing all validators to have this bad committed signatures.

This PR fixes this scenario by:

1. avoid proposing duplicate blocks at the same height for leader.
2. add block height check on the VerifyBlock method.
3. add block existence check by hash when overriding commit signature.
4. do not send committed message if failing to commit the block.

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -353,6 +353,11 @@ func (consensus *Consensus) Start(
 					Uint64("MsgBlockNum", newBlock.NumberU64()).
 					Msg("[ConsensusMainLoop] Received Proposed New Block!")
 
+				if newBlock.NumberU64() < consensus.blockNum {
+					consensus.getLogger().Info().Uint64("newBlockNum", newBlock.NumberU64()).
+						Msg("[ConsensusMainLoop] received old block, abort")
+					continue
+				}
 				// Sleep to wait for the full block time
 				consensus.getLogger().Info().Msg("[ConsensusMainLoop] Waiting for Block Time")
 				<-time.After(time.Until(consensus.NextBlockDue))
@@ -553,6 +558,7 @@ func (consensus *Consensus) preCommitAndPropose(blk *types.Block) error {
 
 		if _, err := consensus.Blockchain.InsertChain([]*types.Block{blk}, true); err != nil {
 			consensus.getLogger().Error().Err(err).Msg("[preCommitAndPropose] Failed to add block to chain")
+			return
 		}
 
 		// if leader successfully finalizes the block, send committed message to validators
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -354,7 +354,7 @@ func (consensus *Consensus) Start(
 					Msg("[ConsensusMainLoop] Received Proposed New Block!")
 
 				if newBlock.NumberU64() < consensus.blockNum {
-					consensus.getLogger().Info().Uint64("newBlockNum", newBlock.NumberU64()).
+					consensus.getLogger().Warn().Uint64("newBlockNum", newBlock.NumberU64()).
 						Msg("[ConsensusMainLoop] received old block, abort")
 					continue
 				}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: consensus/consensus_v2.go [hunk 1]
```diff
@@ -353,6 +353,11 @@ func (consensus *Consensus) Start(
 					Uint64("MsgBlockNum", newBlock.NumberU64()).
 					Msg("[ConsensusMainLoop] Received Proposed New Block!")
 
+				if newBlock.NumberU64() < consensus.blockNum {
+					consensus.getLogger().Info().Uint64("newBlockNum", newBlock.NumberU64()).
```
LeoHChen, line 357: make it an error/warn message

consensus/validator.go
```diff
@@ -299,7 +299,8 @@ func (consensus *Consensus) onCommitted(recvMsg *FBFTMessage) {
 	// has more signatures and if yes, override the old data.
 	// Otherwise, simply write the commit signature in db.
 	commitSigBitmap, err := consensus.Blockchain.ReadCommitSig(blockObj.NumberU64())
-	if err == nil && len(commitSigBitmap) == len(recvMsg.Payload) {
+	blk := consensus.Blockchain.GetBlockByHash(blockObj.Hash())
```
LeoHChen, line 302: can you comment on why this check is necessary?

node/node_newblock.go
```diff
@@ -84,6 +84,11 @@ func (node *Node) WaitForConsensusReadyV2(readySignal chan consensus.ProposalTyp
 					newBlock, err := node.ProposeNewBlock(newCommitSigsChan)
 
 					if err == nil {
+						if blk, ok := node.proposedBlock[newBlock.NumberU64()]; ok {
```
LeoHChen, line 87: what to do with this situation when the same number is proposed?  how to recover from it?

node/node_newblock.go
```diff
@@ -84,6 +84,11 @@ func (node *Node) WaitForConsensusReadyV2(readySignal chan consensus.ProposalTyp
 					newBlock, err := node.ProposeNewBlock(newCommitSigsChan)
```
LeoHChen, line 84: also, since we don't want to propose the same block number, does it make sense to do the checking in ProposeNewBlock function?

consensus/consensus_v2.go
rlan35: Done

consensus/validator.go
rlan35: Done.

node/node_newblock.go
rlan35: We don't want to propose the same block, if it is tried, we just return here. If if a duplicate block is indeed proposed, the other checks about the block number will prevent issues downstream.

node/node_newblock.go
rlan35: proposeNewBlock is more like a utility, it's for the caller to determine how to use it. So adding the check in the caller is fine. There is not much performance concern here. And it will be cumbersome to add check inside proposeNewBlock too due to its lack of easy access to the proposed block number.

LeoHChen (approved):
Please run shard restart test of this PR on STN for a few times to make sure the shard won't stuck.

PR #3487, opened 2020-12-12
[pubsub] change to validation ignore when handling non-consensus topic

JackyWYX:
## Issue

When handling non-consensus-bound topic, should use validation ignore instead of validation reject.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -659,7 +659,7 @@ func (node *Node) Start() error {
 						errChan <- withError{
 							errors.WithStack(errConsensusMessageOnUnexpectedTopic), msg,
 						}
-						return libp2p_pubsub.ValidationReject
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -659,7 +659,7 @@ func (node *Node) Start() error {
 						errChan <- withError{
 							errors.WithStack(errConsensusMessageOnUnexpectedTopic), msg,
 						}
-						return libp2p_pubsub.ValidationReject
+						return libp2p_pubsub.ValidationIgnore
```
LeoHChen, line 662: why? did you spot any issue on the network? how many invalid_bound metrics you found on mainnet?

JackyWYX: Ah, I see, my bad. So side chain's `node.Start` does not handle beacon chain message. I see. Closing

PR #3490, opened 2021-01-13
Add ethereum transaction and abstract out the transaction interface

rlan35:
Add ethereum transaction struct and abstract out common transaction interface.


core/types/eth_transaction.go
```diff
@@ -0,0 +1,341 @@
...
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/rlp"
+)
+
+//go:generate gencodec -type ethTxdata -field-override ethTxdataMarshaling -out gen_eth_tx_json.go
+
+// Shard0ChainID to be reserved unique chain ID for eth compatible chains.
+const Shard0ChainID = 0
```
Daniel-VDM, line 36: I'm a little confused by this (could be the name?), but is this just a correcting factor to have the chain ID map to a shard ID?

Some clarity on how the other shards would be represented (if at all) would be nice. My guess is that we just bump the eth tx's chain ID by `n` to get to shard `n`?

JackyWYX, line 36: Suggest to rename to `Shard0ChainIDDiff` to avoid ambiguity

JackyWYX (approved):

core/types/gen_tx_json.go
```diff
@@ -20,7 +20,7 @@ func (t txdata) MarshalJSON() ([]byte, error) {
 		Price        *hexutil.Big    `json:"gasPrice"   gencodec:"required"`
 		GasLimit     hexutil.Uint64  `json:"gas"        gencodec:"required"`
 		ShardID      uint32          `json:"shardID"    gencodec:"required"`
-		ToShardID    uint32          `json:"toShardID"`
+		ToShardID    uint32          `json:"toShardID"  gencodec:"required"`
```
Daniel-VDM, line 23: Looks like this file was desynced from the source file (changed at commit 02fb5cb7b2de48f546281332c3b83a251d0507d7 on 8/8/19) for a long time (maybe before Cx was a thing?). So wondering if there are any backward compatibility issues? Like when replaying old transactions?

Daniel-VDM (approved):
Seems fine, just some things to consider in my comments.

core/types/gen_tx_json.go
rlan35: should be fine. gencodec won't affect transaction data format.

core/types/eth_transaction.go
rlan35: Yes, that's right. will write more comment once we finalized this part of the logic. Now it's just a placeholder.

rlan35: Will do once we have final ChainID reserved.

PR #3491, opened 2021-01-15
Add new block structure for eth compatible transactions

rlan35:
Add new block body with ethereum transaction in it. 

Register the new block structure as blockV3.

So far these codes are still no-op since it's not added as a fork logic yet.


JackyWYX (approved):
LGTM

PR #3500, opened 2021-01-22
Resolves issues on eth comaptible txns

rlan35:
1. Add Eth<Network>Shard0ChainID in chainConfig so we have the base number for shard calculation or beacon chain reference.
2. Move UpdateEthChainIDByShard before Node() initialization so the worker struct can be initialized with the correct config with correct chainID.
3. Add sanity checks in txn ordering code to avoid nil pointer crash.
4. Fix EIP155Signer.Equal(), we only need to compare the chainID inside the signer as each signer is initialized differently.
5. Add ethSigner in SendRawTransaction for ethereum txns.
6. Correctly initialize beacon worker with the eth shard0 chainID.
7. Add forking logic at block proposal and rpc level for ethereum compatible txn support.

Tested locally with metamask for both shard 0 and shard 1. All worked well.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -277,12 +277,13 @@ func setupNodeAndRun(hc harmonyConfig) {
 		fmt.Fprintf(os.Stderr, "ERROR cannot configure node: %s\n", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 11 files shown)


SebastianJ (approved):
Tested on s0/s1 localnet, working well.  LGTM.

PR #3510, opened 2021-01-28
Add eth txn hash lookup entry

rlan35:
Ethereum toolings such as Metamask will lookup txns based on the hash/iD of txn in the original ethereum format. So we need to also index the eth txns by their "ethereum" hash in our node.

Tested with Metamask in localnet. I was able to use the txn hash in MetaMask to look it up in hmy rpc

### DIFF
--- a/core/rawdb/accessors_indexes.go
+++ b/core/rawdb/accessors_indexes.go
@@ -58,6 +58,14 @@ func WriteBlockTxLookUpEntries(db DatabaseWriter, block *types.Block) error {
 		if err := db.Put(key, val); err != nil {
 			return err
 		}
+		if tx.IsEthCompatible() {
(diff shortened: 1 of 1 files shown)


JackyWYX (approved):
LGTM

rlan35:
> LGTM. My only concern is that for hmy transactions converted to EthTransaction, are these external toolings also checks the signature thus cause further errors?

what do you mean by causing further error? This is just to support looking txn up by the original ethereum txn hash format.

JackyWYX:
> > LGTM. My only concern is that for hmy transactions converted to EthTransaction, are these external toolings also checks the signature thus cause further errors?
> 
> what do you mean by causing further error? This is just to support looking txn up by the original ethereum txn hash format.

Ah right, my bad. :)

PR #3512, opened 2021-01-28
Use and return the Ethereum tx hash for Ethereum txs

rlan35:
Since we store ethereum txn using the translated format of hmy txn on the chain. The txn hash of the two types of txns are not the same. We need streamline the logic to make sure to the developers/users, when they are sending eth txns, they are always handling the txns by the eth version of the txn hash.

Fix the txn lookup checking where txn hash and the lookup hash doesn't match.

Also update all the txn hash usage to use the appropriate version of the hash.


SebastianJ (commented):
Fixes the lookup issues LGTM

JackyWYX (approved):
Looks good to me :)

PR #3518, opened 2021-01-30
[viewchange] always use first participant as base of leader pubkey

LeoHChen:
when view change happended at the first block of the new epoch

### DIFF
--- a/consensus/view_change.go
+++ b/consensus/view_change.go
@@ -184,10 +184,12 @@ func (consensus *Consensus) getNextLeaderKey(viewID uint64) *bls.PublicKeyWrappe
 			// viewchange happened at the first block of new epoch
 			// use the LeaderPubKey as the base of the next leader
 			// as we shouldn't use lastLeader from coinbase as the base.
-			// The LeaderPubKey should be updated to the index 0 of the committee
+			// The LeaderPubKey should be updated to the node of index 0 of the committee
+			// so, when validator joined the view change process later in the epoch block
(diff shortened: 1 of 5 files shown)


rlan35 (approved):

consensus/view_change.go
```diff
@@ -184,10 +184,12 @@ func (consensus *Consensus) getNextLeaderKey(viewID uint64) *bls.PublicKeyWrappe
 			// viewchange happened at the first block of new epoch
 			// use the LeaderPubKey as the base of the next leader
 			// as we shouldn't use lastLeader from coinbase as the base.
-			// The LeaderPubKey should be updated to the index 0 of the committee
+			// The LeaderPubKey should be updated to the node of index 0 of the committee
+			// so, when validator joined the view change process later in the epoch block
+			// it can still sync with other validators.
 			if curHeader.IsLastBlockInEpoch() {
 				consensus.getLogger().Info().Msg("[getNextLeaderKey] view change in the first block of new epoch")
-				lastLeaderPubKey = consensus.LeaderPubKey
```
rlan35, line 190: The change looks good. But curious how did "consensus.LeaderPubKey" have a wrong Leader?

LeoHChen: it's not wrong, in the old code, it changes every time a new view change started in the first block of the epoch.  If it is not the epoch block, it will always find the canonical leader from blockchain, thus no issue.

PR #3533, opened 2021-02-08
Graceful shutdown

JackyWYX:
## Issue

1. Refactored service manager with cleaner interface. 
2. Add prometheus to the service manager.
3. Graceful shutdown of the services (including consensus).
4. Some code refactor regarding consensus graceful shutdown.

## UPDATE 02/13/2021

1. Allow multiple interrupt signal to force quit when node program stuck in the shutdown process.
2. Allow node handle errors when starting / stopping services.
3. Resolve the initial 30s delay before the block proposal service can be stopped.

## Tests done

Passed local tests, all nodes can be shutdown gracefully.

### DIFF
--- a/api/service/blockproposal/service.go
+++ b/api/service/blockproposal/service.go
@@ -22,38 +22,26 @@ func New(readySignal chan consensus.ProposalType, commitSigsChan chan []byte, wa
 	return &Service{readySignal: readySignal, commitSigsChan: commitSigsChan, waitForConsensusReady: waitForConsensusReady}
 }
 
-// StartService starts block proposal service.
-func (s *Service) StartService() {
+// Start starts block proposal service.
+func (s *Service) Start() error {
 	s.stopChan = make(chan struct{})
 	s.stoppedChan = make(chan struct{})
 
-	s.Init()
-	s.Run(s.stopChan, s.stoppedChan)
+	s.run(s.stopChan, s.stoppedChan)
+	return nil
 }
 
-// Init initializes block proposal service.
-func (s *Service) Init() {
-}
-
-// Run runs block proposal.
-func (s *Service) Run(stopChan chan struct{}, stoppedChan chan struct{}) {
+func (s *Service) run(stopChan chan struct{}, stoppedChan chan struct{}) {
 	s.waitForConsensusReady(s.readySignal, s.commitSigsChan, s.stopChan, s.stoppedChan)
 }
 
-// StopService stops block proposal service.
-func (s *Service) StopService() {
+// Stop stops block proposal service.
+func (s *Service) Stop() error {
 	utils.Logger().Info().Msg("Stopping block proposal service.")
 	s.stopChan <- struct{}{}
 	<-s.stoppedChan
 	utils.Logger().Info().Msg("Role conversion stopped.")
-}
-
-// NotifyService notify service
-func (s *Service) NotifyService(params map[string]interface{}) {}
-
-// SetMessageChan sets up message channel to service.
-func (s *Service) SetMessageChan(messageChan chan *msg_pb.Message) {
-	s.messageChan = messageChan
+	return nil
 }
 
 // APIs for the services.
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -300,34 +300,6 @@ func setupNodeAndRun(hc harmonyConfig) {
 		utils.Logger().Warn().Err(err).Msg("Check Local Time Accuracy Error")
 	}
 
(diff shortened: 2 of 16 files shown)


api/service/blockproposal/service.go
```diff
@@ -22,38 +22,26 @@ func New(readySignal chan consensus.ProposalType, commitSigsChan chan []byte, wa
...
 	s.stoppedChan = make(chan struct{})
 
-	s.Init()
-	s.Run(s.stopChan, s.stoppedChan)
-}
-
-// Init initializes block proposal service.
-func (s *Service) Init() {
+	s.run(s.stopChan, s.stoppedChan)
+	return nil
```
LeoHChen, line 31: if there is no error to return, just remove the `error` return type.

api/service/blockproposal/service.go
```diff
@@ -22,38 +22,26 @@ func New(readySignal chan consensus.ProposalType, commitSigsChan chan []byte, wa
...
 	utils.Logger().Info().Msg("Role conversion stopped.")
-}
-
-// NotifyService notify service
-func (s *Service) NotifyService(params map[string]interface{}) {}
-
-// SetMessageChan sets up message channel to service.
-func (s *Service) SetMessageChan(messageChan chan *msg_pb.Message) {
-	s.messageChan = messageChan
+	return nil
```
LeoHChen, line 44: same as above, if always return nil, just remove return `error`.

node/node.go
```diff
@@ -798,39 +811,48 @@ func (node *Node) Start() error {
...
+			for {
+				select {
+				case m := <-msgChanNode:
+					ctx, cancel := context.WithTimeout(node.psCtx, 10*time.Second)
+					msg := m
+					go func() {
+						defer cancel()
+						// AUDIT: DANGEROUS!! Some message might be dropped here!!!
+						// TODO: prevent using TryAcquire
+						if semNode.TryAcquire(1) {
```
LeoHChen, line 823: TryAcquire here is to prevent spamming attack.

cmd/harmony/main.go
```diff
@@ -309,10 +309,9 @@ func setupNodeAndRun(hc harmonyConfig) {
 				utils.Logger().Warn().Str("signal", sig.String()).Msg("Gracefully shutting down...")
 				const msg = "Got %s signal. Gracefully shutting down...\n"
 				fmt.Fprintf(os.Stderr, msg, sig)
-				// stop block proposal service for leader
-				if node.Consensus.IsLeader() {
-					node.ServiceManager().StopService(service.BlockProposal)
-				}
```
LeoHChen, line 315: why removed this part?

LeoHChen (commented):
looks okay to me in general, a few comments

api/service/blockproposal/service.go
JackyWYX: This `Service` type implements `Service` interface, which requires the signature to return an error. This error is returned for sync protocol which will be added in some future PR.

api/service/blockproposal/service.go
JackyWYX: Just interface requirement.

node/node.go
JackyWYX: I See. Will remove the comment. BTW, do you think this may be the cause for low-performance machine cannot process all pub-sub messages?

cmd/harmony/main.go
JackyWYX: Since `serviceManager.Stop()` in `node.ShutDown()` will stop all services registered, including BlockProposal.

cmd/harmony/main.go
```diff
@@ -309,10 +309,9 @@ func setupNodeAndRun(hc harmonyConfig) {
...
+
+				currentNode.ShutDown()
```
rlan35, line 313: ShutDown() will exit() the program. So have to be at the end of the block.

rlan35: or probably you forgot to delete the code below

JackyWYX: The code shall be deleted. Thanks

rlan35 (approved):

node/node.go
LeoHChen: not just low performance, spamming p2p messages will cause OOM

LeoHChen (approved):
please rebase and just merge this PR. There should be no regression of this PR on our testnet.

JackyWYX:
Added several changes. Please review this PR again. Thanks. @rlan35 @LeoHChen

rlan35 (approved):

PR #3545, opened 2021-02-23
[testnet] re-enable 40% ext voting power on shard3

LeoHChen:


consensus/votepower/roster.go
```diff
@@ -185,7 +185,7 @@ func Compute(subComm *shard.Committee, epoch *big.Int) (*Roster, error) {
 	// Testnet incident recovery
 	// Make harmony nodes having 70% voting power for epoch 73314
 	if nodeconfig.GetDefaultConfig().GetNetworkType() == nodeconfig.Testnet && epoch.Cmp(big.NewInt(73305)) >= 0 &&
-		epoch.Cmp(big.NewInt(74330)) <= 0 {
+		epoch.Cmp(big.NewInt(74330)) <= 0 && subComm.ShardID != uint32(3) {
```
rlan35, line 188: need to write the logic so that shard 3 turnup happens at a specific epoch. Otherwise, it will happen at the network update time which is rather random and it will cause sync failures for new nodes.

PR #3535, opened 2021-02-08
Stream sync protocol

JackyWYX:
## Important notice

Since the size of this PR is too large for review, will split this PR to a set of small PRs for review. The breakdown PR will follow a bottom-up sequence for review.

1. [Fix p2p base stream read/write](https://github.com/harmony-one/harmony/pull/3552)
2. [Stream manager](https://github.com/harmony-one/harmony/pull/3554)
3. [Rate limiter](https://github.com/harmony-one/harmony/pull/3553)
4. [Protobuf message](https://github.com/harmony-one/harmony/pull/3561)
5. [Request manager](https://github.com/harmony-one/harmony/pull/3560)

## Description

1. Added stream sync protocol running on P2P Stream.
2. Added downloader module to manage the synchronization task.
3. Stream manager module to manage over streams.
4. Request manager module to manage over stream requests.
5. Flags, commands and misc.

## Tests

1. 1-1 sync pair test - Done
2. 1-3 sync test - Done
3. 1-3 with 1 bad peer, 2 good peer (negative test)  (Almost finished)
4. Large scale test (STN test) - Not started.

## Current known issues

1. In very rare cases got error when establishing stream connection `failed to negotiate security protocol: error reading handshake message: noise: message is too short` (https://github.com/libp2p/go-libp2p-noise/issues/70) This issue is already fixed in libp2p version v0.13.0. Will upgrade libp2p after this release (frozen code).

PR #3552, opened 2021-02-26
[stream] Fix p2p base stream read/write

JackyWYX:
## Description

This is a breakdown PR for https://github.com/harmony-one/harmony/pull/3535. 

* Fixed the read/write method of the BaseStream.
* Updated some BaseStream method: `Close` and `ResetOnClose`.

### Some comments

BaseStream is the wrapper of `libp2p_network.Stream` and defines the basic read/write rules. 

#### Write:

* First write 4 bytes as the size of this message (uint32).
* Write the message payload to the stream.
* The message shall not exceed 20MB in size.

#### Read:

* First read 4 bytes as the size of this message (uint32).
* Read the bytes of exactly the size read.
* The message size shall not exceed 20MB in size.

#### Why 20MB?

Currently, we have an assumption that block size is smaller 2MB. Since the pub-sub message should contain the full block in announce, 2MB is the current upper limit of the block size. 

```
// p2p/host.go:70
MaxMessageSize = 1 << 21
```

For running stream protocol, the upper cap of request `getBlocksByNumbers` and `getBlocksByHashes` contains at most 10 block data, so the upper limit of single message is set to 2MB * 10 = 20MB.

### DIFF
--- a/p2p/stream/types/stream.go
+++ b/p2p/stream/types/stream.go
@@ -1,10 +1,14 @@
 package sttypes
 
 import (
-	"io/ioutil"
+	"bufio"
+	"encoding/binary"
+	"fmt"
+	"io"
 	"sync"
 
 	libp2p_network "github.com/libp2p/go-libp2p-core/network"
+	"github.com/pkg/errors"
 )
 
 // Stream is the interface for streams implemented in each service.
@@ -15,12 +19,14 @@ type Stream interface {
 	ProtoSpec() (ProtoSpec, error)
 	WriteBytes([]byte) error
 	ReadBytes() ([]byte, error)
-	Close() error // Make sure streams can handle multiple calls of Close
+	Close() error
+	ResetOnClose() error
 }
 
 // BaseStream is the wrapper around
 type BaseStream struct {
 	raw libp2p_network.Stream
+	rw  *bufio.ReadWriter
 
 	// parse protocol spec fields
 	spec     ProtoSpec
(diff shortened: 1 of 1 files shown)


p2p/stream/types/stream.go
```diff
@@ -59,20 +67,64 @@ func (st *BaseStream) ProtoSpec() (ProtoSpec, error) {
 
 // Close close the stream on both sides.
 func (st *BaseStream) Close() error {
-	return st.raw.Reset()
+	return st.raw.Close()
 }
 
-// WriteBytes write the bytes to the stream
+const (
+	maxMsgBytes = 20 * 1024 * 1024 // 20MB
```
LeoHChen, line 74: have you tested the performance of max 20MB data in streaming? Do we really need to support this huge size?

LeoHChen (approved):
looks good in general. Just wondering about the performance of 20M message.

p2p/stream/types/stream.go
```diff
@@ -41,7 +49,7 @@ type StreamID string
 
 // Meta return the StreamID of the stream
 func (st *BaseStream) ID() StreamID {
-	return StreamID(st.raw.ID())
+	return StreamID(st.raw.Conn().ID())
```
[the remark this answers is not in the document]
JackyWYX: Because previous one was wrong. I am using this ID to avoid setting up duplicate stream to a node in stream manager. But libp2p can setup multiple streams with different streamID under one connection. Thus stream.ID() cannot serve well to deduplicate streams.

For example, connection ID is 

```
Qwxxxxxxxxxxx
```

And stream IDs can be

```
Qwxxxxxxxxxxx-yyyyyyy
Qwxxxxxxxxxxx-zzzzzzz
```

But since I only want to setup one stream for one peer, thus the connection id is used instead of stream ID.

p2p/stream/types/stream.go
JackyWYX: This is currently the theoretical upper limit. As I said in description, for stream protocol, message is capped at 10 blocks, which implies there is at most 20MB data. For real cases, it's less than 200KB, based on the block size.

rlan35 (approved):

PR #3553, opened 2021-02-26
[stream] added rate limiter

JackyWYX:
## Description

This is a small break down PR for https://github.com/harmony-one/harmony/pull/3535. This PR sits on top of https://github.com/harmony-one/harmony/pull/3554

1. Added a ratelimiter module to limit the rate of stream requests.
2. The ratelimiter is limiting: 1. global requests per second 2. request per second for each stream.
3. In upper function call, global rate is set at 50/s, single stream rate is set at 10/s

### Update 03/01/2021

Added stream removed handling according to comment. Now this PR is on top of [stream manager](https://github.com/harmony-one/harmony/pull/3554)

### DIFF
--- a/p2p/stream/common/ratelimiter/ratelimiter.go
+++ b/p2p/stream/common/ratelimiter/ratelimiter.go
@@ -3,7 +3,11 @@ package ratelimiter
 import (
 	"sync"
 
+	"github.com/ethereum/go-ethereum/event"
+
+	"github.com/harmony-one/harmony/p2p/stream/common/streammanager"
 	sttypes "github.com/harmony-one/harmony/p2p/stream/types"
+	p2ptypes "github.com/harmony-one/harmony/p2p/types"
 	"go.uber.org/ratelimit"
 )
 
@@ -12,6 +16,8 @@ import (
 // for consensus on DDoS attacks.
 type RateLimiter interface {
 	LimitRequest(stid sttypes.StreamID)
+
+	p2ptypes.LifeCycle
 }
 
 // rateLimiter is the implementation of RateLimiter.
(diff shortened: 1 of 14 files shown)


p2p/stream/common/ratelimiter/ratelimiter.go
```diff
@@ -0,0 +1,60 @@
...
+		globalLimiter: ratelimit.New(global),
+
+		streamRate: single,
+		limiters:   make(map[sttypes.StreamID]ratelimit.Limiter),
+	}
+}
+
+func (rl *rateLimiter) LimitRequest(stid sttypes.StreamID) {
+	rl.lock.RLock()
+	limiter, ok := rl.limiters[stid]
```
LeoHChen, line 43: limiter is per stream id, will you release the limiter from the map?

LeoHChen (commented):
please consider to add prometheus metrics support

p2p/stream/common/ratelimiter/ratelimiter.go
JackyWYX: Good point. Let me do it.

JackyWYX: After looking into the prometheus for a while, I think I will add prometheus support after the merge of PRs for the following reason:
1. The prometheus metrics will not effect the key feature. It's a better-to-have feature.
2. Prometheus metrics can be added in a separate PR which will make the changes more clean.
So I created [an issue](https://github.com/harmony-one/harmony/issues/3555) to be executed after the review of this entire merge.

JackyWYX:
> please consider to add prometheus metrics support

After looking into the prometheus for a while, I think I will add prometheus support after the merge of PRs for the following reason:

The prometheus metrics will not effect the key feature. It's a better-to-have feature.
Prometheus metrics can be added in a separate PR which will make the changes more clean.
So I created an issue to be executed after the review of this entire merge.

rlan35 (approved):

p2p/stream/common/ratelimiter/ratelimiter.go
JackyWYX: Resolved. Please review again.

JackyWYX:
@rlan35 @LeoHChen  Added removing stream handling to avoid possible memory leak. Please review again.

rlan35 (approved):

PR #3554, opened 2021-02-26
[stream] added stream manager module

JackyWYX:
## Description

This is a breakdown PR of https://github.com/harmony-one/harmony/pull/3535 and is on top of https://github.com/harmony-one/harmony/pull/3552.

1. Added stream manager module to manage over streams of a stream protocol.

Stream manager do the following job:

1. add a new stream.
2. remove a stream.
3. discover and connect new streams when the number of streams is below threshold.
4. emit stream events to inform other modules.
5. reset all streams on close.

### DIFF
--- a/p2p/stream/common/streammanager/config.go
+++ b/p2p/stream/common/streammanager/config.go
@@ -1,0 +1,25 @@
+package streammanager
+
+import "time"
+
+const (
+	// checkInterval is the default interval for checking stream number. If the stream
+	// number is smaller than softLoCap, an active discover through DHT will be triggered.
+	checkInterval = 30 * time.Second
+	// discTimeout is the timeout for one batch of discovery
+	discTimeout = 10 * time.Second
+	// connectTimeout is the timeout for setting up a stream with a discovered peer
+	connectTimeout = 60 * time.Second
+)
+
+// Config is the config for stream manager
+type Config struct {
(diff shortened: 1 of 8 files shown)


LeoHChen (approved):
LGTM in general.

p2p/stream/common/streammanager/config.go
```diff
@@ -0,0 +1,25 @@
+package streammanager
+
+import "time"
+
+const (
+	// checkInterval is the default interval for checking stream number. If the stream
+	// number is smaller than softLoCap, an active discover through DHT will be triggered.
+	checkInterval = 30 * time.Second
```
rlan35, line 8: how are these numbers chosen?

JackyWYX: By estimation without solid evidence. Basically this is just like estimating the timeout to do a HTTP request. The only trick here is to make discTimeout less than checkInterval to avoid simultaneous host/client stream setup. Do you have other references that these numbers can be analyzed upon?

JackyWYX: We can probably observe the timeout by adding prometheus metrics and change to a more solid value after launch. For now I think the number is good to start with. Do you other references that can be helped to determine these number's values?

rlan35 (approved):

Issue #3557, opened 2021-03-01
[STN] During the view change, node panic and exited.


JackyWYX:
### Describe the bug

When testing on stress net, following panic log is observed during view change and cause the node being terminated. 

```
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]: panic: assignment to entry in nil map
Feb 28 23:15:32 ip-172-26-21-217 systemd[1]: harmony.service: Main process exited, code=exited, status=2/INVALIDARGUMENT
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]: goroutine 1543 [running]:
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]: github.com/harmony-one/harmony/consensus.(*viewChange).InitPayload(0xc000de7200, 0xc00007ceb0, 0x895b, 0x801e, 0xc0002b8710, 0x1, 0x1, 0x60, 0x1f)
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]:         /home/yx/go/src/github.com/harmony-one/harmony/consensus/view_change_construct.go:424 +0x786
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]: github.com/harmony-one/harmony/consensus.(*Consensus).startViewChange(0xc000281b00)
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]:         /home/yx/go/src/github.com/harmony-one/harmony/consensus/view_change.go:264 +0x56b
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]: github.com/harmony-one/harmony/consensus.(*Consensus).Start.func1(0xc000281b00, 0xc000bda780, 0xc0017a1ce0, 0xc000bda5a0, 0xc0017a1c80)
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]:         /home/yx/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:334 +0x5db
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]: created by github.com/harmony-one/harmony/consensus.(*Consensus).Start
Feb 28 23:15:32 ip-172-26-21-217 harmony[917611]:         /home/yx/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:287 +0x67
```

LeoHChen:
Is your code based on the latest code?  This error was seen before I remember.

JackyWYX:
[...]

PR #3565, opened 2021-03-03
[viewchange] protect view change struct during reset

LeoHChen:
This is a further fix of the following issue.
https://github.com/harmony-one/harmony/issues/3557

### DIFF
--- a/consensus/consensus_service.go
+++ b/consensus/consensus_service.go
@@ -101,7 +101,10 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 	consensus.UpdateBitmaps()
 	consensus.ResetState()
 
-	consensus.ResetViewChangeState()
+	// do not reset view change state if it is in view changing mode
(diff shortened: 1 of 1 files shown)


JackyWYX (approved):

consensus/consensus_service.go
```diff
@@ -101,6 +101,9 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys []bls_cosi.PublicKeyWrapper
 	consensus.UpdateBitmaps()
 	consensus.ResetState()
 
+	consensus.mutex.Lock()
```
rlan35, line 104: This will cause deadlocks in many places where UpdateConsensusInformation() is used.

LeoHChen: I see.

That was why I introduced smaller lock on view change data struct before.  Now we have one big lock on big function like UpdateConsensusInformation, and the smaller function like ResetViewChangeState can't be locked.  I have to rework a bit on this.

rlan35 (approved):

Issue #3570, opened 2021-03-08
nil pointer crash found on testnet nodes


LeoHChen:
**Describe the bug**
Found the following crash on testnet shard0 nodes.
```
Mar 08 16:30:10 PTN_S0_1 harmony[16506]: panic: assignment to entry in nil map
Mar 08 16:30:10 PTN_S0_1 harmony[16506]: goroutine 22776 [running]:
Mar 08 16:30:10 PTN_S0_1 harmony[16506]: github.com/harmony-one/harmony/consensus.(*viewChange).ProcessViewChangeMsg(0xc000f362a0, 0xc0001285a0, 0x1eb2c00, 0xc00459a700, 0xc00ac18380, 0x2, 0x2)
Mar 08 16:30:10 PTN_S0_1 harmony[16506]:         /mnt/jenkins/workspace/harmony-release/harmony/consensus/view_change_construct.go:288 +0xa72
Mar 08 16:30:10 PTN_S0_1 harmony[16506]: github.com/harmony-one/harmony/consensus.(*Consensus).onViewChange(0xc0001686c0, 0xc00ac18380)
Mar 08 16:30:10 PTN_S0_1 harmony[16506]:         /mnt/jenkins/workspace/harmony-release/harmony/consensus/view_change.go:395 +0x8dd
Mar 08 16:30:10 PTN_S0_1 harmony[16506]: github.com/harmony-one/harmony/consensus.(*Consensus).HandleMessageUpdate(0xc0001686c0, 0x1e950e0, 0xc0141fc0c0, 0xc00c908d20, 0xc00ec40ba0, 0x1b, 0x15)
Mar 08 16:30:10 PTN_S0_1 harmony[16506]:         /mnt/jenkins/workspace/harmony-release/harmony/consensus/consensus_v2.go:117 +0x2fa
Mar 08 16:30:10 PTN_S0_1 harmony[16506]: github.com/harmony-one/harmony/node.(*Node).StartPubSub.func2.1(0xc00ba96380, 0xc00b7d3180, 0xc0141fc000, 0xc0001402c0, 0x1e950e0, 0xc0141fc0c0, 0x1, 0xc00c840460, 0xc00c908d20, 0x0, ...)
Mar 08 16:30:10 PTN_S0_1 harmony[16506]:         /mnt/jenkins/workspace/harmony-release/harmony/node/node.go:786 +0x4ae
[...]

PR #3571, opened 2021-03-08
Fix crashes found during view change

LeoHChen:
This PR fixes the crash found recent test node shard0 down event.

https://github.com/harmony-one/harmony/issues/3570

Added a few more defensive protection to guard un-initialized map.


rlan35 (approved):

rlan35:
Isn't the last race condition prevented? so potentially other race condition?

JackyWYX (approved):

LeoHChen:
> Isn't the last race condition prevented? so potentially other race condition?

during the fix of the last crash, I didn't do a thorough check of all the map access.

PR #3583, opened 2021-03-12
[Stream]  sync protocol

JackyWYX:
## Description

This is a break down pr of https://github.com/harmony-one/harmony/pull/3535 and is rebased on https://github.com/harmony-one/harmony/pull/3582.

* Added sync stream handling code logic.
* Both client and server side logic is in function `syncStream.run`
* Client side logic is called from upper function calls (downloader module) through methods defined in `client.go`

### Commit signature and bitmaps

Currently, block commit sig and bitmap written in db is defined as the concatenation of `signature` + `bitmap`. But blocks long time ago seems having a different data written in db. So currently the way work around to get the correct commit sig and bitmap is through:

1. Try to get the next header.
2. If the next header exist, get signature and bitmaps from the next header and write this data to client.
3. Else read the signature and bitmap from db and write to client.

Please let me know if there is any problem with this approach. Thanks.


JackyWYX:
Following is the data I collected for block number 1 during testing:

https://gist.github.com/JackyWYX/adba7bf16bccce479045600660d8454b

Had no idea how to interpret the commit sig written in DB.

rlan35:
Always use the sig data from next block's header if the current block is not the latest block.

JackyWYX:
> Always use the sig data from next block's header if the current block is not the latest block.

Cool! Yea, this is exactly the logic implemented. Please kindly review. Thanks.

p2p/stream/protocols/sync/chain.go
```diff
@@ -0,0 +1,169 @@
...
+
+	targetBN := ch.schedule.EpochLastBlock(epoch - 1)
+	res.Header = ch.chain.GetHeaderByNumber(targetBN)
+	if res.Header == nil {
+		// we still don't have the given epoch
+		return res, nil
+	}
+	epochBI := new(big.Int).SetUint64(epoch)
+	if ch.chain.Config().IsPreStaking(epochBI) {
+		// For epoch before preStaking, only hash is stored in header
```
rlan35, line 158: what does this mean? only hash is stored in header?

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,396 @@
...
+	blocks, err := req.getBlocksFromResponse(resp)
+	if err != nil {
+		return nil, stid, err
+	}
+	return blocks, stid, nil
+}
+
+// GetEpochState get the epoch block from querying the remote node running sync stream protocol.
+// Currently, this method is only supported by beacon syncer.
+// Note: use this after epoch chain is implemented.
```
rlan35, line 43: This is only used for sync'ing epoch chain right? it shouldn't be needed more normal blocks sync

p2p/stream/protocols/sync/chain.go
JackyWYX: My bad. This was experimental code that wasn't verified. Let me remove these.

p2p/stream/protocols/sync/client.go
JackyWYX: Right. Let me remove this.

rlan35 (approved):

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	blocks := make([]*types.Block, 0, len(blockBytes))
+	for i, bb := range blockBytes {
+		var block *types.Block
+		if err := rlp.DecodeBytes(bb, &block); err != nil {
+			return nil, errors.Wrap(err, "[GetBlocksByHashesResponse]")
+		}
+		if block != nil {
+			block.SetCurrentCommitSig(sigs[i])
+		}
```
LeoHChen, line 340: what happens if block is nil?  seems you keep adding it?

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+	blocks := make([]*types.Block, 0, len(blockBytes))
+	for i, bb := range blockBytes {
+		var block *types.Block
+		if err := rlp.DecodeBytes(bb, &block); err != nil {
+			return nil, errors.Wrap(err, "[GetBlocksByNumResponse]")
+		}
+		if block != nil {
+			block.SetCurrentCommitSig(sigs[i])
+		}
+		blocks = append(blocks, block)
```
LeoHChen, line 152: if block is nil, keep adding it?

p2p/stream/protocols/sync/const.go
```diff
@@ -0,0 +1,17 @@
+package sync
+
+import "time"
+
+const (
+	// GetBlockHashesAmountCap is the cap of GetBlockHashes reqeust
+	GetBlockHashesAmountCap = 50
+
+	// GetBlocksByNumAmountCap is the cap of request of a single GetBlocksByNum request
+	GetBlocksByNumAmountCap = 10
```
LeoHChen, line 10: can you add some additional comments here why 10?  I remember there was a max size limit.

p2p/stream/protocols/sync/protocol.go
```diff
@@ -0,0 +1,260 @@
...
+	smConfig := streammanager.Config{
+		SoftLoCap: config.SmSoftLowCap,
+		HardLoCap: config.SmHardLowCap,
+		HiCap:     config.SmHiCap,
+		DiscBatch: config.DiscBatch,
+	}
+	sp.sm = streammanager.NewStreamManager(sp.ProtoID(), config.Host, config.Discovery,
+		sp.HandleStream, smConfig)
+
+	sp.rl = ratelimiter.NewRateLimiter(sp.sm, 50, 10)
```
LeoHChen, line 94: comment on the 50, 10? or make them constants?

p2p/stream/protocols/sync/protocol.go
```diff
@@ -0,0 +1,260 @@
...
+		if e != nil {
+			p.logger.Warn().Err(e).Str("protocol", string(pid)).
+				Msg("cannot advertise sync protocol")
+			continue
+		}
+		if nextWait == 0 || nextWait > w {
+			nextWait = w
+		}
+	}
+	if nextWait < minAdvertiseInterval {
```
LeoHChen, line 196: advertise every minute?

p2p/stream/protocols/sync/stream.go
```diff
@@ -0,0 +1,348 @@
...
+		if err != nil {
+			return nil, err
+		}
+		blocksBytes = append(blocksBytes, bb)
+
+		var sig []byte
+		if block != nil {
+			sig = block.GetCurrentCommitSig()
+		}
+		sigs = append(sigs, sig)
```
LeoHChen, line 305: what if block is nil?

LeoHChen (commented):
Sorry for the late review. LGTM in general, just a few points need clarification.  Please also add metrics later when handling each message.

p2p/stream/protocols/sync/chain.go
JackyWYX: Code removed

p2p/stream/protocols/sync/client.go
JackyWYX: Code removed

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+)
+
+// GetBlocksByNumber do getBlocksByNumberRequest through sync stream protocol.
+// Return the block as result, target stream id, and error
+func (p *Protocol) GetBlocksByNumber(ctx context.Context, bns []uint64, opts ...Option) ([]*types.Block, sttypes.StreamID, error) {
+	if len(bns) > GetBlocksByNumAmountCap {
+		return nil, "", fmt.Errorf("number of blocks exceed cap of %v", GetBlocksByNumAmountCap)
+	}
+
+	req := newGetBlocksByNumberRequest(bns)
```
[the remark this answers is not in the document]
JackyWYX: You are right. Better do a sanity check and not send out the nil request if empty. Logics added.

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+	resp, stid, err := p.rm.DoRequest(ctx, req, opts...)
+	if err != nil {
+		// At this point, error can be context canceled, context timed out, or waiting queue
+		// is already full.
+		return nil, stid, err
+	}
+
+	// Parse and return blocks
+	blocks, err := req.getBlocksFromResponse(resp)
```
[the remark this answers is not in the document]
JackyWYX: The function is blocked at 

```
resp, stid, err := p.rm.DoRequest(ctx, req, opts...)
```

And this step is purely extracting the result from the response. Whether it is a sync response is also check within this function. Please take a look.

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	blockBytes, sigs, err := req.parseBlockBytesAndSigs(sResp)
+	if err != nil {
+		return nil, err
+	}
+	blocks := make([]*types.Block, 0, len(blockBytes))
+	for i, bb := range blockBytes {
+		var block *types.Block
+		if err := rlp.DecodeBytes(bb, &block); err != nil {
+			return nil, errors.Wrap(err, "[GetBlocksByNumResponse]")
```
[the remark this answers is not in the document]
JackyWYX: 1. Why the error message is not clear? What is your expectation of the error message? (Currently, the error message is like `[GetBlocksByNumResponse] rlp: some rlp error message)
2. Yes and no. All blocks need to be correctly RLP decoded here. If decode errored here, I won't return any blocks but the criminal stream ID accused of delivering this false message. But the overall validation of the block message and error handling logic is executed in downloader module.

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+	sResp, ok := resp.(*syncResponse)
+	if !ok || sResp == nil {
+		return nil, errors.New("not sync response")
+	}
+	if errResp := sResp.pb.GetErrorResponse(); errResp != nil {
+		return nil, errors.New(errResp.Error)
+	}
+	bhResp := sResp.pb.GetGetBlocksByHashesResponse()
+	if bhResp == nil {
+		return nil, errors.New("response not GetBlocksByHashes")
```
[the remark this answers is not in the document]
JackyWYX: What is your expectation of this error message? 

For example, the remote node delivers GetBlocksByNumber response for GetBlocksByHashes request. So this error reports the response type is not correct. Actually, this error message shall never be triggered if the remote node is honest.

p2p/stream/protocols/sync/client.go
```diff
@@ -0,0 +1,344 @@
...
+		sigs       = bhResp.CommitSig
+	)
+	if len(blockBytes) != len(sigs) {
+		return nil, fmt.Errorf("sig size not expected: %v / %v", len(sigs), len(blockBytes))
+	}
+	blocks := make([]*types.Block, 0, len(blockBytes))
+	for i, bb := range blockBytes {
+		var block *types.Block
+		if err := rlp.DecodeBytes(bb, &block); err != nil {
+			return nil, errors.Wrap(err, "[GetBlocksByHashesResponse]")
```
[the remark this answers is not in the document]
JackyWYX: What is your expectation of error message?

p2p/stream/protocols/sync/client.go
JackyWYX: Yes. Nil block is added as a placeholder. The following equation is strictly defined in verification

```
len(req.bns) == len(resp.blockBytes) == len(resp.sigs)
```

p2p/stream/protocols/sync/client.go
JackyWYX: Yes. nil is a placeholder equivalent to say "I don't have the block message of the given hash/number".

p2p/stream/protocols/sync/const.go
JackyWYX: Right, it is better to add some comments here. BTW, max size limit is actually calculated based on this number, which is  just an arbitrary number I believe reasonable.

p2p/stream/protocols/sync/protocol.go
JackyWYX: Let me make them constants. Thanks

p2p/stream/protocols/sync/protocol.go
JackyWYX: Nope. This is actually the backoff logic to handle error case.

If Advertise returns an error, the nextWait will return 0. If this is the case, the node will go into an infinite loop which is a black hole for CPU resource. So I added this minimum to retry in 1 min.

For normal non-error cases, nextWait will always be 6 hours from libp2p.

p2p/stream/protocols/sync/stream.go
JackyWYX: nil is a placeholder to say: "I don't have the block"

PR #3586, opened 2021-03-13
[Stream] added stream downloader to consensus

JackyWYX:
## Description

This is a break down of https://github.com/harmony-one/harmony/pull/3535, and is on top of https://github.com/harmony-one/harmony/pull/3587.

Added stream downloader to consensus. Now consensus will trigger new downloader logic only if downloader is setup.

### DIFF
--- a/consensus/downloader.go
+++ b/consensus/downloader.go
@@ -1,0 +1,131 @@
+package consensus
+
+import (
+	"github.com/ethereum/go-ethereum/event"
+	"github.com/harmony-one/harmony/core/types"
+	"github.com/pkg/errors"
+)
+
+// downloader is the adapter interface for downloader.Downloader, which is used for
+// 1. Subscribe download finished event to help syncing to the latest block.
+// 2. Trigger the downloader to start working
+type downloader interface {
+	SubscribeDownloadFinished(ch chan struct{}) event.Subscription
+	SubscribeDownloadStarted(ch chan struct{}) event.Subscription
+	DownloadAsync()
+}
+
+// Set downloader set the downloader of the shard to consensus
+// TODO: It will be better to move this to consensus.New and register consensus as a service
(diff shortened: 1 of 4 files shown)


consensus/downloader.go
```diff
@@ -0,0 +1,131 @@
...
+	dh.finishedSub.Unsubscribe()
+}
+
+func (dh *downloadHelper) downloadStartedLoop() {
+	for {
+		select {
+		case <-dh.startedCh:
+			dh.c.BlocksNotSynchronized()
+
+		case err := <-dh.finishedSub.Err():
```
rlan35, line 68: should this be startedCh.Err()?

consensus/downloader.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return nil
+}
+
+func (consensus *Consensus) spinUpStateSync() {
+	if consensus.dHelper != nil {
+		consensus.dHelper.d.DownloadAsync()
+		consensus.current.SetMode(Syncing)
+		for _, v := range consensus.consensusTimeout {
+			v.Stop()
```
rlan35, line 115: will the timer be reset when the sync is finished?

consensus/downloader.go
JackyWYX: Yea, you got me.

consensus/downloader.go
JackyWYX: The timer is restarted in `consensus.Start()` 

```
// consensus/consensus_v2.go:346
case <-consensus.syncReadyChan:
   ...
```

I currently did not refactor this part of code because:

1. Both legacy sync and stream sync touches this part logic. 
2. I want to minimize the interface change related to consensus when adding new feature.

Probably this piece of code can be better refactored when DNS sync is fully removed. Added TODO. Thanks for bring this up.

rlan35 (approved):

PR #3588, opened 2021-03-13
[Stream] Enable downloader module and updated main.go flags.

JackyWYX:
## Description

This is another break down of https://github.com/harmony-one/harmony/pull/3535, and is rebased on https://github.com/harmony-one/harmony/pull/3586.

NOTE: This PR will turn the stream sync protocol on and disable the legacy DNS sync client by default, and will require some extensive tests.

1. Added sync related flags.
2. Register sync service (downloader) to service manager.
3. Added RPC and Rosetta supports.

## Tests

1. Local test passed.
2. Testing with fault injected code on stn. (ON-GOING, code base https://github.com/JackyWYX/harmony/tree/stn_stream_main)

### DIFF
--- a/api/service/manager.go
+++ b/api/service/manager.go
@@ -21,6 +21,7 @@ const (
 	BlockProposal
 	NetworkInfo
 	Prometheus
+	Synchronize
 )
 
 func (t Type) String() string {
@@ -37,6 +38,8 @@ func (t Type) String() string {
 		return "NetworkInfo"
 	case Prometheus:
 		return "Prometheus"
+	case Synchronize:
+		return "Synchronize"
 	default:
 		return "Unknown"
 	}
--- a/api/service/manager.go
+++ b/api/service/manager.go
@@ -82,6 +82,11 @@ func (m *Manager) GetServices() []Service {
 	return m.services
 }
 
+// GetService get the specified service
+func (m *Manager) GetService(t Type) Service {
+	return m.serviceMap[t]
+}
+
 // StartServices run all registered services. If one of the starting service returns
 // an error, closing all started services.
 func (m *Manager) StartServices() (err error) {
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -16,6 +16,9 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/harmony-one/harmony/api/service/synchronize"
+	"github.com/harmony-one/harmony/hmy/downloader"
+
 	ethCommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/log"
 	"github.com/harmony-one/bls/ffi/go/bls"
@@ -204,6 +207,7 @@ func applyRootFlags(cmd *cobra.Command, config *harmonyConfig) {
 	applyDevnetFlags(cmd, config)
 	applyRevertFlags(cmd, config)
 	applyPrometheusFlags(cmd, config)
+	applySyncFlags(cmd, config)
 }
 
 func setupNodeLog(config harmonyConfig) {
@@ -304,12 +308,6 @@ func setupNodeAndRun(hc harmonyConfig) {
 		WSPort:       hc.WS.Port,
 		DebugEnabled: hc.RPCOpt.DebugEnabled,
 	}
-	if nodeConfig.ShardID != shard.BeaconChainShardID {
-		utils.Logger().Info().
-			Uint32("shardID", currentNode.Blockchain().ShardID()).
-			Uint32("shardID", nodeConfig.ShardID).Msg("SupportBeaconSyncing")
-		currentNode.SupportBeaconSyncing()
-	}
 
 	// Parse rosetta config
 	nodeConfig.RosettaServer = nodeconfig.RosettaServerConfig{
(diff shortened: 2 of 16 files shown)


cmd/harmony/main.go
```diff
@@ -699,6 +708,37 @@ func setupPrometheusService(node *node.Node, hc harmonyConfig, sid uint32) {
 	node.RegisterService(service.Prometheus, p)
 }
 
+func setupSyncService(node *node.Node, host p2p.Host, hc harmonyConfig) {
+	blockchains := []*core.BlockChain{
```
rlan35, line 712: what if it's the beacon chain, then there is only one chain that needs sync

cmd/harmony/main.go
```diff
@@ -699,6 +708,37 @@ func setupPrometheusService(node *node.Node, hc harmonyConfig, sid uint32) {
...
+		SmHardLowCap: hc.Sync.DiscHardLowCap,
+		SmHiCap:      hc.Sync.DiscHighCap,
+		SmDiscBatch:  hc.Sync.DiscBatch,
+	}
+	// If we are running side chain, we will need to do some extra works for beacon
+	// sync
+	if node.NodeConfig.ShardID != 0 {
+		dConfig.BHConfig = &downloader.BeaconHelperConfig{
+			BlockC:     node.BeaconBlockChannel,
+			InsertHook: node.BeaconSyncHook,
```
rlan35, line 731: where is this Hook used? why we need this hook?

cmd/harmony/main.go
JackyWYX: Blockchains will be de-duplicated in `NewDownloaders` (hmy/downloader/downloaders.go:14), so finally there will only be one downloader running. 

But it seems this piece of code is causing confusion, let me refactor it a little bit.

cmd/harmony/main.go
JackyWYX: This hook is to broadcast the beacon crosslink if the node is a side-chain leader. It is called in `downloader.BeaconHelper`.

This hook is actually making cross-chain transaction possible. Without it, the cross-shard transaction will be blocked for lacking knowledge of crosslink data. 

And this is the refactor of the old DNS sync logic:

```
// DoBeaconSyncing update received beaconchain blocks and downloads missing beacon chain blocks
func (node *Node) DoBeaconSyncing() {
	if node.NodeConfig.IsOffline {
		return
	}

	go func(node *Node) {
		for beaconBlock := range node.BeaconBlockChannel {
			if node.beaconSync != nil {
				err := node.beaconSync.UpdateBlockAndStatus(
					beaconBlock, node.Beaconchain(), node.BeaconWorker, true,
				)
				if err != nil {
				} else if node.Consensus.IsLeader() {
					// Only leader broadcast crosslink to avoid spamming p2p
					node.BroadcastCrossLink()  <---------------------------------------NOTE THIS LINE
				}
			}
		}
	}(node)
```

rlan35: cross-shard transaction doesn't depends on cross-links. Cross-shard txns only depends on the correct committee information to verify the signatures.

JackyWYX: Actually, the local rpc tests fails at CX transactions before I added this piece of code. Let me dig deeper into the mechanism then.

rlan35: Probably the communication between shard is broken. Though crosslink is not needed. The cross-shard beacon block broadcasting is still needed for other shard to receive the new committee info in time.

JackyWYX: Cross-shard beacon block broadcasting is fine as far as I see. You can start running localnet and check the side chain's epoch changes swiftly as same as beacon epoch, which means beacon committee is synced OK.

rlan35: keep in mind that local cx txn testing is flaky. So it may not be a problem in the code.

api/service/manager.go
```diff
@@ -82,6 +85,11 @@ func (m *Manager) GetServices() []Service {
 	return m.services
 }
 
+// GetService get the specified service
+func (m *Manager) GetService(t Type) Service {
+	return m.serviceMap[t]
```
rlan35, line 90: any lock needed?

cmd/harmony/main.go
```diff
@@ -304,12 +308,6 @@ func setupNodeAndRun(hc harmonyConfig) {
 		WSPort:       hc.WS.Port,
 		DebugEnabled: hc.RPCOpt.DebugEnabled,
 	}
-	if nodeConfig.ShardID != shard.BeaconChainShardID {
-		utils.Logger().Info().
-			Uint32("shardID", currentNode.Blockchain().ShardID()).
-			Uint32("shardID", nodeConfig.ShardID).Msg("SupportBeaconSyncing")
-		currentNode.SupportBeaconSyncing()
```
rlan35, line 311: will legacy sync still work after this is deleted?

api/service/manager.go
JackyWYX: I don't think we will need lock now since the serviceMap is fixed.

cmd/harmony/main.go
JackyWYX: No. So I added this back in https://github.com/harmony-one/harmony/pull/3595. Please take a look

JackyWYX:
Closing. Since all codes in this PR is included in #3595. Please review #3595 instead.

Issue #3589, opened 2021-03-15
STN Stream sync takes CPU after long run 


JackyWYX:
## Describe the bug

After running for 3~4 days of error injected [stream sync code](https://github.com/JackyWYX/harmony/tree/stn_stream_main), nodes are taking a lot of CPU resources.

JackyWYX:
Current observation is that node prints a lot of log in:

```
        // p2p/stream/common/requestmanager/requestmanager.go:291
	st, err := rm.pickAvailableStream(req)
	if err != nil {
		rm.logger.Debug().Err(err).Str("request", req.String()).Msg("No available streams.")
		rm.addRequestToWaitings(req, reqPriorityHigh)
		return nil, nil
	}```

**Comment by JackyWYX** [comment]
But sync are doing well since message about sync completed with inserted blocks is seen in logs, which means it is doing fine.

So here is two guesses that cause node to be stuck in an endless loop:

1. Streams data in request manager is corrupted (The worst case).
2. request manager is doing request more than its capacity. Need to add some rest to the loop function.

We will see the next time this problem happens.

**Comment by JackyWYX** [comment]
Message is observed again but the CPU is not increasing:

```
{"level":"debug","module":"request manager","error":"no more available streams","request":"REQUEST [GetBlocksByHashes: d1e57d5ddaaa816670c8aa9c0e4ed12c490e515af2a4915c4100af238a205ddc]","caller":"/home/yx/go/src/github.com/harmony-one/harmony/p2p/stream/common/requestmanager/requestmanager.go:293","time":"2021-03-16T04:34:26.97720335Z","message":"No available streams."}
```

So CPU bound is caused by some other goroutines instead of request manager. Will do a pprof after CPU spiking is observed again.

JackyWYX:
Propose this fix to address the concern and fix the debug message constantly printing (actually a bug fixed).

> 1. Streams data in request manager is corrupted (The worst case). `<-FIXED`
> 2. request manager is doing request more than its capacity. Need to add some rest to the loop function. `<-ACTUALLY A BUG`

JackyWYX:
Next thing to do is to merge this fix PR, and then inject some error code in stream protocol, deploy and test it out in testnet. Hopefully this unstable codebase will regenerate the CPU issue really quick.

JackyWYX:
Already observed increase of CPU usage on STN machine `54.201.29.75` at 30%~40%. As for contrast, the CPU usage is only at 2~5% for initial start.

[...]

PR #3591, opened 2021-03-16
[Stream] Stream requestmanager fix

JackyWYX:
## Issue

Fix some request manager concerns and bug (https://github.com/harmony-one/harmony/issues/3589).

1. requestmanager's streams is updated with refresh instead of follow events. So that the streams in request manager is not corrupted.
2. Requests will no longer stay in waitings queue once the request is being canceled. So that there shall not be error message with 

```
{"level":"debug","module":"request manager","error":"no more available streams","request":"REQUEST [GetBlocksByHashes: d1e57d5ddaaa816670c8aa9c0e4ed12c490e515af2a4915c4100af238a205ddc]","caller":"/home/yx/go/src/github.com/harmony-one/harmony/p2p/stream/common/requestmanager/requestmanager.go:293","time":"2021-03-16T04:39:32.377309174Z","message":"No available streams."}
```

## Tests

Unit tests added for each scenario.


rlan35 (approved):

p2p/stream/common/requestmanager/requestmanager.go
```diff
@@ -349,26 +348,49 @@ func (rm *requestManager) pickAvailableStream(req *request) (*stream, error) {
...
 	rm.lock.Lock()
 	defer rm.lock.Unlock()
 
-	if _, ok := rm.streams[st.ID()]; !ok {
-		rm.streams[st.ID()] = &stream{Stream: st}
-		rm.available[st.ID()] = struct{}{}
+	added, removed := checkStreamUpdates(rm.streams, rm.sm.GetStreams())
+
+	for _, st := range added {
+		rm.logger.Info().Str("streamID", string(st.ID())).Msg("add new stream")
```
LeoHChen, line 358: may change to debug level , IMO.

p2p/stream/common/requestmanager/requestmanager.go
```diff
@@ -349,26 +348,49 @@ func (rm *requestManager) pickAvailableStream(req *request) (*stream, error) {
...
+		rm.addNewStream(st)
+	}
+	for _, st := range removed {
+		rm.logger.Info().Str("streamID", string(st.ID())).Msg("remove stream")
```
LeoHChen, line 362: may change to debug level , IMO.

p2p/stream/common/requestmanager/requestmanager.go
JackyWYX: These log messages may appear a lot in an unstable network (which is the case of error injected in stn) where peer connection is unstable thus a lot of connect/disconnect operation is executed. For stable network, this shall not appear a lot. Please let me know if this log appears a lot in a stable network.

p2p/stream/common/requestmanager/requestmanager.go
JackyWYX: These log messages may appear a lot in an unstable network (which is the case of error injected in stn) where peer connection is unstable thus a lot of connect/disconnect operation is executed. For stable network, this shall not appear a lot. Please let me know if this log appears a lot in a stable network.

Issue #3590, opened 2021-03-15
[Stream] stream protocol dry run mode


JackyWYX:
## Problem/limitation At Hand

For stream sync protocol deployment on mainnet, we probably need TWO steps for safer production launch:

1. As an internal update, all nodes running new stream sync server, but use legacy DNS sync for client. Deploy another "probe" node running full stream sync logic. Thus the downloader logic can be verified on the probe node while not effecting existing logic.
2. As the testing with probe node finished, we can update all nodes to adopt the new downloader logic and publicly release later.

## Feature request

1. Add hidden flag `--sync.dryrun` to dry-run sync protocol.

LeoHChen:
As to the first step, we may even enable the stream sync server mode on all nodes *by default* including internal and external nodes.

We can deploy one or a few internal nodes to enable downloader logic, and engage a few external validators in downloading mode, to have a better coverage.

* not a dry-run IMO, maybe `--sync.downloader` option

JackyWYX:
> As to the first step, we may even enable the stream sync server mode on all nodes _by default_ including internal and external nodes.
> 
> We can deploy one or a few internal nodes to enable downloader logic, and engage a few external validators in downloading mode, to have a better coverage.
> 
> * not a dry-run IMO, maybe `--sync.downloader` option

Sure, great thanks for your advice.

JackyWYX:
Created branch https://github.com/harmony-one/harmony/pull/3595 and added --sync.downloader feature. 

The code is now merged in stn_stream_main branch and deployed to stn. Note: Unstable consensus feature is disabled since for legacy DNS client, the consensus might be forever stuck since DNS nodes could possibly contain lagging data.

Also ran a local node connecting to stressnet running downloader module. And it shows the node run syncing successfully.

```
./harmony --datadir stn --sync.downloader --sync.legacy.client=false --network stn
``` 

We can set stn to use downloader module by default after stn runs good till tomorrow.

JackyWYX:
Also noted that local node fails sync at block number 14628. This might be caused by some nasty historical "forks" at stressnet. Better clean start stn sometime later.

JackyWYX:
The flags and corresponding updates are already merged in this PR: https://github.com/harmony-one/harmony/pull/3595. Clsoing

PR #3595, opened 2021-03-17
[Stream] Stream downloader flag

JackyWYX:
## Description

Added downloader service in main.go. Also added all flags for syncing. After this PR,

1. By default, nodes will use DNS sync on mainnet and testnet.
2. By default, nodes will run stream sync as a server but will not run as a client.
3. To start a node running only stream sync logic, run with command as follows or update harmony.conf.

```
./harmony --sync.downloader --sync.legacy.client=false
```

Running nodes with downloader as default will be enabled after mainnet internal tests.

## Downloader flags

According to https://github.com/harmony-one/harmony/issues/3590, we will need a binary running 

1. DNS client
2. DNS server
3. Stream server
4. NO downloader (stream client)

So in this PR, flag `--sync.downloader` is added to instruct the node not to start downloader module.

## Test

local test done.

For STN tests, nodes runs good for both --sync.downloader is enabled or not.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -282,6 +282,7 @@ func setupNodeAndRun(hc harmonyConfig) {
 	currentNode := setupConsensusAndNode(hc, nodeConfig)
 	nodeconfig.GetDefaultConfig().ShardID = nodeConfig.ShardID
 	nodeconfig.GetDefaultConfig().IsOffline = nodeConfig.IsOffline
+	nodeconfig.GetDefaultConfig().Downloader = nodeConfig.Downloader
 
 	// Check NTP configuration
 	accurate, err := ntp.CheckLocalTimeAccurate(nodeConfig.NtpServer)
@@ -538,6 +539,7 @@ func createGlobalConfig(hc harmonyConfig) (*nodeconfig.ConfigType, error) {
 	nodeConfig.SetShardID(initialAccounts[0].ShardID) // sets shard ID
 	nodeConfig.SetArchival(hc.General.IsBeaconArchival, hc.General.IsArchival)
 	nodeConfig.IsOffline = hc.General.IsOffline
+	nodeConfig.Downloader = hc.Sync.Downloader
 
 	// P2P private key is used for secure message transfer between p2p nodes.
 	nodeConfig.P2PPriKey, _, err = utils.LoadKeyFromFile(hc.P2P.KeyFile)
@@ -715,6 +717,7 @@ func setupSyncService(node *node.Node, host p2p.Host, hc harmonyConfig) {
 	}
 
 	dConfig := downloader.Config{
+		ServerOnly:   !hc.Sync.Downloader,
 		Network:      nodeconfig.NetworkType(hc.Network.NetworkType),
 		Concurrency:  hc.Sync.Concurrency,
 		MinStreams:   hc.Sync.MinPeers,
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -712,7 +712,7 @@ func setupPrometheusService(node *node.Node, hc harmonyConfig, sid uint32) {
 
 func setupSyncService(node *node.Node, host p2p.Host, hc harmonyConfig) {
 	blockchains := []*core.BlockChain{node.Blockchain()}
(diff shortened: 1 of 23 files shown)


cmd/harmony/main.go
```diff
@@ -699,6 +710,39 @@ func setupPrometheusService(node *node.Node, hc harmonyConfig, sid uint32) {
 	node.RegisterService(service.Prometheus, p)
 }
 
+func setupSyncService(node *node.Node, host p2p.Host, hc harmonyConfig) {
+	blockchains := []*core.BlockChain{node.Blockchain()}
+	if node.NodeConfig.ShardID != 0 {
```
LeoHChen, line 715: better add a function to determine sidechain, as it is widely used in the code I think.

hmy/downloader/beaconhelper.go
```diff
@@ -120,7 +120,9 @@ func (bh *beaconHelper) insertLastMileBlocks() (inserted int, bn uint64, err err
 			bn--
 			return
 		}
-		if err = bh.ih.verifyAndInsertBlock(b); err != nil {
+		// TODO: Instruct the beacon helper to verify signatures. This may require some forks
```
LeoHChen, line 123: no verification now?

hmy/downloader/shortrange.go
```diff
@@ -56,7 +58,7 @@ func (d *Downloader) doShortRangeSync() (int, error) {
 	}
 	n, err := d.ih.verifyAndInsertBlocks(blocks)
 	if err != nil {
-		if !errors.As(err, &sigVerifyError{}) {
+		if !errors.As(err, &emptySigVerifyError) {
```
LeoHChen, line 61: is emptySigVerifyError an pointer?

node/node_syncing.go
```diff
@@ -33,6 +38,14 @@ func init() {
 	rand.Seed(time.Now().UnixNano())
 }
 
+// BeaconSyncHook is the hook function called after inserted beacon in downloader
+// TODO: This is a small misc piece of consensus logic. Better put it to consensus module.
```
LeoHChen, line 42: Can you add a TODO to improve the logic?  Leader is way too busy now, may need to choose non-leader to broadcast.

node/node_syncing.go
```diff
@@ -540,12 +565,49 @@ func (node *Node) getEncodedBlockByHash(hash common.Hash) ([]byte, error) {
...
-func (node *Node) IsOutOfSync(bc *core.BlockChain) bool {
-	return node.stateSync.IsOutOfSync(bc, false)
+// SyncPeers return connected sync peers for each shard
+func (node *Node) SyncPeers() map[string]int {
+	ds := node.getDownloaders()
+	if ds == nil {
+		return nil
+	}
+	nums := ds.NumPeers()
+	res := make(map[string]int)
```
LeoHChen, line 595: any concurrency issue on the map?

hmy/downloader/beaconhelper.go
JackyWYX: No. Since the block from pub-sub message does not contain `commitSigAndBitmaps` field, thus the signature verification of the this block cannot be fulfilled. Better add signature in pub-sub message in beacon client topic later. A network fork is expected.

hmy/downloader/shortrange.go
JackyWYX: It is, and it is OK to use a pointer to pointer here. The second argument need to be a pointer to error interface. Please check out the [example here](https://golang.org/pkg/errors/#example_As).

node/node_syncing.go
JackyWYX: Sure, np

node/node_syncing.go
JackyWYX: No. There cannot be concurrency issue on a local variable.

hmy/downloader/beaconhelper.go
rlan35: insertChain will verify the signature for previous block. But not this block. Definitely it's better to verify for this block before commit like we recently added for the sync logic.

rlan35: But we can do it as another task.

rlan35 (approved):

PR #3599, opened 2021-03-19
return error for invalid address parsing

rlan35:


rpc/transaction.go
```diff
@@ -55,7 +55,10 @@ func (s *PublicTransactionService) GetAccountNonce(
 	blockNum := blockNumber.EthBlockNumber()
 
 	// Response output is the same for all versions
-	addr := internal_common.ParseAddr(address)
+	addr, err := internal_common.ParseAddr(address)
+	if err != nil {
+		return 0, err
```
LeoHChen, line 60: does it make sense to return 0 here?

LeoHChen (commented):
Does this PR change the PRC return value format?  If yes, we need to notify our partners and adjust the explorer code then.

rpc/transaction.go
rlan35: it's returning the error. So this value is not meaningful and shouldn't be used anyway. 0 as a default value is fine.

rlan35:
> Does this PR change the PRC return value format? If yes, we need to notify our partners and adjust the explorer code then.

this PR doesn't change rpc format.

JackyWYX (approved):

PR #3605, opened 2021-03-23
disable early unlock of tokens

rlan35:
After the introduction of redelegation feature, it's convenient for delegators to switch between validators without having to wait for the 7 epoch locking time. We also had the convenient feature in the very beginning to allow immediate unlock of undelegated tokens from validators who were elected in the last 7 epochs (it's also meant for easy switch away from bad validators). Together with this two feature, there is a path to achieve early unlocking by redelegating to an unelected validator and then undelegate from there.

This change disable the early unlock capability from unelected validators to fix this problem. This feature is not very relevant anyway after the introduction of redelegation.

### DIFF
--- a/staking/types/delegation.go
+++ b/staking/types/delegation.go
@@ -178,13 +178,13 @@ func (d *Delegation) DeleteEntry(epoch *big.Int) {
 // RemoveUnlockedUndelegations removes all fully unlocked
 // undelegations and returns the total sum
 func (d *Delegation) RemoveUnlockedUndelegations(
-	curEpoch, lastEpochInCommittee *big.Int, lockPeriod int,
+	curEpoch, lastEpochInCommittee *big.Int, lockPeriod int, noEarlyUnlock bool,
 ) *big.Int {
 	totalWithdraw := big.NewInt(0)
 	count := 0
 	for j := range d.Undelegations {
 		if big.NewInt(0).Sub(curEpoch, d.Undelegations[j].Epoch).Int64() >= int64(lockPeriod) ||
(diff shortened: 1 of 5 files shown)


staking/types/delegation.go
```diff
@@ -178,13 +178,13 @@ func (d *Delegation) DeleteEntry(epoch *big.Int) {
...
 func (d *Delegation) RemoveUnlockedUndelegations(
-	curEpoch, lastEpochInCommittee *big.Int, lockPeriod int,
+	curEpoch, lastEpochInCommittee *big.Int, lockPeriod int, noEarlyUnlock bool,
 ) *big.Int {
 	totalWithdraw := big.NewInt(0)
 	count := 0
 	for j := range d.Undelegations {
 		if big.NewInt(0).Sub(curEpoch, d.Undelegations[j].Epoch).Int64() >= int64(lockPeriod) ||
-			big.NewInt(0).Sub(curEpoch, lastEpochInCommittee).Int64() >= int64(lockPeriod) {
+			(!noEarlyUnlock && big.NewInt(0).Sub(curEpoch, lastEpochInCommittee).Int64() >= int64(lockPeriod)) {
```
LeoHChen, line 187: My understanding is this logic should be removed after the noearlyunlock is enabled, right?  Those data are not on-chain, so this check can be removed safely?  Can you add a TODO to remove this condition check after noearlyunlock rolled out if it makes sense.

LeoHChen (approved):
One comment for code cleanup.

JackyWYX (approved):

staking/types/delegation.go
rlan35: it can't be removed because it's onchain logic. This is a fork. We need to keep it for the history blocks.

PR #3607, opened 2021-03-24
avoid repeated block verification

rlan35:
Verified blocks should always be marked to avoid repeated verification on the same block.

### DIFF
--- a/consensus/validator.go
+++ b/consensus/validator.go
@@ -173,11 +173,11 @@ func (consensus *Consensus) onPrepared(recvMsg *FBFTMessage) {
 		consensus.getLogger().Debug().Msg("[onPrepared] consensus received message before init. Ignoring")
 		return
 	}
-	if err := consensus.BlockVerifier(&blockObj); err != nil {
+
+	if err := consensus.VerifyBlock(&blockObj); err != nil {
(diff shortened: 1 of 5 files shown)


consensus/validator.go
```diff
@@ -173,11 +173,13 @@ func (consensus *Consensus) onPrepared(recvMsg *FBFTMessage) {
 		consensus.getLogger().Debug().Msg("[onPrepared] consensus received message before init. Ignoring")
 		return
 	}
-	if err := consensus.BlockVerifier(&blockObj); err != nil {
-		consensus.getLogger().Error().Err(err).Msg("[OnPrepared] Block verification failed")
-		return
+	if !consensus.FBFTLog.IsBlockVerified(&blockObj) {
```
JackyWYX, line 176: Since this piece of code is used many times, can we extract it to a function?

JackyWYX (approved):

LeoHChen (commented):
agree with Jacky, it is actually would be better to wrap the checking within blockverifier function to simplify the change.

consensus/validator.go
rlan35: done.

rlan35:
Travis, test this please.

LeoHChen (approved):

PR #3610, opened 2021-03-25
remove unnecessary crosslink broadcast in state sync

rlan35:
1. Let 2% of the validators do crosslink broadcast along with the leader.
2. Cealn up Rand seeding.

### DIFF
--- a/hmy/downloader/beaconhelper.go
+++ b/hmy/downloader/beaconhelper.go
@@ -128,9 +128,12 @@ func (bh *beaconHelper) insertLastMileBlocks() (inserted int, bn uint64, err err
 			return
 		}
 		bh.logger.Info().Uint64("number", b.NumberU64()).Msg("Inserted block from beacon pub-sub")
-		if bh.insertHook != nil {
-			bh.insertHook()
-		}
+		// TODO: clean up insertHook since for beacon chain sync, we shouldn't broadcast crosslinks for all sync nodes.
+		// Crosslink broadcasting is a live operation relative to the beacon consensus, which should only
+		// be triggered when there is a new beacon block produced.
+		//if bh.insertHook != nil {
+		//	bh.insertHook()
+		//}
 		inserted++
 		bn++
 	}
--- a/hmy/downloader/beaconhelper.go
+++ b/hmy/downloader/beaconhelper.go
@@ -128,12 +128,10 @@ func (bh *beaconHelper) insertLastMileBlocks() (inserted int, bn uint64, err err
 			return
 		}
 		bh.logger.Info().Uint64("number", b.NumberU64()).Msg("Inserted block from beacon pub-sub")
-		// TODO: clean up insertHook since for beacon chain sync, we shouldn't broadcast crosslinks for all sync nodes.
-		// Crosslink broadcasting is a live operation related to the beacon consensus, which should only
-		// be triggered when there is a new beacon block produced.
-		//if bh.insertHook != nil {
-		//	bh.insertHook()
-		//}
+
+		if bh.insertHook != nil {
+			bh.insertHook()
+		}
 		inserted++
 		bn++
 	}
--- a/hmy/downloader/beaconhelper.go
+++ b/hmy/downloader/beaconhelper.go
@@ -129,7 +129,7 @@ func (bh *beaconHelper) insertLastMileBlocks() (inserted int, bn uint64, err err
 		}
 		bh.logger.Info().Uint64("number", b.NumberU64()).Msg("Inserted block from beacon pub-sub")
 		// TODO: clean up insertHook since for beacon chain sync, we shouldn't broadcast crosslinks for all sync nodes.
-		// Crosslink broadcasting is a live operation relative to the beacon consensus, which should only
+		// Crosslink broadcasting is a live operation related to the beacon consensus, which should only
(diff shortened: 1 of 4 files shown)


node/node_syncing.go
```diff
@@ -208,8 +204,8 @@ func (node *Node) doBeaconSyncing() {
 					)
 					if err != nil {
 						node.beaconSync.AddLastMileBlock(beaconBlock)
-					} else if node.Consensus.IsLeader() {
-						// Only leader broadcast crosslink to avoid spamming p2p
+					} else if node.Consensus.IsLeader() || rand.Intn(100) <= 1 {
+						// Only leader or 2% of validators broadcast crosslink to avoid spamming p2p
```
JackyWYX, line 208: Leader is still broadcasting crosslink crosslink here. Why don't we need this logic in stream sync (a.k.a. insertHook)?

hmy/downloader/beaconhelper.go
```diff
@@ -128,9 +128,12 @@ func (bh *beaconHelper) insertLastMileBlocks() (inserted int, bn uint64, err err
 			return
 		}
 		bh.logger.Info().Uint64("number", b.NumberU64()).Msg("Inserted block from beacon pub-sub")
-		if bh.insertHook != nil {
-			bh.insertHook()
-		}
+		// TODO: clean up insertHook since for beacon chain sync, we shouldn't broadcast crosslinks for all sync nodes.
+		// Crosslink broadcasting is a live operation related to the beacon consensus, which should only
+		// be triggered when there is a new beacon block produced.
+		//if bh.insertHook != nil {
```
JackyWYX, line 134: I don't get it. We are still broadcasting crosslink in `doBeaconSyncing `, right? Why are we removing this code?

JackyWYX: insertLastMileBlocks is exactly dealing with the beacon blocks receiving from `node.BeaconBlockChannel`

LeoHChen (commented):
why do we need to increase the rate of crosslink broadcast?  Is there any missed crosslink? 

Also, since we have an ordered list of all the active committee member, can we use that list to determine who can broadcast the crosslink, instead of the leader, it is deterministic per epoch, right?  We can add a random factor as well in case the node is offline.

node/node_syncing.go
rlan35: Because this is for the live beacon sync. Meaning when there is a new beacon block produced, it's broadcasted live to the other shards and here is when that block is received. This is the time we should broadcast back the crosslink based on the new beacon block.

But for normal catch up sync. it can be from block 0 to 1M. and we shouldn't do the crosslink broadcast at this situation.

hmy/downloader/beaconhelper.go
rlan35: replied above.

node/node_syncing.go
JackyWYX: But the insertHook is called for live beacon sync. It only deals with block received from pub-sub message. It is not called when syncing from block 0 to 1m.

rlan35:
> why do we need to increase the rate of crosslink broadcast? Is there any missed crosslink?
> 
> Also, since we have an ordered list of all the active committee member, can we use that list to determine who can broadcast the crosslink, instead of the leader, it is deterministic per epoch, right? We can add a random factor as well in case the node is offline.

Because I checked the last time we had the crosslink issue, it's because the single node had a bad db record for the last crosslink so it failed to broadcast the next crosslink in continuity. Letting more nodes to do the broadcasting will prevent single point of failure.

We could do that rotation thing. But it can always happen that the chosen validators fails to broadcast it. The good thing about asking leader to do it is that the leader should always be functioning because otherwise, no new block will be produced and there won't be crosslink broadcasting needed anyway.

node/node_syncing.go
rlan35: ok, So the broadcast here is for legacy sync. and the hooked broadcast is for stream sync. And they are doing exactly the same thing?

JackyWYX: Yep, exactly.

rlan35: ok, let me revert the hook change.

rlan35: reverted.

JackyWYX: Thanks

JackyWYX:
> Because I checked the last time we had the crosslink issue, it's because the single node had a bad db record for the last crosslink so it failed to broadcast the next crosslink in continuity. Letting more nodes to do the broadcasting will prevent single point of failure.

All shard chain leaders will broadcast the crosslinks. So we have redundancy of 3 on mainnet or testnet. Do you think we shall get even more redundancy on this?

rlan35:
> > Because I checked the last time we had the crosslink issue, it's because the single node had a bad db record for the last crosslink so it failed to broadcast the next crosslink in continuity. Letting more nodes to do the broadcasting will prevent single point of failure.
> 
> All shard chain leaders will broadcast the crosslinks. So we have redundancy of 3 on mainnet or testnet. Do you think we shall have even more redundancy on this?

For now, 3 is enough. We don't want to have too many repeated messages to clog the network until we have smarter p2p dedupping logic.

JackyWYX (approved):

PR #3612, opened 2021-03-25
[SYNC][ENGINE] Engine refactor and DNS sync verify signature

JackyWYX:
## Description

1. Refactored engine verify signature logic, and get rid of a lot of duplicate code.
2. Remove insert helper in downloader module, use engine signature verification.
3. Added signature verification in DNS sync logic, and updated corresponding protobuf message.

All changes are backward compatible.

### A few more words about engine refactor

1. Engine is now initialized via New method, and all use of global variable of Engine is removed.
2. Engine is now initialized with a shardID, which enables engine itself to determine whether to recalculate state shard when comparing shard ID (case for verifying CX receipts).
3. Added a small cache in engine to avoid duplicate signature verification to save some CPU (`verifiedSigCache`)

## Test

1. Tested locally.
2. Tested on stressnet of DNS sync.
3. Tested on stressnet of stream sync.
4. Tested on mainnet of 1. stream sync 2. epoch 0 blocks.


api/service/legacysync/syncing.go
```diff
@@ -844,7 +873,7 @@ func (ss *StateSync) generateNewState(bc *core.BlockChain, worker *worker.Worker
 		if block == nil {
 			break
 		}
-		err = ss.UpdateBlockAndStatus(block, bc, worker, false)
+		err = ss.UpdateBlockAndStatus(block, bc, worker, true)
```
rlan35, line 876: why change to true for verifyAllSig?

api/service/legacysync/syncing.go
```diff
@@ -822,16 +848,19 @@ func (ss *StateSync) generateNewState(bc *core.BlockChain, worker *worker.Worker
...
 	var err error
-	for {
-		block := ss.getBlockFromOldBlocksByParentHash(parentHash)
+	for block := ss.getBlockFromOldBlocksByParentHash(parentHash); ; {
 		if block == nil {
 			break
 		}
-		err = ss.UpdateBlockAndStatus(block, bc, worker, false)
+		next := ss.getBlockFromOldBlocksByParentHash(block.Hash())
+		// Enforce sig check for the last block in a batch
```
rlan35, line 856: why does this mean it's last block? Also can you explain the logic that involves getBlockFromOldBlocksByParentHash? what is oldBlocks?

api/service/legacysync/syncing.go
JackyWYX: These blocks were designed to be obtained from DNS node's register/push (DownloaderRequest_REGISTER). And these blocks can possibly be the last block of one sync loop after dealing with "commonBlocks". So we need to verify the signature and flush commitSigAndBitmap to db. The amount of this signature verification is not large at all.

api/service/legacysync/syncing.go
JackyWYX: `oldBlocks` is just some legacy function names. oldBlocks (a.k.a. `ss.commonBlocks`) are blocks downloaded from `ss.downloadBlocks`, which makes up the major portion of synced blocks. E.g. For an out of sync node, one sync loop syncs 1000 blocks, and all these 1000 blocks downloaded from remote nodes resides in `ss.commonBlocks`.

This part of logic change is basically aiming to enforce signature verification (also flush commitSig to db) for the last block of downloaded blocks. E.g, A sync loop has downloaded 900 blocks from remote nodes, and 900th block is the latest block. So this block's sig should be verified and inserted into db, which enables consensus to find the commit sig of current block block.

JackyWYX: I did minimum code change but correct the logics since this is the code about to discard. Do you think I shall refactor it in a more readable way?

hmy/downloader/shortrange.go
```diff
@@ -51,21 +49,25 @@ func (d *Downloader) doShortRangeSync() (int, error) {
...
 	}
 
-	n, err := d.ih.verifyAndInsertBlocks(blocks)
+	n, err := verifyAndInsertBlocks(d.bc, blocks)
 	numBlocksInsertedShortRangeHistogramVec.With(d.promLabels()).Observe(float64(n))
 	if err != nil {
-		if !errors.As(err, &emptySigVerifyError) {
+		if sh.blameAllStreams(blocks, n, err) {
 			sh.removeStreams(whitelist) // Data provided by remote nodes is corrupted
+		} else {
```
rlan35, line 65: what's the cases when a peer is blamed? what happen after the blaming?

hmy/downloader/downloader.go
```diff
@@ -262,3 +264,53 @@ func (d *Downloader) finishSyncing() {
...
+		sigBytes, bitmap, err = chain.ParseCommitSigAndBitmap(block.GetCurrentCommitSig())
+		if err != nil {
+			return errors.Wrap(err, "parse commitSigAndBitmap")
+		}
+	}
+
+	if err := bc.Engine().VerifyHeaderSignature(bc, block.Header(), sigBytes, bitmap); err != nil {
+		return &sigVerifyErr{err}
+	}
+	if err := bc.Engine().VerifyHeader(bc, block.Header(), true); err != nil {
```
rlan35, line 309: why we still verifySeal==true when we already verified the signature?

api/service/legacysync/syncing.go
JackyWYX: This reminds to to also update NEWBLOCK request. Thanks.

api/service/legacysync/syncing.go
```diff
@@ -822,16 +848,19 @@ func (ss *StateSync) generateNewState(bc *core.BlockChain, worker *worker.Worker
...
+		enforceSigCheck := next == nil
+
+		err = ss.UpdateBlockAndStatus(block, bc, worker, enforceSigCheck)
 		if err != nil {
 			break
 		}
-		parentHash = block.Hash()
+		block = next
```
rlan35, line 863: parentHash is not updated. This loop doesn't seem to work anymore.

JackyWYX: ParentHash is now only used once. For the rest of the loop, `block.Hash()` is used and keeps iterating.

hmy/downloader/downloader.go
JackyWYX: To verify the signature of previous block. There can be the corner case that 

```
prevBlock.CommitSigAndBitmap != block.LastCommitSig + block.LastCommitBitmap
```

So we need to verify the seal against last block. 

BTW, engine will not do the same signature verification twice because of `engineImpl.verifiedSigCache` in normal cases.

api/service/legacysync/syncing.go
rlan35: I see. this logic seems more convoluted than the previous one. Why change the loop logic? The previous one seems more logical for readers.

JackyWYX: Because I want to check whether the current block is the last one in `commonBlocks`. Sorry for not focusing on the code quality of these codes since I thought the code was about to be discarded. Let me try refactor a little bit.

JackyWYX: Just refactored. Please let me know whether the new logic makes sense to you. Thanks for pointing this out!

api/service/legacysync/syncing.go
rlan35: got it thanks

hmy/downloader/downloader.go
rlan35: ok.

hmy/downloader/shortrange.go
rlan35: ?

rlan35 (approved):

api/service/legacysync/syncing.go
JackyWYX: NEWBLOCK push request also updated with Signature. So this value can be safely set to true. This is to verify the new block signature pushed from DNS nodes. The amount of this blocks is not much compared with `commonBlocks`.

hmy/downloader/shortrange.go
JackyWYX: 1. This is the case where the peer provides the last block in a batch gives a wrong commit signature. Since the commit signature is not include in block hash, it could be the case that other whitelist streams provides the correct hash chain, and only this node is giving wrong data.
2. In other cases, since all block data is deterministic with a fixed block hash, if signature verification or block insertion fails, we can imply that the given hash chain is invalid. Thus all streams providing that hash chain is guilty.
3. Currently, for the guilty screams, we simply close the connection to that stream. Will implement a timed blacklist later.

api/service/legacysync/syncing.go
JackyWYX: Updated logic to fix errors. Need two bools to check the signature values. Please check the latest code again. Thanks!

JackyWYX:
@rlan35 Updated following changes since your last approve:

1. Added commit signature in NEWBLOCK dns sync requests. Also updated REGISTER dns request to accommodate the changes to enable backward compatibility.
2. Bug fixes regarding beacon sync and CX receipt verification for travis tests. 

Please review the code again and let me know if there are still any problems. Thanks!

api/service/legacysync/syncing.go
```diff
@@ -844,7 +905,7 @@ func (ss *StateSync) generateNewState(bc *core.BlockChain, worker *worker.Worker
 		if block == nil {
 			break
 		}
-		err = ss.UpdateBlockAndStatus(block, bc, worker, false)
+		err = ss.UpdateBlockAndStatus(block, bc, true, true)
```
rlan35, line 908: why turn verifyAllSeal to true here? then it will verify signature for all blocks, isn't it slow?

node/node_syncing.go
```diff
@@ -567,6 +595,60 @@ func (node *Node) getEncodedBlockByHash(hash common.Hash) ([]byte, error) {
...
 
+func (node *Node) getEncodedBlockWithSigByHash(hash common.Hash) ([]byte, error) {
+	if b, ok := blockWithSigReqCache.Get(hash); ok {
+		return b.([]byte), nil
+	}
+	blk := node.Blockchain().GetBlockByHash(hash)
+	if blk == nil {
+		return nil, errBlockNotExist
+	}
+	sab, err := node.getCommitSigAndBitmap(blk)
```
rlan35, line 606: Should prefer the commitSig from nextBlock.CommitSig/Bitmap.  Only the last block in the chain rely on the commitSig in DB because it's not saved in the new block yet.

crypto/hash/rlp.go
```diff
@@ -1,22 +1,47 @@
...
+	New: func() interface{} {
+		return sha3.NewLegacyKeccak256()
+	},
+}
+
 // FromRLP hashes the RLP representation of the given object.
 func FromRLP(x interface{}) (h common.Hash) {
-	hw := sha3.NewLegacyKeccak256()
+	hw := kec256Pool.Get().(hash.Hash)
+	defer func() {
```
LeoHChen, line 21: where is this change coming from?  Is it part of upstream change? why mix it here?

hmy/downloader/shortrange.go
```diff
@@ -51,21 +49,25 @@ func (d *Downloader) doShortRangeSync() (int, error) {
...
+			// It is the last block gives a wrong commit sig. Blame the provider of the last block.
+			criminal := stids[len(stids)-1]
```
LeoHChen, line 75: criminal seems too strong , :-)

internal/chain/engine.go
```diff
@@ -438,127 +389,113 @@ func applySlashes(
...
+// as engine
+func (e *engineImpl) needRecalculateStateShard(chain engine.ChainReader, epoch *big.Int, targetShardID uint32) bool {
+	if chain.Config().IsStaking(epoch) {
+		return false
+	}
+	return targetShardID != e.shardID
+}
+
+// Support 512 at most validator nodes
+const bitmapKeyBytes = 64
```
LeoHChen, line 481: at most 512 validators per shard?  Can it be changed later?

LeoHChen (commented):
A big change. some quick comments.

api/service/legacysync/syncing.go
JackyWYX: Here is only a few block to be updated in this case compared to commonBlocks. These blocks is a part of REGISTER / PUSH flow. The logics are:

1. DNS server got a new block in consensus. Block is then pushed to `VerifiedNewBlock` (consensus/consensus_v2.go:712)
2. DNS server send NEWBLOCK request with the new block (with sig) to registered sync peers. (node/node_syncing.go:383)
3. Client node receive NEWBLOCK request, and add blocks to `peerConfig.newBlocks` (api/service/legacysync/syncing.go:253)
4. Blocks added in last step is pulled from `getMaxConsensusBlockFromParentHash` (api/service/legacysync/syncing.go:904), and being handled here.

So blocks updated here are actually small in number compared to commonBlocks. Not a very critical performance issue I think. On the other hand, this enforce signature check will also flush the block's commit sig to db, which ensures that the last block synced from dns sync is committed with a commit sig for consensus to use.

node/node_syncing.go
JackyWYX: Yes, this is exactly how currently the logic works. Please take a look into the function `getCommitSigAndBitmap`. 

Also to make the function name more clear, I renamed the function name `getCommitSigAndBitmapFromChildOrDB.`

crypto/hash/rlp.go
JackyWYX: Ah, this is just to optimize the memory used for RLP hashing in multi-threading. Sorry for mixing it here. Will open another PR for this.

hmy/downloader/shortrange.go
JackyWYX: Do you mean the variable name? Renamed to `stToBlame`. Thanks!

internal/chain/engine.go
JackyWYX: This is simply a sanity field for de-duplication in memory. And sure, it can be changed anytime when needed.

crypto/hash/rlp.go
JackyWYX: Reverted and PR here: https://github.com/harmony-one/harmony/pull/3616. Please take a look.

api/service/legacysync/syncing.go
```diff
@@ -807,6 +858,12 @@ func (ss *StateSync) UpdateBlockAndStatus(block *types.Block, bc *core.BlockChai
 		Str("blockHex", block.Hash().Hex()).
 		Uint32("ShardID", block.ShardID()).
 		Msg("[SYNC] UpdateBlockAndStatus: New Block Added to Blockchain")
+
+	if verifyAllSig && haveCurrentSig {
+		if err := bc.WriteCommitSig(block.NumberU64(), block.GetCurrentCommitSig()); err != nil {
```
rlan35, line 863: no need to explicitly write to DB here. You can just put the commigSigbitmap inside the Block object and the insertChain() will commit the the sig as long as it's present inside the block.

api/service/legacysync/syncing.go
rlan35: cool

rlan35:
overall, looks good.

api/service/legacysync/syncing.go
JackyWYX: I feel a little unconfident for writing signature to db that is not verified. Do you think writing unverified commit sig might cause some problem?

JackyWYX: After discussion, it's safe to say this will not be a problem. This field in DB is only used by leader proposing a new block when bootstrap. So if the leader node has a wrong commit sig in DB, it will only effect itself and a view change will be expected later. Removed `bc.WriteCommitSig` in both downloader and DNS sync logic.

PR #3615, opened 2021-03-26
clean up and todos

rlan35:
Remove some unused code and add a few todos for optimization (based on code in this PR https://github.com/harmony-one/harmony/pull/3612)

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -711,6 +711,7 @@ func (consensus *Consensus) tryCatchup() error {
 			return err
 		}
 		select {
+		// TODO(jacky): check if this is still needed
 		case consensus.VerifiedNewBlock <- blk:
 		default:
 			consensus.getLogger().Info().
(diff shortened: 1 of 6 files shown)


LeoHChen (approved):
better open a github issue with the TODO included.

consensus/consensus_v2.go
```diff
@@ -711,6 +711,7 @@ func (consensus *Consensus) tryCatchup() error {
 			return err
 		}
 		select {
+		// TODO(jacky): check if this is still needed.
```
JackyWYX, line 714: It is still used in DNS sync. But removing will also not harm :-) BTW, I think the default case of this select can be downgraded to Trace since this channel is not used for stream sync.

consensus/validator.go
```diff
@@ -284,6 +278,8 @@ func (consensus *Consensus) onCommitted(recvMsg *FBFTMessage) {
 	}
 	commitPayload := signature.ConstructCommitPayload(consensus.Blockchain,
 		blockObj.Epoch(), blockObj.Hash(), blockObj.NumberU64(), blockObj.Header().ViewID().Uint64())
+
+	// TODO(jacky): cache the result of signature verification to reuse in next block's last-sig verification
```
JackyWYX, line 282: Very good point!

JackyWYX (approved):

PR #3616, opened 2021-03-27
[crypto] Optimize the FromRLP with sync pool

JackyWYX:
## Description

Optimize the `FromRLP` implementation with `sync.Pool`.

## Benchmark

```
> go test -bench=. -benchmem
goos: darwin
goarch: amd64
pkg: github.com/harmony-one/harmony/crypto/hash
BenchmarkFromRLP_MultiThreaded-12                4841481               263 ns/op             560 B/op          5 allocs/op
BenchmarkFromRLPNoPool_MultiThreaded-12          3665227               344 ns/op            1008 B/op          6 allocs/op
BenchmarkFromRLP_SingleThreaded-12               1303647               909 ns/op             560 B/op          5 allocs/op
BenchmarkFromRLPNoPool_SingleThreaded-12         1296238               920 ns/op            1008 B/op          6 allocs/op
PASS
ok      github.com/harmony-one/harmony/crypto/hash      7.441s
```

We can see that FromRLP with sync pool implementation has better memory and CPU performance over the implementation without sync pool. This is because the sync pool implementation will reuse the memory-allocated hasher and results in better performance both in CPU and memory. The huge CPU difference in multi-threading benchmark actually results from sync.Pool's optimization over golang's GC mechanism.


rlan35 (approved):

rlan35:
awesome! Can't wait to see the drop of CPU usage on mainnet.

JackyWYX:
> awesome! Can't wait to see the drop of CPU usage on mainnet.

Well, this is not going to help with CPU issue a lot as far as I see. It's just a very small fix from my previous experience on Golang. ;)

Issue #3618, opened 2021-03-29
High CPU observed on stressnet with downloader


JackyWYX:
**Describe the bug**

High CPU usage observed on Stressnet node running downloader protocol. Observed CPU usage was about 50%

JackyWYX:
CPU profile did several days ago:

![profile](https://user-images.githubusercontent.com/40162696/112877868-2077d200-907c-11eb-83a8-39234e9e2914.png)

Goroutine profile:

![goroutine](https://user-images.githubusercontent.com/40162696/112877883-2372c280-907c-11eb-8430-528ac48c6527.png)

JackyWYX:
From the pprof diagram, we can come to the following conclusion:

1. From CPU profile diagram, we can see that most CPU used is for goroutine coordination.
2. From goroutine diagram, it seems that the node is opening ~124 connection to remote nodes via sync stream.

The root cause is probably the wrong usage of `stream.Close`. In this case, `stream.Reset` might be more appropriate.

Looking deeper into the implementation of these two functions.

JackyWYX:
After running for a day, still observed ~50% cpu usage on non-leader node 54.214.183.56. The number of goroutine running `Stream.Read` is still high.

![goroutine](https://user-images.githubusercontent.com/40162696/113017633-82940e00-9134-11eb-878e-5255eff34532.png)

For CPU profile, `runtime.SiftdownTimer` has greatly reduced. But still not totally sure why CPU is high.

![profile](https://user-images.githubusercontent.com/40162696/113023563-a0fd0800-913a-11eb-8abc-a76ccd00f65b.png)

JackyWYX:
Comparing with a freshly started node, CPU profile looks almost the same:

![profile](https://user-images.githubusercontent.com/40162696/113024153-46b07700-913b-11eb-9135-2c736cb0adcc.png)

JackyWYX:
And the large number of goroutines is also observed on a freshly started node, meaning that it is most likely come from libp2p basic discovery.

JackyWYX:
To this point, we cannot gather any information useful from pprof. Also noticed that mainnet node also have CPU and memory issues after long run. 

So here a baseline on main branch will be made for comparison.

JackyWYX:
When deployed with none-downloader code, the CPU remains < 10% after two days run.

Following is the pprof graph:

[...]

PR #3621, opened 2021-03-29
[stream] Use stream reset for close

JackyWYX:
## Description

Address https://github.com/harmony-one/harmony/issues/3618.

Use `Stream.Reset` on close to avoid remote node dangling in `read`. 

## Test

Tested locally.

Testing on stressnet of CPU performance.


p2p/stream/types/stream.go
```diff
@@ -67,7 +67,7 @@ func (st *BaseStream) ProtoSpec() (ProtoSpec, error) {
 
 // Close close the stream on both sides.
```
rlan35, line 68: update comment?

JackyWYX: Sure. Done. Thanks!

rlan35 (approved):

PR #3626, opened 2021-03-31
[consensus] use verifyHeader signature in onCommitted

JackyWYX:
## Description

Use engine.VerifyHeaderSignature in consensus to verify committed message, which will use the cache within `engineImpl` to avoid duplicated signature verification and save some CPU resources.

## Test

Tested locally.


consensus/validator.go
```diff
@@ -276,17 +267,34 @@ func (consensus *Consensus) onCommitted(recvMsg *FBFTMessage) {
...
 		return
 	}
-	commitPayload := signature.ConstructCommitPayload(consensus.Blockchain,
-		blockObj.Epoch(), blockObj.Hash(), blockObj.NumberU64(), blockObj.Header().ViewID().Uint64())
-
-	// TODO(jacky): cache the result of signature verification to reuse in next block's last-sig verification
-	if !aggSig.VerifyHash(mask.AggregatePublic, commitPayload) {
+	sigBytes, bitmap, err := chain.ParseCommitSigAndBitmap(recvMsg.Payload)
+	if err != nil {
+		consensus.getLogger().Info().Err(err).
```
LeoHChen, line 272: If it is a real error, please use Error() level in logger.

JackyWYX: Good point! thanks!

consensus/validator.go
```diff
@@ -276,17 +267,34 @@ func (consensus *Consensus) onCommitted(recvMsg *FBFTMessage) {
...
+			Uint64("blockNum", recvMsg.BlockNum).
+			Uint64("viewID", recvMsg.ViewID).
+			Str("blockHash", recvMsg.BlockHash.Hex()).
+			Msg("[OnCommitted] Failed to parse commit sigBytes and bitmap")
+		return
+	}
+	if err := consensus.Blockchain.Engine().VerifyHeaderSignature(consensus.Blockchain, blockObj.Header(),
```
rlan35, line 279: This func checks the quorum by recomputing the voting power from scratch again and again: "quorum.NewVerifier". This is costly. And it's doing the same job as below "IsQuorumAchievedByMask". Please reduce redundant work.

JackyWYX: Thanks. Added another cache in engine to resolve duplicate quorum.Verifier initialization.

rlan35 (approved):

PR #3627, opened 2021-03-31
Avoid repeated state creation in validator reading

rlan35:
Refactor the validator reading interface so we don't recreate a new state for each validator reading.

Also optimize undelegation payout logic to avoid repetitive/redundant execution.


JackyWYX (approved):
Looks good to me.

LeoHChen (approved):

PR #3628, opened 2021-04-01
[sync] ease the verification on dns client side, enable server forward compatibility

JackyWYX:
## Issue

Enable the DNS server forward compatibility by easing the verification on DNS client. This is not a security issue for centralized sync.

1. Instead of adding a new request, add some branching logic to handle existing BLOCK request both at server and client side.
2. During verification, commit signature check is not enforced. A block without commit sig is ok to pass validation.

## Tests

Tested locally.

Also tested on mainnet with local node.


rlan35 (approved):
tested on mainnet. it's working.

PR #3633, opened 2021-04-03
[engine] Engine cache for `VerifyCrossLink`

JackyWYX:
## Description

1. Added a new method `VerifyCrossLink` to engine interface.
2. Refactored some code of`engineImpl` to use the same cache for crosslink and headers.
3. Basically, `VerifyCrossLink` `VerifySeal` `VerifyHeaderSignature` will share the same cache to reduce duplicate signature verification.

## Test

tested locally.

### DIFF
--- a/internal/chain/engine.go
+++ b/internal/chain/engine.go
@@ -472,13 +472,11 @@ func (e *engineImpl) getEpochCtxCached(chain engine.ChainReader, shardID uint32,
 	}
 	cached, ok := e.epochCtxCache.Get(ecKey)
 	if ok && cached != nil {
-		if ec := cached.(epochCtx); ec.qrVerifier != nil {
-			return ec, nil
-		}
+		return cached.(epochCtx), nil
 	}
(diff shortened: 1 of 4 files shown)


internal/chain/engine.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
```
rlan35, line 29: Make it 100?

rlan35 (approved):

rlan35:
Looks great. Please run a mainnet node on shard 0 with this binary to see if it can verify blocks without problems

internal/chain/engine.go
JackyWYX: Sure

JackyWYX:
> Looks great. Please run a mainnet node on shard 0 with this binary to see if it can verify blocks without problems

Run a node connecting to latest mainnet node, and crosslink verification success observed with custom print message. I think this PR is good to go.

PR #3648, opened 2021-04-09
[txPool] added gas balance check when validating delegate transaction

JackyWYX:
## Issue

https://github.com/harmony-one/go-sdk/issues/242 (3.1.1 temporary fix)

Added balance check with gas price for delegation staking transaction. 

Some delegation transaction will pass `txPool.validateTx` with amount set equal to balance (not aware of the additional gas fee). This PR is to fix the validation for delegation transaction.

## Tests

Tested locally and resolve the issue.

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -839,9 +839,19 @@ func (pool *TxPool) validateStakingTx(tx *staking.StakingTransaction) error {
 			return err
 		}
 		pendingEpoch := pool.pendingEpoch()
-		_, _, _, err = VerifyAndDelegateFromMsg(
+		_, delegateAmt, _, err := VerifyAndDelegateFromMsg(
 			pool.currentState, pendingEpoch, stkMsg, delegations, pool.chainconfig.IsRedelegation(pendingEpoch))
-		return err
+		if err != nil {
+			return err
+		}
+		// We need to deduct gas price and verify balance since txn.Cost() is not accurate for delegate
+		// staking transaction because of re-delegation.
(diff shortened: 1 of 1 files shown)


core/tx_pool.go
```diff
@@ -839,9 +839,19 @@ func (pool *TxPool) validateStakingTx(tx *staking.StakingTransaction) error {
...
 			pool.currentState, pendingEpoch, stkMsg, delegations, pool.chainconfig.IsRedelegation(pendingEpoch))
-		return err
+		if err != nil {
+			return err
+		}
+		// We need to deduct gas price and verify balance since txn.Cost() is not accurate for delegate
+		// staking transaction because of re-delegation.
+		gasAmt := new(big.Int).Mul(tx.GasPrice(), new(big.Int).SetUint64(tx.GasLimit()))
+		totalAmt := new(big.Int).Add(delegateAmt, gasAmt)
+		if bal := pool.currentState.GetBalance(from); bal.Cmp(totalAmt) < 0 {
```
rlan35, line 851: redelegation will reuse stake locked undelegation period (7 epoch). the check should be delegateAmt + gasAmt <= balance + stake_in_undelegation.  Did you consider this?

rlan35, line 851: very nice catch!. Please also check on other staking txns to make sure there is no verification issue like this where the gas cost is not counted.

JackyWYX: Other staking transactions are good since their costs are calculated correctly and checked with `stakingTransaction.Cost()`. For delegation, the cost need more comprehensive logic handling along with statedb because of redelegation. So I think other staking transactions are just fine and gas fee is calculated.

rlan35 (approved):

Issue #3650, opened 2021-04-11
InSync checking is not working as expected


LeoHChen:
**Describe the bug**
The newly fixed inSync API is not working as expected.

I upgraded 4 mainnet nodes using the latest build.  The node info can be found in the following URL.
https://watchdog.hmny.io/report-mainnet#version-v6979-v4.0.0-46-g077f1071

Then I tried to check the InSync status of all the four nodes and all of them return `false`.

```bash
curl -s --request POST 'http://IP:9500/' --header 'Content-Type: application/json' --data-raw '{ "jsonrpc": "2.0", "method": "hmy_inSync", "params": [], "id": 1}' | jq .
```

Please double check to make sure the InSync API work as expected. Thanks.

JackyWYX:
My BAD. Please take a look at this fix. https://github.com/harmony-one/harmony/pull/3651 Shall be work now.

PR #3651, opened 2021-04-11
[downloader] fix inSync method

JackyWYX:
## Issue

Resolve https://github.com/harmony-one/harmony/issues/3650.

downloaders.active is only set after `config.ServerOnly` check.


rlan35:
is this tested?

LeoHChen (approved):
returned true on mainnet node I tested.

PR #3656, opened 2021-04-13
[txPool] Further fix in transaction pool

JackyWYX:
## Issue

Another fix for https://github.com/harmony-one/go-sdk/issues/242

Enforce rigid transaction cost check during `txPool.Reset` to remove all invalid transactions. 

## Test

Tested locally.

### DIFF
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -328,7 +328,7 @@ func (l *txList) FilterValid(txPool *TxPool, address common.Address, bn uint64) 
 	var invalids types.PoolTransactions
 
 	removed, errs := l.filterPrice(txPool, address)
-	if bn > l.lastStkCheck+stakingTxCheckThreshold {
+	if bn >= l.lastStkCheck+stakingTxCheckThreshold {
 		l.lastStkCheck = bn
 
 		stkRemoved, errRemoved := l.filterStaking(txPool)
--- a/core/tx_list.go
+++ b/core/tx_list.go
@@ -22,12 +22,9 @@ import (
 	"math/big"
 	"sort"
 
-	"github.com/pkg/errors"
-
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/harmony-one/harmony/core/types"
 	"github.com/harmony-one/harmony/internal/utils"
-	staking "github.com/harmony-one/harmony/staking/types"
 )
 
 // nonceHeap is a heap.Interface implementation over 64bit unsigned integers for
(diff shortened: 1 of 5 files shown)


core/tx_list.go
```diff
@@ -123,6 +125,36 @@ func (m *txSortedMap) Filter(filter func(types.PoolTransaction) bool) types.Pool
...
+
+	// Collect all the transactions to filter out
+	for nonce, tx := range m.items {
+		if err := filter(tx); err != nil {
+			removed = append(removed, tx)
+			errs = append(errs, err)
+			delete(m.items, nonce)
+		}
+	}
+	// If transactions were removed, the heap and cache are ruined
```
rlan35, line 145: what does this mean？

rlan35:
how is this PR tested? also is there performance implication, I suspect that for every block all pending txns will be executed/verified again, it will be costly. any caching to do?

core/tx_list.go
JackyWYX: This is the comment from copying `txSortedMap.Filter`. It means when transactions are moved, the order of heap is destroyed. Thus, will need to re-build `m.index` and `m.items` in the following function

JackyWYX:
## Local test

1. Spin up local network.
2. Create a validator with faucet account:

```
./hmy staking create-validator --validator-addr one1zksj3evekayy90xt4psrz8h6j2v3hla4qwz4ur --amount 10000 --bls-pubkeys ea5b6acfb9146ef566bf05bc3a806988180684d7494d9bfe765afacfb2d9467de7864ece64b9ab7150c806c549d9b504 --name "jacky" --identity "jacky" --details "DETAILS" --security-contact "CONTACT" --website "YOUR-WEBSITE.COM" --max-change-rate 0.1 --max-rate 0.1 --rate 0.1 --max-total-delegation 100000000 --min-self-delegation 10000 --passphrase
```

3. Send two consecutive edit-validator transactions in different consoles at the same time.

```
./hmy staking edit-validator --add-bls-key 66d9897f0f53cd149edc7ebe44fad7688ee6434e1f90974d9d8e6a9c73cd220c491bc0b94cdbc089f5c3b04c2169db14 --validator-addr one1zksj3evekayy90xt4psrz8h6j2v3hla4qwz4ur --node=localhost:9502

./hmy staking edit-validator --add-bls-key 66d9897f0f53cd149edc7ebe44fad7688ee6434e1f90974d9d8e6a9c73cd220c491bc0b94cdbc089f5c3b04c2169db14 --validator-addr one1zksj3evekayy90xt4psrz8h6j2v3hla4qwz4ur --node=localhost:9502 --nonce 2
```

4. After one block, query pending transactions with

```
./hmy blockchain pool --node=localhost:9502
./hmy blockchain pool --node=localhost:9500
```

It is observed that both node's transaction pool is empty, which is good.

JackyWYX:
> how is this PR tested? also is there performance implication, I suspect that for every block all pending txns will be executed/verified again, it will be costly. any caching to do?

No caching is expected. This is to verify the transaction against the latest state db (mutates over time), where we shall not use cache. Currently, there is not much performance concern imo, because for leader node, all pending transactions in txPool is packed in new blocks. Thus no extra work is done in validation.

```
{"level":"info","newTxns":0,"newStakingTxns":0,"blockGasLimit":80000000,"blockGasUsed":0,"caller":"/mnt/jenkins/workspace/harmony-release/harmony/node/worker/worker.go:183","time":"2021-04-13T19:15:52.949667012Z","message":"Block gas limit and usage info"}
```

But this will become a concern when transaction numbers grows to the next level, when normal transactions used up all gases, leaving no space for staking transactions. Thus staking transactions will increase in number over time, which will potentially cause performance issue here. No solution without a fork will resolve this issue imo.

rlan35:
> > how is this PR tested? also is there performance implication, I suspect that for every block all pending txns will be executed/verified again, it will be costly. any caching to do?
> 
> No caching is expected. This is to verify the transaction against the latest state db (mutates over time), where we shall not use cache. Currently, there is not much performance concern imo, because for leader node, all pending transactions in txPool is packed in new blocks. Thus no extra work is done in validation.
> 
> ```
> {"level":"info","newTxns":0,"newStakingTxns":0,"blockGasLimit":80000000,"blockGasUsed":0,"caller":"/mnt/jenkins/workspace/harmony-release/harmony/node/worker/worker.go:183","time":"2021-04-13T19:15:52.949667012Z","message":"Block gas limit and usage info"}
> ```
> 
> But this will become a concern when transaction numbers grows to the next level, when normal transactions used up all gases, leaving no space for staking transactions. Thus staking transactions will increase in number over time, which will potentially cause performance issue here. No solution without a fork will resolve this issue imo.

We can do checks every X blocks. No need to immediately remove bad txns every single block. X could be 10

JackyWYX:
> > > how is this PR tested? also is there performance implication, I suspect that for every block all pending txns will be executed/verified again, it will be costly. any caching to do?
> > 
> > 
> > No caching is expected. This is to verify the transaction against the latest state db (mutates over time), where we shall not use cache. Currently, there is not much performance concern imo, because for leader node, all pending transactions in txPool is packed in new blocks. Thus no extra work is done in validation.
> > ```
> > {"level":"info","newTxns":0,"newStakingTxns":0,"blockGasLimit":80000000,"blockGasUsed":0,"caller":"/mnt/jenkins/workspace/harmony-release/harmony/node/worker/worker.go:183","time":"2021-04-13T19:15:52.949667012Z","message":"Block gas limit and usage info"}
> > ```
> > 
> > 
> > But this will become a concern when transaction numbers grows to the next level, when normal transactions used up all gases, leaving no space for staking transactions. Thus staking transactions will increase in number over time, which will potentially cause performance issue here. No solution without a fork will resolve this issue imo.
> 
> We can do checks every X blocks. No need to immediately remove bad txns every single block. X could be 10

Yea, this is reasonable. Let me update this

JackyWYX:
@rlan35 Updated code with some refactor and and staking transaction check buffer. I will do transaction broadcasting in another PR.

rlan35 (approved):

Issue #3662, opened 2021-04-14
./hmy utility metadata returns incorrect  block number


mirrormirage0:
**Describe the bug**
./hmy utility metadata returns stale blocknum.
If I restart the node , then the right blocknum is retrieved. 

**To Reproduce**
1. Run `./hmy utility metadata` on a node
2. It will display an old block number.
3. However, checking the logs, the node is syncing new blocks. 
4. If the node is restarted, then the command will return latest block number. Then any further calls to `./hmy utility metadata` will display the same number 

 
**Expected behavior**
The latest block number of node should be displayed.

**Screenshots**
![image](https://user-images.githubusercontent.com/29485946/114699537-7d88ae80-9d31-11eb-8cd1-e2905ddae18f.png)

**Environment (please complete the following information):**
 - OS: Ubuntu 18.04
 -

LeoHChen:
@JackyWYX can also take a look at the data returned by the RPC.

mirrormirage0:
This issue is back again. Try on **any explorer node.**

![image](https://user-images.githubusercontent.com/29485946/120244161-8e0dcd80-c27a-11eb-86f4-e5869822d642.png)

Block numbers are increasing in the log file.

JackyWYX:
Please take a look at this PR: https://github.com/harmony-one/harmony/pull/3667, and use

```
.hmy utility metadata | grep current-block-number
```

instead.

mirrormirage0:
`current-block-number` works.

PR #3667, opened 2021-04-21
[rpc] added current block number in utility metadata

JackyWYX:
## Issue

Fix https://github.com/harmony-one/harmony/issues/3662.

The block number greped in the issue is actually the block number being processing in consensus module. 

Added another field in metadata for the current block number.

## Example output 

```
{
  "id": "1",
  "jsonrpc": "2.0",
  "result": {
    "blocks-per-epoch": 32768,
    "blskey": [
      "88402a67eaa469c36ba4667f513d811d9c86a2d20579e4cb6580be3a9edc3bfc79b46e78b43babc91af61de162bfc894"
    ],
    "chain-config": {
      "chain-id": 1,
      "cross-link-epoch": 186,
      "cross-tx-epoch": 28,
      "eip155-epoch": 28,
      "eth-compatible-chain-id": 1666600000,
      "eth-compatible-epoch": 442,
      "eth-compatible-shard-0-chain-id": 1666600000,
      "five-seconds-epoch": 230,
      "istanbul-epoch": 314,
      "no-early-unlock-epoch": 530,
      "prestaking-epoch": 185,
      "quick-unlock-epoch": 191,
      "receipt-log-epoch": 101,
      "redelegation-epoch": 290,
      "s3-epoch": 28,
      "sixty-percent-epoch": 530,
      "staking-epoch": 186,
      "two-seconds-epoch": 366
    },
    "consensus": {
      "blocknum": 11954722,
      "finality": 0,
      "mode": "Syncing",
      "phase": "Announce",
      "viewChangeId": 11955062,
      "viewId": 11955062
    },
    "current-epoch": 537,
    "current-number": 11954738,  <-- this number is newly added and different from consensus.blocknum
    "dns-zone": "t.hmny.io",
    "is-archival": false,
    "is-leader": false,
    "network": "mainnet",
    "node-unix-start-time": 1618991540,
    "p2p-connectivity": {
      "connected": 201,
      "not-connected": 177,
      "total-known-peers": 378
    },
    "peerid": "QmeSejjJWSwoWSohEqmwBu6WrCWdPjWLAk4ZoVUofZd9br",
    "role": "Validator",
    "shard-id": 0,
    "sync-peers": {
      "shard-0": 3
    },
    "version": "Harmony (C) 2020. harmony, version v7000-v3.1.4-78-g6705e233 (yx@ 2021-04-21T07:47:54+0000)"
  }
}
```

## Test

Tested locally and on a node connecting to mainnet.


LeoHChen (approved):
So the original issue reported was due to the node wasn't processing any block in consensus?  Do you need to update the hmycli to support this new field?

JackyWYX:
> So the original issue reported was due to the node wasn't processing any block in consensus? 

Yes. There was only `consensus.blocknum` field which is related to block number. So when users try to query block number through metadata:

```
./hmy utility metadata | grep blocknum
```

which gives the block number in consensus. And consensus is expected to not update block number every 1000 blocks, so it will be confused for user to see after starting the node for a while, the block number in metadata is not updating.

> Do you need to update the hmycli to support this new field?

Not really. hmycli directly print the json format data to cmd. So `./hmy utility metadata` just work fine with new added field.

rpc/common/types.go
```diff
@@ -49,22 +49,23 @@ type ConsensusInternal struct {
...
-	Consensus      ConsensusInternal  `json:"consensus"`
-	C              C                  `json:"p2p-connectivity"`
-	SyncPeers      map[string]int     `json:"sync-peers",omitempty`
+	BLSPublicKey    []string           `json:"blskey"`
+	Version         string             `json:"version"`
+	NetworkType     string             `json:"network"`
+	ChainConfig     params.ChainConfig `json:"chain-config"`
+	IsLeader        bool               `json:"is-leader"`
+	ShardID         uint32             `json:"shard-id"`
+	CurrentBlockNum uint64             `json:"current-number"`
```
rlan35, line 58: current-block-number

rlan35 (approved):

Issue #3645, opened 2021-04-08
Harmony binary help shows port 9000 twice


mindstyle85:
```
$ ./harmony --help

harmony is the Harmony node binary file
...
      --dns.port int                    port of customized dns node (default 9000)
      --dns.zone string                 use customized peers from the zone for state syncing
...
      --p2p.ip string                   ip to listen for p2p protocols (default "0.0.0.0")
      --p2p.keyfile string              the p2p key file of the harmony node (default "./.hmykey")
      --p2p.port int                    port to listen for p2p protocols (default 9000)
...
```

THe dns.port should probably be at 6000, not 9000?

JackyWYX:
I will fix this after I have some bandwidth

PR #3669, opened 2021-04-22
[cmd] Syncing port fix

JackyWYX:
## Issue

Fixes https://github.com/harmony-one/harmony/issues/3645

Updated harmony config to differentiate sync.Port to port used at remote and local server.

## Test

1. Tested locally
2. Tested on mainnet with local node with
   * legacy harmony.conf for compatibility (v1.0.4)
   * latest harmony.conf with different settings (v1.0.5)
   * Different setting of choices with flags


LeoHChen:
So the dns port will be fixed to 6000, not the port - 3000 anymore?

JackyWYX:
> So the dns port will be fixed to 6000, not the port - 3000 anymore? 

Most of the port arithmetic has almost been removed at command line overhaul, but except for this one. The DNS port was also used in DNS sync registration and block push flow. So I was taking cautious steps at that time and didn't remove the -3000 stuff. This PR is taking care of this legacy stuff.

1. Before this PR, DNS port was only used to specify remote DNS service port. E.g. `--network.dns.port 9000` will connect to DNS service `t.s{0}.hmny.io:6000`. On the other side, this flag will not effect the dns server port (started locally), which is defined by `p2p.port - 3000`.
2. After this PR, `--network.dns.port 6000` will connect to DNS service at `t.s{0}.hmny.io:6000`. On the other hand, dns server is start at port specified by flag `--sync.legacy.server.port 6000`.

Thus, after this PR change,
1. No more -3000 port arithmetics except for protocol in dns sync
2. Added another configurable option on the port to start the DNS server.

> It may break multiple nodes running on the same host, even though it is not common.

Ah, thanks for you reminder. I just made a fix on this to only log the error but not exit program when the port starting dns server is occupied. So now:

1. As stated above, previously local dns server port is determined by `p2p.Port - 3000`. Changing `network.dns.port` in previous code means the node cannot connect to our sync service, which I think is not possible for a running validator node.
2. If the user update p2p.port while starting multiple node in the same machine, it can happen multiple node programs try to listen to the same port. After the fix, the error is reported in log and will not interrupt normal node execution.

JackyWYX:
@LeoHChen @rlan35 Can you review merge this? There is one bounty that depends on this PR, https://github.com/harmony-one/bounties/issues/34

JackyWYX (commented):
Really appreciate for taking up this bounty. Added quite a few comments. Other than that, is there going to be flag changes as well? Also, do you mind adding some more detailed test scenario and reports please?

Much appreciated.

rlan35 (approved):

LeoHChen (approved):

PR #3678, opened 2021-04-24
Config v2.0.0

OSadovy:
## Issue

https://github.com/harmony-one/bounties/issues/34

### Unit Test Coverage

Before:

```
ok      github.com/harmony-one/harmony/cmd/harmony      0.285s  coverage: 46.0% of statements
```

After:

```
ok      github.com/harmony-one/harmony/cmd/harmony      0.283s  coverage: 47.1% of statements
ok      github.com/harmony-one/harmony/cmd/harmony/config       0.267s  coverage: 36.5% of statements
```

### Test/Run Logs

```
root@07475f39e48e:~/go/src/github.com/harmony-one/harmony# bin/harmony -c testconfig.toml
Loaded config version 1.0.4 which is not latest (2.0.0).
Would you like to upgrade your config to the latest version? (y/n): y
previous config was stored at testconfig.toml.old
ERROR when loading bls key: stat ./.hmy/blskeys: no such file or directory
```

## Operational Checklist

    **NO**

    **NO**

    **NO**


cmd/harmony/config/config.go
```diff
@@ -255,67 +255,102 @@ func parseNetworkType(nt string) nodeconfig.NetworkType {
...
+	var header struct {
+		Version string
+	}
+	if err := toml.Unmarshal(b, &header); err != nil {
+		return HarmonyConfig{}, fmt.Errorf("reading config version: %w", err)
+	}
+	var config HarmonyConfig
+	// lexicographicaly compare
+	if header.Version <= v1TOMLConfigVersion {
+		config, err = loadHarmonyConfigV1(b)
```
JackyWYX, line 303: It would be better to `loadHarmonyConfigV1` returns `harmonyConfigV1` and then parse V1 to V2 explicitly here. Which will increase the readability as well as give convenience for future updates.

cmd/harmony/config/config.go
```diff
@@ -255,67 +255,102 @@ func parseNetworkType(nt string) nodeconfig.NetworkType {
...
+	} else {
+		err = toml.Unmarshal(b, &config)
```
JackyWYX, line 305: Refactor this to `loadHarmonyConfigV2` would increase readability IMO.

cmd/harmony/config/config.go
```diff
@@ -255,67 +255,102 @@ func parseNetworkType(nt string) nodeconfig.NetworkType {
...
 	}
 	return ioutil.WriteFile(file, b, 0644)
 }
+
+// BackupAndUpgradeConfigToTheLatestVersion backups previous version of config file and persists on the disk the latest format.
+// It returns file name of the backup.
+func BackupAndUpgradeConfigToTheLatestVersion(cfg HarmonyConfig, fileName string) (string, error) {
+	oldFileName := fileName + ".old"
+	for {
+		_, err := os.Stat(oldFileName)
```
JackyWYX, line 337: Potential forever loop and eats up all CPU resources. Would you please add a timely check logic and timeout here?

cmd/harmony/config/config_test.go
```diff
@@ -0,0 +1,302 @@
...
+var testBaseDir = ".testdata"
+
+func init() {
+	if _, err := os.Stat(testBaseDir); os.IsNotExist(err) {
+		os.MkdirAll(testBaseDir, 0777)
+	}
+}
+
+func TestV1_0_0Config(t *testing.T) {
+	testConfig := `Version = "1.0.4"
```
JackyWYX, line 37: Would you please make this data a file and add it to test data folder? This will help with future updates.

cmd/harmony/config/default.go
```diff
@@ -0,0 +1,213 @@
+package config
+
+import nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
+
+const TOMLConfigVersion = "2.0.0"
+
+const (
+	DefNetworkType = nodeconfig.Mainnet
```
JackyWYX, line 8: Make items in this file private?

cmd/harmony/main.go
```diff
@@ -157,37 +194,85 @@ func raiseFdLimits() error {
 	return nil
 }
 
-func getHarmonyConfig(cmd *cobra.Command) (harmonyConfig, error) {
+func promptForConfigUpgrade() bool {
+	if !terminal.IsTerminal(syscall.Stdin) {
+		return false
+	}
+	for {
```
JackyWYX, line 201: Would you please add a timeout mechanism if possible?

cmd/harmony/main.go
```diff
@@ -157,37 +194,85 @@ func raiseFdLimits() error {
 	return nil
 }
 
-func getHarmonyConfig(cmd *cobra.Command) (harmonyConfig, error) {
+func promptForConfigUpgrade() bool {
+	if !terminal.IsTerminal(syscall.Stdin) {
```
JackyWYX, line 198: I think it will be more readable if we can put this if statement in upper function calls. What do you think?

cmd/harmony/main.go
```diff
@@ -157,37 +194,85 @@ func raiseFdLimits() error {
...
 		nt := getNetworkType(cmd)
-		config = getDefaultHmyConfigCopy(nt)
+		cfg = config.GetDefaultHmyConfigCopy(nt)
 	}
 	if err != nil {
-		return harmonyConfig{}, err
+		return config.HarmonyConfig{}, err
 	}
-	if config.Version != defaultConfig.Version {
+	if cfg.Version != config.DefaultConfig.Version {
```
JackyWYX, line 256: Refactor to `config.IsLatestVersion(versionStr string)`? Also shall we consider the config version larger than current version?

JackyWYX (commented):
Really appreciate for taking up this bounty. Added quite a few comments. Other than that, is there going to be flag changes as well? Also, do you mind adding some more detailed test scenario and reports please?

Much appreciated.

cmd/harmony/config/bls.go
```diff
@@ -15,8 +15,8 @@ var (
 	onceLoadBLSKey sync.Once
 )
 
-// setupConsensusKeys load bls keys and set the keys to nodeConfig. Return the loaded public keys.
-func setupConsensusKeys(hc harmonyConfig, config *nodeconfig.ConfigType) multibls.PublicKeys {
+// SetupConsensusKeys load bls keys and set the keys to nodeConfig. Return the loaded public keys.
```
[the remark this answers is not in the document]
OSadovy: I thought this functionality is config related (it is converting config-related data structure) and so belongs here. which place do you think is more appropriate? is cmd/harmony/ (the old place) better?

cmd/harmony/config/config.go
```diff
@@ -7,53 +7,48 @@ import (
...
 	"strings"
 
-	"github.com/harmony-one/harmony/internal/cli"
 	nodeconfig "github.com/harmony-one/harmony/internal/configs/node"
 	"github.com/pelletier/go-toml"
-	"github.com/spf13/cobra"
 )
 
-// harmonyConfig contains all the configs user can set for running harmony binary. Served as the bridge
+// HarmonyConfig contains all the configs user can set for running harmony binary. Served as the bridge
```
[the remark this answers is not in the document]
OSadovy: actually, I think that it is better to have the latest config always to be named the same (e.g. HarmonyConfig, not HarmonyConfigV2). this will keep the fact that there might be several config versions private to the config package, and also keep the possible code changes for migrating to new config versions (v3, v4 etc) to the single package.

cmd/harmony/config/config.go
```diff
@@ -167,43 +160,55 @@ type syncConfig struct {
...
 	nodeType := config.General.NodeType
-	accepts = []string{nodeTypeValidator, nodeTypeExplorer}
+	accepts = []string{NodeTypeValidator, NodeTypeExplorer}
 	if err := checkStringAccepted("--run", nodeType, accepts); err != nil {
 		return err
 	}
 
 	netType := config.Network.NetworkType
-	parsed := parseNetworkType(netType)
+	parsed := ParseNetworkType(netType)
```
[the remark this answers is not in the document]
OSadovy: yes, it is atleast used in flags.go in getNetworkType()

cmd/harmony/config/config.go
OSadovy: My idea is that we should have a file per old config version that we want to support, e.g. v1.go, v2.go etc, which should contain the old struct together with the code to convert that struct into the latest HarmonyConfig. that way we do not clutter main load function with every possible converting logic, but rather have that logic contained close to its struct.
I am not sure that having this convert logic for each version in the one place will increase readability, considering that we might have a number of versions in the future.
Or maybe I am missing the point, then correct me, please.

cmd/harmony/config/config.go
OSadovy: without renaming the struct itself I don't see a point in renaming/refactoring this one into standalone func. the whole idea is to have the latest and greatest HarmonyConfig and related functions public and exported, and other legacy structs and code to work with them private and hidden from other code. 

this way the action list for adding a new (backward incompatible) config version would be:
* copy current struct into private struct with version name in it and put it into a standalone file, say v2.go
* change HarmonyConfig to represent the newest version of config
* add a private loadV2Config() function to v2.go to convert old to new, and call it in the LoadHarmonyConfig publicly exported function.

if you have another plan in mind, could you please elaborate how do you see an action list for changing config to the new incompatible version would be with it?

OSadovy:
> Really appreciate for taking up this bounty. Added quite a few comments. 
Many thanks for your thorough review!

>Other than that, is there going to be flag changes as well? 
I changed some flags (sync.legacy.client -> dnssync.legacy.client, sync.legacy.server -> dnssync.legacy.server) with preserving old versions as well. what else flag changes would you like to see? i thought about changing dns.zone and dns.port, but because those flags are not 100% following config naming convention, I dropped the idea (not to change just for sake of changing).

>Also, do you mind adding some more detailed test scenario and reports please?
Do you mean describing what tests I performed manually, or adding more automatic test cases?

> Much appreciated.
Thanks!

cmd/harmony/config/default.go
OSadovy: They are used in flags_test.go unfortunately. another option that I considered is to move flags-related code into config package as well, but I declined the idea, since flags are not 100% following the config.

cmd/harmony/main.go
OSadovy: do you want for this prompt to disappear in some time if user does not take any action? Frankly speaking, I never saw this kind of behavior in cli applications before. or do you want me to limit number of attempts that use can mistype their choice?

IMHO if we know for sure that user is running this in the terminal, this current behavior is what one might expect for interactive cli application. but I am opened to your suggestions, if you elaborate, please.

cmd/harmony/main.go
OSadovy: great idea! What if I add IsLatestVersion() method on the HarmonyConfig struct itself?

cmd/harmony/config/bls.go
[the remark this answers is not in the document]
JackyWYX: As far as I see, this file is more than parsing BLS config data structure. I would suggest you to put it back in `cmd/harmony`

cmd/harmony/config/config.go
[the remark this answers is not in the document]
JackyWYX: Yea, I totally agree with you about abstracting all details about config loading. What I have in mind is:

1. Latest config should always be named as `HarmonyConfig`
2. V2 should be named privately like `harmonyConfigV2`

Something like:

```
// config.go
type HarmonyConfig harmonyConfigV2

// v2.go
type harmonyConfigV2 struct {...}
```

Thus the two points you give in the comment will be still well kept:

1. Different versions are named privately.
2. Latest version is always called as `HarmonyConfig`

Moreover, his code will more explicitly show that the latest HarmonyConfig is of version V2, and will have smaller code changes when updating to a new version (Only need to redefine `type HarmonyConfig harmonyConfigV3` instead of rename the file to V2). 

Or do you have any other concerns about this? We can have a more in depth discussion.

cmd/harmony/config/config.go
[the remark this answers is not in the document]
JackyWYX: Ah I see. Nice!

cmd/harmony/config/config.go
```diff
@@ -255,67 +255,102 @@ func parseNetworkType(nt string) nodeconfig.NetworkType {
...
-		return harmonyConfig{}, err
+	var header struct {
+		Version string
+	}
+	if err := toml.Unmarshal(b, &header); err != nil {
+		return HarmonyConfig{}, fmt.Errorf("reading config version: %w", err)
+	}
+	var config HarmonyConfig
+	// lexicographicaly compare
+	if header.Version <= v1TOMLConfigVersion {
```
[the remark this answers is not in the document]
JackyWYX: I think hashicorp solution should be better IMO: https://github.com/hashicorp/go-version.

I understand that lexicographic solution will work for most of the cases if the version string is legit. But what about the case that user accidentally modified version string and don't notice the change? E.g. ` v2.0.0`. If this is the case, the config will be silently loaded as v1 and then parsed to v2, which might cause loaded config to have an unexpected result. 

So this is my concern about "hacking" with the string comparison. I would encourage a systematic approach to add the go-version package could potentially provide more stability to the system.

cmd/harmony/config/config.go
JackyWYX: Do you have any thoughts on this one?

cmd/harmony/config/config.go
JackyWYX: I am very glad to say that I definitely agree every point you have in the comment. What I have in mind is that, currently `LoadHarmonyConfig` is actually doing two jobs:

1. Load the config from toml file to its own version.
2. Parse the config to the latest version.

So I would love to see this two jobs are defined in two functions:

1. In each version file, load `toml` file to its own version. This will also handle minor version updates, E.g. v1.0.4 to v1.0.5.
2. Create a file `parse.go` to parse the legacy version to the latest.
```
// v1.go
func loadHarmonyConfigV1(file string) harmonyConfigV1 {}

// v2.go
func loadHarmonyConfigV2(file string) harmonyConfigV2 {}

// parse.go
func parseConfigToLatest(version vesion.Version, config interface{}) HarmonyConfig {
    switch version {
    ...
    }
}

func parseConfigV1ToV2(v1 harmonyConfigV1) harmonyConfigV2 {...}
func parseConfigV2ToV3(v2 harmonyConfigV2) harmonyConfigV3 {...}
```

In this case, when a new version of harmonyConfig is launched, there will be minimum update in the legacy file and functions.

cmd/harmony/config/config.go
JackyWYX: > * copy current struct into private struct with version name in it and put it into a standalone file, say v2.go
> * change HarmonyConfig to represent the newest version of config
> * add a private loadV2Config() function to v2.go to convert old to new, and call it in the LoadHarmonyConfig publicly exported function.

Yes, you nailed it! 

And here I say refactor is just to refactor this one line `toml.Unmarshal` to a function to increase readability

cmd/harmony/config/default.go
JackyWYX: Ah I see. Please just keep it as it is, thanks

cmd/harmony/main.go
JackyWYX: There is a legacy way of starting harmony node with shell script `node.sh`: https://github.com/harmony-one/harmony/blob/main/scripts/node.sh

The suggested way of running harmony at that time was to run the shell script within tmux session. Though starting with `node.sh` is not recommended since a long time ago, we still spot some users keeps running nodes with deprecated `node.sh`. Thus the prompt without a timeout will potentially cause trouble in this case. 

Though it's uncommon, I would still suggest you to add a timeout mechanism if possible. Thanks!

JackyWYX:
Thanks!

> I changed some flags (sync.legacy.client -> dnssync.legacy.client, sync.legacy.server -> dnssync.legacy.server) with preserving old versions as well. what else flag changes would you like to see? i thought about changing dns.zone and dns.port, but because those flags are not 100% following config naming convention, I dropped the idea (not to change just for sake of changing).

The legacy flags still need to be supported IMO, since some user may start harmony with some flags in systemd, and the harmony will not work for these legacy flags. I would suggest to copy the legacy flags with exactly the same names and make them backward compatible.

PR #3682, opened 2021-04-29
Add vrf proposal and verification in block logic

rlan35:
Add vrf epoch and the logic to propose vrf in a new block.

Starting from the VRF epoch, leaders are obligated to propose VRF in every block header.

The VRF is computed using the previous block's hash and the leader's private key.

The resulting vrf and proof are concatenated together and put in the vrf byte[] field of the block header.

Block verification now will verify the vrf data if needed.

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -726,49 +726,6 @@ func (consensus *Consensus) GenerateVrfAndProof(newHeader *block.Header) error {
 	return nil
 }
 
-// ValidateVrfAndProof validates a VRF/Proof from hash of previous block
-func (consensus *Consensus) ValidateVrfAndProof(headerObj *block.Header) bool {
-	vrfPk := vrf_bls.NewVRFVerifier(consensus.LeaderPubKey.Object)
-	var blockHash [32]byte
-	previousHeader := consensus.Blockchain.GetHeaderByNumber(
-		headerObj.Number().Uint64() - 1,
-	)
-	if previousHeader == nil {
-		return false
-	}
-
-	previousHash := previousHeader.Hash()
-	copy(blockHash[:], previousHash[:])
-	vrfProof := [96]byte{}
-	copy(vrfProof[:], headerObj.Vrf()[32:])
-	hash, err := vrfPk.ProofToHash(blockHash[:], vrfProof[:])
-
-	if err != nil {
-		consensus.getLogger().Warn().
-			Err(err).
-			Str("MsgBlockNum", headerObj.Number().String()).
-			Msg("[OnAnnounce] VRF verification error")
-		return false
-	}
-
-	if !bytes.Equal(hash[:], headerObj.Vrf()[:32]) {
-		consensus.getLogger().Warn().
-			Str("MsgBlockNum", headerObj.Number().String()).
-			Msg("[OnAnnounce] VRF proof is not valid")
-		return false
-	}
-
-	vrfBlockNumbers, _ := consensus.Blockchain.ReadEpochVrfBlockNums(
-		headerObj.Epoch(),
-	)
-	consensus.getLogger().Info().
-		Str("MsgBlockNum", headerObj.Number().String()).
-		Int("Number of VRF", len(vrfBlockNumbers)).
-		Msg("[OnAnnounce] validated a new VRF")
-
-	return true
-}
-
 // GenerateVdfAndProof generates new VDF/Proof from VRFs in the current epoch
 func (consensus *Consensus) GenerateVdfAndProof(newBlock *types.Block, vrfBlockNumbers []uint64) {
 	//derive VDF seed from VRFs generated in the current epoch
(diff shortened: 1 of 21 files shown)


internal/chain/engine.go
```diff
@@ -135,6 +138,104 @@ func (e *engineImpl) VerifyShardState(
...
+	}
+
+	if bc.Config().IsVRF(header.Epoch()) && len(header.Vrf()) != 128 {
+		return errors.Errorf(
+			"[VerifyVRF] invalid vrf data format or no vrf proposed %x", header.Vrf(),
+		)
+	}
+	if !bc.Config().IsVRF(header.Epoch()) {
+		if len(header.Vrf()) != 0 {
+			errors.Errorf(
```
LeoHChen, line 158: missing return?

internal/chain/engine.go
```diff
@@ -135,6 +138,104 @@ func (e *engineImpl) VerifyShardState(
...
+func (e *engineImpl) VerifyVRF(
+	bc engine.ChainReader, header *block.Header,
+) error {
+	if bc.ShardID() != header.ShardID() {
+		return errors.Errorf(
+			"[VerifyVRF] shardID not match %d %d", bc.ShardID(), header.ShardID(),
+		)
+	}
+
+	if bc.Config().IsVRF(header.Epoch()) && len(header.Vrf()) != 128 {
```
LeoHChen, line 151: hardcoded 128 number, please use constant

consensus/consensus_v2.go
```diff
@@ -774,78 +700,30 @@ func (consensus *Consensus) postCatchup(initBN uint64) {
...
-		return false
-	}
-
-	if !bytes.Equal(hash[:], headerObj.Vrf()[:32]) {
-		consensus.getLogger().Warn().
-			Str("MsgBlockNum", headerObj.Number().String()).
-			Msg("[OnAnnounce] VRF proof is not valid")
-		return false
-	}
+	vrf, proof := sk.Evaluate(previousHash[:])
```
LeoHChen, line 717: any error checking here?

LeoHChen (commented):
LGTM in general, is there any performance benchmarking done? how is it impact on the validator's CPU?

internal/chain/engine.go
rlan35: good catch. Done.

internal/chain/engine.go
rlan35: done.

consensus/consensus_v2.go
rlan35: Added based on the logic in Evaludate()

rlan35:
> LGTM in general, is there any performance benchmarking done? how is it impact on the validator's CPU?

Added benchmark. the performance is similar to BLS signatures.

Time required to sign 1000 times: 0.542673 seconds
Time required to verify sig 1000 times: 1.499797 seconds
Time required to generate vrf 1000 times: 0.525362 seconds
Time required to verify vrf 1000 times: 2.076890 seconds

consensus/consensus_v2.go
```diff
@@ -774,78 +700,34 @@ func (consensus *Consensus) postCatchup(initBN uint64) {
...
+		return errors.New("[GenerateVrfAndProof] failed to generate vrf")
 	}
 
-	if !bytes.Equal(hash[:], headerObj.Vrf()[:32]) {
-		consensus.getLogger().Warn().
-			Str("MsgBlockNum", headerObj.Number().String()).
-			Msg("[OnAnnounce] VRF proof is not valid")
-		return false
-	}
+	newHeader.SetVrf(append(vrf[:], proof...))
```
LeoHChen, line 722: is it possible vrf is nil?

LeoHChen (approved):
lgtm, get review from jacky then

consensus/consensus_v2.go
rlan35: vrf is [32]byte array. it can't be nil.

internal/configs/sharding/mainnet.go
```diff
@@ -17,9 +17,6 @@ const (
 
 	mainnetVdfDifficulty = 50000 // This takes about 100s to finish the vdf
 
-	// TODO: remove it after randomness feature turned on mainnet
-	mainnetRandomnessStartingEpoch = 100000
-
```
AlexiaChen, line 22: according to this comment, Does this note mean that the VRF random function has not been officially enabled on the main network until then? That is, the description of the relevant VRF in the white paper has not been implemented yet?

internal/chain/engine.go
```diff
@@ -135,6 +140,104 @@ func (e *engineImpl) VerifyShardState(
 	return nil
 }
 
+// VerifyVRF implements Engine, checking the vrf is valid
+func (e *engineImpl) VerifyVRF(
```
AlexiaChen, line 144: Why should the check function of VRF be moved from the consensus module to the engine module?

AlexiaChen (approved):
LGTM. Given my familiarity with the source code of this chain, I don't see any problems at the moment.

internal/configs/sharding/mainnet.go
rlan35: Yes, it was an old implementation that's never be launched in mainnet.

internal/chain/engine.go
rlan35: because verifyVRF belongs to the logic for block verification which is part of the engine logic. Not limited to consensus only.

Issue #3696, opened 2021-05-06
[Feature] rate limiter and cache for RPC


AlexiaChen:
# Problem/limitation At Hand

if requests increasing  always，the number of goroutine will increase with unlimited. the node will be killed indeed， cause there is no rate limiter now.

- Need to add rate limiter for rpc
- Add cache for RPC `GetBlockByNumber` （https://medium.com/@vCabbage/go-avoid-duplicate-requests-with-sync-singleflight-311601b3068b#989e）

for preventing OOM crash issue when  requests rapidly increasing

**Which component?**

RPC

PR #3698, opened 2021-05-07
[rpc] Add Rate Limiter for RPCs

AlexiaChen:
## Issue

https://github.com/harmony-one/harmony/issues/3696

- The current PR only supports BlockChain type RPCs, not other RPCs such as TxPool, Contract, etc. Because I don't know if restricting the flow of other RPCs will affect other products（ I'm not sure if I want to add rate limiter to all RPCs）
- I didn't add singleflight to `GetBlockByNumber` because I don't think it has much effect, because this needs to depend on the same key, but for explorer node, a large number of requests are not for the same key, because the block height is different. Unless a large number of requests are for blocks of the same height.

## Usage of rate limiter

- add two flags： `--rpc.ratelimiter` and `--rpc.ratelimit`. Or you can use it in config file, it all the same.
- the flag  `--rpc.ratelimiter` is default enabled, so you can ignore it
- the flag `-rpc.ratelimit` is used for control rate of requests, the value is the number of requests per sencond allowed with rate limiter.the default value is 100 (100/s).

so use rate limiter with shell, need to run as follows:

```bash
./harmony --rpc.ratelimiter --rpc.ratelimit=100
OR
./harmony
```

## Test

Use harmony explorer start a tons of requests, the real rate of requests is equal to config of `--rpc.ratelimit=<requests_per_second>`

## Operational Checklist

    **NO**

    **NO**

  **NO**

### DIFF
--- a/rpc/blockchain.go
+++ b/rpc/blockchain.go
@@ -23,34 +23,39 @@ import (
 	v2 "github.com/harmony-one/harmony/rpc/v2"
 	"github.com/harmony-one/harmony/shard"
 	stakingReward "github.com/harmony-one/harmony/staking/reward"
+
+	lru "github.com/hashicorp/golang-lru"
 )
 
 // PublicBlockchainService provides an API to access the Harmony blockchain.
 // It offers only methods that operate on public data that is freely available to anyone.
 type PublicBlockchainService struct {
-	hmy     *hmy.Harmony
-	version Version
-	limiter *rate.Limiter
+	hmy        *hmy.Harmony
+	version    Version
+	limiter    *rate.Limiter
+	blockCache *lru.Cache
 }
 
-var (
+const (
 	DefaultRateLimiterWaitTimeout = 5 * time.Second
+	blockCacheLimit               = 256
 )
 
 // NewPublicBlockchainAPI creates a new API for the RPC interface
 func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version, limiterEnable bool, limit int) rpc.API {
+	blockCache, _ := lru.New(blockCacheLimit)
 	if limiterEnable {
 		return rpc.API{
 			Namespace: version.Namespace(),
 			Version:   APIVersion,
-			Service:   &PublicBlockchainService{hmy, version, rate.NewLimiter(rate.Limit(limit), 1)},
+			Service:   &PublicBlockchainService{hmy, version, rate.NewLimiter(rate.Limit(limit), 1), blockCache},
 			Public:    true,
 		}
 	} else {
 		return rpc.API{
 			Namespace: version.Namespace(),
 			Version:   APIVersion,
-			Service:   &PublicBlockchainService{hmy, version, nil},
+			Service:   &PublicBlockchainService{hmy, version, nil, blockCache},
 			Public:    true,
 		}
 	}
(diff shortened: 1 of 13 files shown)


rlan35 (approved):

LeoHChen (commented):
LGTM in general, my question is about the default rate of 100/s. 

we need to spend some time tuning the rate.  How did you come up with this number?  Please work with Haodi to get some real-world value from the current mainnet.  We want to limit the request, but we also don't want to slow down the service.  Please check/analyze the existing explorer node, and get some data out of the load balancer and target group.

Thanks.

AlexiaChen:
> 
> 
> LGTM in general, my question is about the default rate of 100/s.
> 
> we need to spend some time tuning the rate. How did you come up with this number? Please work with Haodi to get some real-world value from the current mainnet. We want to limit the request, but we also don't want to slow down the service. Please check/analyze the existing explorer node, and get some data out of the load balancer and target group.
> 
> Thanks.

the default rate is the result of my test on local, it's not real-world rate. If the speed reaches 1000/s, I find that the memory still keeps growing. Because the goroutine is always blocked by the rate limiter's wait method, the number of goroutines still keeps growing cumulatively.

AlexiaChen:
> 
> 
> LGTM in general, my question is about the default rate of 100/s.
> 
> we need to spend some time tuning the rate. How did you come up with this number? Please work with Haodi to get some real-world value from the current mainnet. We want to limit the request, but we also don't want to slow down the service. Please check/analyze the existing explorer node, and get some data out of the load balancer and target group.
> 
> Thanks.

I ask Haodi the real data in the load balancer, It's 18000 requests / min in the real world, So, I set default value is 300/s(18000 / 60 = 300), That's alright ?

rpc/blockchain.go
```diff
@@ -28,16 +30,31 @@ import (
...
+
 // NewPublicBlockchainAPI creates a new API for the RPC interface
-func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version) rpc.API {
-	return rpc.API{
-		Namespace: version.Namespace(),
-		Version:   APIVersion,
-		Service:   &PublicBlockchainService{hmy, version},
-		Public:    true,
+func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version, limiterEnable bool, limit int) rpc.API {
+	if limiterEnable {
```
JackyWYX, line 42: What is the case for not limiting the rate?

rpc/blockchain.go
```diff
@@ -198,6 +225,16 @@ func (s *PublicBlockchainService) GetBlockByNumber(
 func (s *PublicBlockchainService) GetBlockByHash(
 	ctx context.Context, blockHash common.Hash, opts interface{},
 ) (response StructuredResponse, err error) {
+
+	if s.limiter != nil {
```
JackyWYX, line 229: Can we refactor this into a new function?

JackyWYX (commented):
Added some comments. BTW, have you considered migration of harmony config file? Also, it is recommended to have your config migration rebased on top of this PR after merge: https://github.com/harmony-one/harmony/pull/3681.

rpc/blockchain.go
AlexiaChen: good point!! This urgent task about OOM makes me temporarily forget an important principle in software engineering - Do not repeat yourself. Thanks for your reminder!!

rpc/blockchain.go
AlexiaChen: Now for the explorer node for the time being there is no unrestricted scenario, but I think this switch is still exposed as much as possible, this is a functional switch

PR #3699, opened 2021-05-12
[discovery] increase the interval of advertise

LeoHChen:
increase time for stream checking from 30 seconds to 300 seconds


LeoHChen:
The first chart shows a node CPU usage after the PR is applied to the latest build.
image

The 2nd chart shows another node with the original release.
image

JackyWYX:
Hey leo, Is there any sign of dropping CPU with this PR?

PR #3686, opened 2021-05-03
[tracer][rpc] go version of "trace_block"

peekpi:
## Issue

Explorer and some other products need a faster `trace_block` RPC.
The original trace_block calls a JS VM, which is too slow for products, so I rewrite the RPC in pure go language.

## Test

Before:

trace 100 blocks take around 30s

After:

trace 100 blocks take around 4s

### Test Script

```javascript

const Web3 = require('web3')

//const nodeurl = 'https://api.harmony.one'
const nodeurl = 'http://64.227.106.149:9500'

const web3 = new Web3(nodeurl)
web3.extend({
    property: 'trace',
    methods: [{
        name: 'block',
        call: 'trace_block',
        params: 1,
        inputFormatter: [web3.utils.numberToHex]
    }]
});
Time = ()=>Number(new Date())/1000

async function batchMain() {
    const latest = await web3.eth.getBlockNumber()
    console.log(latest)
    return new Promise((resolve, reject)=>{
        const result = []
        const batch = new web3.BatchRequest();
        for(let i = 100; i > 0; i--){
            const bno = latest - i;
            batch.add(web3.trace.block.request(bno, (err, res)=>{
                if(err) console.error(err)
                result.push(res)
                if(i == 1) resolve(result)
            }));
        }
        batch.execute();
    })
}

const time = ()=>Number(new Date())/1000

async function main() {
    const start = time()
    const r = await batchMain()
    console.log("main:", r.length, r.reduce((t,b)=>t+b.length, 0), time()-start)
}

main()

```

## Operational Checklist

1. **Does this PR introduce backward-incompatible changes to the on-disk data structure and/or the over-the-wire protocol?**. (If no, skip to question 8.)

    **NO**

8. **Does this PR introduce backward-incompatible changes *NOT* related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)

    **NO**

11. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**

     **NO**

PR #3703, opened 2021-05-13
Add vrf as a precompile in evm

rlan35:
Adding vrf result from the consensus (stored in header) as a precompile contract in evm.

The precompile will simply return the VRF result stored in the current block header as a 32 bytes array, regardless of the input of the precompile contract.

The address for the the vrf contract is at 0xff to avoid having conflict with potentially new precompiles from Ethereum team.

Tested with a contract using evm assembly:

```
contract VRFReader {
  function vrf() public view returns (bytes32 result) {
  	bytes32 input;
  	assembly {
  		let memPtr := mload(0x40)
                if iszero(staticcall(not(0), 0xff, input, 32, memPtr, 32)) {
        	        invalid()
                }
                result := mload(memPtr)
        }
  }
}
```

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -724,7 +724,7 @@ func (consensus *Consensus) GenerateVrfAndProof(newHeader *block.Header) error {
 	consensus.getLogger().Info().
 		Uint64("BlockNum", newHeader.Number().Uint64()).
 		Uint64("Epoch", newHeader.Epoch().Uint64()).
-		Bytes("VRF+Proof", newHeader.Vrf()).
+		Hex("VRF+Proof", newHeader.Vrf()).
 		Msg("[GenerateVrfAndProof] Leader generated a VRF")
 
 	return nil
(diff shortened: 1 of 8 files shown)


peekpi (approved):
this has a bit impact on #3686 because it uses the Istanbul version of PrecompiledContracts. but easy to fix.

AlexiaChen (approved):
variable `input` is just place holder from `staticcall` OPCODE in the context？  `0xff` is actually the address of VRF contract ? Why this address could avoid potential conflict with ETH.? is there any reference?

rlan35:
> variable `input` is just place holder from `staticcall` OPCODE in the context？ `0xff` is actually the address of VRF contract ? Why this address could avoid potential conflict with ETH.? is there any reference?

Yea, input is for complying with the interface. 0xff is at the other end of the space, while eth is working incrementally from 0 upwards.

rlan35:
> this has a bit impact on #3686 because it uses the Istanbul version of PrecompiledContracts. but easy to fix.

Cool, please have a PR to update your trace block code accordingly. thanks

PR #3711, opened 2021-05-17
[rpc] Add Rate Limiter and cache for RPCs

AlexiaChen:
## Issue

https://github.com/harmony-one/harmony/issues/3696

- The current PR only supports BlockChain type RPCs, not other RPCs such as TxPool, Contract, etc. Because I don't know if restricting the flow of other RPCs will affect other products（ I'm not sure if I want to add rate limiter to all RPCs）
- add LRU cache for `GetBlockByNumber`. if it triggers rate limiter wait timeout, the RPC will return empty, This results in a large number of visits in an instant while reducing the user experience. so need to add cache for json body of block
- combine https://github.com/harmony-one/harmony/pull/3698 into this

## Usage of rate limiter

- add two flags： `--rpc.ratelimiter` and `--rpc.ratelimit`. Or you can use it in config file, it all the same.
- the flag  `--rpc.ratelimiter` is default enabled, so you can ignore it
- the flag `-rpc.ratelimit` is used for control rate of requests, the value is the number of requests per sencond allowed with rate limiter.the default value is 100 (100/s).

so use rate limiter with shell, need to run as follows:

```bash
./harmony --rpc.ratelimiter --rpc.ratelimit=100
OR
./harmony
```

## Test

- Use harmony explorer start a tons of requests, the real rate of requests is equal to config of `--rpc.ratelimit=<requests_per_second>`

- Request same height using RPC, to check if block json is get from cache：

```bash
./hmy  blockchain block-by-number 10000
```

## Operational Checklist

    **NO**

    **NO**

  **NO**

### DIFF
--- a/rpc/blockchain.go
+++ b/rpc/blockchain.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/rpc"
+	"golang.org/x/time/rate"
 
 	"github.com/harmony-one/harmony/hmy"
 	"github.com/harmony-one/harmony/internal/chain"
@@ -21,23 +23,43 @@ import (
 	v2 "github.com/harmony-one/harmony/rpc/v2"
 	"github.com/harmony-one/harmony/shard"
 	stakingReward "github.com/harmony-one/harmony/staking/reward"
+
+	lru "github.com/hashicorp/golang-lru"
 )
 
 // PublicBlockchainService provides an API to access the Harmony blockchain.
 // It offers only methods that operate on public data that is freely available to anyone.
 type PublicBlockchainService struct {
-	hmy     *hmy.Harmony
-	version Version
+	hmy        *hmy.Harmony
+	version    Version
+	limiter    *rate.Limiter
+	blockCache *lru.Cache
 }
 
+const (
+	DefaultRateLimiterWaitTimeout = 5 * time.Second
+	blockCacheLimit               = 256
+)
+
 // NewPublicBlockchainAPI creates a new API for the RPC interface
-func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version) rpc.API {
-	return rpc.API{
-		Namespace: version.Namespace(),
-		Version:   APIVersion,
-		Service:   &PublicBlockchainService{hmy, version},
-		Public:    true,
+func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version, limiterEnable bool, limit int) rpc.API {
+	blockCache, _ := lru.New(blockCacheLimit)
+	if limiterEnable {
+		return rpc.API{
+			Namespace: version.Namespace(),
+			Version:   APIVersion,
+			Service:   &PublicBlockchainService{hmy, version, rate.NewLimiter(rate.Limit(limit), 1), blockCache},
+			Public:    true,
+		}
+	} else {
+		return rpc.API{
+			Namespace: version.Namespace(),
+			Version:   APIVersion,
+			Service:   &PublicBlockchainService{hmy, version, nil, blockCache},
+			Public:    true,
+		}
 	}
+
 }
 
 // ChainId returns the chain id of the chain - required by MetaMask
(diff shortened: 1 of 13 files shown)


rpc/blockchain.go
```diff
@@ -128,8 +150,22 @@ func (s *PublicBlockchainService) BlockNumber(ctx context.Context) (interface{},
...
 	ctx context.Context, blockNumber BlockNumber, opts interface{},
 ) (response StructuredResponse, err error) {
-	// Process arguments based on version
+
 	blockNum := blockNumber.EthBlockNumber()
+	if block, ok := s.blockCache.Get(uint64(blockNum)); ok {
+		return block.(StructuredResponse), nil
+	}
+
+	if s.limiter != nil {
```
JackyWYX, line 159: Can we refactor this into a function since this is heavily used in RPC calls?

rpc/blockchain.go
```diff
@@ -21,23 +23,43 @@ import (
...
 // NewPublicBlockchainAPI creates a new API for the RPC interface
-func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version) rpc.API {
-	return rpc.API{
-		Namespace: version.Namespace(),
-		Version:   APIVersion,
-		Service:   &PublicBlockchainService{hmy, version},
-		Public:    true,
+func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version, limiterEnable bool, limit int) rpc.API {
+	blockCache, _ := lru.New(blockCacheLimit)
+	if limiterEnable {
```
JackyWYX, line 47: What is the case for not enabling rate limiter? Is there going to be a transition period or something?

AlexiaChen: I replied on rate limiter PR

rpc/blockchain.go
AlexiaChen: Yes, it will .

rpc/blockchain.go
AlexiaChen: @JackyWYX  Now for the explorer node for the time being there is no unrestricted scenario, but I think this switch is still exposed as much as possible, this is a functional switch

JackyWYX: OK, sure. But do we really need to expose this to CLI level? @LeoHChen

JackyWYX:
Have you considered migration of harmony config file? Would you please add your config change to config file migration as introduced in #3681?

AlexiaChen:
> 
> 
> Have you considered migration of harmony config file? Would you please add your config change to config file migration of version v1.0.0 as introduced in #3681?

it seems like there is no migrations instruction from now, mention by RJ https://github.com/harmony-one/harmony/pull/3681#issuecomment-842590297.   

Then, I change my test in `config_migrations_test.go` (config 2.0.0) to fix the CI tests, I don't know if this can solve your question

JackyWYX:
For the caching part, I still have some questions. Would you please give me the answer of either of the following questions to prove the caching mechanism will help OOM issue?

1. Golang memory contains stack and queue. Stack memory is collected once it is not used, while queue memory waits for GC to manage over it. So my question is, is the JSON related memory allocated in stack or queue? If all memory taken is allocated in stack, there is actually not much optimization on memory issue with respect to OOM issue IMO.
2. On the other side, is there some data to prove that the code with cache actually have smaller memory in the system?

JackyWYX:
> it seems like there is no migrations instruction from now, mention by RJ [#3681 (comment)](https://github.com/harmony-one/harmony/pull/3681#issuecomment-842590297).
> 
> Then, I change my test in `config_migrations_test.go` (config 2.0.0) to fix the CI tests, I don't know if this can solve your question

Config migration is, user using old config file like v1.0.4 with the latest binary. In this case, the new added field, like `rpc.ratelimit` will be default to 0 when loading config file. Is this the case we expect?

AlexiaChen:
> 
> 
> For the caching part, I still have some questions. Would you please give me the answer of either of the following questions to prove the caching mechanism will help OOM issue?
> 
>     1. Golang memory contains stack and queue. Stack memory is collected once it is not used, while queue memory waits for GC to manage over it. So my question is, is the JSON related memory allocated in stack or queue? If all memory taken is allocated in stack, there is actually not much optimization on memory issue with respect to OOM issue IMO.
> 
>     2. On the other side, is there some data to prove that the code with cache actually have smaller memory in the system?

 the cache is all about User experience  if requests triggers rate limiter wait timeout, the RPC will return empty, This results in a large number of visits in an instant while reducing the user experience. so need to add cache for json body of block.

JackyWYX:
> the cache is all about User experience that I discussed with RJ. Think this case, if requests triggers rate limiter wait timeout, the RPC will return empty, This results in a large number of visits in an instant while reducing the user experience. so need to add cache for json body of block. Because most of the requests on the explorer node are for the same data, the cache is useful for this case. My previous PR also did not include cache, the previous idea is consistent with your current idea, did not take into account this case

I see. Adding cache is not about memory optimization but user experience. Got it. Thanks!

AlexiaChen:
> 
> 
> > it seems like there is no migrations instruction from now, mention by RJ [#3681 (comment)](https://github.com/harmony-one/harmony/pull/3681#issuecomment-842590297).
> > Then, I change my test in `config_migrations_test.go` (config 2.0.0) to fix the CI tests, I don't know if this can solve your question
> 
> Config migration is, user using old config file like v1.0.4 with the latest binary. In this case, the new added field, like `rpc.ratelimit` will be default to 0 when loading config file. Is this the case we expect?

got it. I will try test if it fit the expect

rlan35 (approved):

AlexiaChen:
> 
> 
> > > it seems like there is no migrations instruction from now, mention by RJ [#3681 (comment)](https://github.com/harmony-one/harmony/pull/3681#issuecomment-842590297).
> > > Then, I change my test in `config_migrations_test.go` (config 2.0.0) to fix the CI tests, I don't know if this can solve your question
> > 
> > 
> > Config migration is, user using old config file like v1.0.4 with the latest binary. In this case, the new added field, like `rpc.ratelimit` will be default to 0 when loading config file. Is this the case we expect?
> 
> got it. I will try test if it fit the expect

@JackyWYX    i fixed 1.04 migration for rate limiter config, Thank you very much!!!!! Otherwise there has been a bug in the PR here, that is, when using the configuration file, the rate limiter is not turned on by default, and its value is also 0 by default. The mechanism of the configuration file is not the same as the mechanism of the flag

PR #3712, opened 2021-05-17
Reduce min delegation from 1000 to 100

rlan35:
As per requested in the passed proposal: https://governance.harmony.one/#/staking-mainnet/proposal/Qmf5k8GH3ZBAze6uru8Qk85jpDaMjzqQbXunVXji5upg7N

### DIFF
--- a/core/staking_verifier.go
+++ b/core/staking_verifier.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"math/big"
 
+	"github.com/harmony-one/harmony/internal/params"
+
 	"github.com/harmony-one/harmony/crypto/bls"
 
 	"github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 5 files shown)


core/staking_verifier_test.go
```diff
@@ -1000,9 +1002,42 @@ func TestVerifyAndDelegateFromMsg(t *testing.T) {
...
+		{
+			// 17: small amount v2
+			sdb: makeStateDBForStake(t),
+			msg: func() staking.Delegate {
+				msg := defaultMsgDelegate()
+				msg.Amount = new(big.Int).Mul(big.NewInt(90), oneBig)
+				return msg
+			}(),
+			ds:         makeMsgCollectRewards(),
+			epoch:      big.NewInt(100),
```
AlexiaChen, line 1014: Is there any special meaning of setting `epoch` to 100 here? Or is it just consistent with the min delegation value of 100

core/staking_verifier.go
```diff
@@ -212,15 +217,21 @@ func VerifyAndDelegateFromMsg(
...
+		}
 	}
 
 	updatedValidatorWrappers := []*staking.ValidatorWrapper{}
 	delegateBalance := big.NewInt(0).Set(msg.Amount)
 	fromLockedTokens := map[common.Address]*big.Int{}
 
 	var delegateeWrapper *staking.ValidatorWrapper
-	if redelegation {
+	if chainConfig.IsRedelegation(epoch) {
```
AlexiaChen, line 234: I noticed that there are now functional adjustments about chains, are they all based on epoch Instead of a specific height value? This feels like it would be more convenient, an epoch I remember is 24 hours?

rlan35: all of our hard forks are based on epochs instead of blocks.

core/staking_verifier_test.go
rlan35: It's just a random number which is used to differentiate the epoch before and after the fork (in tests)

LeoHChen (approved):

core/staking_verifier.go
AlexiaChen: sounds great

AlexiaChen (approved):

PR #3714, opened 2021-05-18
extend EPoS bound to 35%

rlan35:
As requested in the passed governance proposal: https://governance.harmony.one/#/staking-mainnet/proposal/QmdmnVwW6ob5UfB9hNYxKsRXAyM52jQpNXsV5Vw4fNbiqa

### DIFF
--- a/staking/effective/calculate.go
+++ b/staking/effective/calculate.go
@@ -15,17 +15,14 @@ import (
 
 // medium.com/harmony-one/introducing-harmonys-effective-proof-of-stake-epos-2d39b4b8d58
 var (
-	two       = numeric.NewDecFromBigInt(big.NewInt(2))
-	c, _      = numeric.NewDecFromStr("0.15")
-	onePlusC  = numeric.OneDec().Add(c)
-	oneMinusC = numeric.OneDec().Sub(c)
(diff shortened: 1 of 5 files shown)


mindstyle85 (approved):
looks good, so as far as i understand code, you added a new variable CV2 which is 35%, then add or substract that from 1 to get the 1.35 and 0.65 bounds.

staking/effective/calculate.go
```diff
@@ -148,13 +145,24 @@ func Compute(
...
+		picks[i].EPoSStake = effectiveStake(min, max, picks[i].RawStake)
 	}
 
 	return median, picks
 }
+
+func effectiveStake(min, max, actual numeric.Dec) numeric.Dec {
+	newMax := numeric.MinDec(max, actual)
+	return numeric.MaxDec(newMax, min)
+}
```
AlexiaChen, line 168: I noticed that the function was just refactored, the algorithm didn't actually change, just the max and min calculations were extracted externally, right? This does make it clearer

AlexiaChen (approved):
LGTM

staking/effective/calculate.go
rlan35: yes correct.

PR #3715, opened 2021-05-19
Add min commission rate logic

rlan35:
As per requested in this governance proposal: https://governance.harmony.one/#/staking-mainnet/proposal/QmYbBQuDSZbMVa6RBi3kzFiNkXxv8PMp9rrT25drkxQnTb

Specifically:
1. Add logic to set first elected epoch for all elected validators starting from the fork epoch (note the first elected epoch setting is not retroactive and only take effect after the fork epoch)
2. Enforce the minimum commission fee of 5% if the current epoch is 100/10 (mainnet/testnet) more than the validator's first elected epoch.
3. Add logic to enforce the minimum commission fee in EditValidator so validators can't go below 5% for commission fee once it's pass 100/10 (mainnet/testnet) epochs after first elected.

### DIFF
--- a/staking/availability/measure.go
+++ b/staking/availability/measure.go
@@ -243,7 +243,7 @@ func UpdateMinimumCommissionFee(
 	if nodeconfig.GetDefaultConfig().GetNetworkType() != nodeconfig.Mainnet {
 		promoPeriod = 10
 	}
-	if firstElectionEpoch.Uint64() != 0 && firstElectionEpoch.Sub(electionEpoch, firstElectionEpoch).Int64() >= int64(promoPeriod) {
+	if firstElectionEpoch.Uint64() != 0 && big.NewInt(0).Sub(electionEpoch, firstElectionEpoch).Int64() >= int64(promoPeriod) {
 		if wrapper.Rate.LT(MinCommissionRate) {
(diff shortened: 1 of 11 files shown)


staking/availability/measure.go
```diff
@@ -215,3 +221,37 @@ func ComputeAndMutateEPOSStatus(
...
+	wrapper, err := state.ValidatorWrapper(addr)
+	if err != nil {
+		return err
+	}
+
+	firstElectionEpoch := state.GetValidatorFirstElectionEpoch(addr)
+
+	promoPeriod := 100
+	if nodeconfig.GetDefaultConfig().GetNetworkType() != nodeconfig.Mainnet {
+		promoPeriod = 10
```
AlexiaChen, line 244: promoPeriod = 10 is random value right? Because except for the main network which is the first 100 epochs(proposal descrition), other networks don't care much about this value

AlexiaChen (approved):

staking/availability/measure.go
rlan35: yes, just want to make it faster so we can test it early.

rlan35:
> All logics are good. Just one minor issue.

Good point done.

Issue #3694, opened 2021-05-06
hmy_getAllDelegationInformation RPC interface returns useless data


jhd2best:
**Describe the bug**
hmy_getAllDelegationInformation RPC interface returns no-use data(delegation which amout is 0). that will cause a lot of additional server bandwidth usage and meaningless performance consumption.

**To Reproduce**
```
curl -d '{
    "jsonrpc":"2.0",
    "method":"hmy_getAllDelegationInformation",
    "params":[-1],
    "id":1
}' -H 'Content-Type:application/json' -X POST 'https://api.s0.t.hmny.io'
```

**Expected behavior**
Filter delegation and undelegation records whose amount is 0

PR #3717, opened 2021-05-20
[rpc] rpc skip delegations with zero amount and empty undelegation

JackyWYX:
## Issue

As requested in https://github.com/harmony-one/harmony/issues/3694, now delegation rpc will not return delegation with zero amount and empty undelegation.

This will potentially break the backward compatibility. Please make sure there is not external dependency on this two functions before going forward.

```
hmy_getDelegationsByValidator
hmy_getAllDelegationInformation
```

## Test

Tested locally with explorer node on mainnet.

Result dumped here: https://gist.github.com/JackyWYX/a77f78f432971aed85a950b76d712fa5

Noticed that there are validators with zero delegation entries, is this expected? Any further change request? @jhd2best

### DIFF
--- a/rpc/staking.go
+++ b/rpc/staking.go
@@ -6,15 +6,13 @@ import (
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
-	staking "github.com/harmony-one/harmony/staking/types"
-
-	"github.com/pkg/errors"
-
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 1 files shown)


jhd2best:
yeah, they should be validators that have been withdrawn. The staking dashboard does not show validators below 10,000, but they do exist. no further change request. thanks

jhd2best:
btw. the request is from this one https://github.com/harmony-one/harmony/issues/3694

jhd2best (approved):

rlan35 (approved):

PR #3728, opened 2021-06-03
[rpc] Improve GetBlocks

AlexiaChen:
<del>have not tested it with @hypnagonia</del>

### DIFF
--- a/rpc/blockchain.go
+++ b/rpc/blockchain.go
@@ -165,6 +165,11 @@ func (s *PublicBlockchainService) GetBlockByNumber(
 		return block.(StructuredResponse), nil
 	}
 
+	err = s.wait(ctx)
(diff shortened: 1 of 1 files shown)


LeoHChen:
any explanation for this improvement?  Please elaborate it a bit more.

AlexiaChen:
@LeoHChen   Because @hypnagonia gave me feedback that the GetBlocks timeout doesn't look like 5 seconds, it looks like minutes. The reason is that the explorer app only calls GetBlocks and not GetBlockByNumber directly, the RPC GetBlocks is to get a height range of blocks and indirectly calls GetBlockByNumber many times. GetBlocks and GetBlockByNumber both have rate limiter wait methods, so the timeout time seems to be longer on the client side

PR #3729, opened 2021-06-04
Fix Ticker cannot stop for releasing resource

AlexiaChen:
Ticker Need to use like this: https://github.com/harmony-one/harmony/blob/b602e8d454694373ab8f68b4e95cfd3a60670bb9/core/tx_pool.go#L317-L324

Do not forget 

```go
defer ticker.Stop()
```

if forget do such thing, maybe leading this meory leak like this: 
- https://github.com/google/goexpect/issues/3
- https://github.com/google/goexpect/pull/4

### DIFF
--- a/rpc/filters/api.go
+++ b/rpc/filters/api.go
@@ -67,6 +67,7 @@ func (api *PublicFilterAPI) isEth() bool {
 func (api *PublicFilterAPI) timeoutLoop() {
 	// TODO ek – infinite loop; add shutdown/cleanup logic
 	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
 	for {
 		<-ticker.C
 		api.filtersMu.Lock()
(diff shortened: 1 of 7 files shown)


rpc/filters/api.go
```diff
@@ -67,6 +67,7 @@ func (api *PublicFilterAPI) isEth() bool {
 func (api *PublicFilterAPI) timeoutLoop() {
 	// TODO ek – infinite loop; add shutdown/cleanup logic
 	ticker := time.NewTicker(5 * time.Minute)
+	defer ticker.Stop()
```
JackyWYX, line 70: Do you think this can be the potential memory leak reason?

JackyWYX (approved):
Thanks for pointing that out. That were definitely some wrong usage of ticker. Correcting them shall be great.

LeoHChen (approved):
I felt this may contribute to the memory leakage

rpc/filters/api.go
AlexiaChen: I just corret them all, Some of this code will not have memory leaks because it is an infinite loop and does not exit.

PR #3746, opened 2021-06-09
[SYNC] Add a hard cap to getBlocks request at server end

JackyWYX:
## Issue

Add a hard cap to getBlocks request at server end. The server will send a maximum of 4MB message over DNS sync service.

 This is a temporary fix to an unusual case where a block sync request exceededing gRPC 4MB message upper limit. 

## Test

Need to be tested on server to see whether this helps a validator with old binary file (still have batch size of 15) to continue syncing.

### DIFF
--- a/node/node_syncing.go
+++ b/node/node_syncing.go
@@ -31,6 +31,10 @@ import (
 // Constants related to doing syncing.
 const (
 	SyncFrequency = 60
+
+	// getBlocksRequestHardCap is the hard capped message size at server side for getBlocks request.
+	// The number is 4MB (default gRPC message size) minus 2k reserved for message overhead.
+	getBlocksRequestHardCap = 4*1024*1024 - 2*1024
 )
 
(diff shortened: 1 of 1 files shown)


JackyWYX:
@LeoHChen Can you help test with this one? Just to see whether validators not updated can continue with the beacon sync.

node/node_syncing.go
```diff
@@ -478,10 +483,14 @@ func (node *Node) CalculateResponse(request *downloader_pb.DownloaderRequest, in
...
 			}
-
-			if err == nil {
-				response.Payload = append(response.Payload, encoded)
+			if err != nil {
+				continue
+			}
+			payloadSize += len(encoded)
+			if payloadSize > getBlocksRequestHardCap {
+				break
```
rlan35, line 491: add log here.

rlan35 (approved):

PR #3764, opened 2021-06-10
add ddos protection at pubsub

JackyWYX:
## Issue

To prevent DDOS attack at beacon pub-sub, add a rate limiter per peer at ddos protection. 

## Test

No test done for now.

### DIFF
--- a/node/ddos.go
+++ b/node/ddos.go
@@ -1,0 +1,104 @@
+package node
+
+import (
+	"sync"
+
+	libp2p_peer "github.com/libp2p/go-libp2p-core/peer"
+	libp2p_pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"golang.org/x/time/rate"
+)
+
+const (
+	pubSubRateLimit  = 3  // 2 messages per second
+	pubSubBurstLimit = 10 // 10 messages at burst
+)
+
+// Preconfigured internal nodes
+// TODO: replace all internal node peer ID here
+var trustedPeers []libp2p_peer.ID
+
+func getDefaultTrustedPeerMap() map[libp2p_peer.ID]struct{} {
+	peerMap := make(map[libp2p_peer.ID]struct{})
+	for _, peer := range trustedPeers {
+		peerMap[peer] = struct{}{}
+	}
+	return peerMap
+}
+
+// TODO: maintain a timed list to avoid memory leak (growing forever)
+// TODO: make each message with a weight
+type psRateLimiter struct {
+	limiters     map[libp2p_peer.ID]*rate.Limiter
+	trustedPeers map[libp2p_peer.ID]struct{}
+
+	lock sync.Mutex
+}
+
+func newPSRateLimiter() *psRateLimiter {
+	return &psRateLimiter{
+		limiters:     make(map[libp2p_peer.ID]*rate.Limiter),
+		trustedPeers: getDefaultTrustedPeerMap(),
+	}
+}
+
+// Allow returns whether a pub-sub message is allowed to be processed
+func (rl *psRateLimiter) Allow(id libp2p_peer.ID) bool {
+	rl.lock.Lock()
+	defer rl.lock.Unlock()
+
+	if rl.isTrusted(id) {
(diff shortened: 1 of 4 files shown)


JackyWYX:
After running on a s0 explorer node for several minutes, here is the new prometheus metrics:

```
# HELP hmy_p2p_rate_blacklisted number of pub-sub message rejected and dropped since the sender is blacklisted
# TYPE hmy_p2p_rate_blacklisted counter
hmy_p2p_rate_blacklisted{topic="harmony/0.0.1/client/beacon"} 875
hmy_p2p_rate_blacklisted{topic="harmony/0.0.1/node/beacon"} 913
# HELP hmy_p2p_rate_limited_msg number of pub-sub message rejected and dropped due to rate limit
# TYPE hmy_p2p_rate_limited_msg counter
hmy_p2p_rate_limited_msg{topic="harmony/0.0.1/client/beacon"} 97
hmy_p2p_rate_limited_msg{topic="harmony/0.0.1/node/beacon"} 55
`

**Review by rlan35 — APPROVED** [review]

**Comment by JackyWYX** [comment]
Added the grace period during bootstrap. With current parameter, there is no evidence of nodes blocked at testnet. Have the testnet node run for a while.

Meanwhile, run the latest code on mainnet shows:

```
hmy_p2p_rate_limited_msg{topic="harmony/0.0.1/client/beacon"} 49
hmy_p2p_rate_limited_msg{topic="harmony/0.0.1/node/beacon"} 60
```

It grows fast. Note that currently it can be explorer nodes being blocked. We shall add all our IPs at the trusted peers.

**Comment by JackyWYX** [comment]
Another finding is that there is a bug in libp2p that causes a panic because of the usage of `WithBlacklist` option (https://github.com/libp2p/go-libp2p-pubsub/issues/426).

The current solution is not to use libp2p blacklist, but use a local one at `StartPubSub`.

After this change, this is the expected prometheus metrics at mainnet machine:

```
# TYPE hmy_p2p_rate_blacklisted counter
hmy_p2p_rate_blacklisted{topic="harmony/0.0.1/client/beacon"} 6112
hmy_p2p_rate_blacklisted{topic="harmony/0.0.1/node/beacon"} 5687
# HELP hmy_p2p_rate_limited_msg number of pub-sub message rejected and dropped due to rate limit
# TYPE hmy_p2p_rate_limited_msg counter
hmy_p2p_rate_limited_msg{topic="harmony/0.0.1/client/beacon"} 13
hmy_p2p_rate_limited_msg{topic="harmony/0.0.1/node/beacon"} 6
```

File: node/ddos.go [hunk 1]
```diff
@@ -0,0 +1,166 @@
...
+	"time"
+
+	"github.com/harmony-one/abool"
+	"github.com/harmony-one/harmony/p2p"
+	libp2p_peer "github.com/libp2p/go-libp2p-core/peer"
+	libp2p_pubsub "github.com/libp2p/go-libp2p-pubsub"
+	"golang.org/x/time/rate"
+)
+
+const (
```
LeoHChen, line 14: all those numbers have to be calibrated on the mainnet, testnet.

node/ddos.go
```diff
@@ -0,0 +1,166 @@
...
+	pubSubBurstLimit = 20 // 20 messages at burst
+
+	// When a node is bootstrapped, it will be flooded with some pub-sub message from the past.
+	// Ease the rate limit at the initial bootstrap.
+	rateLimiterEasyPeriod = 5 * time.Second
+)
+
+// Preconfigured internal nodes
+// TODO: replace all internal node peer ID here
+var trustedPeers = []string{
```
LeoHChen, line 25: no guarantee the trustedPeerID won't change, they are generated randomly.  If we change the validator nodes, they will be different.

did you consider all the bootnodes, explorer nodes, snapshot nodes?

LeoHChen (commented):
LGTM in general, just want to have more test and calibration on testnet/mainnet.

I would also suggest having an external file contains the list of trusted peer so that it would be easier to update the list w/o changing the code.

PR #3768, opened 2021-06-12
stop invalid beacon block broadcast

rlan35:


JackyWYX (approved):

LeoHChen:
So, this PR will make sure the nodes in the beacon chain will also do the checking and validation of the beacon block and stop the spamming if the beacon block is not valid.  Previously the nodes in the beacon chain will not ignore the invalid messages, but not stop the broadcast.

PR #3769, opened 2021-06-12
a few spam protections

rlan35:
Update minimum txn gas price to 1 gwei
Remove temporary block txn size limit
Separate txn size limit from p2p node data limit
Update _gasPrice api to return 10Gwei as recommended gas price.


JackyWYX (approved):
The code looks good to me :) Just need to be verified in community.

PR #3773, opened 2021-06-14
Implement pprof module extension

niklr:
## Issue
@JackyWYX 

[harmony-one/bounties#46](https://github.com/harmony-one/bounties/issues/46)

### Unit Test Coverage

`go test -cover ./cmd/harmony/`

Before:

```
ok  	github.com/harmony-one/harmony/cmd/harmony	0.266s	coverage: 47.8% of statements
```

After:

```
ok  	github.com/harmony-one/harmony/cmd/harmony	0.136s	coverage: 48.5% of statements
```

### Test/Run Logs
When running and stdin is terminal

```
./harmony -c config_1_0_4.toml 
Old config version detected 1.0.4
Do you want to update config to the latest version: [y/N]
y
Original config backed up to config_1_0_4.toml.backup
Successfully migrated config_1_0_4.toml from 1.0.4 to 2.1.0 
```

When stdin is piped

```
echo Y | ./harmony -c config_1_0_4.toml 
Old config version detected 1.0.4
Update saved config with `./harmony config update [config_file]`
```

## Operational Checklist
   **NO**
2. **Does this PR introduce backward-incompatible changes _NOT_ related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)
   **NO**

### pprof basics

pprof lets you collect CPU profiles, traces, and heap profiles. The normal way to use pprof is:

- Run `./harmony --pprof`
- Run `curl localhost:6060/debug/pprof/{profile_name} --output profile.pb.gz` to save a profile
- Use `go tool pprof profile.pb.gz` to analyze said profile

### pprof profiles

A Profile is a collection of stack traces showing the call sequences that led to instances of a particular event, such as allocation. Packages can create and maintain their own profiles; the most common use is for tracking resources that must be explicitly closed, such as files or network connections.

Each Profile has a unique name. A few profiles are predefined:

- goroutine    - stack traces of all current goroutines
- heap         - a sampling of memory allocations of live objects
- allocs       - a sampling of all past memory allocations
- threadcreate - stack traces that led to the creation of new OS threads
- block        - stack traces that led to blocking on synchronization primitives
- mutex        - stack traces of holders of contended mutexes

For more information see [runtime/pprof docs](https://golang.org/pkg/runtime/pprof)

A special profile accepted by `--pprof.profile.names`
- cpu          - profiles the current process

### pprof config

- --pprof                           enable pprof profiling
- --pprof.addr string               listen address for pprof (default "127.0.0.1:6060")
- --pprof.folder string             folder to put pprof profiles (default "./profiles")
- --pprof.profile.names strings     a list of pprof profile names (separated by ,) e.g. cpu,heap,goroutine
- --pprof.profile.intervals ints    a list of pprof profile interval integer values (separated by ,)
- --pprof.profile.debug ints        a list of pprof profile debug integer values (separated by ,)

**Example 1**

`./harmony --pprof.profile.names cpu,heap --pprof.profile.intervals 0,10`

Saves the cpu and heap profiles in the `./profiles` folder. The cpu profile will be saved on shutdown and the heap profile will be saved every 10 seconds.

**Example 2**

`./harmony --pprof.profile.names cpu,heap,goroutine --pprof.profile.intervals 10 --pprof.profile.debug 0,0,2`

Saves the cpu, heap and goroutine profiles in the `./profiles` folder. All profiles will be saved every 10 seconds. 
The goroutine profile will use debug=2 which means to print the goroutine stacks in the same form that a Go program uses when dying due to an unrecovered panic.

### DIFF
--- a/api/service/pprof/service.go
+++ b/api/service/pprof/service.go
@@ -50,6 +50,7 @@ var (
 	initOnce sync.Once
 	svc      = &Service{}
 	cpuFile  *os.File
+	lock     sync.Mutex
 )
 
 // NewService creates the new pprof service
@@ -75,11 +76,6 @@ func newService(cfg Config) *Service {
 	}
 	svc.profiles = profiles
 
-	go func() {
-		utils.Logger().Info().Str("address", cfg.ListenAddr).Msg("starting pprof HTTP service")
-		http.ListenAndServe(cfg.ListenAddr, nil)
-	}()
-
 	return svc
 }
 
@@ -93,6 +89,11 @@ func (s *Service) Start() error {
 	if err != nil {
 		return err
 	}
+
+	go func() {
+		utils.Logger().Info().Str("address", s.config.ListenAddr).Msg("starting pprof HTTP service")
+		http.ListenAndServe(s.config.ListenAddr, nil)
+	}()
 
 	if _, ok := s.profiles[CPU]; ok {
 		// The nature of the pprof CPU profile is fundamentally different to the other profiles, because it streams output to a file during profiling.
@@ -177,6 +178,8 @@ func saveProfile(profile Profile, dir string) error {
 
 // restartCpuProfile stops the current CPU profile, if any and then starts a new CPU profile. While profiling in the background, the profile will be buffered and written to a file.
 func restartCpuProfile(dir string) error {
+	lock.Lock()
+	defer lock.Unlock()
 	stopCpuProfile()
 	f, err := newTempFile(dir, CPU, ".pb.gz")
 	if err != nil {
--- a/api/service/pprof/service.go
+++ b/api/service/pprof/service.go
@@ -49,6 +49,7 @@ type Service struct {
 var (
 	initOnce sync.Once
 	svc      = &Service{}
+	cpuFile  *os.File
 )
 
 // NewService creates the new pprof service
@@ -61,21 +62,21 @@ func NewService(cfg Config) *Service {
 
 func newService(cfg Config) *Service {
 	if !cfg.Enabled {
-		utils.Logger().Info().Msg("Pprof service disabled...")
+		utils.Logger().Info().Msg("pprof service disabled...")
 		return nil
 	}
 
-	utils.Logger().Debug().Str("cfg", cfg.String()).Msg("Pprof")
+	utils.Logger().Debug().Str("cfg", cfg.String()).Msg("pprof")
 	svc.config = cfg
 
-	if profiles, err := cfg.unpackProfilesIntoMap(); err != nil {
-		log.Fatal("Could not unpack pprof profiles into map")
-	} else {
-		svc.profiles = profiles
+	profiles, err := cfg.unpackProfilesIntoMap()
+	if err != nil {
+		log.Fatal("could not unpack pprof profiles into map")
 	}
+	svc.profiles = profiles
 
 	go func() {
-		utils.Logger().Info().Str("address", cfg.ListenAddr).Msg("Starting pprof HTTP service")
+		utils.Logger().Info().Str("address", cfg.ListenAddr).Msg("starting pprof HTTP service")
 		http.ListenAndServe(cfg.ListenAddr, nil)
 	}()
 
@@ -93,8 +94,10 @@ func (s *Service) Start() error {
 		return err
 	}
 
-	if cpuProfile, ok := s.profiles[CPU]; ok {
-		go handleCpuProfile(cpuProfile, dir)
+	if _, ok := s.profiles[CPU]; ok {
+		// The nature of the pprof CPU profile is fundamentally different to the other profiles, because it streams output to a file during profiling.
+		// Thus it has to be started outside of the defined interval.
+		go restartCpuProfile(dir)
 	}
 
 	for _, profile := range s.profiles {
@@ -112,9 +115,12 @@ func (s *Service) Stop() error {
 	}
 	for _, profile := range s.profiles {
 		if profile.Name == CPU {
-			pprof.StopCPUProfile()
+			stopCpuProfile()
 		} else {
-			saveProfile(profile, dir)
+			err := saveProfile(profile, dir)
+			if err != nil {
+				utils.Logger().Error().Err(err).Msg(fmt.Sprintf("could not save pprof profile: %s", profile.Name))
+			}
 		}
 	}
 	return nil
@@ -135,9 +141,15 @@ func scheduleProfile(profile Profile, dir string) {
 				select {
 				case <-ticker.C:
 					if profile.Name == CPU {
-						handleCpuProfile(profile, dir)
+						err := restartCpuProfile(dir)
+						if err != nil {
+							utils.Logger().Error().Err(err).Msg("could not start pprof CPU profile")
+						}
 					} else {
-						saveProfile(profile, dir)
+						err := saveProfile(profile, dir)
+						if err != nil {
+							utils.Logger().Error().Err(err).Msg(fmt.Sprintf("could not save pprof profile: %s", profile.Name))
+						}
 					}
 				}
 			}
@@ -149,30 +161,39 @@ func scheduleProfile(profile Profile, dir string) {
 func saveProfile(profile Profile, dir string) error {
 	f, err := newTempFile(dir, profile.Name, ".pb.gz")
 	if err != nil {
-		utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not save pprof profile: %s", profile.Name))
 		return err
 	}
 	defer f.Close()
-	if profile.ProfileRef != nil {
-		err = profile.ProfileRef.WriteTo(f, profile.Debug)
-		if err != nil {
-			utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not write pprof profile: %s", profile.Name))
-			return err
-		}
-		utils.Logger().Info().Msg(fmt.Sprintf("Saved pprof profile in: %s", f.Name()))
+	if profile.ProfileRef == nil {
+		return nil
 	}
+	err = profile.ProfileRef.WriteTo(f, profile.Debug)
+	if err != nil {
+		return err
+	}
+	utils.Logger().Info().Msg(fmt.Sprintf("saved pprof profile in: %s", f.Name()))
 	return nil
 }
 
-// handleCpuProfile handles the provided CPU profile
-func handleCpuProfile(profile Profile, dir string) {
+// restartCpuProfile stops the current CPU profile, if any and then starts a new CPU profile. While profiling in the background, the profile will be buffered and written to a file.
+func restartCpuProfile(dir string) error {
+	stopCpuProfile()
+	f, err := newTempFile(dir, CPU, ".pb.gz")
+	if err != nil {
+		return err
+	}
+	pprof.StartCPUProfile(f)
+	cpuFile = f
+	utils.Logger().Info().Msg(fmt.Sprintf("saved pprof CPU profile in: %s", f.Name()))
+	return nil
+}
+
+// stopCpuProfile stops the current CPU profile, if any
+func stopCpuProfile() {
 	pprof.StopCPUProfile()
-	f, err := newTempFile(dir, profile.Name, ".pb.gz")
-	if err == nil {
-		pprof.StartCPUProfile(f)
-		utils.Logger().Info().Msg(fmt.Sprintf("Saved pprof CPU profile in: %s", f.Name()))
-	} else {
-		utils.Logger().Error().Err(err).Msg("Could not start pprof CPU profile")
+	if cpuFile != nil {
+		cpuFile.Close()
+		cpuFile = nil
 	}
 }
 
@@ -185,8 +206,8 @@ func (config *Config) unpackProfilesIntoMap() (map[string]Profile, error) {
 	for index, name := range config.ProfileNames {
 		profile := Profile{
 			Name:     name,
-			Interval: 0,
-			Debug:    0,
+			Interval: 0, // 0 saves the profile when stopping the service
+			Debug:    0, // 0 writes the gzip-compressed protocol buffer
 		}
 		// Try set interval value
 		if len(config.ProfileIntervals) == len(config.ProfileNames) {
@@ -203,7 +224,7 @@ func (config *Config) unpackProfilesIntoMap() (map[string]Profile, error) {
 		// Try set the profile reference
 		if profile.Name != CPU {
 			if p := pprof.Lookup(profile.Name); p == nil {
-				return nil, fmt.Errorf("Pprof profile does not exist: %s", profile.Name)
+				return nil, fmt.Errorf("pprof profile does not exist: %s", profile.Name)
 			} else {
 				profile.ProfileRef = p
 			}
@@ -219,7 +240,7 @@ func newTempFile(dir, name, suffix string) (*os.File, error) {
 	currentTime := time.Now().Unix()
 	f, err := os.OpenFile(filepath.Join(dir, fmt.Sprintf("%s%d%s", prefix, currentTime, suffix)), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
 	if err != nil {
-		return nil, fmt.Errorf("Could not create file of the form %s%d%s", prefix, currentTime, suffix)
+		return nil, fmt.Errorf("could not create file of the form %s%d%s", prefix, currentTime, suffix)
 	}
 	return f, nil
 }
--- a/api/service/pprof/service.go
+++ b/api/service/pprof/service.go
@@ -149,17 +149,17 @@ func scheduleProfile(profile Profile, dir string) {
 func saveProfile(profile Profile, dir string) error {
 	f, err := newTempFile(dir, profile.Name, ".pb.gz")
 	if err != nil {
-		utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not save profile: %s", profile.Name))
+		utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not save pprof profile: %s", profile.Name))
 		return err
 	}
 	defer f.Close()
 	if profile.ProfileRef != nil {
 		err = profile.ProfileRef.WriteTo(f, profile.Debug)
 		if err != nil {
-			utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not write profile: %s", profile.Name))
+			utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not write pprof profile: %s", profile.Name))
 			return err
 		}
-		utils.Logger().Info().Msg(fmt.Sprintf("Saved profile in: %s", f.Name()))
+		utils.Logger().Info().Msg(fmt.Sprintf("Saved pprof profile in: %s", f.Name()))
 	}
 	return nil
 }
@@ -170,9 +170,9 @@ func handleCpuProfile(profile Profile, dir string) {
 	f, err := newTempFile(dir, profile.Name, ".pb.gz")
 	if err == nil {
 		pprof.StartCPUProfile(f)
-		utils.Logger().Info().Msg(fmt.Sprintf("Saved CPU profile in: %s", f.Name()))
+		utils.Logger().Info().Msg(fmt.Sprintf("Saved pprof CPU profile in: %s", f.Name()))
 	} else {
-		utils.Logger().Error().Err(err).Msg("Could not start CPU profile")
+		utils.Logger().Error().Err(err).Msg("Could not start pprof CPU profile")
 	}
 }
 
@@ -203,7 +203,7 @@ func (config *Config) unpackProfilesIntoMap() (map[string]Profile, error) {
 		// Try set the profile reference
 		if profile.Name != CPU {
 			if p := pprof.Lookup(profile.Name); p == nil {
-				return nil, fmt.Errorf("Profile does not exist: %s", profile.Name)
+				return nil, fmt.Errorf("Pprof profile does not exist: %s", profile.Name)
 			} else {
 				profile.ProfileRef = p
 			}
@@ -219,7 +219,7 @@ func newTempFile(dir, name, suffix string) (*os.File, error) {
 	currentTime := time.Now().Unix()
 	f, err := os.OpenFile(filepath.Join(dir, fmt.Sprintf("%s%d%s", prefix, currentTime, suffix)), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666)
 	if err != nil {
-		return nil, fmt.Errorf("could not create file of the form %s%d%s", prefix, currentTime, suffix)
+		return nil, fmt.Errorf("Could not create file of the form %s%d%s", prefix, currentTime, suffix)
 	}
 	return f, nil
 }
--- a/api/service/pprof/service.go
+++ b/api/service/pprof/service.go
@@ -1,0 +1,225 @@
+package pprof
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime/pprof"
+	"sync"
+	"time"
+
+	"github.com/ethereum/go-ethereum/rpc"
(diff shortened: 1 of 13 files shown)


niklr:
## Test case 1 - Starting harmony without any arguments or config
Running:
`./harmony`

Result:

```
Staking mode; node key 02c0db3387d1acba29266e087bb45d29ce71d6f9271f7d3c44c3ad346cb999005f0d268850709c91c4d5620770012f99; -> shard 1
Started RPC server at: 127.0.0.1:9500
Started WS server at: 127.0.0.1:9800
```

Problems Found: None

## Test case 2 - Passing --pprof --pprof.folder ./profiles --pprof.profile.names cpu,heap,goroutine --pprof.profile.intervals 10 --pprof.profile.debug 0,0,2
Running:

```
./harmony --pprof --pprof.folder ./profiles --pprof.profile.names cpu,heap,goroutine --pprof.profile.intervals 10 --pprof.profile.debug 0,0,2
```

Result:

```
Staking mode; node key 02c0db3387d1acba29266e087bb45d29ce71d6f9271f7d3c44c3ad346cb999005f0d268850709c91c4d5620770012f99; -> shard 1
Started RPC server at: 127.0.0.1:9500
Started WS server at: 127.0.0.1:9800
```

Problems Found: None

## Test case 3 Running with config 1.0.2
Running:

`./harmony -c config_1_0_2.toml`

Result:

```
Old config version detected 1.0.2
Do you want to update config to the latest version: [y/N]

Staking mode; node key 02c0db3387d1acba29266e087bb45d29ce71d6f9271f7d3c44c3ad346cb999005f0d268850709c91c4d5620770012f99; -> shard 1
Started RPC server at: 127.0.0.1:9500
Started WS server at: 127.0.0.1:9800
```

Problems Found: None

## Test case 4 Running with config 1.0.3
Running:

`./harmony -c config_1_0_3.toml`

Result:

```
Old config version detected 1.0.3
Do you want to update config to the latest version: [y/N]

Staking mode; node key 02c0db3387d1acba29266e087bb45d29ce71d6f9271f7d3c44c3ad346cb999005f0d268850709c91c4d5620770012f99; -> shard 1
Started RPC server at: 127.0.0.1:9500
Started WS server at: 127.0.0.1:9800
```

Problems Found: None

## Test case 5 Running with config 1.0.4
Running:

`./harmony -c config_1_0_4.toml`

Result:

```
Old config version detected 1.0.4
Do you want to update config to the latest version: [y/N]

Staking mode; node key 02c0db3387d1acba29266e087bb45d29ce71d6f9271f7d3c44c3ad346cb999005f0d268850709c91c4d5620770012f99; -> shard 1
Started RPC server at: 127.0.0.1:9500
Started WS server at: 127.0.0.1:9800
```

Problems Found: None

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+}
+
+// handleCpuProfile handles the provided CPU profile by stopping or starting it as specified
+func handleCpuProfile(profile Profile, dir string, shouldStop bool) {
+	if shouldStop {
+		pprof.StopCPUProfile()
+	}
+	f, err := newTempFile(dir, profile.Name, ".pb.gz")
+	if err == nil {
```
JackyWYX, line 168: Close the file after finished using it?

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+	} else {
+		utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not save profile: %s", profile.Name))
+	}
+}
+
+// handleCpuProfile handles the provided CPU profile by stopping or starting it as specified
+func handleCpuProfile(profile Profile, dir string, shouldStop bool) {
+	if shouldStop {
+		pprof.StopCPUProfile()
```
JackyWYX, line 165: If you look into the implementation, you could probably find that there is already a in process CPU profiling check. I think it is safe to remove `shouldStop` check.

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+
+// unpackProfilesIntoMap unpacks the profiles specified in the configuration into a map
+func (config *Config) unpackProfilesIntoMap() (map[string]Profile, error) {
+	result := make(map[string]Profile)
+	if len(config.ProfileNames) > 0 {
+		for index, name := range config.ProfileNames {
+			if _, ok := result[name]; ok {
+				return nil, errors.New(fmt.Sprintf("Pprof profile names contains duplicate: %v", name))
+			} else {
+				profile := &Profile{
```
JackyWYX, line 184: Probably we don't need the pointer in the first place.

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+	for index := 1; index < 10000; index++ {
+		switch f, err := os.OpenFile(joinPath(dir, fmt.Sprintf("%s%03d%s", prefix, index, suffix)), os.O_RDWR|os.O_CREATE|os.O_EXCL, 0666); {
+		case err == nil:
+			return f, nil
+		case !os.IsExist(err):
+			return nil, err
+		}
+	}
+	// Give up
+	return nil, fmt.Errorf("could not create file of the form %s%03d%s", prefix, 1, suffix)
```
JackyWYX, line 234: What about all 10000 indexes are consumed? Node is a long run program and 10000 is very easy to surpass. Have you considered using other indexes like unix time for the specifier?

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+func scheduleProfile(profile Profile, dir string) {
+	go func() {
+		if profile.Interval > 0 {
+			ticker := time.NewTicker(time.Second * time.Duration(profile.Interval))
+			defer ticker.Stop()
+			for {
+				select {
+				case <-ticker.C:
+					if profile.Name == CPU {
+						handleCpuProfile(profile, dir, true)
```
JackyWYX, line 134: What if the interval is greater than the CPU profile duration? Could it happen that a CPU profile is never finished?

JackyWYX (commented):
Sorry for the late reply. Some suggestions in the comments. Please fix. Thanks.

cmd/harmony/config.go
```diff
@@ -95,8 +95,12 @@ type txPoolConfig struct {
 }
 
 type pprofConfig struct {
-	Enabled    bool
-	ListenAddr string
+	Enabled            bool
+	ListenAddr         string
+	Folder             string
+	ProfileNames       []string `toml:",omitempty"`
```
JackyWYX, line 101: I think these values are critical and shall not be hidden from users even the value is empty. Do you think we shall remove the "omitempty" toml flag?

cmd/harmony/config_migrations.go
```diff
@@ -156,4 +156,14 @@ func init() {
 		confTree.Set("Version", "2.0.0")
 		return confTree
 	}
+
+	migrations["2.0.0"] = func(confTree *toml.Tree) *toml.Tree {
+		// Legacy conf missing fields
+		if confTree.Get("Pprof.Folder") == nil {
```
JackyWYX, line 162: Can we also set enabled default to true? Can we also add default value to ProfileNames, ProfileIntervals, and ProfileDebugValues.

cmd/harmony/flags.go
```diff
@@ -955,6 +959,30 @@ var (
...
+	}
+	pprofProfileNamesFlag = cli.StringSliceFlag{
+		Name:     "pprof.profile.names",
+		Usage:    "a list of pprof profile names (separated by ,) e.g. cpu,heap,goroutine",
+		DefValue: defaultConfig.Pprof.ProfileNames,
+		Hidden:   true,
+	}
+	pprofProfileIntervalFlag = cli.IntSliceFlag{
+		Name:     "pprof.profile.intervals",
+		Usage:    "a list of pprof profile interval integer values (separated by ,)",
```
JackyWYX, line 976: The unit is second right? Would you please elaborate a little bit?

cmd/harmony/flags.go
```diff
@@ -955,6 +959,30 @@ var (
...
+		Name:     "pprof.folder",
+		Usage:    "folder to put pprof profiles",
+		DefValue: defaultConfig.Pprof.Folder,
+		Hidden:   true,
+	}
+	pprofProfileNamesFlag = cli.StringSliceFlag{
+		Name:     "pprof.profile.names",
+		Usage:    "a list of pprof profile names (separated by ,) e.g. cpu,heap,goroutine",
+		DefValue: defaultConfig.Pprof.ProfileNames,
+		Hidden:   true,
```
JackyWYX, line 972: I think it shall not be set to hidden. It shall be commonly used for debugging.

cmd/harmony/flags.go
```diff
@@ -955,6 +959,30 @@ var (
...
+	}
+	pprofProfileIntervalFlag = cli.IntSliceFlag{
+		Name:     "pprof.profile.intervals",
+		Usage:    "a list of pprof profile interval integer values (separated by ,)",
+		DefValue: defaultConfig.Pprof.ProfileIntervals,
+		Hidden:   true,
+	}
+	pprofProfileDebugFlag = cli.IntSliceFlag{
+		Name:     "pprof.profile.debug",
+		Usage:    "a list of pprof profile debug integer values (separated by ,)",
```
JackyWYX, line 982: What does each integer mean?

JackyWYX:
Please rebase and solve the conflict introduced by another config upgrade. Thanks!!

niklr:
Thank your for the review. Really good points indeed! I will try to resolve them asap.

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+		utils.Logger().Info().Msg("Pprof service disabled...")
+		return nil
+	}
+
+	utils.Logger().Debug().Str("cfg", cfg.String()).Msg("Pprof")
+	svc.config = cfg
+
+	if profiles, err := cfg.unpackProfilesIntoMap(); err != nil {
+		utils.Logger().Error().Err(err).Msg("Could not unpack profiles into map")
+		return nil
```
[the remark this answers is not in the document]
niklr: Indeed, providing an invalid profile name could cause the code to panic. With log.Fatal we should be better off in this case.

https://github.com/harmony-one/harmony/blob/92568eae89d5cf706b51150ff64e828599d7ed10/api/service/pprof/service.go#L71-L72

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+
+// saveProfile saves the provided profile in the specified directory
+func saveProfile(profile Profile, dir string) {
+	f, err := newTempFile(dir, profile.Name, ".pb.gz")
+	if err == nil {
+		defer f.Close()
+		if p := pprof.Lookup(profile.Name); p == nil {
+			utils.Logger().Error().Msg(fmt.Sprintf("Profile does not exist: %s", profile.Name))
+		} else {
+			runtime.GC() // get up-to-date statistics
```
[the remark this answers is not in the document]
niklr: The default example suggests to call garbage collection as described here https://golang.org/pkg/runtime/pprof/. But I agree it could potentially be expensive -> removed. Maybe making the call optional with an additional flag?

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+				utils.Logger().Info().Msg(fmt.Sprintf("Saved profile in: %s", f.Name()))
+			}
+		}
+	} else {
+		utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not save profile: %s", profile.Name))
+	}
+}
+
+// handleCpuProfile handles the provided CPU profile by stopping or starting it as specified
+func handleCpuProfile(profile Profile, dir string, shouldStop bool) {
```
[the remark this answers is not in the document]
niklr: The duration of the CPU profile can be configured like the other profiles with the intervals flag. 
- Providing `--pprof.profile.names cpu` which is equivalent to `--pprof.profile.names cpu --pprof.profile.intervals 0` will start it on startup and save/stop on shutdown.
- Providing `--pprof.profile.names cpu --pprof.profile.intervals 10` will start and stop/save it every 10 seconds.

api/service/pprof/service.go
```diff
@@ -0,0 +1,235 @@
...
+// Start start the service
+func (s *Service) Start() error {
+	dir := getBasePath(s.config.Folder)
+	err := os.MkdirAll(dir, os.FileMode(0755))
+	if err != nil {
+		return err
+	}
+
+	if cpuProfile, ok := s.profiles[CPU]; ok {
+		handleCpuProfile(cpuProfile, dir, false)
```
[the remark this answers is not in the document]
niklr: The CPU profiling is now started asynchronously to avoid blocking the start of other services:

https://github.com/harmony-one/harmony/blob/92568eae89d5cf706b51150ff64e828599d7ed10/api/service/pprof/service.go#L96-L97

api/service/pprof/service.go
niklr: The file is closed when calling pprof.StopCPUProfile()

api/service/pprof/service.go
niklr: I agree the [implementation taken from Google](https://github.com/google/pprof/blob/a478d1d731e942fbe69291748caafefe749a19ea/internal/driver/tempfile.go#L25) is not optimal -> now using unix timestamp instead :)

https://github.com/harmony-one/harmony/blob/92568eae89d5cf706b51150ff64e828599d7ed10/api/service/pprof/service.go#L219-L224

api/service/pprof/service.go
niklr: Not sure if I understand correctly. The CPU profile is stopped when the service stops or stopped/started based on the specified interval. The pprof HTTP implementation handles it differently with a default duration of 30 seconds but this does not have an impact here.

cmd/harmony/flags.go
niklr: 0 writes the gzip-compressed protocol buffer. 1 writes the legacy text format with comments translating addresses to function names and line numbers, so that no tools are needed to read the profile. But predefined profiles may assign meaning to other debug values. E.g. for the "goroutine" profile 2 means to print the goroutine stacks in the same form that a Go program uses when dying due to an unrecovered panic. More information can be found here: https://golang.org/pkg/runtime/pprof/

https://github.com/harmony-one/harmony/blob/92568eae89d5cf706b51150ff64e828599d7ed10/cmd/harmony/flags.go#L987-L989

cmd/harmony/flags.go
niklr: Yes, the unit is second. Providing for example `--pprof.profile.names cpu,heap --pprof.profile.intervals 0,10` saves the cpu profile on shutdown and the heap profile every 10 seconds. `--pprof.profile.names cpu,heap,goroutine --pprof.profile.intervals 30` saves the cpu, heap and goroutine profile every 30 seconds.

https://github.com/harmony-one/harmony/blob/92568eae89d5cf706b51150ff64e828599d7ed10/cmd/harmony/flags.go#L981-L983

niklr:
Setting the default value in config_migrations.go as follows results in the error `Can't convert []([]string) to []string(slice)`

```
migrations["2.1.0"] = func(confTree *toml.Tree) *toml.Tree {
  ...
  if confTree.Get("Pprof.ProfileNames") == nil {
    confTree.Set("Pprof.ProfileNames", []string{})
  }
  ...
  
  confTree.Set("Version", "2.2.0")
  return confTree
}
```

Changing `[]string{}` to `make([]string, 0)` results in `Can't convert []([]string) to []string(slice)`

Do you know what I'm missing here @JackyWYX ?

niklr:
Replacing it with `make([]interface{}, 0)` did the trick.

niklr:
All mentioned suggestions and remarks should now be covered. Please review again. @JackyWYX @LeoHChen 

Thanks a lot!

api/service/pprof/service.go
```diff
@@ -0,0 +1,225 @@
...
+
+// saveProfile saves the provided profile in the specified directory
+func saveProfile(profile Profile, dir string) error {
+	f, err := newTempFile(dir, profile.Name, ".pb.gz")
+	if err != nil {
+		utils.Logger().Error().Err(err).Msg(fmt.Sprintf("Could not save pprof profile: %s", profile.Name))
+		return err
+	}
+	defer f.Close()
+	if profile.ProfileRef != nil {
```
JackyWYX, line 156: Consider following?

```
if profile.ProfileRef == nil {
    return nil
}
```

api/service/pprof/service.go
JackyWYX: I looked into the implementation of `StopCPUProfile`, and it seems that the function is not closing the file. Check again?

api/service/pprof/service.go
JackyWYX: So what is the duration of execution of the function `handleCpuProfile`? Any reference?

cmd/harmony/flags.go
JackyWYX: This is good description here. Can you add the link to the help message? The E.g. seems to be not very clear.

Also, do you think it is better to have the flag have a default value as 0?

JackyWYX (commented):
Still some comments out there. Please fix. Thanks!

api/service/pprof/service.go
```diff
@@ -0,0 +1,225 @@
...
+		return nil, nil
+	}
+	for index, name := range config.ProfileNames {
+		profile := Profile{
+			Name:     name,
+			Interval: 0,
+			Debug:    0,
+		}
+		// Try set interval value
+		if len(config.ProfileIntervals) == len(config.ProfileNames) {
```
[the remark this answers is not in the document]
niklr: The new default value is 600 as you suggested. However, a 0 for interval means the profile is saved on shutdown. This might be useful if you want to profile the CPU for the entire lifetime of the application by generating just 1 file.

go.mod
```diff
@@ -40,7 +40,7 @@ require (
 	github.com/multiformats/go-multiaddr-dns v0.3.1
 	github.com/natefinch/lumberjack v2.0.0+incompatible
 	github.com/pborman/uuid v1.2.0
-	github.com/pelletier/go-toml v1.2.0
+	github.com/pelletier/go-toml v1.9.3
```
niklr, line 43: Dependency `go-toml` has to be updated otherwise array default values can't be set in `cmd/harmony/config_migrations.go`

api/service/pprof/service.go
niklr: I have renamed the function and updated the comment to hopefully make it a bit more clear.

https://github.com/harmony-one/harmony/blob/e25016d6bcfabe564950232dcb082d20e8fff4b7/api/service/pprof/service.go#L178-L179

api/service/pprof/service.go
```diff
@@ -0,0 +1,225 @@
...
+	if err != nil {
+		return err
+	}
+	err = os.MkdirAll(dir, os.FileMode(0755))
+	if err != nil {
+		return err
+	}
+
+	if cpuProfile, ok := s.profiles[CPU]; ok {
+		go handleCpuProfile(cpuProfile, dir)
```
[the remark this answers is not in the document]
niklr: If we don't start the CPU profile here, we would miss the first interval cycle.

https://github.com/harmony-one/harmony/blob/e25016d6bcfabe564950232dcb082d20e8fff4b7/api/service/pprof/service.go#L97-L101

api/service/pprof/service.go
niklr: You are right. This should be resolved with the new function:

https://github.com/harmony-one/harmony/blob/e25016d6bcfabe564950232dcb082d20e8fff4b7/api/service/pprof/service.go#L191-L198

niklr:
Ready for round 3 @JackyWYX

api/service/pprof/service.go
```diff
@@ -0,0 +1,246 @@
...
+	utils.Logger().Debug().Str("cfg", cfg.String()).Msg("pprof")
+	svc.config = cfg
+
+	profiles, err := cfg.unpackProfilesIntoMap()
+	if err != nil {
+		log.Fatal("could not unpack pprof profiles into map")
+	}
+	svc.profiles = profiles
+
+	go func() {
```
JackyWYX, line 78: Do you mind moving these two lines to `service.Start`?

api/service/pprof/service.go
```diff
@@ -0,0 +1,246 @@
...
+	pprof.StartCPUProfile(f)
+	cpuFile = f
+	utils.Logger().Info().Msg(fmt.Sprintf("saved pprof CPU profile in: %s", f.Name()))
+	return nil
+}
+
+// stopCpuProfile stops the current CPU profile, if any
+func stopCpuProfile() {
+	pprof.StopCPUProfile()
+	if cpuFile != nil {
```
JackyWYX, line 194: Have you tried testing with `go test --race` of this line? It seems there is a data race condition of `cpuFile`

go.mod
JackyWYX: Nice! Thanks!

JackyWYX:
One last small comments about data racing.

niklr:
Fixed data racing using a mutex and moved the two lines @JackyWYX

api/service/pprof/service.go
```diff
@@ -0,0 +1,249 @@
...
+type Service struct {
+	config   Config
+	profiles map[string]Profile
+}
+
+var (
+	initOnce sync.Once
+	svc      = &Service{}
+	cpuFile  *os.File
+	lock     sync.Mutex
```
JackyWYX, line 53: The variable name is too general. Better rename to something like `cpuLock`.

JackyWYX (approved):
@niklr Thanks for all your work! Impressed :) 

@rlan35 @LeoHChen Take a look?

niklr:
Thank you @JackyWYX, that review process was rigorous. :) Makes me appreciate the harmony project even more.

With that, do you consider this bounty as completed? The deadline is approaching quite soon (~4 days).

JackyWYX:
Please resolve the last comment. Thanks!

rlan35 (approved):

rlan35:
looks good to merge if this is backward-compatible. We should do more testing to make sure no race condition and it won't crash the node.

JackyWYX:
> looks good to merge if this is backward-compatible. We should do more testing to make sure no race condition and it won't crash the node.

What kind of testing are you suggesting?

rlan35:
> > looks good to merge if this is backward-compatible. We should do more testing to make sure no race condition and it won't crash the node.
> 
> What kind of testing are you suggesting?

Like extended period of testing longevity test. Like doing frequent pprof every 1s for 2 hours and see if there is any race condition and other problems.

JackyWYX:
> Like extended period of testing longevity test. Like doing frequent pprof every 1s for 2 hours and see if there is any race condition and other problems.

Sound reasonable. @niklr Could you please do this test and give out the machine performance metrics? (Like CPU usage, memory, with / without pprof module)

niklr:
Sure, I will record metrics with the following script tomorrow (getting late here already):

```
#! /bin/bash
echo "ELAPSED %CPU %MEM" >> output.txt
while :
do
    clear
    ps -p {HARMONY_PID} -o etime,%cpu,%mem | tail -n +2 | tee -a output.txt
    sleep 1s
done
```
Or do you have any other preferences @JackyWYX ?

JackyWYX:
> Or do you have any other preferences @JackyWYX ?

Could you do performance benchmark on this PR and previous code to see the performance difference?

niklr:
10 minutes running the previous code + 2 hours running this PR with pprof enabled (--pprof.profile.names cpu,heap --pprof.profile.intervals 1)

Something like that?

JackyWYX:
> 10 minutes running the previous code + 2 hours running this PR with pprof enabled (--pprof.profile.names cpu,heap --pprof.profile.intervals 1)
> 
> Something like that?

Can you do 2 hours running previous code vs 2 hours running with pprof enabled?

niklr:
Harmony performance profiling over 2 hours
- [with previous code](https://gist.github.com/niklr/07161b02aa38c8cdfc35f9900d85850b)
- [with pprof enabled](https://gist.github.com/niklr/a5c429b9d586d7004267e93a07c7d353)

LGTM @JackyWYX @rlan35

PR #3775, opened 2021-06-15
Explorer db commit performance optimization

JackyWYX:
## Issue

This PR is the temporary fix to address the explorer node data fails to catchup with beacon chain. In explorer service, instead of write address transaction every block, commit transaction data to db every 20 blocks. Meanwhile the data not committed will be kept in memory cache for external RPC queries. 

Note this PR is just optimizing the caching mechanism of explorer node, and it does not effect the leveldb schema.

## Test

1. Passed local test.
2. Benchmarked on mainnet explorer node (aws.c5.2xlarge)

Mainnet metrics: (cache and flush address info into db every 100 blocks)
1. Every batch need to write ~820MB data to db.
2. Each batch need ~1m10s to flush to db.
3. On average ~0.7s a block to commit explorer db.
4. Increased CPU usage 20% -> 30% (which is expected since previously all explorer dump are blocked by disk IO).
5. Increased memory usage 6GB -> 8GB.

This is still not good, but it could get the explorer node work and get synced gradually. I will work out another PR to do the explorer DB schema change.

Also, it would be better to confirm the data correctness on testnet.

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -257,8 +257,6 @@ func (storage *Storage) UpdateTxAddressStorage(addr oneAddress, txRecords TxReco
 
 func (storage *Storage) flushLocked() error {
 	storage.lock.Lock()
-	defer storage.lock.Unlock()
-
 	batch := new(leveldb.Batch)
 	for addr, addressInfo := range storage.dirty {
 		key := GetAddressKey(addr)
@@ -273,10 +271,14 @@ func (storage *Storage) flushLocked() error {
 		key := GetCheckpointKey(new(big.Int).SetUint64(bn))
 		batch.Put(key, []byte{})
 	}
+	storage.lock.Unlock()
+
 	if err := storage.db.Write(batch, nil); err != nil {
 		return errors.Wrap(err, "failed to write explorer data")
 	}
 	// clean up
+	storage.lock.Lock()
+	defer storage.lock.Unlock()
 	for addr, addressInfo := range storage.dirty {
 		storage.clean.Add(addr, addressInfo)
 	}
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -203,11 +203,11 @@ func (storage *Storage) DumpCatchupBlock(block *types.Block) {
 }
 
 func (storage *Storage) isBlockComputed(bn uint64) bool {
-	blockCheckpoint := GetCheckpointKey(new(big.Int).SetUint64(bn))
-	if _, err := storage.GetDB().Get(blockCheckpoint, nil); err == nil {
+	if _, isDirty := storage.dirtyBNs[bn]; isDirty {
 		return true
 	}
-	if _, isDirty := storage.dirtyBNs[bn]; isDirty {
+	blockCheckpoint := GetCheckpointKey(new(big.Int).SetUint64(bn))
+	if _, err := storage.GetDB().Get(blockCheckpoint, nil); err == nil {
 		return true
 	}
 	return false
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -256,7 +256,7 @@ func (storage *Storage) UpdateTxAddressStorage(addr oneAddress, txRecords TxReco
 }
 
 func (storage *Storage) flushLocked() error {
-	storage.lock.Lock()
+	storage.lock.RLock()
 	batch := new(leveldb.Batch)
 	for addr, addressInfo := range storage.dirty {
 		key := GetAddressKey(addr)
@@ -271,8 +271,9 @@ func (storage *Storage) flushLocked() error {
 		key := GetCheckpointKey(new(big.Int).SetUint64(bn))
 		batch.Put(key, []byte{})
 	}
-	storage.lock.Unlock()
+	storage.lock.RUnlock()
 
+	// Hack: during db write, dirty is read-only. We can release the lock for now.
 	if err := storage.db.Write(batch, nil); err != nil {
 		return errors.Wrap(err, "failed to write explorer data")
 	}
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -282,6 +282,14 @@ func (storage *Storage) flushLocked() error {
 	return nil
 }
 
+// GetAddressInfo get the address info of the given address
+func (storage *Storage) GetAddressInfo(addr string) (*Address, error) {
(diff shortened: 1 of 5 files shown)


rlan35:
Two questions:
1. How to migrate the old explorer db to this new scheme?
2. Since the data is temporarily cached in memory, will the rpc serve data from the memory too?

JackyWYX:
> 1. How to migrate the old explorer db to this new scheme?

* I will add a specifier such as "version" to decide whether the migration is needed.
* For migration process, I am thinking of two ways to achieve this:
    * Automated. When node starts the explorer service, it will check whether migration is needed. If true, it will suspend the explorer db read/write and do migration automatically. Meanwhile, the RPC related to explorer DB will also be unavailable. The duration will be more clear after implementing the migration logic.
    * Manually rclone. Instead of composing the explorer DB from scratch, we can save the explorer db to rclone, and do rclone before upgrade the node. This is expected to save some time compared with the previous solution.

> 2. Since the data is temporarily cached in memory, will the rpc serve data from the memory too?

Yes. It will. It first try to get data from dirty memory, then clean cache, and finally read from db.

LeoHChen:
Noted, we are doing a migration of explorer nodes from EBS volume to local NVMe drive to boost the IOPS in the meantime.  https://github.com/harmony-one/harmony-ops-priv/issues/37

api/service/explorer/storage.go
```diff
@@ -69,78 +103,239 @@ func (storage *Storage) Init(ip, port string) {
...
+			return &Address{
+				ID: string(addr),
+			}, nil
+		}
+		return nil, err
+	}
+	storage.clean.Add(addr, addrInfo)
+	return addrInfo, nil
+}
+
```
AlexiaChen, line 323: you already added get data from batch cache?

AlexiaChen (approved):
awesome. The solution more and more better than my solution.

LeoHChen:
What's the testing plan for this PR?  Please specify a few test scenarios so that @sophoah can work with p-ops to test it out.

api/service/explorer/storage.go
JackyWYX: Yes You may take a look at the usage of `GetAddressInfo`, which is a public function to be used outside the module.

LeoHChen (approved):
LGTM in general, please add a test plan issue.

api/service/explorer/storage.go
```diff
@@ -69,78 +103,239 @@ func (storage *Storage) Init(ip, port string) {
...
+				utils.Logger().Error().Err(err).Msg("[explorer] handle new block result failed")
+			}
+			continue
+		case <-storage.closeC:
+			return
+		default:
+		}
+		// handles catchupBlock
+		select {
+		case b := <-storage.newBlockC:
```
rlan35, line 157: why this newBlockC logic is repeated here?

JackyWYX: Here is the trick to prioritize handling `newBlockC` over `catchupBlockC`. This is not very clean though. Do you have another idea to prioritizing handling a channel over another?

api/service/explorer/storage.go
```diff
@@ -69,78 +103,239 @@ func (storage *Storage) Init(ip, port string) {
...
-	for address, txRecords := range acntsTxns {
-		storage.UpdateTxAddressStorage(address, txRecords, false /* isStaking */)
+func (storage *Storage) isBlockComputed(bn uint64) bool {
+	blockCheckpoint := GetCheckpointKey(new(big.Int).SetUint64(bn))
+	if _, err := storage.GetDB().Get(blockCheckpoint, nil); err == nil {
+		return true
 	}
-	for address, txRecords := range acntsStakingTxns {
-		storage.UpdateTxAddressStorage(address, txRecords, true /* isStaking */)
+	if _, isDirty := storage.dirtyBNs[bn]; isDirty {
```
rlan35, line 210: shall we check on memory first to avoid potential db read?

api/service/explorer/storage.go
rlan35: If this is to prioritize processing new block, why do we need the line from 145-154?

api/service/explorer/storage.go
JackyWYX: OK. Reasonable.

api/service/explorer/storage.go
JackyWYX: Line 145-154 is to prioritze processing new block. I don't get it. Select case in golang does not have a priority on the competing channels. And this is the hack to achieve this.

api/service/explorer/storage.go
```diff
@@ -69,78 +103,239 @@ func (storage *Storage) Init(ip, port string) {
...
+	addrInfo, err := storage.readAddressInfoFromDB(addr)
+	if err != nil {
+		if errors.Is(err, leveldb.ErrNotFound) {
+			return &Address{
+				ID: string(addr),
+			}, nil
+		}
+		return nil, err
+	}
+	storage.clean.Add(addr, addrInfo)
```
rlan35, line 320: Seems the storage.clean never gets cleared? The clean seems not like cache for data that's already committed to db?

JackyWYX: Clean is an LRU cache, it has it's own evict rules implemented. What do you mean by "seem not like cache"?

rlan35:
so this PR hasn't changed the structure of the indexes?

JackyWYX:
> so this PR hasn't changed the structure of the indexes?

No. No db schema changed. Just adding a caching mechanism

api/service/explorer/storage.go
rlan35: but why do we need to repeat the same thing again in the last select here? 157 - 160?

api/service/explorer/storage.go
rlan35: I see. It is a cache. I thought it's a map like dirty.

api/service/explorer/storage.go
JackyWYX: Because the loop should be both waiting on newBlockC as well as catchupBlockC. If we do not repeat the same here, the loop will not continue if a block is sent to newBlockC, but not catchupBlockC.

JackyWYX: https://stackoverflow.com/questions/11117382/priority-in-go-select-statement-workaround

rlan35: looks good. thanks.

rlan35 (approved):

PR #3785, opened 2021-06-17
WIP: [Core] account abstraction (EIP-2938 ) integration

OSadovy:
Merged changes from https://github.com/quilt/go-ethereumwith appropriate modifications to adapt for harmony evm and transaction pool.

Currently most tests are passing that I can test locally. I haven't managed to run the official test suite from https://github.com/quilt/tests/tree/account-abstraction yet.

Since I am very new to harmony core internals I might be doing something completely wrong. So I welcome early feedback on the approach. Guidance about how to integrate with the official test suite is also most welcome.

## Issue

https://github.com/harmony-one/bounties/issues/35

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**


staking/types/transaction.go
```diff
@@ -328,3 +336,42 @@ func (tx *StakingTransaction) SenderAddress() (common.Address, error) {
...
 	return &tx.from
 }
+
+/*
+// AsMessage returns the transaction as a core.Message.
+//
+// AsMessage requires a signer to derive the sender.
+//
+// XXX Rename message to something less arbitrary?
+func (tx *StakingTransaction) AsMessage(s Signer) (coretypes.Message, error) {
```
rlan35, line 346: why comment out this part?

internal/params/config.go
```diff
@@ -404,6 +404,12 @@ func (c *ChainConfig) IsReceiptLog(epoch *big.Int) bool {
 	return isForked(c.ReceiptLogEpoch, epoch)
 }
 
+// IsAccountAbstraction returns whether num is either equal to the AA fork block or greater.
+func (c *ChainConfig) IsAccountAbstraction(epoch *big.Int) bool {
+	// TODO: Replace with correct block number
+	return isForked(c.IstanbulEpoch, epoch)
```
rlan35, line 410: please create a c.AccountAbstractionEpoch param and set it to be TBD.

internal/common/address.go
```diff
@@ -52,6 +61,16 @@ func IsBech32Address(s string) bool {
 	return true
 }
 
+// IsEntryPointAddress returns true if the Address is the ENTRY_POINT Address.
+func (a Address) IsEntryPoint() bool {
```
rlan35, line 65: why not just hardcode the entry point address and do byte.compare?

core/vm/opcodes.go
```diff
@@ -198,6 +198,11 @@ const (
 	LOG4
 )
 
+// account Abstraction ops.
+const (
+	PAYGAS OpCode = 0xaa + iota
```
rlan35, line 203: should it be 0x49 as in the eip description https://eips.ethereum.org/EIPS/eip-2938 ?

core/vm/instructions.go
```diff
@@ -908,6 +918,25 @@ func makeLog(size int) executionFunc {
 	}
 }
 
+func opPaygas(pc *uint64, interpreter *EVMInterpreter, contract *Contract, memory *Memory, stack *Stack) ([]byte, error) {
```
rlan35, line 921: In the reference implementation, they used callContext, why we don't have it here? https://github.com/quilt/go-ethereum/blob/9c4f387bc356af0e75019764d5730e5ea8b36d7f/core/vm/instructions.go#L914

core/types/transaction_signing.go
```diff
@@ -242,7 +246,7 @@ func (fs FrontierSigner) SignatureValues(tx InternalTransaction, sig []byte) (r,
 func (fs FrontierSigner) Hash(tx InternalTransaction) common.Hash {
 	return hash.FromRLP([]interface{}{
 		tx.Nonce(),
-		tx.GasPrice(),
+		tx.RawGasPrice(),
```
rlan35, line 249: what's the reason to change GasPrice to RawGasPrice？

core/types/eth_transaction.go
```diff
@@ -40,9 +41,10 @@ import (
 type EthTransaction struct {
 	data ethTxdata
 	// caches
-	hash atomic.Value
-	size atomic.Value
-	from atomic.Value
+	hash  atomic.Value
```
rlan35, line 44: Actually in harmony, EthTransaction is exactly the ethereum transaction format. So the code here should be very similar to the reference implementation in /core/types/transaction.go

staking/types/transaction.go
OSadovy: sorry, it seems this is a leftover - the AsMessage is used in validate() which itself is used in the transaction pool, only if the transaction is an AA transaction, and staked transactions can not be AA. so I'll remove it.

internal/common/address.go
OSadovy: this is taken directly from the upstream implementation. does it make any difference?

core/vm/opcodes.go
OSadovy: this was taken from the original upstream implementation. I can change as per EIP specification. what is the rule here, should we follow upstream to be bytecode compatible or follow the specs?

internal/params/config.go
OSadovy: will do

core/vm/instructions.go
OSadovy: because call context seems to be introduced in more recent version ofthe vm than harmony uses currently. but symantically we do exactly the same: call context is a struct with stack, contract and memory, and we pass exactly the same values just as separate arguments.

core/types/transaction_signing.go
OSadovy: good catch! as far as I remember, some harmony tests were failing because the new `GasPrice()` implementation calls `IsAA()` which assumes that transaction is signed, whereas it was not the case in tests. This should be symantically equivalent to what upstream implementation is doing (since they are not calling `GasPrice()`, they do `tx.data.Price` which equals to `RawGasPrice()`).

core/types/eth_transaction.go
OSadovy: yeah, and so it is. I tried to migrate&merge all relevant changes from upstream core/types/transaction.go. If you think there is some code missing here, please elaborate.

AlexiaChen:
thank you for your work. can you fix conflict？ thanks.

core/vm/opcodes.go
rlan35: we should follow the upstream implementation for now.

core/evm.go
```diff
@@ -29,6 +29,11 @@ import (
 	staking "github.com/harmony-one/harmony/staking/types"
 )
 
+type headerGetter interface {
```
rlan35, line 32: why refactor the ChainContext into headerGetter?   if it's not necessary, better keep as is.

OSadovy: Because `NewEVMContext()` is used in tx pool's `validateTx()` and since pool's `blockChain` attribute does not support the ChainContext interface. this way is the least changes to wire it together.

core/state_transition.go
```diff
@@ -210,28 +217,34 @@ func (st *StateTransition) to() common.Address {
 
 func (st *StateTransition) useGas(amount uint64) error {
 	if st.gas < amount {
-		return vm.ErrOutOfGas
+		return ErrIntrinsicGas
```
rlan35, line 220: why change this error to ErrIntrinsicGas?

core/state_transition.go
```diff
@@ -280,9 +312,28 @@ func (st *StateTransition) TransitionDb() (ExecutionResult, error) {
...
+			if vmErr != nil {
+				return ExecutionResult{}, vmErr
+			} else {
+				return ExecutionResult{}, ErrNoPaygas
+			}
+		}
+	}
+	if st.msg.IsAA() {
+		st.gasPrice = st.evm.GasPrice
+		if msg.IsAA() && st.evm.PaygasMode != vm.PaygasNoOp {
```
rlan35, line 330: seems this part from 330-336 is not in the eth code? why it's added here?

core/vm/jump_table.go
```diff
@@ -216,160 +251,183 @@ func newFrontierInstructionSet() JumpTable {
 			maxStack:    maxStack(0, 0),
 			halts:       true,
 			valid:       true,
+			internal:    true,
```
rlan35, line 254: do you know why this new field internal are all true?   if it's all true, what's the point of having this field?

core/tx_pool.go
```diff
@@ -324,7 +342,7 @@ func (pool *TxPool) loop() {
 	defer journal.Stop()
 
 	// Track the previous head headers for transaction reorgs
-	head := pool.chain.CurrentBlock()
+	head := pool.chain.CurrentBlock().Header()
```
rlan35, line 345: why make this change?

rlan35:
I've went through all the codes and put my comments. I think the code in this PR are mostly looking good. You just need to test it with all the available tests to show that it's working as intended. You also need to show us a few examples by running the blockchain and sending AA txns showing that the txns are actually processed correctly and included in the block.

core/state_transition.go
OSadovy: because `TransitionDb()` in the reference implementation returns this error and it is expected in tests. from what I see, harmony implementation refactored the `TransitionDb()` a little bit, moving out some stuff to other methods like `useGas()` . thus I had to change `useGas()` to be consistent with the reference implementation. In general, the intrinsic gas for a transaction is the amount of gas that the transaction uses before any code runs.

core/state_transition.go
OSadovy: Oops, this seems to be a mistake. thanks for catching, I'll remove it!

core/vm/jump_table.go
OSadovy: I can see ~10 instructions that are not market as internal=true, for example: EXTCODEHASH, CREATE2, BALANCE, EXTCODESIZE, EXTCODECOPY, BLOCKHASH, COINBASE, ... Those are not allowed to be used in the AA transaction code before PAYGAS is executed.

core/tx_pool.go
OSadovy: Should be an oversight from some older refactoring from me. will remove it.

adsorptionenthalpy:
A new pull request will be submitted for account abstraction pending further review from the protocol engineering team. Closing.

Issue #3784, opened 2021-06-16
/node-sync checking too strict


LeoHChen:
**Describe the bug**
We use /node-sync API to check if the explorer node is sync'ed up with the blockchain.  If true, it will stay in or add back to the load balancer, if false, it will be removed from the load balancer.

I found this check is too strict, as the `inSyuncThreshold` is set to 0. So, if the explorer node is just missing 1 or 2 blocks, it may report as out of sync and will be taken out of the load balancer, which ends up with no nodes behind the load balancers, and the RPC endpoints become unavailable.  So the nodes were taken out of the load balancer and added back frequently, which is a waste of time as well.

So, my proposal is the relax the out-of-sync checking on the explorer nodes a little bit, to like 5 blocks.  However, this function `IsoutOfSync` is also used in consensus sync-up check, which has less tolerance room.  Needs to be careful to change the parameter here.

https://github.com/harmony-one/harmony/blob/c0414bd1057552fc898c67e4f0d3c570fd85a5a9/api/service/legacysync/syncing.go#L1064

```golang
// IsOutOfSync checks whether the node is out of sync from other peers
func (ss *StateSync) IsOutOfSync(bc *core.BlockChain, doubleCheck bool) (bool, uint64) {
   if ss.syncConfig == nil {
      return true, 0 // If syncConfig is not instantiated, return not in sync
   }
   otherHeight1 := ss.getMaxPeerHeight(false)
   lastHeight := bc.CurrentBlock().NumberU64()
   wasOutOfSync := lastHeight+inSyncThreshold < otherHeight1

   if !doubleCheck {
      utils.Logger().Info().
         Uint64("OtherHeight", otherHeight1).
         Uint64("lastHeight", lastHeight).
         Msg("[SYNC] Checking sync status")
      return wasOutOfSync, otherHeight1
   }
   time.Sleep(1 * time.Second)
   // double check the sync status after 1 second to confirm (avoid false alarm)

   otherHeight2 := ss.getMaxPeerHeight(false)
   currentHeight := bc.CurrentBlock().NumberU64()

   isOutOfSync := currentHeight+inSyncThreshold < otherHeight2
   utils.Logger().Info().
      Uint64("OtherHeight1", otherHeight1).
      Uint64("OtherHeight2", otherHeight2).
      Uint64("lastHeight", lastHeight).
      Uint64("currentHeight", currentHeight).
      Msg("[SYNC] Checking sync status")
[...]

PR #3786, opened 2021-06-17
[sync] add a tolerance number for node-sync checking

LeoHChen:
Fix https://github.com/harmony-one/harmony/issues/3784


LeoHChen:
Tested locally, by changing the tolerance to a huge number, so to test the /node-sync status.  Also tested with additional logs.

JackyWYX (approved):

PR #3787, opened 2021-06-17
[tracer] optimize CPU usage of trace_block RPC

peekpi:
* use single Goroutines to deal trace task
* drop trace task if the request is canceled.

## Issue

 the trace RPC may causing higher CPU usage, especially if there are a lot of malicious transactions. 
This PR does not reduce execution time(nor does not increase it), but it does reduce CPU usage.

### Test/Run Logs
```
curl --request POST 'http://54.189.61.183:9500' --header 'Content-Type: application/json' --data-raw '{
    "jsonrpc": "2.0",
    "method": "trace_block",
    "params": ["0xd6739e"],
    "id": 1
}'
```

## Operational Checklist

    **NO**

    **NO**

    **NO**

## TODO 
need to run it for several days to see the results.

### DIFF
--- a/hmy/tracer.go
+++ b/hmy/tracer.go
@@ -342,10 +342,84 @@ func (hmy *Harmony) TraceChain(ctx context.Context, start, end *types.Block, con
 	return sub, nil
 }
 
+// same as TraceBlock, but only use 1 thread
+func (hmy *Harmony) traceBlockNoThread(ctx context.Context, block *types.Block, config *TraceConfig) ([]*TxTraceResult, error) {
+	// Create the parent state database
+	if err := hmy.BlockChain.Engine().VerifyHeader(hmy.BlockChain, block.Header(), true); err != nil {
+		return nil, err
+	}
+	parent := hmy.BlockChain.GetBlock(block.ParentHash(), block.NumberU64()-1)
+	if parent == nil {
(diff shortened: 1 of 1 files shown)


rlan35 (approved):

hmy/tracer.go
```diff
@@ -342,10 +342,84 @@ func (hmy *Harmony) TraceChain(ctx context.Context, start, end *types.Block, con
 	return sub, nil
 }
 
+// same as TraceBlock, but only use 1 thread
+func (hmy *Harmony) traceBlockNoThread(ctx context.Context, block *types.Block, config *TraceConfig) ([]*TxTraceResult, error) {
```
LeoHChen, line 346: can you add instrumentation to keep track of the execution time for comparison?

LeoHChen (approved):
LGTM in general, it would be better to add some instrumentation to compare the execution time of the call.

hmy/tracer.go
peekpi: It doesn't reduce execution time, it just reduces thread usage.
another bug fixed is: if a client uses a batch RPC to trace some blocks, and the connection broke during the execution of these blocks, the Trace module still completes all the tasks. this wastes a lot of CPUs.

Issue #3740, opened 2021-06-07
Explorer nodes can't catch up with latest blocks


LeoHChen:
**Describe the bug**
Recently, with a higher volume of tx happened on-chain, the explorer nodes, either archival or non-archival nodes, can't catch up with the latest blockchain blocks.

**To Reproduce**
We've seen this problem on this build v6999-v4.0.0-66-g343dbe89 on the mainnet.

The explorer nodes can't catch up with the latest blocks by a few minutes. However, there is no sign of CPU / IO overload on the nodes. After I restarted the harmony process, it can catch up very quickly.   So I think it is not the resource issue, there could be some logic that is slow to catch up.

**Expected behavior**
Explorer nodes should be able to catch up with the latest block when there is no resource overload issue.

rlan35:
on one explorer node I checked, it's CPU overloaded.
![image](https://user-images.githubusercontent.com/1042540/121097148-5e922000-c7a8-11eb-8934-c134b5b5c34d.png)

rlan35:
another node (34.222.182.47) definitely isn't catching up with the latest block and keeps triggering the block sync logic to catch up (which has 60s delay). It's very likely caused by too much CPU/memory load where the blocks broadcasted from consensus aren't processed in time (due to missing parent block). So likely the lastMile block logic need some optimization. @JackyWYX . you can check the log on 34.222.182.47, the issue is clear.

LeoHChen:
@rlan35 , 382% CPU is not CPU overloaded, this node has 8 cores, you may use `htop` to take a took.  The CPU load is under 50%

LeoHChen:
But I agree with you, I doubt the last mile block sync should be a bit more aggressive in order to catch up.  We do have enough CPU power.  Some parameters may need to be tuned in the last mile catch. @JackyWYX

JackyWYX:
My current guess is follows the following logic: The explorer node is heavy in p2p message handling, resulting in committed block handled at `AddNewBlockForExplorer` is out of order. Thus result in unknown ancestor problems. This issue is only for explorer node. For validator node, there is already caching mechanism in consensus module. 

I will first try to confirm my guess, and apply the fix later.

LeoHChen:
heavy p2p message is reasonable as those explorer nodes are also RPC end points, received a lot of RPC requests from users and sent p2p messages to the network.

JackyWYX:
[...]

PR #3790, opened 2021-06-19
Explorer db schema

JackyWYX:
## Issue

An additional fix to https://github.com/harmony-one/harmony/issues/3740.

Previously, explorer db is not scalable. Fix the issue by migrating and implement new explorer DB schema.

1. When explorer node is started, it will check whether the migration is done by reading a specifier as version.
2. If the migration is needed, the explorer will start the migration process. Meanwhile, RPCs related to explorer DB will be unavailable with error `explorer db not ready`. 4 RPCs are effected: `GetTransactionsCount`, `GetTransactionsHistory`, `GetStakingTransactionsHistory`, `GetTransactionsCount`.
3. The migration consist of two steps: 1. Data migration 2. Data integration check. If the data integration fails, the node program exits. The process takes ~30 min on mainnet explorer node.
4. The progress of migration can be viewed through log:
```
tail -f latest/zerolog-harmony.log | grep "explorer DB migration to 1.0.0"
```
5. After the migration, explorer DB write will no longer be a bottle neck for scaling. It can reach ~100ms per block.

## Suggested OPS

To do the migration, it is suggested to follow the following steps

1. Unregister the explorer node from load balancer.
2. Stop harmony service.
3. Copy ~/data/explorer_storage_0.0.0.0_9000 db for backup.
4. Start harmony service.
5. (Option) Use `tail -f latest/zerolog-harmony.log | grep "explorer DB migration to 1.0.0"` to check progress.
6. When migration finished, register the explorer node to load balancer.

## Test

Test passed locally.

Tested on mainnet explorer machine. 
1. The whole migration process takes ~30min for an explorer node to migrate the explorer DB to the latest version. DB data migration takes ~18min, and integrity check takes ~11min.
2. After the migration, the explorer DB processing block is much faster. Basically, it takes <100ms per block at peak.
3. Memory used is also hugely reduced, from 8G to 1.4G.
4. [A complete log grepped during migration process](https://gist.github.com/JackyWYX/7453c48bc5bddcc9765826e5a2c246a1).

Overall the result looks quite promising.


rlan35:
Overall, looks good. As long as the tests pass and the rpc data is consistent with old explorer, we can merge and deploy.

rlan35 (approved):

sophoah:
@JackyWYX on your mainnet explorer node test have you confirmed if the explorer dashboard is able to index the data with no issue ? We need to test the data structure returned by the API are still the same. Let me add a few more notes on #3793 
Once this is approved we can test on testnet

JackyWYX:
> @JackyWYX on your mainnet explorer node test have you confirmed if the explorer dashboard is able to index the data with no issue ? We need to test the data structure returned by the API are still the same. Let me add a few more notes on #3793
> Once this is approved we can test on testnet

I did not confirmed explorer dashboard issue. But the API interface was not changed, so I guess it shall be good.

PR #3795, opened 2021-06-21
[p2p] Try to fix P2P memory issue

AlexiaChen:
## Issue

- orginal issue: https://github.com/harmony-one/harmony-ops-priv/issues/36
- go vet detected context leak:

```bash
# github.com/harmony-one/harmony/p2p/types
p2p/types/peerAddr.go:89:8: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak
```

- go-yamux memory leak issue https://github.com/libp2p/go-yamux/pull/30

Since this library is a dependency of libp2p-kad-dht, choose to upgrade and update libp2p-kad-dht

## Test

According to the [latest analysis and speculation](https://github.com/harmony-one/harmony-ops-priv/issues/36#issuecomment-864555047), there is no evidence that the DNS node memory issue is related to these Fixes, but there is some correlation.   So try to reproduce on my local https://github.com/harmony-one/harmony-ops-priv/issues/36#issuecomment-864986243, test is on going

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/libp2p/go-libp2p-core v0.8.5
 	github.com/libp2p/go-libp2p-crypto v0.1.0
 	github.com/libp2p/go-libp2p-discovery v0.5.0
-	github.com/libp2p/go-libp2p-kad-dht v0.11.1
+	github.com/libp2p/go-libp2p-kad-dht v0.12.1
 	github.com/libp2p/go-libp2p-pubsub v0.4.0
 	github.com/multiformats/go-multiaddr v0.3.1
 	github.com/multiformats/go-multiaddr-dns v0.3.1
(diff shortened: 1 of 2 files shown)


LeoHChen (approved):

LeoHChen:
Can you upgrade the entire libp2p version, not just kad-dht?

JackyWYX (approved):
The code looks good, and it should definitely be resolved. But it seems this is still not the root cause though, since the method is only called in limited times.

Is there anything with libp2p that is possibly related to memory leak?

AlexiaChen:
> The code looks good, and it should definitely be resolved. But it seems this is still not the root cause though, since the method is only called in limited times.
> 
> Is there anything with libp2p that is possibly related to memory leak? 

@JackyWYX  you are definitely right，the root cause is not related to this fix. so another test is on going

PR #3807, opened 2021-06-30
[DO NOT MERGE][pprof] added mutex pprof

JackyWYX:
## Issue

Add the mutex runtime parameter change to enable pprof on mutex. This is the temporary fix to help explorer the archival node out of sync issue.

This PR shall not be merged to main branch since the impact of the parameter is unknown.


JackyWYX:
The usage of the mutex seems quite normal in previous investigation. Closing.

mutex

Issue #3709, opened 2021-05-16
harmony invoked OOM killer


LeoHChen:
**Describe the bug**
Validators recently noticed a higher rate of signature drop on nodes in all shards.

Using ```dmesg | grep harmony``` and you can see oom killer.

```
ubuntu@ip-172-31-36-171:~$ dmesg | grep harmony
[1550230.874887] harmony invoked oom-killer: gfp_mask=0x100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
[1550230.874892] CPU: 1 PID: 42963 Comm: harmony Not tainted 5.4.0-1041-aws #43-Ubuntu
[1550230.874997] [  32209]  1000 32209  1395312   909707  7720960        0             0 harmony
[1550230.874998] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/harmony.service,task=harmony,pid=32209,uid=1000
[1550230.875042] Out of memory: Killed process 32209 (harmony) total-vm:5581248kB, anon-rss:3638828kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:7540kB oom_score_adj:0
[1550231.167912] oom_reaper: reaped process 32209 (harmony), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
[2090066.261310] [  68640]  1000 68640  1400288   908756  7692288        0             0 harmony
[2090066.261313] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/system.slice/harmony.service,task=harmony,pid=68640,uid=1000
[2090066.261360] Out of memory: Killed process 68640 (harmony) total-vm:5601152kB, anon-rss:3635024kB, file-rss:0kB, shmem-rss:0kB, UID:1000 pgtables:7512kB oom_score_adj:0
[2090066.570406] oom_reaper: reaped process 68640 (harmony), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

PR #3808, opened 2021-07-01
[DO NOT MERGE] [p2p] Reduce memory and CPU usage of libp2p

AlexiaChen:
## Issue

- https://github.com/harmony-one/harmony-ops-priv/issues/36
- https://github.com/harmony-one/harmony/issues/3709

According to the test and verification in issue https://github.com/harmony-one/harmony-ops-priv/issues/36, the OOM is indeed caused by the Discovery function of libp2p's pubsub, but this feature cannot be simply disable. Otherwise, it will cause missing feature. After reading the code implementation of libp2p's kad-dht(https://github.com/libp2p/go-libp2p-kad-dht/blob/7cf0c1e05fe3a69cba5d3747358a25b937fcad89/query.go#L282-L300   https://github.com/libp2p/go-libp2p-kad-dht/blob/7cf0c1e05fe3a69cba5d3747358a25b937fcad89/query.go#L344-L349), I found that some Option parameters can reduce the frequency of queryPeer calls and reduce the calls related to `go-yamux's Stream` function. After testing, the memory is currently performing very well.

### Why upgrade kad DHT? 

cause 0.12.2 reduced CPU usage by the FullRT DHT client during bulk providing https://github.com/libp2p/go-libp2p-kad-dht/pull/720 maybe this fix can solve hold high CPU usage for stream protocol made by @JackyWYX  ?

## Test

the node 105.44 Test Without Discovery,  running it for a few days, the memory behaved very well:

the node 105.44 Test With concurrency is 1 and upgrade kad-dht, running for 7 hours the memory is smooth:

图片

### DIFF
--- a/p2p/discovery/option.go
+++ b/p2p/discovery/option.go
@@ -33,6 +33,7 @@ func (opt DHTConfig) getLibp2pRawOptions() ([]libp2p_dht.Option, error) {
 		opts = append(opts, dsOption)
 	}
 
+	opts = append(opts, libp2p_dht.Concurrency(1))
 	return opts, nil
 }
 
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	github.com/libp2p/go-libp2p-core v0.8.5
 	github.com/libp2p/go-libp2p-crypto v0.1.0
 	github.com/libp2p/go-libp2p-discovery v0.5.0
-	github.com/libp2p/go-libp2p-kad-dht v0.12.1
+	github.com/libp2p/go-libp2p-kad-dht v0.12.2
 	github.com/libp2p/go-libp2p-pubsub v0.4.1
 	github.com/multiformats/go-multiaddr v0.3.1
 	github.com/multiformats/go-multiaddr-dns v0.3.1
--- a/p2p/discovery/option_test.go
+++ b/p2p/discovery/option_test.go
@@ -40,14 +40,14 @@ func TestDHTOption_getLibp2pRawOptions(t *testing.T) {
 			opt: DHTConfig{
 				BootNodes: testAddrStr,
 			},
-			expLen: 1,
+			expLen: 2,
 		},
 		{
 			opt: DHTConfig{
 				BootNodes:     testAddrStr,
 				DataStoreFile: &validPath,
 			},
-			expLen: 2,
+			expLen: 3,
 		},
 		{
 			opt: DHTConfig{


AlexiaChen:
Why travis CI not re-run when I pushed new commit ?

p2p/discovery/option.go
```diff
@@ -33,6 +33,7 @@ func (opt DHTConfig) getLibp2pRawOptions() ([]libp2p_dht.Option, error) {
 		opts = append(opts, dsOption)
 	}
 
+	opts = append(opts, libp2p_dht.Concurrency(1))
```
rlan35, line 36: Why set concurrency to 1?

AlexiaChen: Because the size of this parameter directly affects the frequency of calls to queryPeer in Discovery's functional features, and queryPeer is the main person responsible for taking up memory after I read kad-dht implementation code. The default value of this parameter is 10, and changing it to 1 is a relatively safe value for the 4G memory size of the Validator node, which I think is 1. Of course, I can also try to increase this value and test the memory performance again.

rlan35: what's the downside of setting it to 1 v.s. larger number like 10?

AlexiaChen: According to the implementation code of libp2p-kad-dht, it is to affect the number of concurrency of queries on the same path, it reduces the performance of queries on the same path, on the same path, it will go to the number of concurrency of N = 1 Peer to find information. In a nutshell: it affects the performance of Discovery, but it is better than directly Disable Discovery.

rlan35: ok, to test the performance impact of setting it to 1 vs 10. If there is no problem with 1, we can go with 1 then.

JackyWYX (approved):
There shall not be performance issue with the param change.

AlexiaChen:
> 
> 
> There shall not be performance issue with the param change.

Thanks.

Why ? could you explain more detail ? I have not test the performance , but I readed the code, it might  affect  DHT Discovery performance , the code is here  https://github.com/libp2p/go-libp2p-kad-dht/blob/7cf0c1e05fe3a69cba5d3747358a25b937fcad89/query.go#L282-L300    

```go
	// calculate the maximum number of queries we could be spawning.
		// Note: NumWaiting will be updated in spawnQuery
		maxNumQueriesToSpawn := alpha - q.queryPeers.NumWaiting()

		// termination is triggered on end-of-lookup conditions or starvation of unused peers
		// it also returns the peers we should query next for a maximum of `maxNumQueriesToSpawn` peers.
		ready, reason, qPeers := q.isReadyToTerminate(pathCtx, maxNumQueriesToSpawn)
		if ready {
			q.terminate(pathCtx, cancelPath, reason)
		}

		if q.terminated {
			return
		}

		// try spawning the queries, if there are no available peers to query then we won't spawn them
		for _, p := range qPeers {
			q.spawnQuery(pathCtx, cause, p, ch)
		}
```

JackyWYX:
I mean, the slow discovery will not cause trouble at our usage of discovery.

rlan35 (approved):

rlan35:
> There shall not be performance issue with the param change.

sounds good.

PR #3824, opened 2021-07-12
Add logs for elapsed time for block processing; Add bls agg sig benchmark test code.

rlan35:
As in the title.

### DIFF
--- a/api/service/legacysync/syncing.go
+++ b/api/service/legacysync/syncing.go
@@ -836,9 +836,12 @@ func (ss *StateSync) UpdateBlockAndStatus(block *types.Block, bc *core.BlockChai
 			if err != nil {
 				return errors.Wrap(err, "parse commitSigAndBitmap")
 			}
+
+			startTime := time.Now()
 			if err := bc.Engine().VerifyHeaderSignature(bc, block.Header(), sig, bitmap); err != nil {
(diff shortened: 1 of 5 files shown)


api/service/legacysync/syncing.go
```diff
@@ -836,9 +836,12 @@ func (ss *StateSync) UpdateBlockAndStatus(block *types.Block, bc *core.BlockChai
 			if err != nil {
 				return errors.Wrap(err, "parse commitSigAndBitmap")
 			}
+
+			startTime := time.Now()
 			if err := bc.Engine().VerifyHeaderSignature(bc, block.Header(), sig, bitmap); err != nil {
 				return errors.Wrapf(err, "verify header signature %v", block.Hash().String())
 			}
+			utils.Logger().Debug().Int64("elapsed time", time.Now().Sub(startTime).Milliseconds()).Msg("[Sync] VerifyHeaderSignature")
```
JackyWYX, line 844: Suggest to use `.Str("elapsed time", time.Since(startTime).String())`

JackyWYX (approved):

api/service/legacysync/syncing.go
rlan35: Acknowledged. Will keep as is since this code will eventually be deleted.

PR #3825, opened 2021-07-13
[metrics] Change wrong metrics discription

AlexiaChen:
Just discussed with @sophoah , it must be milliseconds in metrics description. 

The changes here are related to the stream_sync metrics, which @JackyWYX  can also look at

### DIFF
--- a/p2p/stream/common/streammanager/metric.go
+++ b/p2p/stream/common/streammanager/metric.go
@@ -62,7 +62,7 @@ var (
 			Namespace: "hmy",
 			Subsystem: "stream",
 			Name:      "setup_stream_duration",
-			Help:      "duration in seconds of setting up connection to a discovered peer",
+			Help:      "duration in milliseconds of setting up connection to a discovered peer",
 			// buckets: 20ms, 40ms, 80ms, 160ms, 320ms, 640ms, 1280ms, +INF
 			Buckets: prometheus.ExponentialBuckets(0.02, 2, 8),
 		},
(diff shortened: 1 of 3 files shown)


JackyWYX (commented):
I am not very sure why it is in unit milliseconds. I looked at the implementation

```
func (t *Timer) ObserveDuration() time.Duration {
	d := time.Since(t.begin)
	if t.observer != nil {
		t.observer.Observe(d.Seconds())
	}
	return d
}
```

And it is observing seconds right? Is there strong evidence showing that millisecond is used instead?

AlexiaChen:
> 
> 
> I am not very sure why it is in unit milliseconds. I looked at the implementation
> 
> ```
> func (t *Timer) ObserveDuration() time.Duration {
> 	d := time.Since(t.begin)
> 	if t.observer != nil {
> 		t.observer.Observe(d.Seconds())
> 	}
> 	return d
> }
> ```
> 
> And it is observing seconds right? Is there strong evidence showing that millisecond is used instead?

Yes, Duration is in seconds, but do you remember that you used `Buckets: prometheus.ExponentialBuckets(0.02, 2, 8)`,?    The factor is `0.02` ` 0.02s = 20ms`. So the start of the buckets is 20ms.     and the Duration time shwon @sophoah  sent to me:

unknown

AlexiaChen:
Sorry, I am totally wrong

AlexiaChen:
@JackyWYX 

here is the picture @sophoah  sent to me:

unknown

AlexiaChen:
I am totally wrong.

Issue #3821, opened 2021-07-09
store used gas as separate entry


hypnagonia:
## Summary

currently getting usedGas is only possible from tx receipt. which requires an extra query for each tx for explorer and increases load on rpc/slowing the syncing speed down

## Current Design
tx receipt is only available only in tax receipt payload

## Problems
requires an extra query for each tx for explorer and increases load on RPC

## Proposal
modify the explorer node to save that data as a separate entry and serve usedGas in getBlockByNumber (with full tx info flag) response

AlexiaChen:
I just look through your proposal, you just want to usedGas for each tx in RPC GetBlockByNumber repsonse to you ?   or extra new RPC for query usedGas for each tx ?   I want to confirm your requirement again. Thanks.

AlexiaChen:
Hey .@hypnagonia,  the RPC `GetBlockByNumber` reponse body has 3 version: v1, v2, eth.    all three block versions need to support `usedGas` for each tx with `FullTx` flag ?

AlexiaChen:
@hypnagonia  @gupadhyaya 

Although the PR is almost complete, I have thought about the problems with the design here.

- Storing a redundant usedGas field in this way will increase the storage pressure on levelDB, and the pressure on writing and reading explorer db will also increase
- explorer db may be gradually deprecated 


Is this design reasonable?

## My Proposal

 In fact there is no need to store usedGas for each Tx, because there is already an interface like `GetReceiptsByHash` in the blockchain, which reads all Receipts of the whole block in bulk, probably faster than \<txhash, usedGas\> which is a DB entries query. I can meet the requirements described by Jenya based on this interface

gupadhyaya:
agree with @AlexiaChen it is better to utilize `GetReceiptsByHash` which is just one additional rpc call per block. not a good idea to introduce additional read/write logic in the explorer db.

Frozen:
Implemented through the bulk block-receipts RPC added by #4077, following the alternative agreed in this discussion. Closing as completed.

PR #3826, opened 2021-07-14
[rpc]  Add New RPC GetReceiptsByBlockHash for quering all receipts of the block in bulk

AlexiaChen:
## Issue

https://github.com/harmony-one/harmony/issues/3821

## Test

No test with @hypnagonia 

### Local Test

- [x] tested hmy
- [x] tested hmyv2
- [x] tested eth

Use RPC to comapre reponse data for same blockHash and TxHash is consistent:

`curl -d '{ "jsonrpc":"2.0", "method":"hmy_getBlockByHash", "params":[ "0x73852a261acd91e6ee0b72e4b39eb3ee42ec31cfb13a5b1b0fdb676e9dbc358c", true], "id":1 }' -H "Content-Type:application/json" -X POST http://127.0.0.1:9500`

response data:

[data.zip](https://github.com/harmony-one/harmony/files/6839998/data.zip)

`curl -d '{ "jsonrpc":"2.0", "method":"hmy_getReceiptsByBlockHash", "params":["0x73852a261acd91e6ee0b72e4b39eb3ee42ec31cfb13a5b1b0fdb676e9dbc358c"], "id":1 }' -H "Content-Type:application/json"
 -X POST http://127.0.0.1:9500`

reponse data:

[bb.zip](https://github.com/harmony-one/harmony/files/6840012/bb.zip)

`curl -d '{ "jsonrpc": "2.0", "method": "hmy_getTransactionReceipt", "params": [ "0xcd2634fdaaff70ab8c538e99b8cbf44ee56fd2073f54ceeeeeb9962b447691a8"], "id": 1}' -H "Content-Type:application/json" -X POST http: //127.0.0.1:9500`

reponse data:

[uu.zip](https://github.com/harmony-one/harmony/files/6840026/uu.zip)

### DIFF
--- a/rpc/eth/types.go
+++ b/rpc/eth/types.go
@@ -53,6 +53,28 @@ type BlockWithFullTx struct {
 	Signers      []string       `json:"signers,omitempty"`
 }
 
+// BlockReceipts represents a block all receipts that will serialize to the RPC representation.
+type BlockReceipts struct {
+	TxReceipts []*TxReceipt `json:"txReceipts"`
+}
+
+// TxReceipt represents a transaction receipt that will serialize to the RPC representation.
+type TxReceipt struct {
+	BlockHash         common.Hash     `json:"blockHash"`
+	TransactionHash   common.Hash     `json:"transactionHash"`
+	BlockNumber       hexutil.Uint64  `json:"blockNumber"`
+	TransactionIndex  hexutil.Uint64  `json:"transactionIndex"`
+	GasUsed           hexutil.Uint64  `json:"gasUsed"`
+	CumulativeGasUsed hexutil.Uint64  `json:"cumulativeGasUsed"`
+	ContractAddress   *common.Address `json:"contractAddress"`
+	Logs              []*types.Log    `json:"logs"`
+	LogsBloom         ethtypes.Bloom  `json:"logsBloom"`
+	From              common.Address  `json:"from"`
+	To                *common.Address `json:"to"`
+	Root              hexutil.Bytes   `json:"root"`
+	Status            hexutil.Uint    `json:"status"`
+}
+
 // Transaction represents a transaction that will serialize to the RPC representation of a transaction
 type Transaction struct {
 	BlockHash        *common.Hash    `json:"blockHash"`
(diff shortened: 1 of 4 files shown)


JackyWYX (commented):
I am not very sure about this design, but why can't we simply use `Receipt.GasUsed`? Is there anything missing here?

AlexiaChen:
@JackyWYX 

From @hypnagonia 's description, the browser usually calls RPC GetBlocks, and GetBlocks calls GetBlockByNumber one by one from start height to end height. the information returned here does not have the usedGas of each tx. maybe @hypnagonia needs to get each tx directly from the return message here And indeed there is no similar interface like GetReceiptByTxHash, only GetReceiptsByBlockHash

JackyWYX (commented):
I don't understand whether there is need to modify APIs. This operation will add some extra burden to already heavily used `GetBlockByNumber`. Would you please add some lines of description of the necessity of modifying APIs instead of optimizing at DAPP?

AlexiaChen:
@JackyWYX 

Do you mean to add a new RPC to serve the query usedGas? Instead of putting it inside the return body of GetBlockByNumber improve payload pressure?

Actually, here's what I was thinking

- First, @hypnagonia 's original requirement is to put it in GetBlockByNumber（maybe this is necessity for explorer? ）, maybe this is more convenient for his business processing, because adding FullTx will get the usedGas list directly, adding RPC will have more changes to his explorer business code

- Second, is it really necessary to add an RPC to query the usedGas list? Because the usedGas payload is just transferred to this new RPC, the pressure on the chain's own requests may not change, because the usedGas list should not occupy much, after all, the FullTx parameters returned payload should not be small, even if the usedGas list of payload is also irrelevant


Perhaps, you have other better suggestions?

JackyWYX:
> Perhaps, you have other better suggestions?

Why don't you implement a new RPC method called `GetReceiptsByBlockHash`?

AlexiaChen:
> 
> 
> > Perhaps, you have other better suggestions?
> 
> Why don't you implement a new RPC method called `GetReceiptsByBlockHash`?

As I mentioned above, adding a new RPC is indeed one way to do this. So it mostly depends on @hypnagonia 's needs. If he feels that this approach will not cause too many changes to his business code, I can do that.

AlexiaChen:
Hi，@JackyWYX  could you pleaese review it again？

rlan35 (approved):

JasonTy:
Hi,  Did the gas returned through hmy_getTransactionReceipt experience hex twice?

PR #3828, opened 2021-07-20
Add logic to do aggregated reward distribution once every 64 blocks

rlan35:
Implement the logic to aggregate the block reward distribution so it only happens every 64 blocks. After the fork epoch, for every 64 blocks, the first 63 blocks won't execute any block reward or signature counting logic, and at the 64th block, it will go back and process all the crosslinks/signatures for the previous 64 blocks (including itself). This way shard 0 block time can be reduced by an amortized time of around 0.5s. 

The new logic keeps the accounting of validator reward the same as before. The part that's improving performance is the delegator reward distribution where previously, for each block, the reward is distributed to all delegators, which is a huge computation cost as there are tens of thousands of delegators. Right now, for every 64 blocks, the reward for each validators are tallied and then distributed to their delegators only once, saving lots of computation.

Previously, it takes around 0.5-0.6 second to do the block reward distribution for every shard 0 block. Now, it takes similar amount of time but it only happens on the 64th block.

The reason for choosing 64 blocks is that the computation time is not significantly higher than the current situation (~0.5s). Following are some data points:

Doing reward distribution for 64 blocks takes around 0.5 second.
Doing reward distribution for 256 blocks takes around 1.8 second (which slows down the block time for too much).

Note: The old block reward logic is intact but only refactored on the crosslink processing logic.

Tested locally and also on mainnet node to see the old logic is compatible with current chain and new logic execute correctly.

### DIFF
--- a/core/offchain.go
+++ b/core/offchain.go
@@ -228,31 +228,26 @@ func (bc *BlockChain) CommitOffChainData(
 			); err != nil {
 				return NonStatTy, err
 			}
-			for _, paid := range [...][]reward.Payout{
-				roundResult.BeaconchainAward, roundResult.ShardChainAward,
-			} {
-				for i := range paid {
-					stats, ok := tempValidatorStats[paid[i].Addr]
-					if !ok {
-						stats, err = bc.ReadValidatorStats(paid[i].Addr)
-						if err != nil {
-							utils.Logger().Info().Err(err).
-								Str("addr", paid[i].Addr.Hex()).
-								Str("bls-earning-key", paid[i].EarningKey.Hex()).
-								Msg("could not read validator stats to update for earning per key")
-							continue
-						}
-						tempValidatorStats[paid[i].Addr] = stats
+			for _, paid := range roundResult.Payouts {
+				stats, ok := tempValidatorStats[paid.Addr]
+				if !ok {
+					stats, err = bc.ReadValidatorStats(paid.Addr)
+					if err != nil {
+						utils.Logger().Info().Err(err).
+							Str("addr", paid.Addr.Hex()).
+							Str("bls-earning-key", paid.EarningKey.Hex()).
+							Msg("could not read validator stats to update for earning per key")
+						continue
 					}
-					for j := range stats.MetricsPerShard {
-						if stats.MetricsPerShard[j].Vote.Identity == paid[i].EarningKey {
-							stats.MetricsPerShard[j].Earned.Add(
-								stats.MetricsPerShard[j].Earned,
-								paid[i].NewlyEarned,
-							)
-						}
+					tempValidatorStats[paid.Addr] = stats
+				}
+				for j := range stats.MetricsPerShard {
+					if stats.MetricsPerShard[j].Vote.Identity == paid.EarningKey {
+						stats.MetricsPerShard[j].Earned.Add(
+							stats.MetricsPerShard[j].Earned,
+							paid.NewlyEarned,
+						)
 					}
-
 				}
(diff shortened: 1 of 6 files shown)


core/offchain.go
```diff
@@ -228,31 +228,26 @@ func (bc *BlockChain) CommitOffChainData(
...
-						if err != nil {
-							utils.Logger().Info().Err(err).
-								Str("addr", paid[i].Addr.Hex()).
-								Str("bls-earning-key", paid[i].EarningKey.Hex()).
-								Msg("could not read validator stats to update for earning per key")
-							continue
-						}
-						tempValidatorStats[paid[i].Addr] = stats
+			for _, paid := range roundResult.Payouts {
+				stats, ok := tempValidatorStats[paid.Addr]
```
rlan35, line 232: Note: no logic change within the loop. Only the loop is changed to a merged list.

staking/network/reward.go
```diff
@@ -53,38 +52,28 @@ func NewPreStakingEraRewarded(totalAmount *big.Int) reward.Reader {
...
-			ShardChainAward:  shardP,
+			Total:   totalPayout,
+			Payouts: payouts,
 		},
-		missingSigners: mia,
 	}
 }
 
-// MissingSigners ..
-func (r *stakingEra) MissingSigners() shard.SlotList {
```
rlan35, line 84: this is not used anywhere.

internal/chain/reward.go
```diff
@@ -506,3 +298,343 @@ func waitForCommitSigs(sigsReady chan bool) error {
...
+	newRewards, payouts :=
+		big.NewInt(0), []reward.Payout{}
+
+	allPayables := []slotPayable{}
+	curBlockNum := header.Number().Uint64()
+
+	allCrossLinks := types.CrossLinks{}
+	startTime := time.Now()
+	for i := curBlockNum - RewardFrequency + 1; i <= curBlockNum; i++ {
+		if i < 0 {
```
AlexiaChen, line 313: Because there is this guarantee: ` if bc.Config().IsAggregatedRewardEpoch(header.Epoch())` It should be impossible for i here to be less than 0？

AlexiaChen (commented):
LGTM.

internal/chain/reward.go
rlan35: for local network where the epoch happens at very early block number such as 20. It is possible.

LeoHChen (commented):
Is there a test plan for such a fundamental change?

rlan35:
> Is there a test plan for such a fundamental change?

The test plan will be straight-forward since the logic changed is very self-contained: just to make sure for every 64 blocks, the block reward for validators are credited correctly. We can also re-test the situation on delegation, block reward collection to make sure it's not affected. I will draft a test plan before deployment on testnet. We can do the test on testnet with Pops validators.

For backward-compatibility, I briefly tested it on mainnet, I can run the binary to sync more blocks before launch.

JackyWYX (approved):
I am OK with all the code changes. I question I have is that is there some proof or metrics to prove that the large block time comes from the reward distribution?

rlan35:
> I am OK with all the code changes. I question I have is that is there some proof or metrics to prove that the large block time comes from the reward distribution?

yup. I did experiment on a instrumented mainnet nodes calculating the time spent on reward distribution and also on each subcomponent of it. It turns out that calculating the shares for all delegators is the most time consuming. So if we aggregate all the rewards on each validator and then distribute to delegators only once, it will save a lot of time.

PR #3829, opened 2021-07-21
[config] Add p2p.disc.concurrency flag

AlexiaChen:
## Issue

Because of the block time problem of all shards, the parameter of concurrency needs to be exposed for adjustment. If it is set to 1, it may currently go wrong and affect the consensus. Set to 1, according to the paper, KAD DHT query time will downgrade to Chord‘s algorithm

图片

 NOTE： Add p2p.disc.concurrency flag and default value with libp2p dht

## Test

Passed Local test

```bash
./harmony --p2p.disc.concurrency 5
```

### DIFF
--- a/cmd/harmony/config_migrations.go
+++ b/cmd/harmony/config_migrations.go
@@ -185,6 +185,9 @@ func init() {
 		if confTree.Get("Pprof.ProfileDebugValues") == nil {
 			confTree.Set("Pprof.ProfileDebugValues", defaultConfig.Pprof.ProfileDebugValues)
 		}
+		if confTree.Get("P2P.DiscConcurrency") == nil {
+			confTree.Set("P2P.DiscConcurrency", defaultConfig.P2P.DiscConcurrency)
+		}
 
 		confTree.Set("Version", "2.2.0")
 		return confTree
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -55,7 +55,7 @@ var (
 		p2pIPFlag,
 		p2pKeyFileFlag,
 		p2pDHTDataStoreFlag,
-
+		p2pDiscoveryConcurrencyFlag,
 		legacyKeyFileFlag,
 	}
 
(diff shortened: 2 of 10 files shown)


internal/configs/node/network.go
```diff
@@ -54,6 +54,8 @@ const (
 	DefaultWSPort = 9800
 	// DefaultPrometheusPort is the default prometheus port. The actual port used is 9000+900
 	DefaultPrometheusPort = 9900
+	// DefaultP2PConcurrency is the default P2P concurrency, 0 means is set the default value of P2P Discovery, the actual value is 10
```
rlan35, line 57: what is the default value from libp2p?

rlan35:
please fix travis test.

cmd/harmony/config_migrations.go
```diff
@@ -103,6 +103,10 @@ func init() {
 			confTree.Set("P2P.IP", defaultConfig.P2P.IP)
 		}
 
+		if confTree.Get("P2P.Concurrency") == nil {
```
JackyWYX, line 106: This is a new field, and you need to update the config version. Also it is recommended that you rebase on top of https://github.com/harmony-one/harmony/pull/3773

cmd/harmony/flags.go
```diff
@@ -513,6 +513,11 @@ var (
 		DefValue:   defaultConfig.P2P.KeyFile,
 		Deprecated: "use --p2p.keyfile",
 	}
+	p2pDiscoveryConcurrencyFlag = cli.IntFlag{
+		Name:     "p2p.concurrency",
```
JackyWYX, line 517: Suggest `p2p.disc.concurrency`

p2p/discovery/discovery.go
```diff
@@ -42,6 +42,13 @@ func NewDHTDiscovery(host libp2p_host.Host, opt DHTConfig) (Discovery, error) {
 	if err != nil {
 		return nil, err
 	}
+
```
JackyWYX, line 45: Suggest move to `getLibp2pRawOptions`

internal/configs/node/network.go
AlexiaChen: comment mentioned,  10 is default

cmd/harmony/config_migrations.go
AlexiaChen: need to wait the PR you mentioed merged, and I can rebase on the top of the PR.  Otherwise this PR may still be changed

LeoHChen (commented):
please don't do two things in one PR.  You may either downgrade or change the default value in the flag.  we have no evidence that the libp2p upgrade will cause the issue.

AlexiaChen:
> 
> 
> please don't do two things in one PR. You may either downgrade or change the default value in the flag. we have no evidence that the libp2p upgrade will cause the issue.

Removed downgrade kad-dht code in `go.mod`, let makes only one change.

rlan35 (approved):

PR #3830, opened 2021-07-21
[rpc] add limitation on GetBlocks and GetLogs. Also some other small changes

JackyWYX:
## Issue

Add limitation to GetBlocks and GetLogs.

### Unit Test Coverage

Passed local tests.

Passed mainnet RPC tests.

```
curl -s --request POST 'http://localhost:9500/' --header 'Content-Type: application/json' --data-raw '{ "jsonrpc": "2.0", "method": "hmyv2_getBlocks", "params": [1, 1025, {"withSigners": false, "fullTx": false,"inclStaking": false}], "id": 1}'
> .... # Some data

curl -s --request POST 'http://localhost:9500/' --header 'Content-Type: application/json' --data-raw '{ "jsonrpc": "2.0", "method": "hmyv2_getBlocks", "params": [1, 1026, {"withSigners": false, "fullTx": false,"inclStaking": false}], "id": 1}'
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"GetBlocks query must be smaller than size 1024"}}
```

### DIFF
--- a/rpc/blockchain.go
+++ b/rpc/blockchain.go
@@ -42,7 +42,8 @@ type PublicBlockchainService struct {
 
 const (
 	DefaultRateLimiterWaitTimeout = 5 * time.Second
-	blockCacheLimit               = 256
+	blockCacheLimit               = 2048
+	rpcGetBlocksLimit             = 1024
 )
 
 // NewPublicBlockchainAPI creates a new API for the RPC interface
@@ -68,7 +69,6 @@ func NewPublicBlockchainAPI(hmy *hmy.Harmony, version Version, limiterEnable boo
 			Public:    true,
 		}
 	}
-
 }
 
 // ChainId returns the chain id of the chain - required by MetaMask
@@ -188,7 +188,6 @@ func (s *PublicBlockchainService) GetBlockByNumber(
 			return nil, err
 		}
 	}
-	blockArgs.InclTx = true
 
 	blockNum := blockNumber.EthBlockNumber()
 	if blockNum != rpc.PendingBlockNumber {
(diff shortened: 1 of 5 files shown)


rpc/filters/filter.go
```diff
@@ -220,6 +220,11 @@ func (f *Filter) unindexedLogs(ctx context.Context, end uint64) ([]*types.Log, e
 	var logs []*types.Log
 
 	for ; f.begin <= int64(end); f.begin++ {
+		select {
+		case <-ctx.Done():
```
rlan35, line 224: so we didn't have timeout on these ranging rpcs?

rpc/common/types.go
```diff
@@ -13,9 +13,8 @@ import (
 // BlockArgs is struct to include optional block formatting params.
 type BlockArgs struct {
 	WithSigners bool     `json:"withSigners"`
-	InclTx      bool     `json:"inclTx"`
 	FullTx      bool     `json:"fullTx"`
-	Signers     []string `json:"signers"`
+	Signers     []string `json:"-"`
```
rlan35, line 17: why change to -?

rpc/blockchain.go
```diff
@@ -188,7 +188,6 @@ func (s *PublicBlockchainService) GetBlockByNumber(
 			return nil, err
 		}
 	}
-	blockArgs.InclTx = true
```
rlan35, line 191: was this a bug that all query got txn included in the return?

rlan35 (approved):

rpc/filters/filter.go
JackyWYX: No. Even now we do not have timeout mechanism in this context. I would suggest to add a default timeout for all RPCs after Alex migrated Ethereum RPC module.

rpc/common/types.go
JackyWYX: The field is just a field to store in-memory computation result, and is not used in RPC calls. Thus I disabled the json serialization of this field. 

Reference at https://api.hmny.io/#52f8a4ce-d357-46f1-83fd-d100989a8243.

rpc/blockchain.go
JackyWYX: This looks like a deprecated field and it is not used in any cases. Thus I removed this field.

Issue #3843, opened 2021-08-09
Do not save validator wrapper state data for every block in archival node


rlan35:
## Summary and Current Design

The archival node is meant to be saving all state data for all history blocks so that you can query state from a history block. Currently, our archival node saves all state data for validator state changes (due to block reward on every block). This causes huge storage burden for archival nodes because for each block, the modified validator data are stored again. The reason is that currently validator data are stored as a single data structure and any single change such as block reward change, will cause the whole validator data structure (ValidatorWrapper) to change, which causes the storage of a new copy of the data in the state db.

## Proposal

Modify the archival node state storage logic so that the validator wrapper data are excluded from the storage for every block. This way, the archival node won't waste the db space for storing mostly identical data (except the block reward part). 

Related codes:

validator wrapper data structure: https://github.com/harmony-one/harmony/blob/main/staking/types/validator.go#L89
logic about state commitment on new blocks: https://github.com/harmony-one/harmony/blob/main/core/blockchain.go#L1171

lcgogo:
when will the code release? my explorer node now have 12TB!!! 

./harmony --run=explorer --run.shard=0

Should I need resync or the new version will decrease the useless data automaticlly?

AlexiaChen:
Hey @lcgogo  Need to wait, this improved PR is being tested, the specific release time need to consult @rlan35

lcgogo:
> Hey @lcgogo Need to wait, this improved PR is being tested, the specific release time need to consult @rlan35

Hi Alexia, is this PR included in the latest 4.2.1 ?

sophoah:
@lcgogo it is not

lcgogo:
> @lcgogo it is not

Does 4.3.0 contain this PR?

lcgogo:
> > @lcgogo it is not
> 
> Does 4.3.0 contain this PR?

@rlan35 @AlexiaChen @sophoah

AlexiaChen:
> 
> 
> > > @lcgogo it is not
> > 
> > 
> > Does 4.3.0 contain this PR?
> 
> @rlan35 @AlexiaChen @sophoah

Hey @lcgogo    FYI: https://github.com/harmony-one/harmony/pull/3844#issuecomment-956829146

lcgogo:
> > > > @lcgogo it is not
> > > 
> > > 
> > > Does 4.3.0 contain this PR?
> > 
> > 
> > @rlan35 @AlexiaChen @sophoah
> 
[...]

PR #3844, opened 2021-08-12
[core] Do not save ValidatorWrapper state for every block

AlexiaChen:
## Issue

https://github.com/harmony-one/harmony/issues/3843

Save ValidatorWrapper for every N blocks， In other words, in fact, in the archvial node is that not every block processed needs to be committed, which would cause an explosion of state

## Test

<del>have not passed unit test</del>

- [x] local test

1. use rclone to sync shard0  data 

```bash
rclone -P sync release:pub.harmony.one/mainnet.min/harmony_db_0 harmony_db_0
```

2. run harmony with archival:

```bash
 ./harmony --run=explorer --run.beacon-archive=true --run.archive=true --run.shard=0
Started Explorer server at: 0.0.0.0:5000
Started RPC server at: 127.0.0.1:9500
Started WS server at: 127.0.0.1:9800
```

3.  use hmy to check lastheaders of s0 comapred with explorer on mainnet

```bash
 ./hmy blockchain latest-headers
```
```json
{
  "id": "1",
  "jsonrpc": "2.0",
  "result": {
    "beacon-chain-header": {
      "difficulty": "0x0",
      "epoch": 672,
      "extraData": "0x",
      "gasLimit": "0x4c4b400",
      "gasUsed": "0x8ee371",
      "hash": "0x3a12b5ea8dbd4f0df728023e7be526837ffac17ed0ffebb1a4814c6c4b5be078",
      "logsBloom": "0x00000002000010008000000000000000001000000410003000000000000100000000000000000000000000000000000008080000400000000000000000000000000000000000100000000008400000010000000000000000000000000000200000000010820000000000000000000800000000000000000000000010000000100000000000000000002000000000000000000000000000040000000200040000000000000000000000000000000000000000000000004040000000001000000000000102000000010000000000000804000000000000000000080000240020000000000000010000000100080000000000000800400000000000000000000100",
      "miner": "0xe5dc9dbd0b93231e5d0bf998669463fe990a7211",
      "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "nonce": "0x0000000000000000",
      "number": "0xf9bb3b",
      "parentHash": "0xa5c498f08cbcb3a3ba9fe4daa9b860840caac8428cae529f477ff0296b7d1ca5",
      "receiptsRoot": "0xc7434bccf50e05d63a7539ee58325731eaf8ec759b50c0d28fb7ad2a669851a9",
      "sha3Uncles": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "shardID": 0,
      "stateRoot": "0xf3b0707d83670df68952de20c1a3babf107364cc01c30a255a6a34bfbaf0adcd",
      "timestamp": "0x61245bf9",
      "transactionsRoot": "0xbadee2dcc97132c00efaeb45af8879e9ab35a7e1b82270a2387ed2cb5d571541",
      "viewID": 16366978
    },
    "shard-chain-header": {
      "difficulty": "0x0",
      "epoch": 672,
      "extraData": "0x",
      "gasLimit": "0x4c4b400",
      "gasUsed": "0x8ee371",
      "hash": "0x3a12b5ea8dbd4f0df728023e7be526837ffac17ed0ffebb1a4814c6c4b5be078",
      "logsBloom": "0x00000002000010008000000000000000001000000410003000000000000100000000000000000000000000000000000008080000400000000000000000000000000000000000100000000008400000010000000000000000000000000000200000000010820000000000000000000800000000000000000000000010000000100000000000000000002000000000000000000000000000040000000200040000000000000000000000000000000000000000000000004040000000001000000000000102000000010000000000000804000000000000000000080000240020000000000000010000000100080000000000000800400000000000000000000100",
      "miner": "0xe5dc9dbd0b93231e5d0bf998669463fe990a7211",
      "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "nonce": "0x0000000000000000",
      "number": "0xf9bb3b",
      "parentHash": "0xa5c498f08cbcb3a3ba9fe4daa9b860840caac8428cae529f477ff0296b7d1ca5",
      "receiptsRoot": "0xc7434bccf50e05d63a7539ee58325731eaf8ec759b50c0d28fb7ad2a669851a9",
      "sha3Uncles": "0x0000000000000000000000000000000000000000000000000000000000000000",
      "shardID": 0,
      "stateRoot": "0xf3b0707d83670df68952de20c1a3babf107364cc01c30a255a6a34bfbaf0adcd",
      "timestamp": "0x61245bf9",
      "transactionsRoot": "0xbadee2dcc97132c00efaeb45af8879e9ab35a7e1b82270a2387ed2cb5d571541",
      "viewID": 16366978
    }
  }
}
```

#### Test results

- local shard 0(number 16366406) catched up near latest block(number 16404268) shown on explorer https://explorer.harmony.one/

- `./hmy blockchain validator all-information-by-block-number 0 0xfa89d6` and `./hmy blockchain validator all-information-by-block-number 0 0xfa8a57` and ` ./hmy blockchain validator all-information-by-block-number 0 0xfa8ad8` can get the info related to Validator

[res-json.zip](https://github.com/harmony-one/harmony/files/7059397/res-json.zip)

In the history blocks, some of them can check the validator's information, and some of them can't, because not every bock is committed to Disk, now it is committing the state to Disk once every 128 blocks interval, and the latest state is kept in memory.

- `./hmy blockchain validator all-information-by-block-number 0 latest`

it can always get the info with last , the latest validatorwrapper state always keep in memory 

- storage test result:
 Storage growth rate of archived nodes is more than twice that of non-archived nodes, because other state change(exclude ValidatorWrapper) still commited into disk

run `./hmy blockchain latest-header && du -b ./harmony_db_0`

1. non-archival node (my PR)

|  height   |  bytes  |
|  ----  | ----  |
| 16472034     | 215361268349 |
| 16472084     | 215350751961 |
| 16472158     | 215372972324 |
| 16472835     | 215407678480 |
| 16473577     | 215451929641 |

from above result , summery as follows:

- block range 1543(height 16472034   to  16473577   ):  storage improved: 86MB 

- improved 0.05MB for every block

- improved 86MB for 51 minutes

2. archival node  (my PR)

|  height   |  bytes  |
|  ----  | ----  |
| 16473594        | 215430229194 |
| 16473771        | 215484376000 |
| 16474138        | 215530991142 |
| 16474450        | 215573004597 |
| 16475163        | 215634054394 |

from above result , summery as follows:

- block range 1569((height  16473594    to  16475163   ):  storage improved: 194MB 

- improved 0.12MB for every block

- improved 194MB for 52.3 minutes

3. archival node (main branch)

|  height   |  bytes  |
|  ----  | ----  |
| 16475787          | 215687030143 |
| 16475897          | 215971599919 |
| 16476250          | 216811970983 |
| 16476837        | 218224117296 |
| 16476890        | 218384751733 |
| 16477236        | 219204198410 |

from above result , summery as follows:

- block range 1449 (height 16477236 to 16475787): storage improved: 3354MB
- improved 2.31MB for every block
- improved 3354MB for 48.3 minutes

4. non-archive node (main branch)

|  height   |  bytes  |
|  ----  | ----  |
| 16477238          | 219179679116 |
| 16477849          | 219260393849 |
| 16477935          | 219271328497 |
| 16478138        |   219262309140|
| 16478294        |   219304492248 |
| 16478574        | 219349872133 |

from above result , summery as follows:

- block range 1336 (height 16477238 to 16478574): storage improved: 162MB
- improved 0.12MB for every block
- improved 162MB for 44.5 minutes

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -119,10 +119,9 @@ type BlockChain struct {
 	chainConfig *params.ChainConfig // Chain & network configuration
 	cacheConfig *CacheConfig        // Cache configuration for pruning
 
-	db              ethdb.Database // Low level persistent database to store final content in
-	triegc          *prque.Prque   // Priority queue mapping block numbers to tries to gc
-	gcproc          time.Duration  // Accumulates canonical block processing for trie dumping
-	validatorBlkCnt uint64         // validator wrapper block counter
+	db     ethdb.Database // Low level persistent database to store final content in
+	triegc *prque.Prque   // Priority queue mapping block numbers to tries to gc
+	gcproc time.Duration  // Accumulates canonical block processing for trie dumping
 
 	hc            *HeaderChain
 	rmLogsFeed    event.Feed
@@ -210,7 +209,6 @@ func NewBlockChain(
 		cacheConfig:                   cacheConfig,
 		db:                            db,
 		triegc:                        prque.New(nil),
-		validatorBlkCnt:               0,
 		stateCache:                    state.NewDatabase(db),
 		quit:                          make(chan struct{}),
 		shouldPreserve:                shouldPreserve,
@@ -1174,13 +1172,11 @@ func (bc *BlockChain) WriteBlockWithState(
 	// Commit state object changes to in-memory trie
 	var root common.Hash
 	if bc.cacheConfig.Disabled && !block.IsLastBlockInEpoch() {
-		if bc.validatorBlkCnt >= triesInMemoryBlocksInterval {
+		if block.NumberU64()%triesInMemoryBlocksInterval == 0 {
 			root, err = bc.commitWithValidatorWrapper(state, block.Epoch())
-			bc.validatorBlkCnt = 0
 		} else {
 			root, err = bc.commitWithoutValidatorWrapper(state.Copy(), block.Epoch())
 			bc.commitWithValidatorWrapper(state, block.Epoch())
-			bc.validatorBlkCnt++
 		}
 	} else {
 		root, err = bc.commitWithValidatorWrapper(state, block.Epoch())
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -1178,8 +1178,7 @@ func (bc *BlockChain) WriteBlockWithState(
 			root, err = bc.commitWithValidatorWrapper(state, block.Epoch())
 			bc.validatorBlkCnt = 0
 		} else {
-			stateCopy := state.Copy()
-			root, err = bc.commitWithoutValidatorWrapper(stateCopy, block.Epoch())
+			root, err = bc.commitWithoutValidatorWrapper(state.Copy(), block.Epoch())
 			bc.commitWithValidatorWrapper(state, block.Epoch())
 			bc.validatorBlkCnt++
 		}
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -76,6 +76,7 @@ const (
 	maxTimeFutureBlocks                = 30
 	badBlockLimit                      = 10
 	triesInMemory                      = 128
+	triesInMemoryBlocksInterval        = 128
 	shardCacheLimit                    = 10
 	commitsCacheLimit                  = 10
 	epochCacheLimit                    = 10
@@ -118,9 +119,10 @@ type BlockChain struct {
 	chainConfig *params.ChainConfig // Chain & network configuration
 	cacheConfig *CacheConfig        // Cache configuration for pruning
 
-	db     ethdb.Database // Low level persistent database to store final content in
-	triegc *prque.Prque   // Priority queue mapping block numbers to tries to gc
-	gcproc time.Duration  // Accumulates canonical block processing for trie dumping
+	db              ethdb.Database // Low level persistent database to store final content in
+	triegc          *prque.Prque   // Priority queue mapping block numbers to tries to gc
+	gcproc          time.Duration  // Accumulates canonical block processing for trie dumping
+	validatorBlkCnt uint64         // validator wrapper block counter
 
 	hc            *HeaderChain
 	rmLogsFeed    event.Feed
@@ -208,6 +210,7 @@ func NewBlockChain(
 		cacheConfig:                   cacheConfig,
 		db:                            db,
 		triegc:                        prque.New(nil),
+		validatorBlkCnt:               0,
 		stateCache:                    state.NewDatabase(db),
 		quit:                          make(chan struct{}),
 		shouldPreserve:                shouldPreserve,
@@ -1141,6 +1144,14 @@ func (bc *BlockChain) WriteBlockWithoutState(block *types.Block, td *big.Int) (e
 	return nil
 }
 
+func (bc *BlockChain) commitWithoutValidatorWrapper(state *state.DB, epoch *big.Int) (root common.Hash, err error) {
+	return state.Commit(bc.chainConfig.IsS3(epoch), false)
+}
+
+func (bc *BlockChain) commitWithValidatorWrapper(state *state.DB, epoch *big.Int) (root common.Hash, err error) {
+	return state.Commit(bc.chainConfig.IsS3(epoch), true)
+}
+
 // WriteBlockWithState writes the block and all associated state to the database.
 func (bc *BlockChain) WriteBlockWithState(
 	block *types.Block, receipts []*types.Receipt,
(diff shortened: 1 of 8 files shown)


core/blockchain.go
```diff
@@ -880,6 +883,11 @@ func (bc *BlockChain) Stop() {
 		if size, _ := triedb.Size(); size != 0 {
 			utils.Logger().Error().Msg("Dangling trie nodes after full cleanup")
 		}
+	} else {
+		triedb := bc.stateCache.TrieDB()
+		for !bc.trieRef.Empty() {
+			triedb.Dereference(bc.trieRef.PopItem().(common.Hash))
```
rlan35, line 889: Why do garbage collect here? This part of the code is executed before the node is shut down. No need to garbage collect any memory right?

core/blockchain.go
```diff
@@ -1168,8 +1176,33 @@ func (bc *BlockChain) WriteBlockWithState(
 
 	// Flush trie state into disk if it's archival node or the block is epoch block
 	triedb := bc.stateCache.TrieDB()
+
 	if bc.cacheConfig.Disabled || block.IsLastBlockInEpoch() {
-		if err := triedb.Commit(root, false); err != nil {
+		if bc.cacheConfig.Disabled && !block.IsLastBlockInEpoch() {
```
rlan35, line 1181: We still need to flush the state into disk for archival node, except for validatorWrapper data. Other state changes still need to be flushed to disk.

core/blockchain.go
AlexiaChen: This part of code is idea from here(non-archival node in Stop() ) https://github.com/harmony-one/harmony/blob/d50073f49c7d1210c5502ca2c4de11d3f8959ffa/core/blockchain.go#L880-L882 , Does this mean that there's actually no need to do garbage collection here either, and it's also redundant code?

core/blockchain.go
AlexiaChen: I see, the other states need to be committed every block, while the ValidatorWrapper does not need to be committed per block, but keeps the current commit every N blocks(it still commit finally, It only reduces the frequency of its commit). right ?

core/blockchain.go
rlan35: It’s necessary when the node still runs. But when we exit the node, all memory will be gone. So no need to worry about gc

core/blockchain.go
rlan35: Yes

AlexiaChen:
Hi @rlan35,    Please review it again.   I used a copy of stateDB to modify the state of the deleted ValidatorWrapper, and then commit Flush into disk, and use full state hold in memory to serve RPC like `getValidatorInformation`,  but I feel that the Copy method of stateDB is heavy, is there any optimization suggestions?

core/state/statedb.go
```diff
@@ -760,6 +761,11 @@ func (db *DB) Commit(deleteEmptyObjects bool) (root common.Hash, err error) {
 		if obj := db.stateObjects[addr]; !obj.deleted {
 			// Write any contract code associated with the state object
 			if obj.code != nil && obj.dirtyCode {
+				if !commitValidatorWrapper && obj.IsValidator(db.db) {
+					db.deleteStateObject(obj)
+					obj.deleted = true
```
JackyWYX, line 766: Unfortunately, from my knowledge, if you delete the state object, all information of the address will be deleted, which will make the address unrecoverable. Have you tested the case whether the address is still accessible after this operation?

Also, since the commit block for archive node is not in consensus, could it happen that different archive nodes will have different validatorWrapper commit block, thus resulting in different block hash?

JackyWYX (commented):
Two questions proposed at the comment. The approach is systematically risky as far as I see.

core/state/statedb.go
AlexiaChen: Yes. the address still accessible after this operation, because latest state will keep in memory. So for the Flush to disk operation, I used another copy of the state to modify and commit, but the latest state trie was kept in memory. But for history blocks, you can't rely on the state trie in memory, you need to see if the block is committed to disk. Otherwise, it will be empty.  you can see my test. Definitely does not affect the block hash, because this root value is calculated and only used to commit the state, and does not modify the root in the block header.    Please wait , I will check agian!

JackyWYX: From my knowledge, the state to be stored actually is a different state with a different root from the state in the block. I doubt whether any part of the state from block is recoverable after committing a different state like this. Would you please write a unit test to address my concern?

AlexiaChen: Yes, I know that the state to be stored actually is a different state with a different root from the state in the block, but the goal is that do not save validatorwrapper, it must be different root state from state in the block. if keep root state consistent with block, the validatorwrapper will commited to disk for every block

AlexiaChen: Yes I will write unit test to check

JackyWYX: My understanding is, if the states of 127/128 blocks in archival node are not in storage, thus not retrievable, what is the point of starting an archival node in the first place?

AlexiaChen: you are right! The goal of this issue is essentially a compromise on the archive node state explosion, a workaround, to do perfection is unlikely. A large number of Tx state changes are not discardable. As discussed with RJ before, the best way is to store the members of ValidatorWrapper separately, but that would be complicated. So I have discussed with RJ, and the simple idea is to swipe the disk once every N blocks, but then the state of the Validator cannot be checked on some blocks, my concern is: I don't know if the test can cover all, because the root state is really inconsistent, there may be a corner case that I have not considered
 
do you have any better approaches  to handle  state explosion on archive node?

core/blockchain.go
```diff
@@ -1161,21 +1170,36 @@ func (bc *BlockChain) WriteBlockWithState(
...
 
 	// Commit state object changes to in-memory trie
-	root, err := state.Commit(bc.chainConfig.IsS3(block.Epoch()))
+	var root common.Hash
+	if bc.cacheConfig.Disabled && !block.IsLastBlockInEpoch() {
+		if block.NumberU64()%triesInMemoryBlocksInterval != 0 {
+			root, err = bc.commitWithValidatorWrapper(state, block.Epoch())
+		} else {
+			root, err = bc.commitWithoutValidatorWrapper(state.Copy(), block.Epoch())
+			bc.commitWithValidatorWrapper(state, block.Epoch())
```
LeoHChen, line 1179: no error checking here?

rlan35:
Closing this PR as now we only have updates every 64 blocks.

PR #3850, opened 2021-08-24
[WIP] add history vrf reading capability on vrf precompile

rlan35:


core/evm.go
```diff
@@ -107,6 +108,46 @@ func GetHashFn(ref *block.Header, chain ChainContext) func(n uint64) common.Hash
 	}
 }
 
+// GetVRFFn returns a GetVRFFn which retrieves header vrf by number
+func GetVRFFn(ref *block.Header, chain ChainContext) func(n uint64) common.Hash {
```
LeoHChen, line 112: do you have any limit on n?  what if n is a huge number?

rlan35: The limit is 256 blocks, it's using similar logic as GetHashFn

core/evm.go
```diff
@@ -107,6 +108,46 @@ func GetHashFn(ref *block.Header, chain ChainContext) func(n uint64) common.Hash
...
+			cache = map[uint64]common.Hash{
+				ref.Number().Uint64(): curVRF,
+			}
+		}
+		// Try to fulfill the request from the cache
+		if hash, ok := cache[n]; ok {
+			return hash
+		}
+		// Not cached, iterate the blocks and cache the hashes
+		for header := chain.GetHeader(ref.ParentHash(), ref.Number().Uint64()-1); header != nil; header = chain.GetHeader(header.ParentHash(), header.Number().Uint64()-1) {
```
gupadhyaya, line 133: what is the use case for caching vrf's all previous blocks vs just pulling the requested block number vrf and caching?

rlan35: It’s following the same logic as in getHashFn. I guess it’s safer to trace back blocks rather than directly query by block number.

gupadhyaya (approved):

AlexiaChen (approved):

PR #3858, opened 2021-08-29
Add fork condition for datacopy bug in evm

rlan35:


core/vm/instructions.go
```diff
@@ -767,7 +767,9 @@ func opCall(pc *uint64, interpreter *EVMInterpreter, contract *Contract, memory
 		stack.push(interpreter.intPool.get().SetUint64(1))
 	}
 	if err == nil || err == ErrExecutionReverted {
-		ret = common.CopyBytes(ret)
+		if contract.WithDataCopyFix {
```
sophoah, line 770: @rlan35 I don't understand how delaying `ret = common.CopyBytes(ret)` later after the epoch activation helped with in fixing bug ? additionally `ret = common.CopyBytes(ret)` is already active in our codebase today, can this delay cause any issue with the current evm state when receiving contract call

sophoah: thanks @rlan35 i am adding that https://github.com/harmony-one/harmony/pull/3856/files so we have a reference on when `ret = common.CopyBytes(ret)` was pushed.

sophoah (approved):
LGTM

PR #3876, opened 2021-09-10
Add gas price suggestion on GasPrice rpc

rlan35:
Previously our GasPrice rpc returns a fixed number of 10Gwei as suggested gas price.

Now we are adding a more reasonable gas suggestion logic based on transactions in recent blocks. The main logic are exactly following ethereum's gas suggestion mechanism but with a minimum suggested gas price of 10Gwei.

### DIFF
--- a/hmy/gasprice.go
+++ b/hmy/gasprice.go
@@ -19,12 +19,13 @@ package hmy
 import (
 	"context"
 	"fmt"
-	"github.com/ethereum/go-ethereum/rpc"
-	"github.com/harmony-one/harmony/block"
-	"github.com/harmony-one/harmony/internal/utils"
 	"math/big"
 	"sort"
 	"sync"
+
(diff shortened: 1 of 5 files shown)


JackyWYX (approved):
LGTM!

sophoah:
tested on testnet after the code upgrade : 
curl --location --request POST "https://api.s0.b.hmny.io" \
> --header 'Content-Type: application/json' \
> --data-raw '{
>     "jsonrpc": "2.0",
>     "id": 1,
>     "method": "hmyv2_gasPrice",
>     "params": []
> }'
{"jsonrpc":"2.0","id":1,"result":10000000000}

we'll have to see on mainnet during congestion what would be that number.

sophoah:
@rlan35 would that also work with web3 eth estimateGas  ?

Issue #3881, opened 2021-09-22
Pending transaction subscription not working


gupadhyaya:
Subscribing to pending transactions is not working.

web3.js code
```
const Web3 = require("web3");
const web3 = new Web3("wss://ws.s0.t.hmny.io");
var subscription = web3.eth.subscribe('pendingTransactions', function(error, result){
  if (!error)
      console.log(result);
})
.on("data", function(transaction){
  console.log(transaction);
});
```

similarly, harmony-js sdk code
```
const { Harmony } = require("@harmony-js/core");
const { ChainID, ChainType } = require("@harmony-js/utils");

const hmy_ws = new Harmony("wss://ws.s0.t.hmny.io", {
  chainType: ChainType.Harmony,
  chainId: ChainID.HmyMainnet,
});

let pendingTxns = hmy_ws.blockchain.newPendingTransactions();
pendingTxns.on("data", function(transaction){
  console.log(transaction);
});
```

However, pending txns rpc returns output correctly:
```
curl --location --request POST 'https://rpc.s0.t.hmny.io' \
--header 'Content-Type: application/json' \
--data-raw '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "hmyv2_pendingTransactions",
    "params": []
}'
```

rlan35:
fixed in https://github.com/harmony-one/harmony/pull/3885

sophoah:
code has been pushed to testnet :
and
```
const Web3 = require("web3");
const web3 = new Web3("wss://ws.s0.t.hmny.io");
var subscription = web3.eth.subscribe('pendingTransactions', function(error, result){
  if (!error)
      console.log(result);
})
.on("data", function(transaction){
  console.log(transaction);
});
```
is now working
```
node web3-Pendingtx-ws.js
0x769520008066520063131436e0eb9bdb
0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143
0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143

0x712d2f3f25692363726e352ffb377ba7592e2566cae89250afe4dd749524efa2
0x712d2f3f25692363726e352ffb377ba7592e2566cae89250afe4dd749524efa2
```

PR #3885, opened 2021-09-28
add pending txn subscription support

rlan35:
Feed promoted pending txns into the pending txn subscription channel, so all subscribed source can get the new pending txns.

### DIFF
--- a/core/events.go
+++ b/core/events.go
@@ -22,7 +22,7 @@ import (
 )
 
 // NewTxsEvent is posted when a batch of transactions enter the transaction pool.
-type NewTxsEvent struct{ Txs []*types.Transaction }
+type NewTxsEvent struct{ Txs []types.PoolTransaction }
 
 // PendingLogsEvent is posted pre mining and notifies of pending logs.
 type PendingLogsEvent struct {
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -1290,7 +1290,7 @@ func (pool *TxPool) removeTx(hash common.Hash, outofbound bool) {
 // invalidated transactions (low nonce, low balance) are deleted.
 func (pool *TxPool) promoteExecutables(accounts []common.Address) {
 	// Track the promoted transactions to broadcast them at once
-	var promoted types.PoolTransactions
+	var promoted []types.PoolTransaction
 	logger := utils.Logger().With().Stack().Logger()
 
 	// Gather all the accounts potentially needing updates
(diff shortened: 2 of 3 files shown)


gupadhyaya (approved):

sophoah:
@rlan35 was the PR been tested ? can you attached the test logs if you did ? Also since we have type change what is the impact of our RPC usage ? do we need to any special communication regarding that change ?

gupadhyaya:
> @rlan35 was the PR been tested ? can you attached the test logs if you did ? Also since we have type change what is the impact of our/dapp RPC usage ? do we need to any special communication regarding that change ?

this should not have any regressions. the test logs are in the issue: https://github.com/harmony-one/harmony/issues/3881
should be ok to release.

sophoah:
see test done on testnet here : https://github.com/harmony-one/harmony/issues/3881#issuecomment-930808246

sophoah:
@rlan35 not sure if normal but the test i did shows a duplicate tx each time.
```
node web3-Pendingtx-ws.js
0x769520008066520063131436e0eb9bdb
0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143
0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143
0x712d2f3f25692363726e352ffb377ba7592e2566cae89250afe4dd749524efa2
0x712d2f3f25692363726e352ffb377ba7592e2566cae89250afe4dd749524efa2
```

it's a minor issue but just wanted to let you know

rlan35:
> @rlan35 not sure if normal but the test i did shows a duplicate tx each time.
> 
> ```
> node web3-Pendingtx-ws.js
> 0x769520008066520063131436e0eb9bdb
> 0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143
> 0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143
> 0x712d2f3f25692363726e352ffb377ba7592e2566cae89250afe4dd749524efa2
> 0x712d2f3f25692363726e352ffb377ba7592e2566cae89250afe4dd749524efa2
> ```
> 
> it's a minor issue but just wanted to let you know

@gupadhyaya is this normal, did you notice this in your local testing?

damiensgit:
Just a note that whilst pending txes worked on my local build as above, it appears that getting the transaction for the hash always return nil (on go-etherum client) e.g.

`tx, is_pending, _ := client.TransactionByHash(context.Background(), transactionHash)`

is_pending is true but tx is always nil

damiensgit:
A web3 exmple of my earlier comment:

const web3 = new Web3("testnet"); 
var subscription = web3.eth.subscribe("pendingTransactions", function(err, result) {
    if (!err){
        console.log(result)
    }
}).on("data", function(transaction){
    console.log(transaction)
    web3.eth.getTransaction(transaction).then(console.log);
})


0xd085ab5e305a7f224aca5650f5ee0345d8a58e77729ea5205407cef15b7d6143
null

orlandodiaz:
Im having the same issue as damien.  I also get null transaction data on pending transactions. This only happens on my private rpc node and not on the hmy public one though

PR #3891, opened 2021-09-30
turn on sha3-256 on mainnet and testnet

sophoah:
turning on sha3-256
- mainnet at epoch 725
- testnet at epoch 74570


sophoah:
duplicate with 3890

PR #3894, opened 2021-10-06
Turn pprof default on with local saved files

JackyWYX:
## Issue

Currently, the pprof module is default off. Need to turn it on by default to enable history pprof data persisting on disk.

For `harmony.conf`, it would be better to have `ProfileNames` set to `[]string{"cpu", "heap", "goroutine"}` so that the specified profiles will be persisted into disk.

## Test

Tested and run locally.

### DIFF
--- a/cmd/harmony/config_migrations_test.go
+++ b/cmd/harmony/config_migrations_test.go
@@ -266,7 +266,7 @@ Version = "1.0.4"
   Port = 9000
 
 [Pprof]
-  Enabled = false
+  Enabled = true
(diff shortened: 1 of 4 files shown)


sophoah (approved):
LGTM

sophoah:
@JackyWYX can we add a configurable flag for a pprof log rotation, we could decide to keep only 1 day or 15 days of pprof log. Few reasons :
1) 8.16MB per day can become big over time
2) avoid unnecessary disk maintenance by the node runner

PR #3908, opened 2021-10-26
add prune beacon chain feature

LuttyYang:
## Issue

https://github.com/harmony-one/bounties/issues/84

## Operational Checklist

    NO

    NO

    NO


core/blockchain.go
```diff
@@ -87,6 +87,8 @@ const (
 	pendingCrossLinksCacheLimit        = 2
 	blockAccumulatorCacheLimit         = 256
 	maxPendingSlashes                  = 512
+	// pruneBeaconChainBefore Keep the block for nearly 3 days
+	pruneBeaconChainBefore = 129600
```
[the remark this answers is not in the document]
LuttyYang: fixed

core/rawdb/accessors_chain.go
```diff
@@ -419,3 +419,22 @@ func FindCommonAncestor(db DatabaseReader, a, b *block.Header) *block.Header {
 	}
 	return a
 }
+
+func IteratorBlocks(iterator DatabaseIterator, cb func(blockNum uint64, hash common.Hash) bool) {
+	iter := iterator.NewIteratorWithPrefix(headerPrefix)
+	defer iter.Release()
+
+	for iter.Next() {
```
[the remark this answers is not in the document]
LuttyYang: It will not iterating through all blocks, Every time we prune beacon chain data within a limited time(currently it's 200ms).

See https://github.com/harmony-one/harmony/pull/3908/files#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R1347 and https://github.com/harmony-one/harmony/pull/3908/files#diff-53d5f4b8a536ec2a8c8c92bf70b8268f1d77ad77e9f316e6f68a2bcae5303215R1372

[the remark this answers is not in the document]
LuttyYang: currently the pruning operator is not blocking

LuttyYang:
@JackyWYX You are right, the major contributor to the storage is validator snapshot (in testnet shard 1 beacon chain, it almost 76%)

I added prune validator snapshot feature, But this requires @JackyWYX to evaluate whether it is safe to delete.


testnet shard 1 beacon chain data constitute

LuttyYang:
In testnet test: db size from 43GB to 8.7GB

image

Every pruneBeaconChain took almost 300ms in ssd disk
image

More infomation will update in next day

rlan35:
> @JackyWYX You are right, the major contributor to the storage is validator snapshot (in testnet shard 1 beacon chain, it almost 76%, mainnet snapshot is damaged, calculation later)
> 
> I added prune validator snapshot, tx and cx feature, But this requires @rlan35 to evaluate whether it is safe to delete.
> 
> **testnet shard 1 beacon chain data constitute(yellow line is prune data, almost 90% size)** image

Testnet doesn't have many transactions and testnet epoch is every 4 hours. So you will see the validatorSnapshot taking more spaces. But in mainnet. epoch is every 18 hours and there is lots of txns. So I doubt the situation will be a lot different there. Do you mind analyzing the storage cost for mainnet instead?

core/blockchain.go
```diff
@@ -1255,6 +1261,16 @@ func (bc *BlockChain) WriteBlockWithState(
 		return NonStatTy, err
 	}
 
+	if bc.chainConfig.IsEnablePruneBeaconChainFeature() {
+		nowBlockNum := block.Number().Uint64()
+		maxBlockNum := big.NewInt(0).Sub(block.Number(), big.NewInt(pruneBeaconChainBlockBefore)).Uint64()
+		maxEpoch := big.NewInt(0).Sub(block.Epoch(), big.NewInt(pruneBeaconChainBeforeEpoch))
+		if rand.Intn(100) == 1 && nowBlockNum > pruneBeaconChainBlockBefore && maxEpoch.Uint64() > pruneBeaconChainBeforeEpoch {
```
rlan35, line 1268: why not prune 1 block for each new block inserted?

rlan35: These two conditions seems not quite reasonable "nowBlockNum > pruneBeaconChainBlockBefore && maxEpoch.Uint64() > pruneBeaconChainBeforeEpoch"

rlan35: And These two conditions seems not quite reasonable "nowBlockNum > pruneBeaconChainBlockBefore && maxEpoch.Uint64() > pruneBeaconChainBeforeEpoch"

cmd/harmony/main.go
```diff
@@ -290,6 +290,10 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 	// Update ethereum compatible chain ids
 	params.UpdateEthChainIDByShard(nodeConfig.ShardID)
 
+	if hc.Feature != nil && hc.Feature.EnablePruneBeaconChain {
+		nodeConfig.Feature.EnablePruneBeaconChain = hc.Feature.EnablePruneBeaconChain
```
rlan35, line 294: Please make sure the beacon chain pruning config only works on non shard 0 validators.

core/rawdb/accessors_chain.go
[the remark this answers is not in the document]
rlan35: I don't quite understand. What's the range that this iterator is iterating on?

LuttyYang:
@rlan35 

You are rignt! We can see I see a lot of unknown (other) data in the database, The length of their keys is 32 bytes, I haven't found their source in the code yet. I guess they are a transactions data, But when I hex them, I can’t find these transactions.

Here are some samples:
Keys Hex: `6ab3de58182b6dcf0c8b31253391448c7935ccc59243302f925164c2155c3094`
Value Hex: `f39f3540171b6c0c960b71a7020d9f60077f6af931a8bbf590da0223dacf75c7af929114d4596f5b8cb26d1bad2cdd431e51d651`

Keys Hex: `4c8eef247a6311b6e7856e427dda2a996f61bce864544f99711e57a8a8bd930e`
Value Hex: `f9013180a004d64835b820db52817e07de5fb7c3b3ae270530f565786fdbaedc40cea4fab4a0fc92b3b82ca72aec1ddac408475526a96e2d91b780978a132d02c14dbe620b5da0e10d74b13bf4e71631f0e0e727af875787310406ff5030fdbd7bab80e8bf7ada808080a035be599177102c05d6871a732494d274258a44b9f6c60cb1bb2e672099e0f9c8a06389d0f4f2dfc45dd7c0d09fa221d90f760b2f98e41aa77715fec5080618269680a062857465df814e3f47663c3a4adebc05813217266a77615e875a0375af5ae2afa0707a1a18b56c87f75649155acea750f357e4c9a18bbfe5db5202e46734a4b7bb8080a016e154a3f57f49ee4fd430465c2b9db9a9d905e2a5b93afa47074e0ff7a72316a0f80154a8d5b2a94bb815f1dbb2a7d9ca4fa79ea02166cbee3b5f56aba5f39fd980`

Keys Hex: `57b4246adf4b54e875dc93eab4110248d0567794a700b92781015af4efb29ee5`
Value Hex: `f90211a08e778cd54bc7ba0d8ffc3c38969b704da48c62e27273f70fae3176feb28df708a05d9760eb6ef776e266ecd1da9740bc76bbc0fe1384b74728919bbe075299ec6ea0f46d9c99a2f3eae5d7dcd89817be6162e6011dd0a70df1fa757f0538fa7c0822a05ba0429828e80db268d016202ea6db9d4decb2d2297bc68f620aee8cfbd47d28a0b336865f8bb1228cc63778962d3b65f3f923273206e089d5c7bc9fcc51b69f80a053abf54a189e1f03067778456628f284324ac80d33db997e9573ee81ecdb6ccaa0358f56aa7deebac13ce473e48e654c475e9a605337e608c09b9b9ceae972795da084ddbf05b94e8e37f8fe6520889c844546634853245b533e3400400939cbba6ca068648ba8e09cd3e08c395774ac7d22b22bf8c86730397ef20e948e03fb4a70c3a03cd32572f2eb6a3246006e7f406e72dcfd1604772d1ceb66f016857e575b8c53a0227703cbed7f8faec7362c884a179301ad999d80095f4eb78b340136d6437fafa05d39cd0083c712ad87072610ab94f34eaf4b7f9117a0aae802650cafd63d75a1a0baeac6ad013c36ea449b17ea099645f7559522ea4d564922c76768360e2ef85aa0700876751332561790f298106e2448f67e68b6e269889c7670d12bc43c604555a050e55e16459e614cb749feea79301d5add78fb43a5d7c46e22dffa630fbb50c3a07a0b63459c2e5092e85eb4c41ad34aa48ea2179d5edebacbb3ad29527746508d80`


**mainnet shard 1 beacon chain data constitute(yellow line is prune data, almost 60% size)**

image

core/blockchain.go
LuttyYang: use `rand.Intn(100) == 1` has two advantages:
1. Each time the data of up to 10,000 keys is deleted, this is enough to keep the deletion to the latest with this probability. At the same time, it can reduce the number of IOs and minimize the impact, The Compact that contains the following uses a lower probability, which is the same reason.
2. If delete every new block added, this is equivalent to directly increasing the IO of the add new block, which will affect every node. If we use probabilization, it will at least not be at the same time.

This is prune 1 block for each new block, and compact every 10 blocks:
image

it used avg 23.9M/s disk write.

This is prune 1 block for every 100 blocks, and compact every 1000 blocks:
image

it used avg 8.7M/s disk write.

LuttyYang: And These two conditions "nowBlockNum > pruneBeaconChainBlockBefore && maxEpoch.Uint64() > pruneBeaconChainBeforeEpoch", because `Uint64` Is unsigned, If a user does not use the snapshot to start a node,  Deletion will not start until 16 epoch are synchronized

cmd/harmony/main.go
LuttyYang: This is already limited, see:

https://github.com/harmony-one/harmony/pull/3908/files#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618R159

core/blockchain.go
rlan35: I don't get this "This is prune 1 block for every 100 blocks, and compact every 1000 blocks(in testnet shard 1):" shouldn't you prune 100 blocks for every 100 blocks? 

The difference here is whether you do small work for each block, or you batch all the work together every X blocks. IMO, spreading the work evenly for every block is less impactful to the node. But of course, you can prove me wrong with testing.

rlan35:
> @rlan35
> 
> You are rignt! We can see situation will be a lot different there.
> 
> **mainnet shard 1 beacon chain data constitute(yellow line is prune data, almost 60% size)**
> 
> image
> 
> And we can see I see a lot of unknown (other) data in the database, The length of their keys is 32 bytes, I haven't found their source in the code yet. I guess they are a transactions data, But when I hex them, I can’t find these transactions.
> 
> Here are some samples: Keys Hex: `6ab3de58182b6dcf0c8b31253391448c7935ccc59243302f925164c2155c3094` Value Hex: `f39f3540171b6c0c960b71a7020d9f60077f6af931a8bbf590da0223dacf75c7af929114d4596f5b8cb26d1bad2cdd431e51d651`
> 
> Keys Hex: `4c8eef247a6311b6e7856e427dda2a996f61bce864544f99711e57a8a8bd930e` Value Hex: `f9013180a004d64835b820db52817e07de5fb7c3b3ae270530f565786fdbaedc40cea4fab4a0fc92b3b82ca72aec1ddac408475526a96e2d91b780978a132d02c14dbe620b5da0e10d74b13bf4e71631f0e0e727af875787310406ff5030fdbd7bab80e8bf7ada808080a035be599177102c05d6871a732494d274258a44b9f6c60cb1bb2e672099e0f9c8a06389d0f4f2dfc45dd7c0d09fa221d90f760b2f98e41aa77715fec5080618269680a062857465df814e3f47663c3a4adebc05813217266a77615e875a0375af5ae2afa0707a1a18b56c87f75649155acea750f357e4c9a18bbfe5db5202e46734a4b7bb8080a016e154a3f57f49ee4fd430465c2b9db9a9d905e2a5b93afa47074e0ff7a72316a0f80154a8d5b2a94bb815f1dbb2a7d9ca4fa79ea02166cbee3b5f56aba5f39fd980`
> 
> Keys Hex: `57b4246adf4b54e875dc93eab4110248d0567794a700b92781015af4efb29ee5` Value Hex: `f90211a08e778cd54bc7ba0d8ffc3c38969b704da48c62e27273f70fae3176feb28df708a05d9760eb6ef776e266ecd1da9740bc76bbc0fe1384b74728919bbe075299ec6ea0f46d9c99a2f3eae5d7dcd89817be6162e6011dd0a70df1fa757f0538fa7c0822a05ba0429828e80db268d016202ea6db9d4decb2d2297bc68f620aee8cfbd47d28a0b336865f8bb1228cc63778962d3b65f3f923273206e089d5c7bc9fcc51b69f80a053abf54a189e1f03067778456628f284324ac80d33db997e9573ee81ecdb6ccaa0358f56aa7deebac13ce473e48e654c475e9a605337e608c09b9b9ceae972795da084ddbf05b94e8e37f8fe6520889c844546634853245b533e3400400939cbba6ca068648ba8e09cd3e08c395774ac7d22b22bf8c86730397ef20e948e03fb4a70c3a03cd32572f2eb6a3246006e7f406e72dcfd1604772d1ceb66f016857e575b8c53a0227703cbed7f8faec7362c884a179301ad999d80095f4eb78b340136d6437fafa05d39cd0083c712ad87072610ab94f34eaf4b7f9117a0aae802650cafd63d75a1a0baeac6ad013c36ea449b17ea099645f7559522ea4d564922c76768360e2ef85aa0700876751332561790f298106e2448f67e68b6e269889c7670d12bc43c604555a050e55e16459e614cb749feea79301d5add78fb43a5d7c46e22dffa630fbb50c3a07a0b63459c2e5092e85eb4c41ad34aa48ea2179d5edebacbb3ad29527746508d80`
> 
> These data make our compact require more IO and longer time, Do you know what these are

These are all the defined keys for the data entry: https://github.com/harmony-one/harmony/blob/main/core/rawdb/schema.go#L31. Don't see any keys as plane 32 bytes hash. That's weird.

core/blockchain.go
rlan35: The condition is confusing. Better just make it directly readable. Don't rely on uint overflow for condition, it's easily causing mistakes. For example, "maxEpoch.Uint64() > pruneBeaconChainBeforeEpoch" this one seems wrong. When maxEpoch is negative, it works. When it's positive. It means it is only true when maxEpoch > 16? Then block.Epoch() have to be 32?

core/blockchain.go
```diff
@@ -1272,6 +1288,156 @@ func (bc *BlockChain) WriteBlockWithState(
 	return CanonStatTy, nil
 }
 
+func (bc *BlockChain) pruneBeaconChain(db ethdb.Database, maxBlockNum uint64, maxEpoch *big.Int) {
+	// check is start
+	if atomic.CompareAndSwapInt32(&bc.isPruneBeaconChainRunning, 0, 1) {
```
rlan35, line 1293: Can you implement the same logic with just Mutex? It's simpler to understand and less error-prone.

LuttyYang: This equal `TryLock`, golang 'sync/mutex' package currently no this method, We can import other packages to solve this problem, but I don’t think it is necessary.

If we use `Lock` method, If the processing is slow, a waiting goroutine will create, this may cause memory leaks

core/blockchain.go
LuttyYang: I corrected the conditions. and And I re-observed, most of the IO in compact stage. It's prune 1 block for each new block inserted now.

LuttyYang:
@rlan35 

Mainnet shard 1, This reduces almost 110GB of space.
image

There are still possibilities that can be reduced, but first we need to know what unknown are

core/blockchain.go
JackyWYX: You may make this a function, e.g. `isPruning`. Also works for defer

core/blockchain.go
```diff
@@ -1272,6 +1287,156 @@ func (bc *BlockChain) WriteBlockWithState(
...
+
+		// limit time spent
+		startTime := time.Now()
+		deletedBlockCount := 0
+		deletedValidatorSnapshot := 0
+		deleter := db.NewBatch()
+		deleteRange := make([][2][]byte, 0)
+
+		defer func() {
+			if err := deleter.Write(); err != nil {
```
JackyWYX, line 1303: Looks like the deleter batch is a little too overwhelmed. Have you done metric on memory usage?

core/blockchain.go
LuttyYang: @JackyWYX Good idea, I optimized it

core/blockchain.go
LuttyYang: leveldb is very good for this optimization, but I also added a memory limit now, it will not use more than 4MB of memory

rlan35:
Please make sure that the feature is turned off by default and only enabled if specifically configured. And also fix the Travis tests.

LuttyYang:
> Please make sure that the feature is turned off by default and only enabled if specifically configured. And also fix the Travis tests.

It's turned off by default, Travis Has been successed

LuttyYang:
@rlan35 

I think the unknown data is the `Ethereum Merkle Patricia Trie Data`

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+	deletedValidatorSnapshot int
+	skipValidatorSnapshot    int
+}
+
+func newBlockchainPruner() *blockchainPruner {
+	return &blockchainPruner{}
+}
+
+func (bp *blockchainPruner) initPruneBeaconChain(db ethdb.Database) error {
+	if bp.batchWriter == nil {
```
JackyWYX, line 50: Why don't you just put this code in `newBlockchainPruner`? Lazy initialization is not really needed in this scenario.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+func (bp *blockchainPruner) runWhenNotPruning(cb func()) {
+	// check is start
+	if atomic.CompareAndSwapInt32(&bp.isPruneBeaconChainRunning, 0, 1) {
+		defer atomic.StoreInt32(&bp.isPruneBeaconChainRunning, 0)
+
+		cb()
+	}
+}
+
+func (bp *blockchainPruner) StartPruneBeaconChain(db ethdb.Database, maxBlockNum uint64, maxEpoch *big.Int) {
```
JackyWYX, line 119: Recommend to rename to `Start`.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+	maxDeleteBlockOnce = 10000
+	// maxMemoryUsage max memory usage is 4 MB
+	maxMemoryUsage = 4 * 1024 * 1024
+	// compactProportion compact proportion is thousandth
+	compactProportion = 1000
+)
+
+type blockchainPruner struct {
+	isPruneBeaconChainRunning int32 // isPruneBeaconChainRunning  must be called atomically
+
```
JackyWYX, line 35: Suggest to add `core.BlockChain`  in the struct.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+		cb()
+	}
+}
+
+func (bp *blockchainPruner) StartPruneBeaconChain(db ethdb.Database, maxBlockNum uint64, maxEpoch *big.Int) {
+	bp.runWhenNotPruning(func() {
+		err := bp.initPruneBeaconChain(db)
+		if err != nil {
+			utils.Logger().Info().Err(err).Msg("pruneBeaconChain init error")
```
JackyWYX, line 123: This is actually not a very good coding practice in harmony code base. I suggest return the error and let caller function to handle/log the errors in lower modules.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+			return false
+		}
+
+		// skip genesis block
+		if blockNum == 0 {
+			return true
+		}
+
+		blockInfo := rawdb.ReadBlock(reader, hash, blockNum)
+		err := bp.deleteBlockInfo(blockInfo)
```
JackyWYX, line 180: Suggest once pruned, add a flag in db. So that when the next time prune is started, the same block does not need to prune twice.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+				Msg("pruneBeaconChain delete block error")
+			return false
+		}
+
+		// limit time spent
+		bp.deletedBlockCount++
+		return bp.deletedBlockCount < maxDeleteBlockOnce
+	})
+
+	bp.deleteRange = append(bp.deleteRange, [2][]byte{minKey, maxKey})
```
JackyWYX, line 203: I would personally suggest to return the slice instead of adding the result to structure. [Purely functional programming](https://en.wikipedia.org/wiki/Purely_functional_programming) shall be preferred over stateful ones.

internal/configs/harmony/harmony.go
```diff
@@ -27,6 +27,7 @@ type HarmonyConfig struct {
 	Revert     *RevertConfig     `toml:",omitempty"`
 	Legacy     *LegacyConfig     `toml:",omitempty"`
 	Prometheus *PrometheusConfig `toml:",omitempty"`
+	Feature    *FeatureConfig    `toml:",omitempty"`
```
JackyWYX, line 30: I would suggest to add it to `GeneralConfig`. We shall make each config in `HarmonyConfig` for a micro-service within the node codebase. In this case, `Pruning` itself is not a service, but a part of `BlockChain`.

internal/params/config.go
```diff
@@ -638,3 +643,12 @@ func (c *ChainConfig) Rules(epoch *big.Int) Rules {
 		IsSHA3:       c.IsSHA3(epoch),
 	}
 }
+
+func (c *ChainConfig) EnablePruneBeaconChainFeature() {
+	c.pruneBeaconChainEnable = true
```
JackyWYX, line 648: This shall not be part of chainConfig. Keep it within `HarmonyConfig`. `ChainConfig` is all about chain fork params, pruning is not effecting chain config IMO.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+			// limit time spent
+			bp.deletedValidatorSnapshot++
+		} else {
+			bp.skipValidatorSnapshot++
+		}
+
+		return bp.deletedValidatorSnapshot < maxDeleteBlockOnce
+	})
+
+	bp.deleteRange = append(bp.deleteRange, [2][]byte{minKey, maxKey})
```
JackyWYX, line 255: Could you specify a little more about how the deletion is worked?

core/blockchain_pruner.go
```diff
@@ -0,0 +1,256 @@
...
+					Msg("pruneBeaconChain compact db success")
+			}
+		}()
+
+		bp.pruneBeaconChainBlock(db, maxBlockNum)
+		bp.pruneBeaconChainValidatorSnapshot(db, maxEpoch)
+	})
+}
+
+func (bp *blockchainPruner) pruneBeaconChainBlock(reader ethdb.Database, maxBlockNum uint64) {
```
JackyWYX, line 168: Looks like this will iterate through the whole beaconchain for each start. Any more elegant solution?

core/blockchain.go
```diff
@@ -1255,6 +1257,15 @@ func (bc *BlockChain) WriteBlockWithState(
 		return NonStatTy, err
 	}
 
+	if bc.chainConfig.IsEnablePruneBeaconChainFeature() {
+		if block.Number().Cmp(big.NewInt(pruneBeaconChainBlockBefore)) > 0 && block.Epoch().Cmp(big.NewInt(pruneBeaconChainBeforeEpoch)) > 0 {
+			// probability of 1 in 100 to prune
+			maxBlockNum := big.NewInt(0).Sub(block.Number(), big.NewInt(pruneBeaconChainBlockBefore)).Uint64()
+			maxEpoch := big.NewInt(0).Sub(block.Epoch(), big.NewInt(pruneBeaconChainBeforeEpoch))
+			go bc.blockchainPruner.StartPruneBeaconChain(bc.db, maxBlockNum, maxEpoch)
```
JackyWYX, line 1265: Looks messy. So there will be multiple goroutines doing the prune at the same time, and the same block looks like will be pruned multiple times... This is not optimal. You should write some scheduler logic for scheduling the pruning tasks.

JackyWYX:
Welcome for discussion for the comments BTW.

core/blockchain_pruner.go
LuttyYang: Removed lazy initialization.

core/blockchain_pruner.go
LuttyYang: Renamed

core/blockchain_pruner.go
LuttyYang: You are right, edited.

core/blockchain_pruner.go
LuttyYang: You are right, edited.

internal/configs/harmony/harmony.go
LuttyYang: You are right, edited.

core/blockchain_pruner.go
LuttyYang: Add `core.BlockChain` for what

core/blockchain_pruner.go
LuttyYang: the same block does not prune twice. After each block is deleted, it will not appear again in the next iteration. And iterating will be faster than confirming the existence of multiple times.

core/blockchain_pruner.go
LuttyYang: It's not `deleteRange`, It's `compactRange`, After levelDB deletes the data, the space cannot be released immediately, there is to release the deleted space quickly

core/blockchain_pruner.go
LuttyYang: It does not directly iterate the entire beaconchain, it will quickly find the first block that has not been deleted (same as Get), and then read sequentially backwards, This is a feature of LevelDB

core/blockchain.go
LuttyYang: As above, the same block will not be iterated or deleted multiple times, and there is an atomic lock inside, so only one is executing at the same time

LuttyYang:
You can use the following of the code to verify that iteration is faster and simpler than mark in db.

```
package main

import (
	"encoding/binary"
	"github.com/syndtr/goleveldb/leveldb"
	"github.com/syndtr/goleveldb/leveldb/util"
	"os"
	"testing"
)

const (
	testDir = "test_db"
	testOnce = 1000
)

func dbName(i int) []byte {
	key := make([]byte, 4)
	binary.LittleEndian.PutUint32(key, uint32(i))

	return append([]byte{'H'}, key...)
}

func prepareDB(testTimes int) *leveldb.DB {
	db, err := leveldb.OpenFile(testDir, nil)
	if err != nil {
		panic(err)
	}

	// first, fill data
	for i := 0; i < testTimes; i++ {
		key := dbName(i)
		err := db.Put(key, key, nil)
		if err != nil {
			panic(err)
		}
	}

	return db
}

func clearDB(db *leveldb.DB) {
	err := db.Close()
	if err != nil {
		panic(err)
	}

	err = os.RemoveAll(testDir)
	if err != nil {
		panic(err)
	}
}

func BenchmarkDBName(b *testing.B) {
	for i := 0; i < b.N; i++ {
		dbName(i)
	}
}

func BenchmarkMarkInDB(b *testing.B) {
	db := prepareDB(b.N * testOnce)
	defer clearDB(db)

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		var lastBlockNum, j int
		lastBlockByte, err := db.Get([]byte("Last"), nil)
		if err != nil {
			if err != leveldb.ErrNotFound {
				panic(err)
			}
			lastBlockNum = 0
		} else {
			lastBlockNum = int(binary.LittleEndian.Uint32(lastBlockByte))
		}

		for j = lastBlockNum; j < lastBlockNum+testOnce; j++ {
			key := dbName(j)

			err := db.Delete(key, nil)
			if err != nil {
				panic(err)
			}
		}

		lastBlockByte = make([]byte, 4)
		binary.LittleEndian.PutUint32(lastBlockByte, uint32(j))
		err = db.Put([]byte("Last"), lastBlockByte, nil)
		if err != nil {
			panic(err)
		}
	}
}

func BenchmarkIterDB(b *testing.B) {
	db := prepareDB(b.N * testOnce)
	defer clearDB(db)

	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		j := 0

		iterator := db.NewIterator(util.BytesPrefix([]byte{'H'}), nil)
		for j < testOnce {
			j++

			err := db.Delete(iterator.Key(), nil)
			if err != nil {
				panic(err)
			}
		}
		iterator.Release()
	}
}

```

image

OneUnitedPowerValidator:
Is this still in consideration to be implemented?

LuttyYang:
@OneUnitedPowerValidator Currently under review

LeoHChen:
This feature needs a configuration flag to enable it, and disable it by default.  Please add a configuration flag.

LuttyYang:
@LeoHChen There is an optional configuration to open, But once opened, if you want to close, you need to rclone again

LeoHChen:
a few questions,

1) does a validator need to sync from the very beginning to use this new feature?
2) if this feature is turned on existing db, will it prune existing db, or just reduce the size on new db syncing?
3) can we create a new rclone repo with this feature on, from the genesis block so that new validators may rclone a much smaller db?

LuttyYang:
@LeoHChen 

1. No, Can be opened and closed at any time
2. When Opened, it prune existing db
3. Yes, After this test on mainnet has no problems for a period of time, This is of course the best

soltrac:
Do we have any ETA on this PR?

jhd2best:
hey @LuttyYang @rlan35 @LeoHChen 

tested this PR as the validator on shard1 of mainnet, and the result as follows:

first of all, our normal node's db size of shard1 is `929G`
#### Node 1
using binary build by this PR, synced from latest block(at that time), and the storage used now
image

#### Node 2
using binary build by this PR, synced from 0 block, and the storage used now
image

both nodes can sign blocks normally (48 hours of attention). It seems that if we start syncing from block 0, we will save a lot of space. Next, the database synchronized from block 0 can be made into a snapshot to provide validator download

upload_to_test.sh
```diff
@@ -0,0 +1,8 @@
+rm -rf bin
```
LeoHChen, line 1: remove this test script please

internal/params/config.go
LeoHChen: Can you reorg it based on Jacky's comment?

core/rawdb/accessors_offchain.go
```diff
@@ -184,6 +184,27 @@ func DeleteValidatorSnapshot(db DatabaseDeleter, addr common.Address, epoch *big
 	return nil
 }
 
+func IteratorValidatorSnapshot(iterator DatabaseIterator, cb func(addr common.Address, epoch *big.Int) bool) (minKey []byte, maxKey []byte) {
+	iter := iterator.NewIteratorWithPrefix(validatorSnapshotPrefix)
+	defer iter.Release()
+
+	minKey = headerPrefix
+	for iter.Next() {
+		// validatorSnapshotKey = validatorSnapshotPrefix + addr bytes (20 bytes) + epoch bytes
```
LeoHChen, line 193: remove the comment please

core/rawdb/accessors_offchain.go
```diff
@@ -184,6 +184,27 @@ func DeleteValidatorSnapshot(db DatabaseDeleter, addr common.Address, epoch *big
...
+		key := iter.Key()
+
+		maxKey = key
+		addressBytes := key[len(validatorSnapshotPrefix) : len(validatorSnapshotPrefix)+20]
+		epochBytes := key[len(validatorSnapshotPrefix)+20:]
```
LeoHChen, line 198: where is the 20 coming from? please comment it

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+
+	return nil
+}
+
+func (bp *blockchainPruner) pruneBeaconChainValidatorSnapshot(maxEpoch *big.Int) (minKey []byte, maxKey []byte) {
+	return rawdb.IteratorValidatorSnapshot(bp.db, func(addr common.Address, epoch *big.Int) bool {
+		if epoch.Cmp(maxEpoch) < 0 {
+			err := rawdb.DeleteValidatorSnapshot(bp, addr, epoch)
+			if err != nil {
+				utils.Logger().Info().
```
LeoHChen, line 200: use Logger().Error() here

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+			utils.Logger().Info().
+				Uint64("blockNum", blockNum).
+				AnErr("err", err).
+				Msg("pruneBeaconChain delete block info error")
+			return false
+		}
+
+		err = rawdb.DeleteBlock(bp, hash, blockNum)
+		if err != nil {
+			utils.Logger().Info().
```
LeoHChen, line 155: use Logger().Error()

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+
+		// skip genesis block
+		if blockNum == 0 {
+			return true
+		}
+
+		blockInfo := rawdb.ReadBlock(bp.db, hash, blockNum)
+		err := bp.deleteBlockInfo(blockInfo)
+		if err != nil {
+			utils.Logger().Info().
```
LeoHChen, line 146: Use Logger().Error()

consensus/votepower/roster.go
```diff
@@ -103,7 +103,7 @@ type topLevelRegistry struct {
 type Roster struct {
 	Voters map[bls.SerializedPublicKey]*AccommodateHarmonyVote
 	topLevelRegistry
-	ShardID uint32
+	ShardID      uint32
```
LeoHChen, line 106: please don't change the whitespace here

core/blockchain.go
```diff
@@ -1255,6 +1257,21 @@ func (bc *BlockChain) WriteBlockWithState(
 		return NonStatTy, err
 	}
 
+	if bc.chainConfig.IsEnablePruneBeaconChainFeature() {
+		if block.Number().Cmp(big.NewInt(pruneBeaconChainBlockBefore)) > 0 && block.Epoch().Cmp(big.NewInt(pruneBeaconChainBeforeEpoch)) > 0 {
+			// probability of 1 in 100 to prune
```
LeoHChen, line 1262: why use a 1/100 probability to prune?

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+				Uint64("blockNum", blockNum).
+				AnErr("err", err).
+				Msg("pruneBeaconChain delete block error")
+			return false
+		}
+
+		// limit time spent
+		bp.deletedBlockCount++
```
LeoHChen, line 163: please add some promethesus metrics of the deletedBlockCount

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	for _, stx := range info.StakingTransactions() {
+		err := rawdb.DeleteTxLookupEntry(bp, stx.Hash())
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
```
LeoHChen, line 192: can you add a debug message?

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+				utils.Logger().Info().
+					Str("addr", addr.Hex()).
+					Uint64("epoch", epoch.Uint64()).
+					AnErr("err", err).
+					Msg("pruneBeaconChain delete validator snapshot error")
+				return false
+			}
+
+			// limit time spent
+			bp.deletedValidatorSnapshot++
```
LeoHChen, line 209: metrics please

core/blockchain_pruner.go
```diff
@@ -0,0 +1,220 @@
...
+		} else {
+			bp.skipValidatorSnapshot++
```
LeoHChen, line 211: metrics please

core/gen_genesis.go
```diff
@@ -5,6 +5,7 @@ package core
 import (
 	"encoding/json"
 	"errors"
+
```
LeoHChen, line 8: please remove the whitespace change

LeoHChen (commented):
final round of review comments,

core/blockchain_pruner.go
LuttyYang: added

internal/params/config.go
LuttyYang: fixed

LeoHChen (approved):

JackyWYX (approved):

core/blockchain_pruner.go
```diff
@@ -0,0 +1,227 @@
...
+		}
+
+		// limit time spent
+		bp.deletedBlockCount++
+		return bp.deletedBlockCount < maxDeleteBlockOnce
+	})
+}
+
+func (bp *blockchainPruner) deleteBlockInfo(info *types.Block) error {
+	for _, cx := range info.IncomingReceipts() {
```
JackyWYX, line 176: Can info be a nil pointer?

core/blockchain_pruner.go
```diff
@@ -0,0 +1,227 @@
...
+		if blockNum >= maxBlockNum {
+			return false
+		}
+
+		// skip genesis block
+		if blockNum == 0 {
+			return true
+		}
+
+		blockInfo := rawdb.ReadBlock(bp.db, hash, blockNum)
```
JackyWYX, line 150: Looks like if a certain block is pruned in last iteration, block Info can be nil. Could this possibly happen? If we reach a block that is pruned previously, does the prune iteration continue or break?

JackyWYX (commented):
I wonder how long it takes for the first iteration of pruning? 

If the first iteration is interrupted with Ctrl+C, how will it effect the second iteration?

For example, current block number = 10000. For the first pruning, I shut down the node when it finished pruning for 0~1000. The next time I start the node, is it true that the pruning will panic when pruning block 0?

core/blockchain_pruner.go
LuttyYang: optimized it

node/node.go
```diff
@@ -155,6 +155,12 @@ func (node *Node) Beaconchain() *core.BlockChain {
 	if err != nil {
 		utils.Logger().Error().Err(err).Msg("cannot get beaconchain")
 	}
+	// only available in validator node and shard 1-3
+	isEnablePruneBeaconChain := node.HarmonyConfig != nil && node.HarmonyConfig.General.EnablePruneBeaconChain
```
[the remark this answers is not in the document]
LuttyYang: optimized it

LuttyYang:
> I wonder how long it takes for the first iteration of pruning?
> 
> If the first iteration is interrupted with Ctrl+C, how will it effect the second iteration?
> 
> For example, current block number = 10000. For the first pruning, I shut down the node when it finished pruning for 0~1000. The next time I start the node, is it true that the pruning will panic when pruning block 0?

It takes about a week for the first iteration.

The iteration does not start from the first block every time, but starts from the last deleted block.

PR #3914, opened 2021-10-27
Implement beacon chain pruning

Rhythm-ONE:
## Issue

https://github.com/harmony-one/bounties/issues/84

## Test

I created a new validator on testnet and ran it on a very weak (1cpu core) vm. I tested that it is able to still sign transactions while continuing to delete old blocks in the background. I checked for BINGOs and was getting them as expected.
	The staking dashboard for the validator can be found at: https://staking.harmony.one/validators/testnet/one1ru8vrz7y5axhe96slwyh3tws8z8800ul5p4f44

### Test/Run Logs
Sample log output (logs every 10,000th deleted block)
https://gist.github.com/Rhythm-ONE/6c7a3cadbb82c83b05bf34bb5266c996

Sample BINGOs from transaction signing on testnet
https://gist.github.com/Rhythm-ONE/0cfdefe976c6dbea9cfa35070cf9374b

## Operational Checklist

    **NO**
    **NO**
9. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**
    **NO**

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -391,10 +391,9 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 		if currentNode.NodeConfig.ShardID != shard.BeaconChainShardID && currentNode.HarmonyConfig.General.PruneBeaconchain {
 			beaconchain := currentNode.Beaconchain()
 			beaconchain.Config().ShouldPrune = true
-			pruningStartingBlockNum := core.DetermineInitialPruningStartingBlockNumber(
-				beaconchain.ChainDb(), beaconchain.CurrentBlock().NumberU64(), core.PreserveBlockAmount)
-			beaconchain.SetInitialPruningStartBlockNum(pruningStartingBlockNum)
-			go beaconchain.InitialBlockPruning()
+			pruningStartBlockNum, pruningEndBlockNum := core.GetInitialPruningRange(beaconchain.ChainDb(), beaconchain.CurrentBlock().NumberU64())
+			beaconchain.SetInitialPruningEndBlockNum(pruningEndBlockNum)
+			go beaconchain.InitialBlockPruning(pruningStartBlockNum, pruningEndBlockNum)
 		}
 	} else if currentNode.NodeConfig.Role() == nodeconfig.ExplorerNode {
 		currentNode.RegisterExplorerServices()
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -387,10 +387,15 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 	}
 	if currentNode.NodeConfig.Role() == nodeconfig.Validator {
 		currentNode.RegisterValidatorServices()
+		// set up initial block pruning for beacon chain
 		if currentNode.NodeConfig.ShardID != shard.BeaconChainShardID && currentNode.HarmonyConfig.General.PruneBeaconchain {
-			currentNode.Beaconchain().Config().ShouldPrune = true
-			// Prune blocks before syncing starts so we can prune them one-at-a-time later
-			go currentNode.Beaconchain().PruneBlocks(core.PreserveBlockAmount)
+			beaconchain := currentNode.Beaconchain()
+			beaconchain.Config().ShouldPrune = true
+			utils.Logger().Info().Msgf("ShardID %d currentBlock %d", beaconchain.ShardID(), beaconchain.CurrentBlock().NumberU64())
+			pruningStartingBlockNum := core.DetermineInitialPruningStartingBlockNumber(
+				beaconchain.ChainDb(), beaconchain.ShardID(), beaconchain.CurrentBlock().NumberU64(), core.PreserveBlockAmount)
+			beaconchain.SetInitialPruningStartBlockNum(pruningStartingBlockNum)
+			go beaconchain.InitialBlockPruning()
 		}
 	} else if currentNode.NodeConfig.Role() == nodeconfig.ExplorerNode {
 		currentNode.RegisterExplorerServices()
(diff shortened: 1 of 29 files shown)


Rhythm-ONE:
> > > Can you do a benchmark of this PR on how many storage we are saving for mainnet beacon？
> > 
> > 
> > This is a difficult question to answer, and something I was hoping to ask you guys about. After deleting blocks in levelDB the disk space isn't freed up immediately. I read a little about it and it seems like after some period of time or a given amount of writes to the db, there's some sort of compaction that happens to free up the unused space. I haven't looked into how that's configured or whether there's anything we can safely do in code to trigger that compaction, and was kinda concerned about the performance ramifications of doing so. My current db0 folder for testnet has still been slowly growing even with these changes despite most of the blocks being gone.
> 
> You can try run a node on shard 1 for mainnet and see what is the beacon chain db size.

I chatted a bit with @JackyWYX and am going to focus my efforts on researching the compaction issue with levelDB. I've pushed some updates I made that delete other data related to the blocks being pruned, but I haven't addressed the race conditions yet. No matter how much data I delete the disk space does not go down at all. If we can't solve the compaction issue then my solution won't be viable anyway. Here are some issues I found related to the problem I'm having with disk space:
* https://github.com/google/leveldb/issues/593
* https://github.com/google/leveldb/issues/783
* https://github.com/ideawu/ssdb/issues/1332

Per Jacky's suggestion, I'm going to write my own code to research and test the levelDB compaction issue since testing the changes as part of the node service is time consuming and difficult. I'll report back whatever I find.

Rhythm-ONE:
> > > Can you do a benchmark of this PR on how many storage we are saving for mainnet beacon？
> > 
> > 
> > This is a difficult question to answer, and something I was hoping to ask you guys about. After deleting blocks in levelDB the disk space isn't freed up immediately. I read a little about it and it seems like after some period of time or a given amount of writes to the db, there's some sort of compaction that happens to free up the unused space. I haven't looked into how that's configured or whether there's anything we can safely do in code to trigger that compaction, and was kinda concerned about the performance ramifications of doing so. My current db0 folder for testnet has still been slowly growing even with these changes despite most of the blocks being gone.
> 
> You can try run a node on shard 1 for mainnet and see what is the beacon chain db size.

I added some commits Friday to do manual compaction and it got the db down from 46gigs to 34.2. Manual compaction seems like the way to go, but I think we need to delete more data.

cmd/harmony/main.go
```diff
@@ -387,6 +387,16 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 	}
 	if currentNode.NodeConfig.Role() == nodeconfig.Validator {
 		currentNode.RegisterValidatorServices()
+		// set up initial block pruning for beacon chain
+		if currentNode.NodeConfig.ShardID != shard.BeaconChainShardID && currentNode.HarmonyConfig.General.PruneBeaconchain {
+			beaconchain := currentNode.Beaconchain()
+			beaconchain.Config().ShouldPrune = true
+			utils.Logger().Info().Msgf("ShardID %d currentBlock %d", beaconchain.ShardID(), beaconchain.CurrentBlock().NumberU64())
+			pruningStartingBlockNum := core.DetermineInitialPruningStartingBlockNumber(
+				beaconchain.ChainDb(), beaconchain.ShardID(), beaconchain.CurrentBlock().NumberU64(), core.PreserveBlockAmount)
```
rlan35, line 396: beaconchain.ShardID() should always be 0. So seems redundant here.

core/blockchain_pruner.go
```diff
@@ -0,0 +1,127 @@
...
+			numDeletedBlocksThisBatch = 0
+		}
+	}
+
+	utils.Logger().Info().Msgf("Finished pruning blocks\n")
+}
+
+// PruneBlock deletes the block with a given block number. Returns bool for whether a block was deleted.
+func (bc *BlockChain) PruneBlock(blockNumber uint64) bool {
+	if blockNumber == 0 { // do not prune genesis block
```
rlan35, line 54: you can add a mutex for PruneBlock to avoid race conditions.

rlan35:
Please run this code on a validator node on shard 1 and see how much space the beacon chain will take eventually after pruning.

Rhythm-ONE:
> Please run this code on a validator node on shard 1 and see how much space the beacon chain will take eventually after pruning.

I'll set one up on mainnet and get back to you. Also updating the code based on your suggestions above. I should have that update pushed shortly.

core/blockchain_pruner.go
JackyWYX: PruneBlock is not a easy fix. Adding a lock will not solve everything.  One approach I recommend is to have 
1. one bool flag for each block
2. a block number before which all blocks are checked to be pruned.

Rhythm-ONE:
> > Please run this code on a validator node on shard 1 and see how much space the beacon chain will take eventually after pruning.
> 
> I'll set one up on mainnet and get back to you. Also updating the code based on your suggestions above. I should have that update pushed shortly.

I haven't made too much headway on this in the past week. I did some testing and discovered some performance issues, but I need to continue digging into that to figure out what can be done about it. Given that it's my first time working with golang everything is taking a bit longer than it would otherwise. I've learned a little about pprof and a couple other diagnostic tools. I'm gonna be out of town for a bit soon so unless I figure things out this week I probably won't have any updates until later this month or early next month.

Rhythm-ONE:
I haven't been able to pick this back up as I've been very busy since I got back from my trip. Testing was very time consuming and fairly expensive since I needed to rent a server in order to be able to quickly rclone and test changes, so I doubt I'll be able to justify working on this given my other commitments. I'll close this out and pray for lutty's solution to work out for us.

PR #3916, opened 2021-10-29
Deep Copy Validator Wrapper in StateDB

Tw33t3r:
## Issue

Copy function in statedb.go does not deeply copy stateValidators.

This push implements and tests correctness of these deep copies.

## Test

Tests were built matching the structure for existing testCopy tests:

A generic []validatorWrapper was built and stored in state.
State was copied using statedb.Copy() into variable copy.
The copy was copied again into ccopy.
Each *big.Int field in each value of []validatorWrapper in original, copy, and ccopy were modified and tested for expected values.
### Unit Test Coverage

Before:

```
root@9dbbf7e849b7:~/go/src/github.com/harmony-one/harmony/core/state# go test -cover
PASS
coverage: 68.3% of statements
ok      github.com/harmony-one/harmony/core/state       9.996s
```

After:

```
root@9dbbf7e849b7:~/go/src/github.com/harmony-one/harmony/core/state# go test -cover
PASS
coverage: 70.2% of statements
ok      github.com/harmony-one/harmony/core/state       9.929s
```

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -37,6 +37,7 @@ import (
 	"github.com/harmony-one/harmony/staking"
 	"github.com/harmony-one/harmony/staking/effective"
 	stk "github.com/harmony-one/harmony/staking/types"
+	staketest "github.com/harmony-one/harmony/staking/types/test"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


core/state/statedb.go
```diff
@@ -633,6 +633,15 @@ func (db *DB) Copy() *DB {
 		}
 		state.stateObjectsDirty[addr] = struct{}{}
 	}
+	for addr := range db.stateValidators {
```
JackyWYX, line 636: You may take a look at the Deep copy method already implemented in the test code: `staking/types/test/copy.go:12`

Tw33t3r: Ah, I must have missed this when I was initially looking at the code. Is there a reason not to import and use the function you mention? All libraries staking/types/test/ uses are already being imported in statedb.go

rlan35: You can definitely reuse the existing code. Just double check the code and make sure there is enough unit tests to cover the new code change.

rlan35 (approved):

PR #3919, opened 2021-10-29
[RPC] General solution of prometheus metrics for all RPC methods

JackyWYX:
## Issue

Previously we do not have the general solution for all RPC metrics. This PR add the general metric to all RPC methods.

The counters per method is now available at `hmy_rpc2_request_count` via `localhost:9900/metrics`

## Test

Tested locally. Example output for metrics result

```
curl localhost:9900
```

[Example output](https://gist.github.com/JackyWYX/874e571400d2362816e28afb5653dc09#file-metrics-txt)

## TODO:

1. Create the prometheus metric page at grafana using new metrics.
2. Retire the current prometheus RPC metric.

### DIFF
--- a/api/service/blockproposal/service.go
+++ b/api/service/blockproposal/service.go
@@ -3,7 +3,6 @@ package blockproposal
 import (
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
 	"github.com/harmony-one/harmony/consensus"
-	"github.com/harmony-one/harmony/eth/rpc"
 	"github.com/harmony-one/harmony/internal/utils"
 )
 
@@ -41,10 +40,5 @@ func (s *Service) Stop() error {
 	s.stopChan <- struct{}{}
 	<-s.stoppedChan
 	utils.Logger().Info().Msg("Role conversion stopped.")
-	return nil
-}
-
-// APIs for the services.
-func (s *Service) APIs() []rpc.API {
 	return nil
 }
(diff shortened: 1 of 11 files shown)


AlexiaChen (approved):
LGTM.

rlan35 (approved):

sophoah:
@JackyWYX the grafana page will be updated not too long before the the updates are pushed to our node

sophoah:
@JackyWYX @AlexiaChen can we align the metrics name returned by the RPCs, seems we have now 2 different sets of metric returned : https://github.com/harmony-one/harmony/blob/main/eth/rpc/metrics.go and https://github.com/harmony-one/harmony/blob/main/rpc/metrics.go 
Keeping rpc (for hmy rpc and others) and rpc2 (eth related rpc call) is fine but having the metric name returned the same is important so there are less confusion when implementing the dashboard. It's also easier to follow
See the metric defined which are different:
- original rpc metric has rpcRateLimitCounterVec, rpcQueryInfoCounterVec, rpcRequestDurationVec, rpcRequestDurationGaugeVec vs
- new eth rpc2 metric has requestCounterVec, requestErroredCounterVec, requestDurationHistVec 

cc @LeoHChen @rlan35

sophoah:
Also I am testing on a testnet node version :
```
./harmony -V
Harmony (C) 2020. harmony, version v7205-v4.3.0-31-g4433832f (jenkins@ 2021-11-14T13:56:20+0000)
```

and it doesn't seems to work 
```
curl --location --request POST 'localhost:9500' --header 'Content-Type: application/json' --data-raw '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "eth_chainId",
    "params": []
}' && curl -s http://localhost:9900/metrics | grep rpc2
{"jsonrpc":"2.0","id":1,"result":"0x6357d2e0"}
```

PR #3939, opened 2021-11-20
Resolve harmony-one/bounties#90: Add revert mechanism for UpdateValidatorWrapper

MaxMustermann2:
1. Merge ValidatorWrapper and ValidatorWrapperCopy to let callers ask for either a copy, or a pointer to the cached object. Additionally, give callers the option to not deep copy delegations (which is a heavy process). Copies need to be explicitly committed (and thus can be reverted), while the pointers are (auto-)committed when Finalise is called.
1. Add a UpdateValidatorWrapperWithRevert function, which is used by staking txs `Delegate`, `Undelegate`, and `CollectRewards`. Other 2 types of staking txs and `db.Finalize` continue to use UpdateValidateWrapper without revert, again, to save memory
1. Add unit tests which check
   1. Revert goes through
   1. Wrapper is as expected after revert
   1. State is as expected after revert

## Issue
harmony-one/bounties#90

### Unit Test Coverage

Before:

```
core/state: 70.0%
```

After:

```
core/state: 71.9%
```

### Test/Run Logs
```golang
$ go test -run TestValidatorRevert
=== Testing validator wrapper revert ===
=== Creating new validator and wrapper ===
=== Writing validator and wrapper to state ===
=== Validator successfully written to state ===
=== Reverting validator to None ===
=== Revert successful, according to ValidatorWrapper ===
=== Revert successful, according to checkEqual ===
=== Adding delegation to validator in stateDB ===
=== Writing validator with delegation to state ===
=== Validator successfully written to state ===
=== Reverting delegation from validator ===
=== Revert successful, according to ValidatorWrapper ===
=== Revert successful, according to checkEqual ===
PASS
ok  	github.com/harmony-one/harmony/core/state	0.024s
```

## Operational Checklist

No.

No.

No. See [comment](https://github.com/harmony-one/harmony/pull/3939#issuecomment-985894599).

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -23,6 +23,8 @@ import (
 	"sort"
 	"time"
 
+	lru "github.com/hashicorp/golang-lru"
+
 	"github.com/ethereum/go-ethereum/metrics"
 
 	"github.com/ethereum/go-ethereum/common"
@@ -52,6 +54,9 @@ var (
 
 	// emptyCode is the known hash of the empty EVM bytecode.
 	emptyCode = crypto.Keccak256Hash(nil)
+
+	// number of validators stored in the LRU cache
+	validatorWrapperCacheLimit = 1000
 )
 
 type proofList [][]byte
@@ -78,7 +83,7 @@ type DB struct {
 	stateObjects        map[common.Address]*Object
 	stateObjectsPending map[common.Address]struct{} // State objects finalized but not yet written to the trie
 	stateObjectsDirty   map[common.Address]struct{}
-	stateValidators     map[common.Address]*stk.ValidatorWrapper
+	stateValidators     *lru.Cache
 
 	// DB error.
 	// State objects are used by the consensus core and VM which are
@@ -120,13 +125,14 @@ func New(root common.Hash, db Database) (*DB, error) {
 	if err != nil {
 		return nil, err
 	}
+	stateValidators, _ := lru.New(validatorWrapperCacheLimit)
 	return &DB{
 		db:                  db,
 		trie:                tr,
 		stateObjects:        make(map[common.Address]*Object),
 		stateObjectsPending: make(map[common.Address]struct{}),
 		stateObjectsDirty:   make(map[common.Address]struct{}),
-		stateValidators:     make(map[common.Address]*stk.ValidatorWrapper),
+		stateValidators:     stateValidators,
 		logs:                make(map[common.Hash][]*types.Log),
 		preimages:           make(map[common.Hash][]byte),
 		journal:             newJournal(),
@@ -155,7 +161,7 @@ func (db *DB) Reset(root common.Hash) error {
 	db.stateObjects = make(map[common.Address]*Object)
 	db.stateObjectsPending = make(map[common.Address]struct{})
 	db.stateObjectsDirty = make(map[common.Address]struct{})
-	db.stateValidators = make(map[common.Address]*stk.ValidatorWrapper)
+	db.stateValidators.Purge()
 	db.thash = common.Hash{}
 	db.bhash = common.Hash{}
 	db.txIndex = 0
@@ -590,13 +596,14 @@ func (db *DB) ForEachStorage(addr common.Address, cb func(key, value common.Hash
 // Snapshots of the copied state cannot be applied to the copy.
 func (db *DB) Copy() *DB {
 	// Copy all the basic fields, initialize the memory ones
+	stateValidators, _ := lru.New(validatorWrapperCacheLimit)
 	state := &DB{
 		db:                  db.db,
 		trie:                db.db.CopyTrie(db.trie),
 		stateObjects:        make(map[common.Address]*Object, len(db.journal.dirties)),
 		stateObjectsPending: make(map[common.Address]struct{}, len(db.stateObjectsPending)),
 		stateObjectsDirty:   make(map[common.Address]struct{}, len(db.journal.dirties)),
-		stateValidators:     make(map[common.Address]*stk.ValidatorWrapper),
+		stateValidators:     stateValidators,
 		refund:              db.refund,
 		logs:                make(map[common.Hash][]*types.Log, len(db.logs)),
 		logSize:             db.logSize,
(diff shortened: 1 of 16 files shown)


JackyWYX:
The problem is that, if you simply disable the cache, RLP decoding validatorWrapper from code field will definitely take a lot of CPU and memory resource. That's why a benchmark comparison is expected. The revert shall revert the cache as well.

MaxMustermann2:
> The problem is that, if you simply disable the cache, RLP decoding validatorWrapper from code field will definitely take a lot of CPU and memory resource. That's why a benchmark comparison is expected. The revert shall revert the cache as well.

Your assessment is correct; the memory usage had more than doubled. I will modify the code to revert the cache, and get back to you.

MaxMustermann2:
Using the shell script in [#3773](https://github.com/harmony-one/harmony/pull/3773#issuecomment-884534843), I obtained the CPU and memory usage for this PR compared to the existing Harmony code at d5a89692a6964f497442401248184a449b6b5beb (the base for my branch) for ~50 minutes. The charts are plotted below:

### Memory Usage
Memory

### CPU Usage
CPU

JackyWYX (approved):
@LeoHChen @rlan35 The RP looks good to me. Please review the PR code in detail. Thanks

JackyWYX:
@LeoHChen @rlan35 Please check out this PR to unblock the staking precompiles issue.

core/state/statedb.go
```diff
@@ -78,7 +83,7 @@ type DB struct {
 	stateObjects        map[common.Address]*Object
 	stateObjectsPending map[common.Address]struct{} // State objects finalized but not yet written to the trie
 	stateObjectsDirty   map[common.Address]struct{}
-	stateValidators     map[common.Address]*stk.ValidatorWrapper
+	stateValidators     *lru.Cache
```
rlan35, line 86: This stateValidators is like a staged representation of the validatorWrappers in memory for easy modification without having to serialize/deserialize from the account.code field everything it's modified. This map keeps track of all the modifications which will needs to be committed to stateDB eventually. Changing it to cache with a limit could potentially lose modificaiton data for the validatorWrappers.

rlan35:
> Using the shell script in [#3773](https://github.com/harmony-one/harmony/pull/3773#issuecomment-884534843), I obtained the CPU and memory usage for this PR compared to the existing Harmony code at [d5a8969](https://github.com/harmony-one/harmony/commit/d5a89692a6964f497442401248184a449b6b5beb) (the base for my branch) for ~50 minutes. The charts are plotted below:
> 
> ### Memory Usage
> Memory
> 
> ### CPU Usage
> CPU

NIce!

core/state/statedb.go
MaxMustermann2: Good point. I propose setting `validatorWrapperCacheLimit` to 4,000 (currently the PR has it set to 1,000) to work around this issue. The rationale for doing so is that the block gas limit is 80,000,000, and the minimum cost of a staking transaction can be ~20,000 (conservatively; it is often higher). This means that at most 4,000 validator wrapper modifications can occur in a block. Alternatively, I am happy to change it back to a dictionary format. Let me know what you think.

JackyWYX: Could you please change stateValidators to a data structure with something like

```
type validatorCache struct {
    dirty map[common.Address]*ValidatorWrapper
    cache *lru.Cache
}
```

This will make it more readable and will not add unnecessary number of caches to the data structure.

MaxMustermann2:
I attach the graphs for the updated PR below. Since I was using a slightly different system configuration, I re-ran the graph for the existing code base too.
## Memory
Memory
## CPU
CPU

core/state/statedb.go
rlan35: Based on the new profiling, seems there isn't much benefit in using LRU. Blockchain system requires strong deterministic guarantee, please revert to the old way of using maps. Since each block state will be cleared after the block is processed, there shouldn't be memory leak issue on the old way.

rlan35: I saw you added the dirty/cache struct. Since it's not improving any memory performance. Let's not complicate the existing code (may introduce new bugs and it's risky without extensive testing). Sorry for letting you change the code back and forth. (And also it's too complicated to have three boolean in the ValidatorWrapper() method.)

MaxMustermann2:
> Based on the new profiling, seems there isn't much benefit in using LRU. Blockchain system requires strong deterministic guarantee, please revert to the old way of using maps. Since each block state will be cleared after the block is processed, there shouldn't be memory leak issue on the old way.

Please see below the profiling.

## Memory
Memory

## CPU
CPU

core/state/statedb_test.go
```diff
@@ -984,3 +988,72 @@ func makeBLSPubSigPair() blsPubSigPair {
 
 	return blsPubSigPair{shardPub, shardSig}
 }
+
+func TestValidatorRevert(t *testing.T) {
```
rlan35, line 992: Can you add more test cases? To cover more situations like modify,modify,revert; modify,modify,revert,revert. Or updating other fields rather than just the delegations. The test coverage right now is not high.

rlan35:
Please also run a mainnet node with this new code; you can use the rcloned blockchain and let the node running to make sure it can synchronize all the new blocks without problems.

MaxMustermann2:
> Please also run a mainnet node with this new code; you can use the rcloned blockchain and let the node running to make sure it can synchronize all the new blocks without problems.

The results with the memory and CPU usage are from a mainnet node, as specified in the bounty requirements. Do you need anything else from these runs?

core/state/statedb_test.go
MaxMustermann2: I have added more tests, please see `TestValidatorMultipleReverts` in particular. The coverage for new code in `statedb.go` is fairly high.

rlan35:
> 

Ok, that's good. Are the nodes being able to sync to latest block and stay in sync all the time?

MaxMustermann2:
> Ok, that's good. Are the nodes being able to sync to latest block and stay in sync all the time?

Yes, although I had to merge into my build the main branch and #3976 to get the sync to catch up from the rclone base. The "catching up" lasted ~5 hours to cover a difference of ~18,250 blocks between the rcloned database and the mainnet. The node stayed in sync (according to the block number as well as `hmyv2_inSync`) for the next ~3 hours, the duration of my testing. 

For the record, I used a [storage optimized Digital Ocean droplet](https://www.digitalocean.com/blog/introducing-storage-optimized-droplets-with-nvme-ssds/) (_not_ dedicated) with 8 cores, 64 GB RAM and 1.17 TB SSD. This decision was made because Harmony [requirements](https://docs.harmony.one/home/network/validators/server-setup/requirements) recommend using an 8-core server if it's shared, and I needed at least 750 GB for the rclone.

core/state/statedb.go
```diff
@@ -853,17 +865,42 @@ func (db *DB) UpdateValidatorWrapper(
...
+// a given validator (including delegation info)
+func (db *DB) UpdateValidatorWrapperWithRevert(
+	addr common.Address, val *stk.ValidatorWrapper,
+) error {
+	if err := val.SanityCheck(); err != nil {
+		return err
+	}
+	// a copy of the existing store can be used for revert
+	// since we are replacing the existing with the new anyway
+	prev, err := db.ValidatorWrapper(addr, true, false)
```
rlan35, line 889: Should this be "db.ValidatorWrapper(addr, false, true)"? Since you want a copy to be stored as the change journal?

MaxMustermann2: This is because the caller is sending a copy, which is being added to the `db`, while the original is in the `journal`.

rlan35: I see. make sense, the original is replaced with the new one, so it's safe to directly use it without copying. thanks

rlan35 (approved):

PR #3960, opened 2021-12-21
bump min gas price to 30 gwei

rlan35:
Bump min gas price to 30 gwei to deter spamming.

Currently the min gas price is 1 gwei and suggested gas price is 10 gwei. The spamming txns in mainnet blocks are using gas price randomly but mostly close to 1 gwei. 

Assuming on average, the attackers gas price is 2 gwei. With a mainnet block with 100 txns (https://explorer.harmony.one/block/0x41ce7ec54ee61607f959cd891bceead274cb4efc1136cec6cb7cf6c217f43670), 16852987 gas is used, which is 16852987 * 0.000000002 ONE ~ = 0.034 ONE (~ $0.0067). So the cost to spam one block is $0.0067 and the cost to spam for a whole day is 24 * 1800 * 0.0067 = $289.44. 

By changing the min gas price to 30 gwei, the cost of spamming our network for 1 day will be $289.44 * (30/2) = $4341.6.

Let's assume normal users are using 10 gwei as gas price since it's suggested by the network, a normal txn interacting with Dex (e.g. https://explorer.harmony.one/tx/0x8001d74c0b6269f69e2aa2b577ed748fb961dfe7a0390f76dca30c69045d6364) would pay 0.0019 ONE ($0.00095). With the update to min gas price of 30 gwei, the gas fee will be $0.00095 * (30/10) = **$0.00285**. It's still very cheap (still 100x cheaper than ethereum) to use Harmony to interact with Dapps frequently as a normal user.

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 1e9, // 1 Gwei/Nano
+	PriceLimit: 1e11, // 100 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 5e10, // 100 Gwei/Nano
+	PriceLimit: 5e10, // 50 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 5e10, // 50 Gwei/Nano
+	PriceLimit: 3e10, // 30 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 1e11, // 100 Gwei/Nano
+	PriceLimit: 5e10, // 100 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool_test.go
+++ b/core/tx_pool_test.go
@@ -52,7 +52,7 @@ var (
 	testBLSPubKey    = "30b2c38b1316da91e068ac3bd8751c0901ef6c02a1d58bc712104918302c6ed03d5894671d0c816dad2b4d303320f202"
 	testBLSPrvKey    = "c6d7603520311f7a4e6aac0b26701fc433b75b38df504cd416ef2b900cd66205"
 
-	gasPrice       = big.NewInt(1e9)
+	gasPrice       = big.NewInt(3e10)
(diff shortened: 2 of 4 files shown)


LeoHChen (approved):

NOCTLJRNE:
well I don't think I like this idea as a solution to help with the shard0 db size, how do you differentiate spamming transactions with normal transactions ? by increasing the minimum gwei nomal users are the ones who will get hurt not the bad actors, for instance DeFi Kingdoms players. For example this is a transaction from completing  a DFK quest
https://hmny-explorer.netlify.app/#/tx/0xdb3c81982d643433ed1b26d56a1f02deb43915193232d97a4ff9b42b4330b9c4
As you can see I paid 0.006493724 ONE for it, at 1 gwei price, if gwei price was to be at 50, then I would pay 0.3 ONE per quest .
In daily total I would pay no less than 1 ONE if min gwei is set to 50 .

friskyfoxdk:
I'm concerned about this as well. We currently use the suggested gas price of 10 for our transactions. It seems like raising this to a minimum of 10 would be a good incremental change that will still help deter spam.

countvidal:
This is bad for DFK, we need cheap transactions.

Johan22:
As a validator (ONE4All) and a DefiKingdoms player, I can see both sides here.

But I will be very careful raising the fees too high, because it will affect the experience of the players of DefiKingdoms, like NOCTLJRNE explained very well already. At 50 gwei minimum, it will cost some one everyday to do the quests.

I'll be also more in favor of 10 gwei minimum.

image

rlan35:
> I'm concerned about this as well. We currently use the suggested gas price of 10 for our transactions. It seems like raising this to a minimum of 10 would be a good incremental change that will still help deter spam.

Currently, the 10 gwei price is already very cheap ($0.00095). Increasing it to 50gwei don't significantly change the nature of the txn cost from cheap to high, but it will significantly increase the network stability by removing spams. 

But I am hearing the feedback from you, let's find a middle ground and use 30gwei as a incremental step to deter spams. We need to upgrade the network with this asap since it's affecting the user experience now.

NOCTLJRNE:
I don't know why such an important change like this isn't decided by a governance proposal instead, & I'm not sure increasing minimum gwei that much will improve users experience, changing the suggested gas price is sufficient, users who don't mind paying high gas price will just click through, those who don't want to pay high gas price will readjust the price to the point where it's affordable for them w/o being affected by the spamming transactions.  
About users experience issue, I think most of it come from the problem of Harmony default RPC nodes keeps failing/out of sync, and as of now the issue is still not fixed.

rlan35:
closing this PR since the commit is a bit messed up. Opening a new one here: https://github.com/harmony-one/harmony/pull/3960

blockchainlobster:
> Currently, the 10 gwei price is already very cheap ($0.00095). Increasing it to 50gwei don't significantly change the nature of the txn cost from cheap to high, but it will significantly increase the network stability by removing spams.
> 
> But I am hearing the feedback from you, let's find a middle ground and use 30gwei as a incremental step to deter spams. We need to upgrade the network with this asap since it's affecting the user experience now.

I just want to add that while 10 was recommended I think a fair few number of users were manually opting for 1 gwei. Now while going from 10 recommended to 30 minimum doesn't seem like a large increase I should like to point out that that's a 3x of previously recommended gwei but only in periods of low traffic. Traffic spikes and associated bidding wars will of course drive the prices further.

antho686:
> I just want to add that while 10 was recommended I think a fair few number of users were manually opting for 1 gwei. Now while going from 10 recommended to 30 minimum doesn't seem like a large increase I should like to point out that that's a 3x of previously recommended gwei but only in periods of low traffic. Traffic spikes and associated bidding wars will of course drive the average prices further.

While I agree that upgrading min fee won't impact users much on  swaps and liquidity providing etc... but when upgrading Heroes in DFK, or sending multiple Heroes on quests, the number of write operations that happens in a single transaction is significant enough so that at every gwei bump in price makes it more obvious that min fees are up.
but I'm also certain that min cost should be higher to prevent spams. 30 gwei dosn't seem that bad, I usally pay anywhere between 25gwei (in low volume times) and 60gwei(with higher traffic) per transaction ATM.

PR #3969, opened 2022-01-03
bump min gas price to 30 gwei

rlan35:
A refresh of the original PR: https://github.com/harmony-one/harmony/pull/3960

Bump min gas price to 30 gwei to deter spamming.

Currently the min gas price is 1 gwei and suggested gas price is 10 gwei. The spamming txns in mainnet blocks are using gas price randomly but mostly close to 1 gwei.

Assuming on average, the attackers gas price is 2 gwei. With a mainnet block with 100 txns (https://explorer.harmony.one/block/0x41ce7ec54ee61607f959cd891bceead274cb4efc1136cec6cb7cf6c217f43670), 16852987 gas is used, which is 16852987 * 0.000000002 ONE ~ = 0.034 ONE (~ $0.0067). So the cost to spam one block is $0.0067 and the cost to spam for a whole day is 24 * 1800 * 0.0067 = $289.44.

By changing the min gas price to 30 gwei, the cost of spamming our network for 1 day will be $289.44 * (30/2) = $4341.6.

Let's assume normal users are using 10 gwei as gas price since it's suggested by the network, a normal txn interacting with Dex (e.g. https://explorer.harmony.one/tx/0x8001d74c0b6269f69e2aa2b577ed748fb961dfe7a0390f76dca30c69045d6364) would pay 0.0019 ONE ($0.00095). With the update to min gas price of 30 gwei, the gas fee will be $0.00095 * (30/10) = $0.00285. It's still very cheap (still 100x cheaper than ethereum) to use Harmony to interact with Dapps frequently as a normal user.

### DIFF
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 1e9, // 1 Gwei/Nano
+	PriceLimit: 1e11, // 100 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 5e10, // 100 Gwei/Nano
+	PriceLimit: 5e10, // 50 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -171,7 +171,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Rejournal: time.Hour,
 
 	PriceLimit: 3e10, // 30 Gwei or Nano
-	PriceBump:  10,
+	PriceBump:  3,
 
 	AccountSlots: 16,
 	GlobalSlots:  4096,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 3e10, // 30 Gwei/Nano
+	PriceLimit: 3e10, // 30 Gwei or Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 5e10, // 50 Gwei/Nano
+	PriceLimit: 3e10, // 30 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool.go
+++ b/core/tx_pool.go
@@ -170,7 +170,7 @@ var DefaultTxPoolConfig = TxPoolConfig{
 	Journal:   "transactions.rlp",
 	Rejournal: time.Hour,
 
-	PriceLimit: 1e11, // 100 Gwei/Nano
+	PriceLimit: 5e10, // 100 Gwei/Nano
 	PriceBump:  10,
 
 	AccountSlots: 16,
--- a/core/tx_pool_test.go
+++ b/core/tx_pool_test.go
@@ -977,7 +977,7 @@ func testTransactionQueueGlobalLimiting(t *testing.T, nolocals bool) {
 	keys := make([]*ecdsa.PrivateKey, 5)
 	for i := 0; i < len(keys); i++ {
 		keys[i], _ = crypto.GenerateKey()
-		pool.currentState.AddBalance(crypto.PubkeyToAddress(keys[i].PublicKey), big.NewInt(1000000000000000))
+		pool.currentState.AddBalance(crypto.PubkeyToAddress(keys[i].PublicKey), big.NewInt(30000000000000000))
 	}
 	local := keys[len(keys)-1]
 
--- a/core/tx_pool_test.go
+++ b/core/tx_pool_test.go
@@ -710,7 +710,7 @@ func TestTransactionDropping(t *testing.T) {
 	defer pool.Stop()
 
 	account, _ := deriveSender(transaction(0, 0, 0, key))
-	pool.currentState.AddBalance(account, big.NewInt(100000000000000))
+	pool.currentState.AddBalance(account, big.NewInt(30000000000000))
 
 	// Add some pending and some queued transactions
 	var (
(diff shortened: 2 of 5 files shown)


WarrenParks:
Very concerned about this change and what it will start to cost users of DFK specifically around questing. Questing was already starting to get expensive with default of 10 gwei which is why a lot were suggesting changing to 1 gwei. This change will make it 30x more expensive for us that were adjusting to 1. It seems like a more smaller incremental change to set minimum to 10 and maybe default to 30. This would be a much more digestible and testable change then increasing the minimum by 30x.

NOCTLJRNE:
I don't know why such an important change like this isn't decided by a governance proposal instead, & I'm not sure increasing minimum gwei that much will improve users experience, changing the suggested gas price is sufficient, users who don't mind paying high gas price will just click through, those who don't want to pay high gas price will readjust the price to the point where it's affordable for them while still being prioritized over the spamming transactions low gas price.
About users experience issue, I think most of it comes from the problem of Harmony default RPC nodes keeps failing/out of sync, and as of now the issue is still not fixed. I play DeFi Kingdoms using POKT network RPC & rarely run into issues.
I hope this is just a short term solution & the team could figure out or implement a long term solution, otherwise should Harmony grow, it will be just a matter of time before it become victim of its own success, just like how BTC & ETH fee used to be cheap as well.
Sorry for keep repeating what I already said on the old PR though.

rlan35:
> I don't know why such an important change like this isn't decided by a governance proposal instead, & I'm not sure increasing minimum gwei that much will improve users experience, changing the suggested gas price is sufficient, users who don't mind paying high gas price will just click through, those who don't want to pay high gas price will readjust the price to the point where it's affordable for them while still being prioritized over the spamming transactions low gas price. About users experience issue, I think most of it comes from the problem of Harmony default RPC nodes keeps failing/out of sync, and as of now the issue is still not fixed. I play DeFi Kingdoms using POKT network RPC & rarely run into issues. I hope this is just a short term solution & the team could figure out or implement a long term solution, otherwise should Harmony grow, it will be just a matter of time before it become victim of its own success, just like how BTC & ETH fee used to be cheap as well. Sorry for keep repeating what I already said on the old PR though.

Thanks for the feedback. Yes, the fee hike is more of a temporary solution to avoid spam txns overloading the system unnecessarily, causing burden/cost to validators, downgrading normal user's experience. The long term solution enabling cross-shard smart contract support, which we are working on now, to let more apps to use the other shards.

countvidal:
This is very concerning. Harmony blocks arent even full. Raising the gas price 30x is insane for dfk questers. This is a MAJOR decision that should be brought open in a governance discussion.

bearaujus:
Instead of changing the transaction price,
it might be a good idea to try to check the "content" of a transaction. Because usually the attacker only uses empty transactions.

rlan35:
> Instead of changing the transaction price, it might be a good idea to try to check the "content" of a transaction. Because usually the attacker only uses empty transactions.

Unfortunately, it is a permissionless network and we can't filter out txns based on it's content, because it's not our job to judge on what's a good or spam txns. But the fee model is the universal solution to avoid the network getting too crowded all the time.

rlan35:
> This is very concerning. Harmony blocks arent even full. Raising the gas price 30x is insane for dfk questers. This is a MAJOR decision that should be brought open in a governance discussion.

What's the gas cost for DFK questers? Normal dex txns shouldn't cost much. If DFK quest is costing too much, there may be room to reduce the cost on making the smart contract more efficient.

Tailchakra:
> > This is very concerning. Harmony blocks arent even full. Raising the gas price 30x is insane for dfk questers. This is a MAJOR decision that should be brought open in a governance discussion.
> 
> What's the gas cost for DFK questers? Normal dex txns shouldn't cost much. If DFK quest is costing too much, there may be room to reduce the cost on making the smart contract more efficient.

Except DFK Quests have nothing to do with DEX txn. Albeit the current DFK Quest contracts aren't big, limiting to what is available now removes the entire road to innovation. Comparing DEX txn to semi-VRF calculations makes no sense. Apart from the fact that raising gas price so much should be a votable feat first, it also destroys the micro-txn feat Harmony has at the moment. Instantly going to a 20c tx fee for complex calls is game breaking for quite a few upcoming projects.

DylanCkawalec:
> > > This is very concerning. Harmony blocks arent even full. Raising the gas price 30x is insane for dfk questers. This is a MAJOR decision that should be brought open in a governance discussion.
> > 
> > 
> > What's the gas cost for DFK questers? Normal dex txns shouldn't cost much. If DFK quest is costing too much, there may be room to reduce the cost on making the smart contract more efficient.
> 
> Except DFK Quests have nothing to do with DEX txn. Albeit the current DFK Quest contracts aren't big, limiting to what is available now removes the entire road to innovation. Comparing DEX txn to semi-VRF calculations makes no sense. Apart from the fact that raising gas price so much should be a votable feat first, it also destroys the micro-txn feat Harmony has at the moment. Instantly going to a 20c tx fee for complex calls is game breaking for quite a few upcoming projects.

Where is the attack happening?  If the attack is targeting shard 0, and the foundation is concerned with validators dropping then I think this was why resharding was requested. 

I believe this is the best thing the foundation can do for the time being before the resharding occurs. If attack are happening to shards, then perhaps limiting the fee to shards instead of n shards is perfered? 

I dont know how viable that is as a solution.. if fees are more expensive on some shards vs others.

michaelotis:
> 

Maximum would be 2.4 cents, 80 mil (max gasLimit) * 0.00000003. Nobody on the network has such complex contracts that are using 80 mil gas. 

This seems like a reasonable stop gap until resharding is implemented imo. @rlan35 what would be the plan if the spamming didn't stop?

Tailchakra:
> > > > This is very concerning. Harmony blocks arent even full. Raising the gas price 30x is insane for dfk questers. This is a MAJOR decision that should be brought open in a governance discussion.
> > > 
> > > 
> > > What's the gas cost for DFK questers? Normal dex txns shouldn't cost much. If DFK quest is costing too much, there may be room to reduce the cost on making the smart contract more efficient.
> > 
> > 
> > Except DFK Quests have nothing to do with DEX txn. Albeit the current DFK Quest contracts aren't big, limiting to what is available now removes the entire road to innovation. Comparing DEX txn to semi-VRF calculations makes no sense. Apart from the fact that raising gas price so much should be a votable feat first, it also destroys the micro-txn feat Harmony has at the moment. Instantly going to a 20c tx fee for complex calls is game breaking for quite a few upcoming projects.
> 
> Maximum would be 2.4 cents, 80 mil (max gasLimit) * 0.00000003. Nobody on the network has such complex contracts that are using 80 mil gas.
> 
> This seems like a reasonable stop gap until resharding is implemented imo. @rlan35 what would be the plan if the spamming didn't stop?

80m gas limit is 2.4 ONE, not 2.4 cents. Which at this rate could be 2.4 USD in one or two months. The DFK quests I mentioned can already hit 600k gasLimit default in a non-clogged state at times (albeit this is not the eventual usage of the tx). Just saying, maybe it's better to either a; bring it to a governance vote (Harmony is supposed to be pro-voting anyway) or b; lower the size and let the chain organically clog its way up so it also calms its way down when the cross shard transactions have been finished.

True, $0,00448 per tx is close to nothing on a 600k gasLimit, but just a couple of months ago we could do thousands and thousands of transactions with a single ONE, this has already gone down to a hundred. What happens if it really starts to blow up in the next few months? This chain is full of gaming projects and when the decentralized games actually start to hit they will be transaction-heavy.

anykeyh:
I concur with the previous comments, arguing that raising the gas price is not a great solution. Shouldn't the network be able to run at 100% load?

Usually, if bots are spamming transactions and filling the blocks, it should drive the transaction cost higher by supply/demand effect, by itself. 
Here the problem is that a shard (or the RPC applying effects) is not capable of handling 2000tx / block; don't you think those problems will arise again if the shard 0 is used at full capacity, this time by genuine/legit application?

In this case, isn't a fix instead to limit the number of transactions and/or the total gas allowed per block? And relay solely on sharding in the future to balance the load on the network?

I really believe the gas price rise is a bad response to the problem.

sophoah (approved):
LGTM

mindstyle85:
guys onewallet is hardcoded to 1 gwei so no delegator, staker and validator will be able to use it for staking transactions, you urgently need to update the onewallet app too!!!

PR #3971, opened 2022-01-05
add rate limit on PendingTransactions, Call, and GetDelegationsByValidator rpc

rlan35:
With more digging, there are more costly rpcs that's dragging nodes down. One of them is PendingTransactions:
image

this node got out of memory at the same time:
image

during that time, the pending txns in the node is "totalPending:3820". That pendingTransaction rpc is returning all of 3820 txns in each call...
Similarly GetDelegationsByValidator is also called extensively, and it could be costly if the validator has thousands of delegators.
image


sophoah (approved):
LGTM

LeoHChen:
I think it's better to do pagination on pendingtransaction and getdelegationsbyvalidator RPC calls.

PR #3982, opened 2022-01-11
[misc] add a general purpose rate limiter

JackyWYX:
## Issue

Add a general purpose rate limiter to be used by RPC and SYNC module. Here are the features of the module:

1. Provide limit by ID (IP).
2. Self maintained loops to remove inactive IPs to avoid memory leak.

Note: This was part of the PR https://github.com/harmony-one/harmony/pull/3938/files. Handle sync rate limit later.

## Test

Implemented unit test case.

### DIFF
--- a/internal/rate/limiter.go
+++ b/internal/rate/limiter.go
@@ -1,0 +1,114 @@
+package rate
+
+import (
+	"container/list"
+	"context"
+	"sync"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+type (
+	limiterPerID struct {
+		evictList *list.List
+		items     map[string]*list.Element
+
+		lock sync.Mutex
+		t    timer
+		c    configInt
+	}
+
+	entry struct {
+		id       string
+		limiter  *rate.Limiter
+		lastCall time.Time
+	}
+)
+
+// NewLimiterPerID creates a limit limiter based on ID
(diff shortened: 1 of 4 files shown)


internal/rate/limiter.go
```diff
@@ -0,0 +1,114 @@
...
+
+// NewLimiterPerID creates a limit limiter based on ID
+func NewLimiterPerID(rate rate.Limit, burst int, c *Config) *limiterPerID {
+	lpi := &limiterPerID{
+		evictList: list.New(),
+		items:     make(map[string]*list.Element),
+		t:         timerImpl{},
+		c:         toConfigInt(rate, burst, c),
+	}
+	go lpi.maintainLoop()
```
rlan35, line 37: every single IP will have a go routine to maintain it? This could lead to blowup of go routine right? if there is lots of ips. How about just use a single go routine to maintain the whole pool of ips?

JackyWYX: Nope. This is exactly the implementation you are saying. There will only be one maintain loop for handling http server rates.

rlan35:
looks good in general.

rlan35 (approved):

internal/rate/limiter.go
rlan35: I see. looks good.

PR #3984, opened 2022-01-12
[RPC] Add a general RPC rate limiter per IP

JackyWYX:
## Issue

Add a rate limiter that rate limit the incoming RPC requests per IP. This PR is rebased on top of https://github.com/harmony-one/harmony/pull/3983, which use harmony/eth/rpc as customized rpc module.

1. The rate limit will limit both websocket traffic and HTTP requests at server side.
2. The rate limit is IP wise.
3. The rate limit will not be applied to 127.0.0.1.
4. The current rate limit setup is at rate=2 req/second, and burst = 10. We may adjust these parameters after some test.

### Update 1/12

1. Added a default timeout (2s) to rate limiter. If the timeout expires, a "too many request" error will be returned. 
2. Added a prometheus metric to track the counter of rate limiter being hit.

```
# HELP hmy_rpc2_rate_limited Number of request being rate limited
# TYPE hmy_rpc2_rate_limited counter
hmy_rpc2_rate_limited 0
```

## Test

Tested locally. Intense amount of RPC calls will be dropped significantly.

### TODO

1. Test on a node behind a load balancer.
2. Add rate limit related parameters to harmony config, including rate, burst and whitelist.
2. Retire some of the existing rate limiter at RPCs to avoid degrading of use experience.

### DIFF
--- a/eth/rpc/client.go
+++ b/eth/rpc/client.go
@@ -98,6 +98,8 @@ type Client struct {
 	reqInit     chan *requestOp  // register response IDs, takes write lock
 	reqSent     chan error       // signals write completion, releases write lock
 	reqTimeout  chan *requestOp  // removes response IDs when call timeout expires
+
+	limiter *rateLimiter // limit the request if the connection serves as a server
 }
 
 type reconnectFunc func(ctx context.Context) (ServerCodec, error)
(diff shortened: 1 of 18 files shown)


rlan35 (approved):

rlan35:
looks good!

JackyWYX:
1. Added a default timeout (2s) to rate limiter. If the timeout expires, a "too many request" error will be returned. 
2. Added a prometheus metric to track the counter of rate limiter being hit.

```
# HELP hmy_rpc2_rate_limited Number of request being rate limited
# TYPE hmy_rpc2_rate_limited counter
hmy_rpc2_rate_limited 0
```

PR #3987, opened 2022-01-13
send committed message before block insertion

rlan35:
Previously, the committed message is only sent out after the leader successfully inserted the block into chain DB. If the leader has trouble inserting the block into DB, then the committed message would be delayed which could lead to validators believing the leader is malfunctioning and starting view change prematurely.

The better way to do this is to send out the committed message before doing block insertion, which totally makes sense because it's never a bad idea to send out 2/3 committed message as soon as they are collected. This PR does that so block insertion wouldn't mess up the committed message broadcasting.


LeoHChen (approved):

LeoHChen:
LGTM

PR #3989, opened 2022-01-14
cache sig verification result

rlan35:
cache the message sig verification result if it's success.

### DIFF
--- a/consensus/checks.go
+++ b/consensus/checks.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"encoding/binary"
 
+	lru "github.com/hashicorp/golang-lru"
+
 	protobuf "github.com/golang/protobuf/proto"
 	libbls "github.com/harmony-one/bls/ffi/go/bls"
 	msg_pb "github.com/harmony-one/harmony/api/proto/message"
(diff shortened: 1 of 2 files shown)


consensus/checks.go
```diff
@@ -13,38 +15,55 @@ import (
...
+func verifyMessageSig(senderKey *bls.SerializedPublicKey, message *msg_pb.Message) error {
 	signature := message.Signature
 	message.Signature = nil
 	messageBytes, err := protobuf.Marshal(message)
 	if err != nil {
 		return err
 	}
 
+	// Look up cache first
+	MessageSigKey := append(senderKey[:], messageBytes[:]...)
```
JackyWYX, line 38: Message could contain the block message that can be several megabytes. Saving the content string as key may require memory of several gigabytes, which is not resource friendly. Suggest to use the hash of the message instead of the content itself.

consensus/checks.go
```diff
@@ -13,38 +15,55 @@ import (
...
+
 	msgSig := libbls.Sign{}
 	err = msgSig.Deserialize(signature)
 	if err != nil {
 		return err
 	}
 	msgHash := hash.Keccak256(messageBytes)
-	if !msgSig.VerifyHash(signerPubKey, msgHash[:]) {
+	if !msgSig.VerifyHash(pubkey, msgHash[:]) {
 		return errors.New("failed to verify the signature")
```
JackyWYX, line 56: Shall we also cache the error result?

JackyWYX (approved):

consensus/checks.go
rlan35: no need for now.

PR #3992, opened 2022-01-14
reject old viewid and higher beacon blocks.

rlan35:


node/node.go
```diff
@@ -425,6 +426,10 @@ func (node *Node) validateNodeMessage(ctx context.Context, payload []byte) (
 					utils.Logger().Debug().Uint64("receivedNum", block.NumberU64()).
 						Uint64("currentNum", curBeaconHeight).Msg("beacon block sync message rejected")
 					return nil, 0, errors.New("beacon block height smaller than current height beyond tolerance")
+				} else if block.NumberU64()-beaconBlockHeightTolerance > curBeaconHeight {
```
JackyWYX, line 429: This is risky. If a node is lagging behind, it will reject all legit beacon block messages, potentially preventing a legit beacon block broadcast to the network.

JackyWYX (commented):
This PR looks too risky for me. If a node is lagging behind, it will block all legit block messages. This may even make our network unrecoverable if large number of nodes are lagging behind.

node/node.go
rlan35: resolved offline.

PR #4007, opened 2022-01-16
Send 2/3 commit before block insertion in final commit; do not include txns in epoch block

rlan35:
Send 2/3 commit before block insertion in final commit; do not include txns in epoch block.
And revert https://github.com/harmony-one/harmony/commit/79947fb167a514035eb907d7bda187c86d9b7e9c


LeoHChen (approved):

JackyWYX (approved):
Looks good.

PR #4025, opened 2022-01-21
[RPC] add print rate stat logs

JackyWYX:
## Issue

Add RPC rate stats in harmony zerolog.

```
{"level":"info","ip":"127.0.0.1","rate":73.34,"caller":"/Users/yanxiangwang/go/src/github.com/harmony-one/harmony/eth/rpc/handler.go:341","time":"2022-01-21T09:43:22.8244-08:00","message":"[RPC] rate stats"}
```

The stats will collect the RPC request rate (req/s) per IP averaged overtime since node start, and print the rate for all known IPs every 10s. Note this is the code only used for testing purpose, and it shall be removed after metrics collected. 

## Test

Tested locally.

### DIFF
--- a/eth/rpc/handler.go
+++ b/eth/rpc/handler.go
@@ -19,11 +19,15 @@ package rpc
 import (
 	"context"
 	"encoding/json"
+	"net"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/harmony-one/harmony/internal/utils"
 
 	"github.com/ethereum/go-ethereum/log"
 )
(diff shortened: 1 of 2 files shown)


AlexiaChen:
First, I would like to know the background of the addition of the directory /eth/rpc, which I see was added three months ago, during the time I left harmony. The source code inside this directory feels like a replacement for the http package that was previously imported from ETH. Is this direct addition of source code because it allows more flexibility in hooking sensitive information（like IP address） or more flexibility in RPC metrics? Because I searched the history, I didn't find any PRs or issues about the /eth/rpc directory source code being added, so I'm not sure about the background

PR #4026, opened 2022-01-22
Make blockNumber rpc more efficient with atomic uint64

rlan35:
Previous _blockNumber rpc is retrieving the latest block, copying the header and returning the number from the header, which is very unnecessary and expensive.

Adding a uint64 atomic value to store the latest block number and directly turning it.

### DIFF
--- a/core/blockchain.go
+++ b/core/blockchain.go
@@ -139,6 +139,7 @@ type BlockChain struct {
 	pendingSlashingCandidatesMU sync.RWMutex // pending slashing candidates
 
 	currentBlock     atomic.Value // Current head of the block chain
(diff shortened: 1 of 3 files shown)


AlexiaChen (approved):

JackyWYX (commented):
How much resources we are saving for blockNumber requests? Is there a benchmark showing the improvement?

rlan35:
> How much resources we are saving for blockNumber requests? Is there a benchmark showing the improvement?

No benchmark yet. @AlexiaChen is working on a loadtest and benchmark tool.

ONECasey:
I really like this PR. I will be resubmitting.

Issue #4028, opened 2022-01-24
Ed25519 Precompile


gupadhyaya:
This precompile is required for crosschain lightclients of chains that use ed25519 signatures for block validation. The verification of Ed25519 cryptographic signatures in smart contracts using current EVM bytecodes is super expensive. This precompile will help reduce the gas cost drastically. Note that, there exists several references of this precompile in other blockchains like ethereum, polygon, celo, etc. e.g., https://github.com/celo-org/celo-blockchain/pull/1256 

This is required for cosmos bridge.

PR #4041, opened 2022-01-29
[WIP] Add ed25519 Precompile

fzgem18:
## Issue

ed25519 Precompile Contracts Verification #4028 

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**

### DIFF
--- a/core/vm/contracts.go
+++ b/core/vm/contracts.go
@@ -112,6 +112,7 @@ var PrecompiledContractsSHA3FIPS = map[common.Address]PrecompiledContract{
 	common.BytesToAddress([]byte{9}):   &blake2F{},
 	common.BytesToAddress([]byte{255}): &vrf{},
 
+	common.BytesToAddress([]byte{252}): &ed25519Verify{},
 	common.BytesToAddress([]byte{253}): &sha3fip{},
 	common.BytesToAddress([]byte{254}): &ecrecoverPublicKey{},
 }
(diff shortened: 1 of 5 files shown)


core/vm/contracts.go
```diff
@@ -668,3 +670,37 @@ func (c *ecrecoverPublicKey) Run(input []byte) ([]byte, error) {
 
 	return pubKey, nil
 }
+
+// ed25519Verify implements a native Ed25519 signature verification.
+type ed25519Verify struct{}
+
+// RequiredGas returns the gas required to execute the pre-compiled contract.
+func (c *ed25519Verify) RequiredGas(input []byte) uint64 {
```
rlan35, line 678: Are these codes directly copied from celo? Any modifications that's worth mentioning?

internal/params/protocol_params.go
```diff
@@ -175,6 +175,9 @@ const (
 	Sha3FipsGas     uint64 = 30 // Once per SHA3-256 operation.
 	Sha3FipsWordGas uint64 = 6  // Once per word of the SHA3-256 operation's data.
 
+	Ed25519VerifyGas   uint64 = 1500             // Gas needed for and Ed25519 signature verification
```
rlan35, line 178: Is 1500 reasonable? @gupadhyaya

gupadhyaya: yes. same as celo.

core/vm/contracts.go
gupadhyaya: I think there are no major changes required. so, it should be good.

PR #4049, opened 2022-02-08
[stream] Enable downloader

JackyWYX:
## Issue

Enable default config to turn on stream protocol at localnet and testnet. Also with some bug fixes.

```
Default Testnet:
 sync.server = true
 sync.client = false
 dnsSync.server = true 
 dnsSync.client = true

Localnet
 sync.Enabled = true
 sync.Downloader = true
 dnsSync.server = true
 dnsSync.client = true
```

### Unit Test Coverage

Passed all unit tests and local tests. Run on stressnet (resource limited) for a while, and need to have it roll out in testnet.

### DIFF
--- a/hmy/downloader/downloader.go
+++ b/hmy/downloader/downloader.go
@@ -203,6 +203,7 @@ func (d *Downloader) loop() {
 					time.Sleep(5 * time.Second)
 					trigger()
 				}()
+				time.Sleep(1 * time.Second)
 				continue
 			}
 			d.logger.Info().Int("block added", addedBN).
--- a/p2p/stream/common/streammanager/cooldown.go
+++ b/p2p/stream/common/streammanager/cooldown.go
@@ -27,11 +27,13 @@ func newCoolDownCache() *coolDownCache {
 // Has check and add the peer ID to the cache
 func (cache *coolDownCache) Has(id peer.ID) bool {
 	has := cache.timeCache.Has(string(id))
-	cache.timeCache.Add(string(id))
+	if !has {
+		cache.timeCache.Add(string(id))
+	}
 	return has
 }
 
-// Reset reset the cooldown cache
+// Reset the cool down cache
 func (cache *coolDownCache) Reset() {
 	cache.timeCache.Q = list.New()
 	cache.timeCache.M = make(map[string]time.Time)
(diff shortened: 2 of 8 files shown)


sophoah:
This PR has been tested in testnet for more than 3 weeks now. Since the last commit, there has been some instances of node that got stuck which could be due to insufficient stream sync server happening from time to time, a restart of the node fixes it.

There is a really slow OOM issue happening but was already there before streamsync was enabled

Test performed and succeeded: 
enable feature and making sure consensus works as expected
stopped node for a while and making sure it can catchup
brand new node was started from scratch and could sync up
monitor resources for 2 weeks and highlight any issues

@JackyWYX and I agreed on :
1) keep this feature as experimental and merge this PR and make it available in the next release
2) test on mainnet with a few sets of nodes. External validator will be welcome to test and create any issue
3) continue with bug fix until the feature is stable with progressive rollout to more internal node

hmy/downloader/downloader.go
```diff
@@ -203,6 +203,7 @@ func (d *Downloader) loop() {
 					time.Sleep(5 * time.Second)
 					trigger()
 				}()
+				time.Sleep(1 * time.Second)
```
rlan35, line 206: why is the 1s sleep added?

p2p/stream/common/streammanager/cooldown.go
```diff
@@ -0,0 +1,40 @@
...
+	"container/list"
+	"time"
+
+	"github.com/libp2p/go-libp2p-core/peer"
+
+	"github.com/whyrusleeping/timecache"
+)
+
+const (
+	coolDownPeriod = 1 * time.Minute
```
rlan35, line 13: Can you add more explanation about the reason for adding the cooldown period and why it's 1 min?

hmy/downloader/downloader.go
JackyWYX: 1s sleep is to avoid infinite loop when having not enough peers. It is observed on testnet that the node 

1. Node start syncing process.
2. Node found it does not have enough peers and exit syncing with an error.
3. Node start another syncing because consensus / trigger 5s before.

This will cause an infinite loop that consumes CPU resource. Thus add a sleep mechanism here to avoid infinite loop.

p2p/stream/common/streammanager/cooldown.go
JackyWYX: Thanks. Added comments.

rlan35 (approved):

PR #4066, opened 2022-02-23
leveldb shard

LuttyYang:
Add new leveldb shard db mode, significantly improved read and write performance

during the testing for rosetta check data, the bottleneck of performance was found in the leveldb database.

afte we did an in-depth performance investigation on leveldb and found that there are 2 problems as follows:
1. when leveldb has reach more than 1T large amount of data, the read and write performance is greatly reduced, so the overall performance is degraded
2. when leveldb is multi-threaded concurrently reading, there is a lot of lock overhead, resulting in a decrease in competition performance too

so we upgraded the db of harmony node underlying architecture, evolved into a new model, which split the original one leveldb database into 32 to improve db performance

in the new mode, it will take up an additional 1-2GB of memory:
- about 1GB is used by memdb of leveldb
- 512MB is used by local cache

### DIFF
--- a/internal/shardchain/dbfactory.go
+++ b/internal/shardchain/dbfactory.go
@@ -13,6 +13,11 @@ import (
 	"github.com/ethereum/go-ethereum/ethdb"
 )
 
+const (
+	LDBDirPrefix      = "harmony_db"
+	LDBShardDirPrefix = "harmony_sharddb"
+)
+
 // DBFactory is a blockchain database factory.
 type DBFactory interface {
 	// NewChainDB returns a new database for the blockchain for
@@ -27,7 +32,7 @@ type LDBFactory struct {
 
 // NewChainDB returns a new LDB for the blockchain for given shard.
 func (f *LDBFactory) NewChainDB(shardID uint32) (ethdb.Database, error) {
-	dir := path.Join(f.RootDir, fmt.Sprintf("harmony_db_%d", shardID))
+	dir := path.Join(f.RootDir, fmt.Sprintf("%s_%d", LDBDirPrefix, shardID))
 	return rawdb.NewLevelDBDatabase(dir, 256, 1024, "")
 }
 
@@ -48,7 +53,7 @@ type LDBShardFactory struct {
 
 // NewChainDB returns a new memDB for the blockchain for given shard.
 func (f *LDBShardFactory) NewChainDB(shardID uint32) (ethdb.Database, error) {
-	dir := filepath.Join(f.RootDir, fmt.Sprintf("harmony_sharddb_%d", shardID))
+	dir := filepath.Join(f.RootDir, fmt.Sprintf("%s_%d", LDBShardDirPrefix, shardID))
 	shard, err := leveldb_shard.NewLeveldbShard(dir, f.DiskCount, f.ShardCount)
 	if err != nil {
 		return nil, err
--- a/internal/shardchain/dbfactory.go
+++ b/internal/shardchain/dbfactory.go
@@ -2,10 +2,11 @@ package shardchain
 
 import (
 	"fmt"
-	"github.com/harmony-one/harmony/internal/shardchain/leveldb_shard"
-	"github.com/harmony-one/harmony/internal/shardchain/local_cache"
 	"path"
 	"path/filepath"
+
+	"github.com/harmony-one/harmony/internal/shardchain/leveldb_shard"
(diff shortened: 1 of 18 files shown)


jhd2best:
here is the performance test log: 
[db-performance-test-log.xlsx](https://github.com/harmony-one/harmony/files/8124426/db-performance-test-log.xlsx)

LeoHChen:
1. looks good to me in general, regarding the performance test, looks like pebbleDB had outperformed leveldb even with 32 shards, we may need to explore the pebble db approach later then.

2. what's the db directory structure when we use 32 shard db? please provide some example.
3. will it work with existing db or the node has to sync from the beginning to use the shard db?

internal/shardchain/dbfactory.go
```diff
@@ -34,3 +38,21 @@ type MemDBFactory struct{}
...
+// LDBShardFactory is a merged Multi-LDB-backed blockchain database factory.
+type LDBShardFactory struct {
+	RootDir    string // directory in which to put shard databases in.
+	DiskCount  int
+	ShardCount int
+}
+
+// NewChainDB returns a new memDB for the blockchain for given shard.
+func (f *LDBShardFactory) NewChainDB(shardID uint32) (ethdb.Database, error) {
+	dir := filepath.Join(f.RootDir, fmt.Sprintf("harmony_sharddb_%d", shardID))
```
LeoHChen, line 51: make the sharddb prefix configurable

internal/shardchain/leveldb_shard/shard.go
```diff
@@ -0,0 +1,200 @@
...
+				if db != nil {
+					_ = db.Close()
+				}
+			}
+
+			shard = nil
+		}
+	}()
+
+	levelDBOptions := &opt.Options{
```
LeoHChen, line 46: make those numbers configurable with default values

internal/shardchain/local_cache/hack.go
```diff
@@ -0,0 +1,22 @@
+package local_cache
```
LeoHChen, line 1: why this file called "hack"?

internal/shardchain/local_cache/local_cache_database.go
```diff
@@ -0,0 +1,112 @@
...
+	ethdb.KeyValueStore
+
+	enableReadCache bool
+
+	deleteMap map[string]bool
+	readCache *cacheWrapper
+}
+
+func NewLocalCacheDatabase(remoteDB ethdb.KeyValueStore) *LocalCacheDatabase {
+	config := bigcache.DefaultConfig(10 * time.Minute)
```
LeoHChen, line 34: make them configurable so that we can tune the parameters later

LeoHChen (approved):
Looks good to me in general.  Good work.

Just a few minor comments, please make many hard-coded numbers parameters so that they can be tuned.  DB usually needs to be tuned on different scenarios.  So that to make them tuneable is an essential requirement. Thanks.

internal/shardchain/dbfactory.go
LuttyYang: My suggestion is to use a fixed value, because the two data formats are incompatible with the other, it can also prevent users from making changes by mistake, and it is more convenient to write subsequent documents

internal/shardchain/leveldb_shard/shard.go
LuttyYang: My suggestion is to use those fixed value, These parameters have been tested. Too large or too small configurable will affect memory usage and performance.

internal/shardchain/local_cache/hack.go
LuttyYang: This is because the unsafe package is used to improve performance, which may cause incompatibility in future go version upgrades (although most likely not)

internal/shardchain/local_cache/local_cache_database.go
LuttyYang: We will make the configuration later here

internal/shardchain/dbfactory.go
LeoHChen: okay

LeoHChen (approved):

AustinHDev:
Hello! I enabled this sharded database structure on a new Shard0 RPC server. I quickly realized that it was syncing an entirely fresh database.

> 3\. can not work with existing db, but we can mitigate transferred data by rsync

Has this been implemented? Is there an available database to clone? Or can I transform my local shard 0 database?

Thanks!

PR #4068, opened 2022-02-23
feat: Prune stale staking data hard fork + onwards

MaxMustermann2:
Stale staking data is defined as delegations to a validator which have
no rewards, no amount, and no undelegations. This definition, however,
excludes the (inactive) validator's self delegation since it is expected
to be at index 0, and a validator cannot be deleted at the moment. The
process works as below:

 - At the last block of every epoch including the hard fork epoch,
   obtain a list of validators. For each validator, load the
   ValidatorWrapper and iterate through Delegations to find stale
   delegations. Drop such delegations, and make a map of delegators as
   keys with validator addres(es) as the values.
 - Using the map obtained in (1), clear the delegations by delegator in
   offchain data. The map is therefore cached in the ProcessorResult for
   ease of access (between state_processor and node_newblock).
 - Since the objective of this change is to save some space following
   the hard fork, do not allow small undelegations. This means that
   undelegations must be made in such a manner that the remaining amount
   after the undelegation has been made is at least a 100 ONEs, or zero.
 - Lastly, since the removal results in different lengths between
   delegations of validator snapshots and validator wrappers, modify the
   algorithm for AddReward to be based on DelegatorAddress rather than
   the loop index. See core/state/statedb.go

The impact of this change on block processing speed is small. I ran a
node with this change at epoch 881 on mainnet using the rclone database
and saw that pruneStaleStakingData takes 18ms to complete (if each prune
is not logged individually, it takes 9ms).

The PR has been tested locally for integration testing (the undelegation
bit), and unit tests for pruning stale data, remaining undelegation, and
AddReward have been added. Goimports is included.

## Issue
Described above.

### Unit Test Coverage
```bash
for file in $(git diff --name-status main..clear-stale-staking-data | sed s/\^..//); do go test -cover ./$(dirname $file)/; done | sort -f | uniq -i
```
Before:

```
?   	github.com/harmony-one/harmony/consensus/engine	[no test files]
?   	github.com/harmony-one/harmony/hmy	[no test files]
?   	github.com/harmony-one/harmony/internal/chain	[no test files]
?   	github.com/harmony-one/harmony/internal/params	[no test files]
?   	github.com/harmony-one/harmony/test/chain/reward	[no test files]
ok  	github.com/harmony-one/harmony/core	(cached)	coverage: 31.2% of statements
ok  	github.com/harmony-one/harmony/core/state	(cached)	coverage: 71.9% of statements
ok  	github.com/harmony-one/harmony/core/vm	(cached)	coverage: 42.4% of statements
ok  	github.com/harmony-one/harmony/hmy/downloader	(cached)	coverage: 75.9% of statements
ok  	github.com/harmony-one/harmony/node/worker	(cached)	coverage: 27.1% of statements
ok  	github.com/harmony-one/harmony/staking	(cached)	coverage: 89.1% of statements
ok  	github.com/harmony-one/harmony/staking/types	(cached)	coverage: 62.9% of statements
```

After:

```
?   	github.com/harmony-one/harmony/consensus/engine	[no test files]
?   	github.com/harmony-one/harmony/hmy	[no test files]
?   	github.com/harmony-one/harmony/internal/params	[no test files]
?   	github.com/harmony-one/harmony/test/chain/reward	[no test files]
ok  	github.com/harmony-one/harmony/core	(cached)	coverage: 31.1% of statements
ok  	github.com/harmony-one/harmony/core/state	(cached)	coverage: 75.6% of statements
ok  	github.com/harmony-one/harmony/core/vm	(cached)	coverage: 42.4% of statements
ok  	github.com/harmony-one/harmony/hmy/downloader	(cached)	coverage: 76.0% of statements
ok  	github.com/harmony-one/harmony/internal/chain	(cached)	coverage: 2.8% of statements
ok  	github.com/harmony-one/harmony/node/worker	(cached)	coverage: 27.0% of statements
ok  	github.com/harmony-one/harmony/staking	(cached)	coverage: 89.1% of statements
ok  	github.com/harmony-one/harmony/staking/types	(cached)	coverage: 63.2% of statements
```

Per-line test coverage for files modified by this PR is [available here](https://gist.github.com/MaxMustermann2/9b89d7150e652888903ef79d28024011)

### Test/Run Logs
```
{'blockNum': 23232511,
 'caller': '/home/user/go/src/github.com/harmony-one/harmony/internal/chain/engine.go:364',
 'elapsed time': 18,
 'epoch': 881,
 'level': 'info',
 'message': 'pruneStaleStakingData',
 'time': '2022-02-23T08:43:54.259281751Z'}
```
[Full logs](https://gist.github.com/MaxMustermann2/617ba40dd05e53eec553fbeb38bc406f) show that stale data was removed from 472 validators (there were 676 overall).

## Operational Checklist

No.

No.

No, a bulk pruning takes up only 18ms extra at the hard fork epoch. It is reasonable to expect continuous pruning at the end of each epoch to not take more time than that.


core/blockchain.go
```diff
@@ -2908,12 +2918,14 @@ func (bc *BlockChain) addDelegationIndex(
 	// If there is an existing delegation, just return
 	validatorAddressBytes := validatorAddress.Bytes()
 	for _, delegation := range delegations {
+		// this will happen when someone delegates more to the same validator
+		// since a delegationIndex only stores the blockNum, validator address and block number
```
rlan35, line 2922: stores the index in the delegation list.

core/state/statedb.go
```diff
@@ -973,24 +982,70 @@ func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int, shareLo
...
+			if percentage.IsZero() { // stale delegation
+				continue
+			}
+			// try to find in wrapper
+			// this needs to be O(N^2) because the order is not guaranteed
+			// for example, snapshot is A / B / C / D / E where C is a stale delegation
+			// wrapper then becomes A / B / D / E
+			// if C then delegates to this validator, the wrapper becomes A / B / D / E / C
+			found := false
+			for j := 0; j < len(curValidator.Delegations) && !found; j++ {
```
rlan35, line 1027: This could be too costly if the number of delegation in the validator is high (there are some with thousands of delegations). There could be a way to optimize the complexity. Since the two array are only differing by removing some elements, the start index of the searching doesn't have to always be 0, right?

core/blockchain.go
MaxMustermann2: Will update the comment but my intention was to say that it doesn't store the `Amount`, and it is only representative of the Index in the delegation list.

core/state/statedb.go
MaxMustermann2: My original code (that I did not push out) had that assumption. However, it would not necessarily work in the case where a stale delegation is removed and then added by the delegator (to the same validator) later. Let me think of another idea though because I agree with your opinion.

core/state/statedb.go
```diff
@@ -973,24 +982,83 @@ func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int, shareLo
...
+			// for i in [0, 1] j = i - 0 (offset) works well
+			// for i = 2, offset becomes 1
+			// for i in [3, 4] use j = i - 1 (offset)
+			// other cases are
+			// (1) snapshot = A / B / C / D and wrapper = A / C / D (just remove B)
+			// (2) snapshot = A / B / C / D and wrapper = A / C / D / E (remove B and add E)
+			// (3) snapshot = A / B / C / D and wrapper = A / B / C / D / E (just add E)
+			// (4) snapshot and wrapper equal (no effort needed)
+			//		 even if a stale delegation is removed from the end and re-added this works
+			// (5) snapshot = A / B / C / D / E and wrapper = A / B / D / E / C / F (remove and re-add + add)
```
rlan35, line 1038: Great job coming up with this nice solution!

core/state/statedb.go
```diff
@@ -973,24 +982,83 @@ func (db *DB) AddReward(snapshot *stk.ValidatorWrapper, reward *big.Int, shareLo
...
+		// those added later to curValidator are new delegations which do not receive rewards immediately
+		// those removed from curValidator are stale delegations which do not receive rewards anyway
+		offset := 0
+		for i := 0; i < len(snapshot.Delegations); i++ {
+			delegationFromSnapshot := snapshot.Delegations[i]
+			percentage, ok := shareLookup[delegationFromSnapshot.DelegatorAddress]
+			if !ok {
+				return errors.Wrapf(err, "missing delegation shares for reward distribution")
+			}
+			if percentage.IsZero() { // stale delegation
```
rlan35, line 1019: So the assumption here is that all the 0 delegation are removed, right?

rlan35:
Did you implement the part that when a delegation has all the undelegations gone and has 0 balance in it, the delegation should also be removed immediately? I think this PR only have the batch removal, right? And the batch removal should only happen once, after which the on-spot removal should take the job.

MaxMustermann2:
I have set it up such that those removals take place at the end of each epoch including and following the hard fork epoch. So the first such pruning is a batch removal, and the next such pruning, which occurs at the end of the next epoch, is a part of continuous removal.
https://github.com/harmony-one/harmony/blob/494fc3c53c64e9d97d7acfdc96617ae3dd801abe/internal/chain/engine.go#L430-L439
The reason I chose this schedule is that undelegations are paid out during the last block of every epoch. Only after undelegations get paid out can such a pruning take place (otherwise the `Undelegations` structure is not `nil` in `wrapper.Delegations[ j ]`).
https://github.com/harmony-one/harmony/blob/494fc3c53c64e9d97d7acfdc96617ae3dd801abe/internal/chain/engine.go#L422-L428

core/state/statedb.go
MaxMustermann2: Yes

core/state/statedb.go
rlan35: So the expected complexity is still O(N) right? How much slower is expected?

internal/chain/engine.go
```diff
@@ -402,6 +427,78 @@ func IsCommitteeSelectionBlock(chain engine.ChainReader, header *block.Header) b
...
 
+// shouldPruneStaleStakingData checks (1) we are in the beacon chain, and
+// (2) ChainConfig returns true for IsNoNilDelegations
+func shouldPruneStaleStakingData(
+	chain engine.ChainReader,
+	header *block.Header,
+) bool {
+	return header.ShardID() == shard.BeaconChainShardID &&
+		header.IsLastBlockInEpoch() &&
+		chain.Config().IsNoNilDelegations(header.Epoch())
```
rlan35, line 438: Note the pruning only happens at the end of the IsNoNilDelegations epoch. So the new reward calculation logic only need to happen in the blocks with IsNoNilDelegations epoch + 1

internal/chain/reward.go
```diff
@@ -220,9 +220,24 @@ func AccumulateRewardsAndCountSigs(
...
 		} else {
-			return distributeRewardBeforeAggregateEpoch(bc, state, header, beaconChain, defaultReward, sigsReady)
+			return distributeRewardBeforeAggregateEpoch(bc,
+				state,
+				header,
+				beaconChain,
+				defaultReward,
+				sigsReady,
+				// pruning only occurs at the end of the epoch
+				bc.Config().IsNoNilDelegations(header.Epoch()) && header.IsLastBlockInEpoch() && header.ShardID() == shard.BeaconChainShardID,
```
rlan35, line 239: The reward happens every 64 blocks, so any reward logic that happens after the pruning will need to have the new logic running right?

node/worker/worker.go
```diff
@@ -35,20 +35,21 @@ import (
...
+	txs                 []*types.Transaction
+	stakingTxs          []*staking.StakingTransaction
+	receipts            []*types.Receipt
+	logs                []*types.Log
+	reward              reward.Reader
+	outcxs              []*types.CXReceipt       // cross shard transaction receipts (source shard)
+	incxs               []*types.CXReceiptsProof // cross shard receipts and its proof (desitinatin shard)
+	slashes             slash.Records
+	stakeMsgs           []staking.StakeMsg
+	delegationsToRemove map[common.Address][]common.Address
```
rlan35, line 52: is delegationsToRemove in worker actually used somewhere?

MaxMustermann2: Yes, it's used when caching the processor result, in line with `stakeMsgs`

internal/chain/engine.go
MaxMustermann2: Thanks, fixed.

internal/chain/reward.go
MaxMustermann2: Agreed and updated.

internal/chain/engine.go
```diff
@@ -327,16 +327,41 @@ func (e *engineImpl) Finalize(
...
 	} else if len(doubleSigners) > 0 {
-		return nil, nil, errors.New("slashes proposed in non-beacon chain or non-staking epoch")
+		return nil, nil, nil, errors.New("slashes proposed in non-beacon chain or non-staking epoch")
+	}
+
+	// must be done after slashing and paying out rewards
+	// to avoid conflicts with snapshots
+	// any nil (un)delegations are eligible to be removed
+	// bulk pruning happens at the end of the NoNilDelegationsEpoch
+	// and then continuous pruning happens at the end of every epoch thereafter
```
rlan35, line 346: Do we have a separate continuous pruning besides bulk pruning? It's the same logic that happens at the end of epochs right?

MaxMustermann2: The same logic is applied at the end of each epoch. Regular / continuous pruning can occur only after undelegations are paid out; therefore it must occur at the end of the epoch.

staking/types/delegation.go
```diff
@@ -196,3 +214,102 @@ func (d *Delegation) RemoveUnlockedUndelegations(
...
+				}
+			}
+		}
+		// insert delegator into top
+		top[delegator] = topDelegator
+	}
+	return top
+}
+
+// func PrintDelegationsToAlter(
```
rlan35, line 254: remember to remove once not used.

rlan35, line 254: remember to remove once not used.

staking/types/delegation.go
```diff
@@ -196,3 +214,102 @@ func (d *Delegation) RemoveUnlockedUndelegations(
...
+	} else {
+		if delegationIndex.Index == 0 {
+			// since this is referring to self delegation
+			// and it is never removed
+			// no address comparison is needed
+			return &wrapper.Delegations[0], 0, nil
+		}
+		var startingPoint uint64
+		if delegationIndex.Index >= uint64(len(wrapper.Delegations)) {
+			// we never remove the (inactive) validator
```
rlan35, line 293: what does it mean "we never remove the (inactive) validators? And please add more comment here to explain how the startingPoint is chosen.

staking/types/delegation.go
```diff
@@ -196,3 +214,102 @@ func (d *Delegation) RemoveUnlockedUndelegations(
...
+			// so this will never overflow
+			startingPoint = uint64(len(wrapper.Delegations)) - 1
+		} else {
+			startingPoint = delegationIndex.Index
+		}
+		// delegationIndex is as of previous block
+		// and between the two blocks only removal takes place
```
rlan35, line 300: why are only removal takes places? New delegation can be added between blocks right?

rlan35, line 300: why are only removal takes places? New delegation can be added between blocks right?

core/staking_verifier.go
```diff
@@ -519,33 +569,40 @@ func VerifyAndMigrateFromMsg(
...
+			delegationIndex,
+			chainConfig,
+			epoch,
+		); err != nil {
+			return nil, nil, err
+		}
+		if delegation == nil {
+			continue
+		}
+		if delegation.Reward.Cmp(common.Big0) > 0 {
```
rlan35, line 603: Just realized that we should also check whether the delegation reward is "0" after CollectReward() right?

rlan35, line 603: Just realized that we should also check whether the delegation reward is "0" after CollectReward() right?

core/staking_verifier.go
```diff
@@ -285,8 +305,34 @@ func VerifyAndDelegateFromMsg(
...
+					delegation.Amount.Cmp(common.Big0) == 0 &&
+					delegation.Reward.Cmp(common.Big0) == 0 &&
+					chainConfig.IsPostNoNilDelegations(epoch)
+				if shouldRemove {
+					modification := make(map[common.Address]uint64)
+					// mark for removal
+					modification[wrapper.Address] = math.MaxUint64
+					delegationsToAlter[delegation.DelegatorAddress] = modification
+					// mark the following ones with offset 1
+					for j := actualIndex + 1; j < uint64(len(wrapper.Delegations)); j++ {
```
rlan35, line 319: The continuous pruning seems is getting very complicated and also incurs extra cost by itself (such as this looping). And also, have you consider the case when more than 1 delegate/undelegate/collectreward txn happens in the same block? If the complicated is too high and the gain compared with batch pruning is not very high. We may not want to do the continuous pruning then.

rlan35, line 319: The continuous pruning seems is getting very complicated and also incurs extra cost by itself (such as this looping). And also, have you consider the case when more than 1 delegate/undelegate/collectreward txn happens in the same block? If the complicated is too high and the gain compared with batch pruning is not very high. We may not want to do the continuous pruning then.

staking/types/delegation.go
MaxMustermann2: Agreed, I will update the comment. However, the logic is still valid since deletion can take place before the `delegationIndex.Index` - which means we have to go backwards to find the `actualIndex`. Addition takes place on the other side, so it does not affect the logic.

core/staking_verifier.go
MaxMustermann2: Agreed; this is fairly trivial with the continuous pruning mechanism that has been built.

core/staking_verifier.go
MaxMustermann2: > And also, have you consider the case when more than 1 delegate/undelegate/collectreward txn happens in the same block?
Yes, that has been attended to. This is a large reason for the complication.

internal/chain/engine.go
MaxMustermann2: For a machine which inserts blocks at 200-600 ms, for an average of 120 blocks per minute:
- Bulk pruning took 41ms for the validator wrappers on chain. The number of validator wrappers thus modified were 490.
- Updating the delegation indices (off chain data) took 1061ms. The number of delegators impacted was 49000.
Total extra time at hard fork block is 1.1s. If we keep only bulk pruning, the time at each epoch block after the hard fork block is bound to be lower, and we can have code that is less complicated to maintain.
Total extra time spent per epoch is 1.1s, or lower.

On the same machine, continuous pruning results are as per below.
- Continuous pruning requires the modification of validators 1000 times per 2800 blocks (spread across multiple transactions), and thus per epoch approximately ~12,000 validators. This modification takes an additional time of roughly 960ms per epoch.
- This continuous pruning impacted 26,000 delegators across 2800 blocks. Over an epoch this is 304,000 changes. The time taken then, per epoch, is additional 6.8s.
Total extra time per epoch is thus 7.8s.

Based on the above information, it makes sense to go for solely the bulk pruning approach at the end of each epoch. I will push that out after I am done with the two precompiles.

internal/chain/engine.go
```diff
@@ -327,16 +327,41 @@ func (e *engineImpl) Finalize(
...
 		}
 	} else if len(doubleSigners) > 0 {
-		return nil, nil, errors.New("slashes proposed in non-beacon chain or non-staking epoch")
+		return nil, nil, nil, errors.New("slashes proposed in non-beacon chain or non-staking epoch")
+	}
+
+	// must be done after slashing and paying out rewards
+	// to avoid conflicts with snapshots
+	// any nil (un)delegations are eligible to be removed
+	// bulk pruning happens at the first block of the NoNilDelegationsEpoch + 1
```
rlan35, line 345: based on the current logic, it happens at the first block of epoch NoNilDelegationsEpoch

internal/chain/engine.go
```diff
@@ -402,6 +427,86 @@ func IsCommitteeSelectionBlock(chain engine.ChainReader, header *block.Header) b
...
+	header *block.Header,
+	state *state.DB,
+	delegationsToRemove map[common.Address][]common.Address,
+) error {
+	validators, err := chain.ReadValidatorList()
+	if err != nil {
+		return err
+	}
+	for _, validator := range validators {
+		wrapper, err := state.ValidatorWrapper(validator, true, false)
```
rlan35, line 469: to confirm, this retrieve the validatorWrapper reference and any modification on it will be automatically committed to state eventually right?

internal/chain/reward.go
```diff
@@ -394,20 +395,36 @@ func distributeRewardAfterAggregateEpoch(bc engine.ChainReader, state *state.DB,
 		if err != nil {
 			return network.EmptyPayout, err
 		}
-		if err := state.AddReward(snapshot.Validator, due, shares); err != nil {
+		if err := state.AddReward(
+			snapshot.Validator,
+			due,
+			shares,
+			// epoch prior to no nil delegations
```
rlan35, line 402: so we are all good here on the reward logic and you checked, right?

internal/chain/engine.go
MaxMustermann2: Yes.

ONECasey:
@MaxMustermann2 , can you please update this PR or close?

Issue #4052, opened 2022-02-09
Optimize beacon chain sync by only syncing the epoch blocks.


rlan35:
# Problem/limitation At Hand

Beacon chain or shard 0 is the chain every nodes in the shard 1 2 3 needs to synchronize with in order to know the latest sharding structure or committee configuration. Currently, nodes in shard 1 2 3 are simply utilizing the existing blockchain sync mechanism to fully download and verify all the blocks in beacon chain, which is very costly both for computation and storage (shard 0 is taking hundreds of GB of disk).

Validators in shard 1 2 3 are complaining that they are storing too much data as they need to store both their own chain in shard 1 2 3 and the beacon chain. Besides, beacon chain is broadcasting every new block to the whole network so nodes can catch up to the latest blocks real time, which also causes lots of p2p network traffic.

In fact, only the epoch block of beacon chain is needed for the nodes in other shards to know about the epoch transition and committee configuration, because this information only changes at the epoch block (between epochs).

# Proposed Solution

Instead of synchronizing with every single block in beacon chain, only download and synchronize the epoch blocks for beacon chain (which happens every 18.2 hours/epoch).

1. Change the beacon block broadcasting logic to only broadcast epoch blocks.
2. For nodes in shard 1 2 3, only listen to epoch blocks and verify the epoch block based on the signature on the block. No need to do the formal state transition check here since we are not verifying all the blocks. Once the block is verified, read the "shard state" or the committee configuration for the new epoch and do epoch transition just as before using the new shard state.
3. Update the chain sync logic to be able to synchronize the beacon chain epoch blocks from genesis. Basically download all the epoch blocks from genesis and then verify them in sequence until it hits the latest epoch block in beacon chain.

MaxMustermann2:
#4070

PR #4070, opened 2022-02-25
Epoch syncing - syncing only epoch blocks.

Frozen:
## Issue
Optimize beacon chain sync by only syncing the epoch blocks. 
https://github.com/harmony-one/harmony/issues/4052

## Test

https://gist.github.com/Frozen/fd0ffb735e94f7424416d5421c9341e1

## Operational Checklist

    **NO**

    **NO**


api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	// TODO(k.potapov) validate blocks
+
+	for _, block := range decoded {
+		_, err := bc.StoreShardStateBytes(block.Header().Epoch(), block.Header().ShardState())
+		if err != nil {
+			return err
+		}
+
+		err = bc.WriteBlockWithoutState(block, common.Big1)
```
JackyWYX, line 232: Have you tested locally or on mainnet? Is there any error returned from this line?

JackyWYX: There is a lot of messages written to leveldb other than the state. Furthermore, it seems there are not enough check on the block message.

Frozen: Yes, it works locally, i'm going to test it on testnet. There are no checks on block inside WriteBlockWithoutState, i'll add signature validation, other impossible without state.

JackyWYX: Please take a look at Finalize function at `internal/chain/engine.go:276`. You will at least need the state shard for the signature check.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	for _, block := range decoded {
+		sig, bitmap, err := chain.ParseCommitSigAndBitmap(block.GetCurrentCommitSig())
+		if err != nil {
+			return errors.Wrap(err, "parse commitSigAndBitmap")
+		}
+
+		// Signature validation.
+		err = bc.Engine().VerifyHeaderSignature(bc, block.Header(), sig, bitmap)
```
Frozen, line 232: I've added signature validation, but i'm not sure is it correct and enough? Maybe i should add some extra validation?

api/service/legacysync/epoch_syncing.go
Frozen: Yea, i checked that function, but it seems my changes does not affect it. Shard state functions called only for beacon chain, and my code works only for non-beacon.

api/service/legacysync/epoch_syncing.go
rlan35: Signature verification is enough and we can't validate other data since we don't have the consecutive blocks and state.

JackyWYX:
For best engineering practice, I would personally suggest to write a `EpochChain` which maintains a header of epoch blocks and validators data. It can be a component within the `HeaderChain` to avoid doing nasty stuff outside the core module.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,426 @@
...
+			utils.Logger().Error().Err(err).
+				Msgf("[SYNC] ProcessStateSync failed (isBeacon: %t, ShardID: %d, otherHeight: %d, currentHeight: %d)",
+					isBeacon, bc.ShardID(), maxHeight, height)
+			return 2
+		}
+	}
+}
+
+// getMaxPeerHeight gets the maximum blockchain heights from peers
+func (ss *EpochSync) getMaxPeerHeight(isBeacon bool) uint64 {
```
rlan35, line 161: Please refactor the code so no duplicated code like this exists.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,426 @@
...
+// If the last result is expired, ask the remote DNS nodes for latest height and return the result.
+func (ss *EpochSync) GetSyncStatus() SyncCheckResult {
+	return ss.syncStatus.Get(func() SyncCheckResult {
+		return ss.isInSync(false)
+	})
+}
+
+// isInSync query the remote DNS node for the latest height to check what is the current
+// sync status
+func (ss *EpochSync) isInSync(doubleCheck bool) SyncCheckResult {
```
rlan35, line 50: Whether epoch blocks are in sync should depends on checking the latest epoch block right? Seems this part of the code is just a copy and paste of the logic that's checking the latest block of the chain?

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,426 @@
...
+		curEpoch := block.Epoch()
+		for i := 0; i < int(SyncLoopBatchSize); i++ {
+			epochLastBlock := shard.Schedule.EpochLastBlock(curEpoch.Uint64())
+			if epochLastBlock > maxHeight {
+				break
+			}
+			if height != epochLastBlock {
+				heights = append(heights, epochLastBlock)
+			}
+			curEpoch = curEpoch.Add(curEpoch, common.Big1)
```
rlan35, line 142: what is curEpoch used for?

api/service/legacysync/epoch_syncing.go
rlan35: You need to call WriteShardStateBytes() or something related to write the shardState to DB. Take a look at WriteBlockWithState and see how it works, especially the part about offChain data. Then you will know why WriteShardStateBytes() is needed. Because otherwise, the shard state can be read from the DB.

Frozen: I use StoreShardStateBytes above, that function is a wrapper for `WriteShardStateBytes` and calls it internally. So, yes, i store shardState to DB. `WriteBlockWithoutState` i need for height in order to understand where i am right now and what to do next. Without `WriteBlockWithoutState` it starts to panic in different places, it's better not to skip that.

api/service/legacysync/epoch_syncing.go
Frozen: Right, now it depends on last epoch.

api/service/legacysync/epoch_syncing.go
Frozen: fixed

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,383 @@
...
+		if otherHeight1 < lastHeight {
+			heightDiff = 0 //
+		}
+		onTop := epochLastBlock > otherHeight1
+		utils.Logger().Info().
+			Uint64("OtherHeight", otherHeight1).
+			Uint64("lastHeight", lastHeight).
+			Msg("[SYNC] Checking sync status")
+		return SyncCheckResult{
+			IsInSync:    !onTop,
```
rlan35, line 69: onTop means our height is bigger than other height, so it means in sync right? So here it should be:

IsInSync: onTop.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,383 @@
...
+	"github.com/harmony-one/harmony/internal/chain"
+	"github.com/harmony-one/harmony/internal/utils"
+	"github.com/harmony-one/harmony/node/worker"
+	"github.com/harmony-one/harmony/p2p"
+	"github.com/harmony-one/harmony/shard"
+	"github.com/pkg/errors"
+)
+
+type EpochSync struct {
+	blockChain         *core.BlockChain
```
rlan35, line 23: Since we are talking about epoch block sync for beacon chain, this blockChain should be beaconChain right?

api/service/legacysync/epoch_syncing.go
rlan35: I think for checking whether the epoch blocks are latest in sync with the beacon chain, the check logic should be:

1. Get the latest epoch block I have: myEpoch.
2. Get maxPeerHeight: otherHeight
3. using shard schedule to calculate what epoch is the "otherHeight" belong and get: otherEpoch.
4. Compare otherEpoch vs myEpoch to decide whether I am in sync or not.

Frozen: OnTop means i have the highest block that possible. In epoch context it means that i have the highest epoch. I can calculate last block for this epoch, and if it doesn't exists yet means i'm idling. Otherwise i'm synchronizing. I do the same above, but in another words, without converting into epoch, but in heights.

api/service/legacysync/epoch_syncing.go
Frozen: onTop means that our possible height for last block is higher than the highest known block from network.

Frozen: Fixed

api/service/legacysync/epoch_syncing.go
Frozen: i've realized that "in sync" means already synchronized but not "in synchronizing process", exactly opposite meaning. Fixed.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,309 @@
...
+}
+
+// isInSync query the remote DNS node for the latest height to check what is the current
+// sync status
+func (ss *EpochSync) isInSync(doubleCheck bool) SyncCheckResult {
+	if ss.syncConfig == nil {
+		return SyncCheckResult{} // If syncConfig is not instantiated, return not in sync
+	}
+	otherHeight1 := getMaxPeerHeight(ss.syncConfig)
+	curBlock := ss.beaconChain.CurrentBlock()
```
rlan35, line 53: which block does this return? With epoch sync'ing, we don't have the full beacon chain, right? we only have all the epoch blocks of the beacon chain.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+
+	for _, block := range decoded {
+		sig, bitmap, err := chain.ParseCommitSigAndBitmap(block.GetCurrentCommitSig())
+		if err != nil {
+			return errors.Wrap(err, "parse commitSigAndBitmap")
+		}
+
+		// Signature validation.
+		err = bc.Engine().VerifyHeaderSignature(bc, block.Header(), sig, bitmap)
```
rlan35, line 195: Actually, this verification won't work, because the commit/sig stored in the block N is for the last block N-1, to get the commit/sig for block N, you need to find them in the block N+1.

rlan35: Have you tested this part of the code? I bet the verification of the sig should fail.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,309 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// getConsensusHashes gets all hashes needed to download.
+func (ss *EpochSync) getConsensusHashes(startHash []byte, size uint32) error {
```
rlan35, line 222: this and CreateSyncConfig method are all copy/paste, please use inheritance to refactor them so it's reused.

api/service/legacysync/helpers.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"sync"
+
+	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/harmony-one/harmony/internal/utils"
+)
+
+// getMaxPeerHeight gets the maximum blockchain heights from peers
+func getMaxPeerHeight(syncConfig *SyncConfig) uint64 {
```
rlan35, line 11: same here, duplicate code shouldn't exist.

core/blockchain.go
```diff
@@ -334,15 +360,18 @@ func (bc *BlockChain) loadLastState() error {
...
-	if _, err := state.New(currentBlock.Root(), bc.stateCache); err != nil {
-		// Dangling block without a state associated, init from scratch
-		utils.Logger().Warn().
-			Str("number", currentBlock.Number().String()).
-			Str("hash", currentBlock.Hash().Hex()).
-			Msg("Head state missing, repairing chain")
-		if err := bc.repair(&currentBlock); err != nil {
-			return err
+
+	if !bc.options.SkipInitialStateValidation {
```
rlan35, line 364: Why is this needed? please add comment to explain.

core/blockchain.go
```diff
@@ -1912,6 +1945,22 @@ func (bc *BlockChain) WriteShardStateBytes(db rawdb.DatabaseWriter,
 	return decodeShardState, nil
 }
 
+func (bc *BlockChain) StoreShardStateBytes(epoch *big.Int, shardState []byte) (*shard.State, error) {
```
rlan35, line 1948: can you reuse the existing WriteShardStateBytes() instead ?

rlan35:
Is this code working for mainnet data? Please test by running a shard 1 explorer node.

api/service/legacysync/epoch_syncing.go
Frozen: Last successfully inserted block.

core/blockchain.go
Frozen: Technically yes, but it would be inconvenient.

core/blockchain.go
Frozen: Added comment. 
	// Skip state check on initialization.
	// Should be enabled for epoch chain because it doesn't contain valid state.

api/service/legacysync/epoch_syncing.go
Frozen: fixed

node/node.go
```diff
@@ -151,7 +152,11 @@ func (node *Node) Blockchain() *core.BlockChain {
 
 // Beaconchain returns the beaconchain from node.
 func (node *Node) Beaconchain() *core.BlockChain {
-	bc, err := node.shardChains.ShardChain(shard.BeaconChainShardID)
+	return node.EpochChain()
```
rlan35, line 155: Please don't change the existing Beaconchain() which is used somewhere else which may require the full chain. Beaconchain should mean the full chain while EpochChain means the chain with only epoch blocks. Beaconchain can be used as a EpochChain, but EpochChain can't be used as Beaconchain.

node/node_handler.go
```diff
@@ -66,25 +66,7 @@ func (node *Node) HandleNodeMessage(
 	case proto_node.Block:
 		switch blockMsgType := proto_node.BlockMessageType(msgPayload[0]); blockMsgType {
 		case proto_node.Sync:
-			blocks := []*types.Block{}
```
rlan35, line 69: We may still need some logic in this part because instead of broadcasting every block from beacon chain, now only the epoch blocks need to be broadcasted and processed here.

node/node.go
Frozen: Fixed

sophoah:
@Frozen in regards of the recent S1/S2 consensus loss, can you explain how the s0 last epoch sync process starts ? is it possible for a non S0 validator node currently up to date, later lagging behind to the last s0 block, to start a sync when new epoch starts ?

This will help to minimize the time for a shard to recover the consensus when the node is lagging behind at epoch change

cc @rlan35

sophoah:
With the latest commit, as of now :
- node will sync the last epoch block of the previous epoch
- node will be stuck at that block and won't further sync
- and it happens at each new epoch 

is that an intended behavior ? or shouldn't we sync all the block in the current epoch including the last block of the previous ?

sophoah:
question : would the code work as well with streamsync enabled ?

Frozen:
No, it doesn't support streamsync.

Frozen:
Right now it downloads all epoch blocks only, mean last blocks of epoch, with shard state info. i'm going to change it to download only last block of previous epoch, it will reduce lag between node node is ready and not to validate new blocks.

rlan35:
> Right now it downloads all epoch blocks only, mean last blocks of epoch, with shard state info. i'm going to change it to download only last block of previous epoch, it will reduce lag between node node is ready and not to validate new blocks.

All the epoch blocks starting from genesis should be downloaded because you need to be able to verify each of them. Without the previous epoch block, you don't have the committee information to verify the block successfully. So I think you need to make sure the verification of a new epoch block is based on committee of last epoch block.

sophoah:
Hit an issue with the last epoch change, not sure if that happened before as well but the test node is currently over 500 blocks after the new epoch blocks (since epoch change) and the node hasn't downloaded the new epoch block.

here is a log that might be a trigger of this issue :
```json
{"level":"error","error":"empty payload","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:145","time":"2022-04-10T23:57:17.257374782-05:00","message":"[SYNC] ProcessStateSync failed (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 25100287)"}
{"level":"info","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:121","time":"2022-04-10T23:57:19.260001879-05:00","message":"[SYNC] Node is OUT OF SYNC (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 25100287)"}
{"level":"error","error":"empty payload","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:145","time":"2022-04-10T23:57:19.260091004-05:00","message":"[SYNC] ProcessStateSync failed (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 25100287)"}
{"level":"info","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:121","time":"2022-04-10T23:57:21.26087363-05:00","message":"[SYNC] Node is OUT OF SYNC (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 25100287)"}
{"level":"error","error":"empty payload","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:145","time":"2022-04-10T23:57:21.260957516-05:00","message":"[SYNC] ProcessStateSync failed (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 25100287)"}
{"level":"info","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:121","time":"2022-04-10T23:57:23.261946835-05:00","message":"[SYNC] Node is OUT OF SYNC (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 25100287)"}
```

sophoah:
@PkayJava could you help to test that PR on a testnet S0 node and observe the behavior ?
- build a testnet s0 node and validator (I can send you testnet token to create the validator)
- start it with an empty shard 0 DB
- observe sync and consensus
report here any issue. Thanks

PkayJava:
@sophoah - noted with okay.

PkayJava:
@Konstantin - This PR look good.

{"level":"info","blockNum":24600427,"epochNum":75664,"ViewId":24606416,"blockHash":"0xc1a4f16aec3131347ee4424213986f3e85fdcfe23eb2d20b6470b3b6f5e67769","numTxns":1,"numStakingTxns":0,"numSignatures":1,"caller":"/opt/go-path/src/github.com/harmony-one/harmony/node/node_handler.go:377","time":"2022-04-28T15:07:25.345778224Z","message":"BINGO !!! Reached Consensus"}

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+func (ss *EpochSync) isInSync(doubleCheck bool) SyncCheckResult {
+	if ss.syncConfig == nil {
+		return SyncCheckResult{} // If syncConfig is not instantiated, return not in sync
+	}
+	otherHeight1 := getMaxPeerHeight(ss.syncConfig)
+	curBlock := ss.beaconChain.CurrentBlock()
+	epochLastBlock := shard.Schedule.EpochLastBlock(curBlock.Epoch().Uint64())
+	lastHeight := ss.beaconChain.CurrentBlock().NumberU64()
+
+	if !doubleCheck {
```
rlan35, line 56: Whether the epoch chain is in sync should be determined by this:

1. Get my latest epoch block (note since it's epoch chain, the CurrentBlock() will return an epoch block (the last block of an epoch). The epoch of this block is my_epoch.
2. Get the highest block from Peer. The epoch of this block is their_epoch.
3. Compare whether my_epoch == their_epoch-1.  Note it's almost impossible to have my_epoch == their_epoch because I only have the epoch block which is the last block of an epoch. So when I sync'ed the epoch block of an epoch, the beacon chain should already moved to the next epoch.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+			Uint64("OtherHeight", otherHeight1).
+			Uint64("lastHeight", lastHeight).
+			Msg("[EPOCHSYNC] Checking sync status")
+		return SyncCheckResult{
+			IsInSync:    inSync,
+			OtherHeight: otherHeight1,
+			HeightDiff:  heightDiff,
+		}
+	}
+	// double check the sync status after 1 second to confirm (avoid false alarm)
```
rlan35, line 72: No need for double check since epoch block only increase every 18.5hours. There is almost no chance of false alarm.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+		timeout = ss.syncLoop(bc, worker, isBeacon, consensus)
+	}
+}
+
+func (ss *EpochSync) syncLoop(bc *core.BlockChain, worker *worker.Worker, isBeacon bool, _ *consensus.Consensus) (timeout int) {
+	maxHeight := getMaxPeerHeight(ss.syncConfig)
+	for {
+		block := bc.CurrentBlock()
+		height := block.NumberU64()
+		if height >= maxHeight {
```
rlan35, line 113: again, this sync check doesn't work. The epoch block will always be less than the maxHeight since it's the last block of an epoch. We should check my_epoch >= theirEpoch - 1.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+
+		var heights []uint64
+		curEpoch := block.Epoch()
+		for i := 0; i < int(SyncLoopBatchSize); i++ {
+			epochLastBlock := shard.Schedule.EpochLastBlock(curEpoch.Uint64())
+			if epochLastBlock > maxHeight {
+				break
+			}
+			if height != epochLastBlock {
+				heights = append(heights, epochLastBlock)
```
rlan35, line 132: when i = 0, we are downloading the latest epoch block we already had "block := bc.CurrentBlock()". This is an epoch block and if we get EpochLastBlock of it's epoch. it returns the block itself. So we should skip this first one.

api/service/legacysync/helpers.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"github.com/ethereum/go-ethereum/common/math"
+	"github.com/harmony-one/harmony/api/service/legacysync/downloader"
+	"github.com/harmony-one/harmony/internal/utils"
+	"github.com/harmony-one/harmony/p2p"
+	"github.com/pkg/errors"
+)
+
+// getMaxPeerHeight gets the maximum blockchain heights from peers
+func getMaxPeerHeight(syncConfig *SyncConfig) uint64 {
```
rlan35, line 15: Just to confirm. the code in this file helpers.go is directly copy from the original code without functional change, right?

api/service/legacysync/epoch_syncing.go
sophoah: just a side node on this comment @rlan35, it happened on the mainnet i am testing that we have this logs constantly :
```
{"level":"info","caller":"/home/sp/harmony/harmony/api/service/legacysync/epoch_syncing.go:121","time":"2022-06-19T18:56:04.933824999-05:00","message":"[EPOCHSYNC] Node is OUT OF SYNC (isBeacon: true, ShardID: 0, otherHeight: 18446744073709551615, currentHeight: 27951103)"}
```
until a restart of the node is initiated, and the otherHeight is correct

Frozen: 18446744073709551615 is max value for uint64

It started with this value, other peers should overwrite it, but it doesn't. Seems like no other peers or no even one peer responses correctly.

api/service/legacysync/helpers.go
Frozen: yes, copy paste

api/service/legacysync/epoch_syncing.go
Frozen: removed

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,231 @@
...
+			Int("Peers count", ss.syncConfig.PeersCount()).
+			Msg("[EPOCHSYNC] No peers for get height")
+		return SyncCheckResult{}
+	}
+	curEpoch := ss.beaconChain.CurrentBlock().Epoch().Uint64()
+	otherEpoch := shard.Schedule.CalcEpochNumber(otherHeight1).Uint64()
+	normalizedOtherEpoch := otherEpoch - 1
+	inSync := curEpoch == normalizedOtherEpoch
+
+	heightDiff := normalizedOtherEpoch - curEpoch
```
rlan35, line 66: This is not height diff anymore. it's epoch diff. Please change the naming.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	utils.Logger().Info().
+		Uint64("OtherEpoch", otherEpoch).
+		Uint64("CurrentEpoch", curEpoch).
+		Msg("[EPOCHSYNC] Checking sync status")
+	return SyncCheckResult{
+		IsInSync:    inSync,
+		OtherHeight: otherHeight1,
+		HeightDiff:  heightDiff,
```
rlan35, line 78: EpochDiff.

node/node_syncing.go
```diff
@@ -224,32 +220,18 @@ func (node *Node) doBeaconSyncing() {
...
-						}
-						err := node.beaconSync.UpdateBlockAndStatus(
-							beaconBlock, node.Beaconchain(), true,
-						)
-						if err != nil {
-							node.beaconSync.AddLastMileBlock(beaconBlock)
-						}
-					}
-				}
+			for _ = range node.BeaconBlockChannel {
```
rlan35, line 223: Why is this loop empty now?

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,231 @@
...
+		utils.Logger().Info().
+			Msgf("[EPOCHSYNC] Node is OUT OF SYNC (isBeacon: %t, ShardID: %d, otherEpoch: %d, currentEpoch: %d, peers count %d)",
+				isBeacon, bc.ShardID(), otherEpoch, curEpoch, ss.syncConfig.PeersCount())
+
+		var heights []uint64
+		loopEpoch := curEpoch + 1
+		for len(heights) < int(SyncLoopBatchSize) {
+			epochLastBlockHeight := shard.Schedule.EpochLastBlock(loopEpoch)
+			if epochLastBlockHeight > maxHeight {
+				break
```
rlan35, line 123: Change the condition to check on epoch: loopEpoch >= otherEpoch

internal/utils/crosslink_storage.go
```diff
@@ -0,0 +1,27 @@
+package utils
+
+import "sync"
+
+type CrosslinkStorage struct {
```
rlan35, line 5: What's CrosslinkStorage for in this PR? Please delete if it's out of date.

node/node_cross_link.go
```diff
@@ -58,7 +62,82 @@ func (node *Node) ProcessCrossLinkHeartbeatMessage(msgPayload []byte) {
...
+		}
+	}
+
+	if !keyExists {
+		return errors.New("pub key doesn't exist")
+	}
+
+	utils.Logger().Info().
+		Msgf("[ProcessCrossLinkHeartbeatMessage] storing latestSentCrosslink to %d", hb.LatestContinuousBlockNum)
+	atomic.StoreUint64(&latestSentCrosslink, hb.LatestContinuousBlockNum)
```
rlan35, line 139: latestSentCrosslink is just a temporary in-memory cache. The real lastContinuousCrossLink should be stored in DB.  Take a look at "ReadShardLastCrossLink" in the original code.

node/node_handler.go
```diff
@@ -84,9 +84,11 @@ func (node *Node) HandleNodeMessage(
 						if block.ShardID() == 0 {
 							utils.Logger().Info().
 								Msgf("Beacon block being handled by block channel: %d", block.NumberU64())
-							go func(blk *types.Block) {
-								node.BeaconBlockChannel <- blk
-							}(block)
+							if block.IsLastBlockInEpoch() {
+								go func(blk *types.Block) {
+									node.BeaconBlockChannel <- blk
```
rlan35, line 89: Note to also change the beacon chain block broadcasting logic to only broadcast epoch block.

node/node_handler.go
```diff
@@ -177,6 +179,9 @@ func (node *Node) BroadcastCrossLinkFromShardsToBeacon() { // leader of 1-3 shar
 	if node.IsRunningBeaconChain() {
 		return
 	}
+	if !(node.Consensus.IsLeader() || rand.Intn(100) == 0) {
```
rlan35, line 182: Change to do 2%. "rand.Intn(100) <= 1"

Frozen: fixed

internal/utils/crosslink_storage.go
Frozen: removed

api/service/legacysync/epoch_syncing.go
Frozen: fixed

api/service/legacysync/epoch_syncing.go
Frozen: SyncCheckResult is struct with field HeightDiff, and it used in other places too, for example in legasy sync, where 
HeightDiff is appropriate name. I can't just rename it, it will lead to other changes. Suggest leave it as is.

node/node_syncing.go
Frozen: everything inside this loop is moved to another place or removed as useless. But we still need to read from channel, or it will block. I will consider removing this channel, but not necessarily in the current pr.

node/node_cross_link.go
rlan35: If you don't store the latestContinuousBlockNum in DB, you should update the logic involving ReadShardLastCrossLink() since that part of the code is still reading that data from DB.

rlan35 (approved):

rlan35:
Looks good. Please fix travis, and then let's test on testnet first.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,196 @@
...
+		if maxHeight == 0 || maxHeight == math.MaxUint64 {
+			utils.Logger().Info().
+				Msgf("[EPOCHSYNC] No peers to sync (isBeacon: %t, ShardID: %d, peerscount: %d)",
+					isBeacon, bc.ShardID(), ss.syncConfig.PeersCount())
+			return 10
+		}
+
+		curEpoch := bc.CurrentBlock().Epoch().Uint64()
+		otherEpoch := shard.Schedule.CalcEpochNumber(maxHeight).Uint64()
+		if otherEpoch-1 <= curEpoch {
```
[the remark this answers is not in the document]
rlan35: Good point. better change it for peace of mind.

p2p/stream/protocols/sync/chain.go
```diff
@@ -50,23 +50,20 @@ func (ch *chainHelperImpl) getBlockHashes(bns []uint64) []common.Hash {
 }
 
 func (ch *chainHelperImpl) getBlocksByNumber(bns []uint64) ([]*types.Block, error) {
-	var (
-		blocks = make([]*types.Block, 0, len(bns))
-	)
+	curBlock := ch.chain.CurrentHeader().Number().Uint64()
```
rlan35, line 53: Is this change a bug fix?

rlan35 (approved):

rlan35:
Finally looks good overall!

api/service/legacysync/epoch_syncing.go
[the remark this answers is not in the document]
Frozen: fixed

MaxMustermann2 (approved):

PR #4073, opened 2022-03-01
change rosetta DefaultGasPrice

jhd2best:


rosetta/services/construction.go
```diff
@@ -19,7 +19,7 @@ import (
 
 const (
 	// DefaultGasPrice ..
-	DefaultGasPrice = denominations.Nano
+	DefaultGasPrice = 40 * denominations.Nano
```
LeoHChen, line 22: the default is 30, right?  why change to 40?

jhd2best: default is 1

Issue #4076, opened 2022-03-02
Add hmyv2_getBlockReceipts RPC V2 API


bbist:
We need to fetch all transaction receipts for a given block. We are currently fetching all transactions of given block and using hundreds of concurrent calls to get receipt for each transaction. This puts unnecessary latency on our (relay) application as well as stress on harmony node. It would make sense to add an endpoint `hmyv2_getBlockReceipts` RPC call that returns all receipts of a specified block hash.
Note: The issue with high load and latency is severe when icon chain is out of sync with harmony chain, and relay has to synchronize many old blocks.

**Which component?**

HMYV2 RPC API

# Proposed Solution

Implement `hmyv2_getBlockReceipts` RPC call as specified above.

PR #4077, opened 2022-03-02
bbist/hmyv2_getBlockReceipts

bbist:
## Issue
https://github.com/harmony-one/harmony/issues/4076

~~Possible breaking changes include the changes to TxReceipt and StakingTxReceipt RPC response structs to make them compatible with ethereum clients, which I suppose should have been the goal anyway.~~

Reverted back the changes to v2 receipt types, and choosing to use `hmy_getBlockReceipts` RPC v1 API instead.

### DIFF
--- a/rpc/blockchain.go
+++ b/rpc/blockchain.go
@@ -415,6 +415,75 @@ func (s *PublicBlockchainService) GetBlockSignerKeys(
 	return s.helper.GetBLSSigners(bn)
 }
 
+// GetBlockReceipts returns all transaction receipts for a particular block.
+func (s *PublicBlockchainService) GetBlockReceipts(
+	ctx context.Context, blockHash common.Hash,
+) ([]StructuredResponse, error) {
+	timer := DoMetricRPCRequest(GetBlockReceipts)
+	defer DoRPCRequestDuration(GetBlockReceipts, timer)
+
+	block, err := s.hmy.GetBlock(ctx, blockHash)
+	if err != nil {
+		return nil, err
+	}
+
+	receipts, err := s.hmy.GetReceipts(ctx, blockHash)
+	if err != nil {
+		return nil, err
+	}
+
+	rmap := make(map[common.Hash]*types.Receipt, len(receipts))
+	for _, r := range receipts {
+		rmap[r.TxHash] = r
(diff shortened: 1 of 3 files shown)


rlan35 (approved):

bbist:
@rlan35,
Can you please review this again?
We've been using this endpoint on our localnet without any issue. But today we tried to do a testnet tests, and this RPC endpoint has started to return "method handler crashed" error. It is so frequent that the RPC endpoint is almost unusable.

bbist:
I get the results most of the time. But once it starts crashing it keeps on crashing for some blocks and works for other blocks.
I get the "method handler crashed" from direct RPC call as well.
image

The postman snapshot was for the block receipts for block number 23672403 as specified in the last line of the log file above.

At the time of this testing, it gave no block receipts but it started working after a while.
image

bbist:
Here is a script to replicate the issue. You need `jq` and `curl` installed on your system.
```bash
#!/usr/bin/env bash

btp_hmny_uri=https://rpc.s0.b.hmny.io

function hmny_jsonrpc() {
    curl "$btp_hmny_uri" -s -X POST \
        -H 'Content-Type:application/json' \
        -d "$(jq <<<{} -c \
            '.id=1|.jsonrpc="2.0"|.method=$method|.params=$params' \
            --arg method "$1" --argjson params "$2")"
}

block_number=$(hmny_jsonrpc hmyv2_blockNumber '[]' | jq -r .result)

for ((i = $block_number; i < $((block_number + 1000)); i++)); do
    block_hash=$(hmny_jsonrpc hmyv2_getBlockByNumber "[$i,{}]" | jq -r .result.hash)
    if [[ "$block_hash" == "null" ]]; then
        ((i--))
        continue
    fi
    resp=$(hmny_jsonrpc hmy_getBlockReceipts "[\"$block_hash\"]")
    error=$(echo "$resp" | jq -r .error)
    msg=""
    if [[ "$error" == "null" ]]; then
        msg=$(echo "$resp" | jq -r .result | jq -c .)
    else
        msg=$(echo "$error" | jq -c .)
    fi
    echo "$i: $block_hash: ${msg:0:100}"
done
```

PR #4084, opened 2022-03-07
add rpc getShardState and getShardStateRLPHex by epoch number

bbist:
Added following RPC methods:
1. `hmyv2_getShardState`
    Return shardState for given epoch number.
2. `hmyv2_getShardStateRLPHex`
    Return RLP -> Hex encoded shardState for given epoch number.

### DIFF
--- a/rpc/blockchain.go
+++ b/rpc/blockchain.go
@@ -807,6 +807,48 @@ func toHexSlice(b [][]byte) []string {
 	return r
 }
 
+func (s *PublicBlockchainService) GetShardState(
+	ctx context.Context, epoch *big.Int,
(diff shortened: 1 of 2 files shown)


rlan35:
Curious to know what's the use case to add these rpcs?

rlan35 (approved):

bbist:
> Curious to know what's the use case to add these rpcs?

I apologize for a late reply.

We needed to get the validators used to sign blocks of a specific epoch. But we found RPC endpoint epochLastBlock to get the last block number of a specific epoch. So we can extract the shardState from the full block header of epochLastBlock.

Hence this PR is not needed anymore.

PR #4095, opened 2022-03-11
Consensus extra commit logic to avoid waiting for 100% sigs

rlan35:
This PR implement the extra commit signature logic which allow leaders to only wait for 2/3 of the commit signature in current block and include the rest of the signatures (extra commit sig) in next block. Specifically, following changes will take effect after the extra commit fork epoch:

1. ExtraCommitSig and ExtraCommitBitmap field is added to block header v4 which only take effect after ExtraCommitEpoch.
2. Leaders will immediately finalizing the block once it received more than 2/3 of the voting powers on commit sig.
3. In the next block, the leader will accept the rest of the commit signatures from the previous block and store it as extra commit signatures in DB and include it in the proposal in the new block. 
4. Validators is allowed to sign commit signature on the previous block so the signature can be included as a extra commit sig even if the validator is slow and didn't make it in the first 2/3 of the signatures.
5. After the fork epoch, crosslinks and block rewards are calculated by combing the extra commit sig and the normal 2/3 of the sig together. The block reward distribution logic stay the same.

Fully tested this change on local network and was able to see extra commit signature being signed and accepted, crosslinks constructed/accepted and block rewards distributed correctly.

### DIFF
--- a/block/v4/header.go
+++ b/block/v4/header.go
@@ -68,7 +68,7 @@ type headerFields struct {
 	Vdf                  []byte   `json:"vdf"`
 	ShardState           []byte   `json:"shardState"`
 	CrossLinks           []byte   `json:"crossLink"`
-	Slashes              []byte   `json:slashes`
+	Slashes              []byte   `json:"slashes"`
 }
 
 // ParentHash is the header hash of the parent block.  For the genesis block


consensus/checks.go
```diff
@@ -67,6 +67,18 @@ func (consensus *Consensus) isRightBlockNumAndViewID(recvMsg *FBFTMessage,
 	return true
 }
 
+func (consensus *Consensus) isPreviousBlockNumAndViewID(recvMsg *FBFTMessage,
+) bool {
+	if recvMsg.ViewID != consensus.GetCurBlockViewID()-1 || recvMsg.BlockNum != consensus.blockNum-1 {
```
rlan35, line 72: Previous block's viewID should be fetched from the block itself, not calculating with curBlock.viewID -1.

consensus/consensus_service.go
```diff
@@ -227,7 +232,8 @@ func (consensus *Consensus) checkViewID(msg *FBFTMessage) error {
 		return nil
 	} else if msg.ViewID > consensus.GetCurBlockViewID() {
 		return consensus_engine.ErrViewIDNotMatch
-	} else if msg.ViewID < consensus.GetCurBlockViewID() {
+	} else if msg.ViewID < consensus.GetCurBlockViewID()-1 {
```
rlan35, line 235: Update to compare with the viewID of last bock.

consensus/quorum/one-node-one-vote.go
```diff
@@ -83,6 +83,11 @@ func (v *uniformVoteWeight) IsAllSigsCollected() bool {
 	return v.SignersCount(Commit) == v.ParticipantsCount()
 }
 
+// IsAllSigsCollected ..
+func (v *uniformVoteWeight) IsAllSigsCollectedInPreviousBlock() bool {
+	return v.SignersCount(CommitInPreviousBlock) == v.ParticipantsCount()
```
rlan35, line 88: Make sure this logic works during epoch transition.

block/v4/header.go
```diff
@@ -0,0 +1,448 @@
...
+	Time                *big.Int       `json:"timestamp"        gencodec:"required"`
+	Extra               []byte         `json:"extraData"        gencodec:"required"`
+	MixDigest           common.Hash    `json:"mixHash"          gencodec:"required"`
+	// Additional Fields
+	ViewID               *big.Int `json:"viewID"           gencodec:"required"`
+	Epoch                *big.Int `json:"epoch"            gencodec:"required"`
+	ShardID              uint32   `json:"shardID"          gencodec:"required"`
+	LastCommitSignature  [96]byte `json:"lastCommitSignature"  gencodec:"required"`
+	LastCommitBitmap     []byte   `json:"lastCommitBitmap"     gencodec:"required"` // Contains which validator signed
+	ExtraCommitSignature [96]byte `json:"extraCommitSignature"  gencodec:"required"`
```
MaxMustermann2, line 65: Add comment to mark v3 and v4 diff please.

consensus/consensus_v2.go
```diff
@@ -209,6 +209,22 @@ func (consensus *Consensus) finalCommit() {
 	}
 
 	block.SetCurrentCommitSig(commitSigAndBitmap)
+
+	commitSigBitmap := CommitSigBitmaps{CommitSigBitmap: commitSigAndBitmap}
+	parentHeader := consensus.Blockchain.GetBlockByHash(block.ParentHash())
+	if consensus.Blockchain.Config().IsExtraCommit(parentHeader.Epoch()) && parentHeader.Number().Uint64() > 1 {
```
MaxMustermann2, line 215: Why are you using the condition `parentHeader.Number().Uint64() > 1` ?

rlan35: Added "// Extra commit sig logic only works starting from block 2 since genesis block doesn't have signatures."

block/v4/header.go
rlan35: done.

consensus/leader.go
```diff
@@ -195,8 +195,25 @@ func (consensus *Consensus) onPrepare(recvMsg *FBFTMessage) {
...
+			Msg("[OnCommit] Failed finding a matching block for committed message")
+		return
+	}
+
 	//// Read - Start
-	if !consensus.isRightBlockNumAndViewID(recvMsg) {
+	if !consensus.isCurrentBlockNumAndViewID(recvMsg) {
+		if consensus.Blockchain.Config().IsExtraCommit(blockObj.Epoch()) &&
+			consensus.isLatestCommittedBlockNumAndViewID(recvMsg) && !blockObj.IsLastBlockInEpoch() {
+			// If it's the block in extra epoch epoch and not the epoch block, process it for the extra commit sig.
```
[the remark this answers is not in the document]
rlan35: done

core/blockchain.go
```diff
@@ -166,6 +167,7 @@ type BlockChain struct {
 	futureBlocks                  *lru.Cache     // future blocks are blocks added for later processing
 	shardStateCache               *lru.Cache
 	lastCommitsCache              *lru.Cache
+	extraCommitsCache             *lru.Cache
```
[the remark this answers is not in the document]
rlan35: It's a 10 slot cache, which will be close to nothing to store 10 signatures. So not a concern. But good to check.

MaxMustermann2 (approved):

Issue #4014, opened 2022-01-18
[Feature] Enhancement to hmyv2_getValidatorInformation


FortuneV13:
API enhancement to hmyv2_getValidatorInformation

I've noticed last epoch in  committee can get out of sync with the signed data especially around the beginning of the epoch.

I have also noticed that on subsequent calls the data can be older than that of the previous. Both myself (vstats)  and smart stake have had these issues. 

I would like to proposed two variables to be added to current-epoch-signing-percent:

1) Epoch number which signed data relates to. 
2) Block number  

This would make it much easier to grab correct data when polling for various stats. 
Let me know if you need anymore info.

 "result": {
        "active-status": "active",
        "booted-status": null,
        "current-epoch-performance": {
            "current-epoch-signing-percent": {
                "current-epoch-signed": 905065,
                "current-epoch-signing-percentage": "1.000000000000000000",
                "current-epoch-to-sign": 905065,
               ** "epoch": xxxxx**,
               ** "block": xxxxx**
            }
        },

PR #4096, opened 2022-03-12
Add epoch, block number to hmyv2_getValidatorInformation response

harjas27:
## Issue
https://github.com/harmony-one/harmony/issues/4014

### DIFF
--- a/hmy/staking.go
+++ b/hmy/staking.go
@@ -326,12 +326,12 @@ func (hmy *Harmony) GetValidatorInformation(
 	lastBlockOfEpoch := shard.Schedule.EpochLastBlock(hmy.BeaconChain.CurrentBlock().Header().Epoch().Uint64())
 
 	computed.BlocksLeftInEpoch = lastBlockOfEpoch - hmy.BeaconChain.CurrentBlock().Header().Number().Uint64()
-	computed.Block = hmy.BeaconChain.CurrentBlock().Header().Number().Uint64()
-	computed.Epoch = hmy.BeaconChain.CurrentBlock().Header().Epoch().Uint64()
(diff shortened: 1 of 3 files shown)


staking/types/validator.go
```diff
@@ -109,6 +109,8 @@ type Computed struct {
 	BlocksLeftInEpoch uint64      `json:"-"`
 	Percentage        numeric.Dec `json:"current-epoch-signing-percentage"`
 	IsBelowThreshold  bool        `json:"-"`
+	Epoch             uint64      `json:"epoch"`
+	Block             uint64      `json:"block"`
```
MaxMustermann2, line 113: In my opinion, it is probably better to change either of the following structures:
https://github.com/harmony-one/harmony/blob/5163c682911deb60566c15bccad5561a6644030a/staking/types/validator.go#L144-L160

MaxMustermann2 (approved):

hmy/staking.go
```diff
@@ -330,6 +330,8 @@ func (hmy *Harmony) GetValidatorInformation(
 	if defaultReply.CurrentlyInCommittee {
 		defaultReply.Performance = &staking.CurrentEpochPerformance{
 			CurrentSigningPercentage: *computed,
+			Epoch:                    hmy.BeaconChain.CurrentBlock().Header().Number().Uint64(),
+			Block:                    hmy.BeaconChain.CurrentBlock().Header().Epoch().Uint64(),
```
Frozen, line 334: There is a small chance that block could change from previous call and getting twice from mutex doesn't make it faster.

Frozen (approved):

PR #4097, opened 2022-03-14
feat: Prune validator rewards in archival node...

MaxMustermann2:
...before aggregate epoch. Since such rewards were given out every block
originally, this led to an explosion in state size. By not writing
validator wrappers to disk each time the reward is modified, we save the
space used by archival nodes.

In other words, this change prevents the writing of validator wrappers
to disk for blocks that are part of pre staking epoch and onwards, until
the start of aggregate reward epoch. However, rewards are written to
disk at every 64th block (63, 127,...., 32767). In case the binary is
shut down early (and gracefully), the latest set of validator wrappers
is written to disk for preservation.

Lastly, the change adds a flag called "--prune.rewards", which when set
will iterate over the blocks saved in the archival node and attempt to
prune the reward set. Once done, it will exit with error code 1.

## Issue
#3843 
### Unit Test Coverage

Before (c055bc1fd869e6096d0111df8826dd6d64750e28)

```
?   	github.com/harmony-one/harmony/internal/configs/harmony	[no test files]
ok  	github.com/harmony-one/harmony/cmd/harmony	0.064s	coverage: 43.1% of statements
ok  	github.com/harmony-one/harmony/core	(cached)	coverage: 30.8% of statements
ok  	github.com/harmony-one/harmony/core/state	8.706s	coverage: 70.9% of statements
```

After:

```
?   	github.com/harmony-one/harmony/internal/configs/harmony	[no test files]
ok  	github.com/harmony-one/harmony/cmd/harmony	0.062s	coverage: 42.7% of statements
ok  	github.com/harmony-one/harmony/core	(cached)	coverage: 30.9% of statements
ok  	github.com/harmony-one/harmony/core/state	(cached)	coverage: 71.0% of statements
```

### Test/Run Logs
Not applicable

## Operational Checklist

No.

No.

No.


core/state/statedb.go
```diff
@@ -684,10 +684,16 @@ func (db *DB) GetRefund() uint64 {
...
 func (db *DB) Finalise(deleteEmptyObjects bool) {
+	db.finalise(deleteEmptyObjects, false)
+}
+
+func (db *DB) finalise(deleteEmptyObjects bool, skipValidators bool) {
 	// Commit validator changes in cache to stateObjects
 	// TODO: remove validator cache after commit
-	for addr, wrapper := range db.stateValidators {
-		db.UpdateValidatorWrapper(addr, wrapper)
+	if !skipValidators {
```
rlan35, line 693: This seems to be off. We don't have to ignore validatorWrapper change in the state DB. We only want to not store the validatorWrapper history. By adding this condition to ignore validatorWrapper change in state calculation, it's actually modifying the structure of the latest state trie (thus changing the state root).

core/blockchain.go
```diff
@@ -1198,7 +1212,19 @@ func (bc *BlockChain) WriteBlockWithState(
...
 
 	// Commit state object changes to in-memory trie
-	root, err := state.Commit(bc.chainConfig.IsS3(block.Epoch()))
+	var root common.Hash
+	skipValidators := bc.cacheConfig.Disabled && // archival node
+		!bc.chainConfig.IsAggregatedRewardEpoch(block.Epoch()) && // before aggregated reward epoch
+		block.NumberU64()%chain.RewardFrequency != chain.RewardFrequency-1 && // keep every 64th block
+		block.ShardID() == shard.BeaconChainShardID && // beacon chain
+		bc.chainConfig.IsPreStaking(block.Epoch()) // is actually staking
+	if skipValidators {
```
rlan35, line 1221: similar comments as the other one. We don't want to modify the state calculation logic when doing the state commitment for the latest block, that's going to cause problem (missing validatorWrapper in the latest state). The last state should always have all the information and the logic for it shouldn't change. The only thing we want to change is to remove the history state data for validatorWrappers.

rlan35: As a example: https://medium.com/codechain/ethereums-state-trie-pruning-45ea73ed2c78#:~:text=In%20state%20trie%20pruning%2C%20it,the%20DB%20are%20not%20pruned.    we don't want to modify those blue blocks, we just want to remove the red blocks.

core/state/statedb.go
MaxMustermann2: Thanks for explaining this to me today. I agree with your assessment, and I am working on a pruning mechanism that doesn't work on the latest block to avoid altering the state root; only on prior blocks. We don't expect this change to impact the state root of the old blocks, so nodes going for this strategy are eligible for participating in the syncing process of other nodes.

cmd/harmony/main.go
```diff
@@ -367,6 +369,59 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
...
+	if shouldPrune {
+		chain := currentNode.Beaconchain()
+		curNum := chain.CurrentBlock().NumberU64()
+		if curNum == 0 {
+			fmt.Println("No rewards to prune because we are at genesis block")
+			os.Exit(1)
+		}
+		chainConfig := chain.Config()
+		// max number to be last block of epoch "AggregatedRewardEpoch - 1"
+		var maxNum uint64
```
rlan35, line 384: Actually we want to prune to the last block because even with aggregated reward logic, validatorWrapper may still change by other actions such as delegation, undelegation, collect rewards and many more. So many history records may still exist after aggregated reward epoch, though much less than before.

core/blockchain.go
```diff
@@ -105,8 +106,9 @@ const (
 	blockAccumulatorCacheLimit         = 256
 	maxPendingSlashes                  = 512
 	// BlockChainVersion ensures that an incompatible database forces a resync from scratch.
-	BlockChainVersion = 3
-	pendingCLCacheKey = "pendingCLs"
+	BlockChainVersion            = 3
+	pendingCLCacheKey            = "pendingCLs"
+	PruneValidatorWrappersBefore = 32768
```
rlan35, line 111: please add a comment for this field, I don't quite get what's it's purpose.

core/blockchain.go
```diff
@@ -3167,3 +3187,52 @@ func isUnrecoverableErr(err error) bool {
...
+// re-creates the state, removes the validator wrappers and writes the state to disk
+func (bc *BlockChain) PruneValidatorWrappersAtBlock(number uint64, checked bool) error {
+	if number == 0 {
+		return errors.New("Cannot trim rewards for genesis block")
+	}
+	block := bc.GetBlockByNumber(number)
+	if block != nil {
+		clearValidatorWrappers := checked || // if already checked, skip
+			(!bc.chainConfig.IsAggregatedRewardEpoch(block.Epoch()) && // before aggregated reward epoch
+				number%chain.RewardFrequency != chain.RewardFrequency-1 && // keep every 64th block
```
rlan35, line 3201: We actually don't need to keep the wrapper every 64 block. We can remove all the history.

core/blockchain.go
```diff
@@ -3167,3 +3187,52 @@ func isUnrecoverableErr(err error) bool {
 	isTooManyOpenFiles := strings.Contains(err.Error(), tooManyOpenFilesErrStr)
 	return isLeveldbErr && !isTooManyOpenFiles
 }
+
+// PruneValidatorWrappersAtBlock takes a block number which must be already available
+// re-creates the state, removes the validator wrappers and writes the state to disk
+func (bc *BlockChain) PruneValidatorWrappersAtBlock(number uint64, checked bool) error {
```
rlan35, line 3193: what is "checked" mean?

core/blockchain.go
```diff
@@ -3167,3 +3187,52 @@ func isUnrecoverableErr(err error) bool {
...
+		if clearValidatorWrappers {
+			// Reconstruct the state using the parent state as reference point
+			parent := bc.GetBlock(block.ParentHash(), block.NumberU64()-1)
+			// on non-archival nodes this will mostly fail with missing trie node
+			// but for archival nodes, state is written for each block
+			state, err := state.New(parent.Root(), bc.stateCache)
+			if err != nil {
+				return err
+			}
+			_, _, _, _, _, _, newState, err := bc.processor.Process(
```
rlan35, line 3213: why is process() called? it's too computation-heavy.

core/state/statedb.go
```diff
@@ -765,7 +769,18 @@ func (db *DB) Commit(deleteEmptyObjects bool) (root common.Hash, err error) {
...
-				db.db.TrieDB().InsertBlob(common.BytesToHash(obj.CodeHash()), obj.code)
+				// for validators - clear the code if asked to
+				// this is better than deleting the state object because recovery then is difficult
+				// this way we save space while retaining functionality
+				// IMPORTANT - note the order here for short-circuiting must not be changed
+				// otherwise TestSnapshot2 will fail because obj.IsValidator
+				// will cache the result in obj.originStorage
+				if clearValidatorWrappers && obj.IsValidator(db.db) {
+					// emptyCode a misnomer, should be emptyCodeHash
+					db.db.TrieDB().InsertBlob(emptyCode, nil)
```
rlan35, line 780: This doesn't seem right. Inserting nil at emptyCodeHash is just nil'ing the trie node at emptyCodeHash. It's not deleting the validatorWrapper's data, right?

cmd/harmony/main.go
```diff
@@ -811,9 +840,7 @@ func setupSyncService(node *node.Node, host p2p.Host, hc harmonyconfig.HarmonyCo
 	node.RegisterService(service.Synchronize, s)
 
 	d := s.Downloaders.GetShardDownloader(node.Blockchain().ShardID())
-	if hc.Sync.Downloader {
```
rlan35, line 814: why remove this condition? This is the condition to enable stream sync (@JackyWYX can confirm.)

core/blockchain.go
```diff
@@ -3158,3 +3186,75 @@ func isUnrecoverableErr(err error) bool {
...
+					return errors.Wrapf(
+						err,
+						"Could not commit empty code for %s to disk",
+						address.Hex(),
+					)
+				}
+			}
+		}
+	}
+	//// The below can be used to check that the deletion works
```
rlan35, line 3242: Remove this before merge.

rlan35 (approved):

rlan35:
please work with @sophoah  to test this out on an archival node and see how it performs. Need to make sure that after the pruning, other things such as block sync won't have issue.

cmd/harmony/main.go
MaxMustermann2: Was removed accidentally. Restored it.

sophoah:
Test build v7432-v4.3.6-93-g9a135bab:
node started to the pruning work with the following logs
```
Cleared validator wrappers for block 13593 successfully
Cleared validator wrappers for block 13594 successfully
Cleared validator wrappers for block 13595 successfully
Cleared validator wrappers for block 13596 successfully
Cleared validator wrappers for block 13597 successfully
^C
```
pruning worked at a speed of 10k blocks per 15m starting from the earlier blocks, I estimated the pruning work to last for 25 days to clear 23.9M blocks. The machine had 6vcpu and 16G ram.

After discussing with @MaxMustermann2 we decided to stop the test and check if we can improve the speed

sophoah:
Additionally that PR will break https://api.hmny.io/#8ce13bda-e768-47b9-9dbe-193aba410b0a and https://api.hmny.io/#a229253f-ca76-4b9d-88f5-9fd96e40d583 for archival node that ran the pruning or was synced from 0, hence another PR will be required to update those RPC

rlan35 (approved):

core/blockchain.go
```diff
@@ -1261,6 +1286,16 @@ func (bc *BlockChain) WriteBlockWithState(
 		}
 	}
 
+	if bc.cacheConfig.Disabled { // archival node
+		laggedNumber := currentBlock.NumberU64() - ClearValidatorWrappersLag
```
rlan35, line 1290: Based on soph's data, it takes roughly 100ms to clear one block, which is a lot for continuous pruning every block. One optimization is to check whether the pruned block has any staking txn, or whether the block is the block reward distribution block (every 64 blocks), if none is true, then no pruning is needed (no validator wrapper change)

PR #4112, opened 2022-03-28
add script to check snapdb

peekpi:
Add a script to check that the snapdb(#4096) downloaded by users is complete.
Usage:
`snapdb_checker.sh hash ./harmony_db_0` calculate hash of given DB
`snapdb_checker.sh check ./harmony_db_0` check that the DB is complete.

When a new snapdb is published, the hash should be calculated and added to the DBHashList map of the script. The key of DBHashList is block height, and the value is hash.

### DIFF
--- a/scripts/snapdb_checker.sh
+++ b/scripts/snapdb_checker.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+
+set -e
+
+declare -a DBHashList
+DBHashList=(
+    [22816574]=f5894991e83cff54b215cda93f18bc813bb9bf9a45643c8751f09891aad1f091
+)
+
+DB_FILES="CURRENT MANIFEST* *.log *.ldb"
+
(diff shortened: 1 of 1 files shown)


scripts/snapdb_checker.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+set -e
+
+declare -a DBHashList
+DBHashList=(
+    [22816574]=f5894991e83cff54b215cda93f18bc813bb9bf9a45643c8751f09891aad1f091
+)
+
+DB_FILES="CURRENT MANIFEST* *.log" # *.ldb"
```
LeoHChen, line 10: why not check the ldb files?

peekpi: oh, it's a misstake, fixed.

LeoHChen (approved):

PR #4118, opened 2022-04-04
[core] fix: Use time-based ordering to avoid spam

MaxMustermann2:
Closes harmony-one/harmony#4113 by sorting transactions by time received
instead of using a hashmap based transaction ordering. This sorting is
on top of the gas price and nonce sorting already implemented.

Effectively, this allows the production of almost a deterministic order
of transaction ordering, as opposed to a heap-based hash map ordering
which is affected by Golang's internal operations. Consequently,
arbitrageurs, who spam the network with a view to exist around
arbitrag-able transactions, will have to focus on latency instead of
network spamming.

See also bnb-chain/bsc#269 and ethereum/go-ethereum#21358 for related
issues in other chains.

## Issue
#4113

### Unit Test Coverage

Before:

```
core/test: 19.4%
```

After:

```
core/test: 19.6%
```

### Test/Run Logs
Ran the following contract + brownie script to observe that time based ordering is dependent on latency and that all transactions go through as expected.
```Solidity

//SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.0;

contract SampleContract {
    address deployer;
    address setter;
    event SampleEvent(bool result);

    constructor() {
        deployer = msg.sender;
        setter = msg.sender;
    }

    function sampleFunction() public {
        bool result = false;
        if (setter == deployer) {
            setter = msg.sender;
            result = true;
        }
        emit SampleEvent(result);
    }
}
```
```python
from brownie import (
    accounts,
    SampleContract,
    web3
)
from pyhmy import (
    signing,
    transaction,
    util
)
import multiprocessing
import pytest
from eth_utils import to_checksum_address
import time

@pytest.fixture( scope = "module", autouse = True )
def setup():
    accounts.add( '1f84c95ac16e6a50f08d44c7bde7aff8742212fda6e4321fde48bf83bef266dc' )
    print( "Deploying contract" )
    SampleContract.deploy( { 'from': accounts[ 0 ] } )
    print( "Contract ready" )

def send_tx( signed_tx ):
    hash = transaction.send_raw_transaction( signed_tx.rawTransaction.hex() )
    return hash

def check_result( tx_hash ):
    receipt = web3.eth.wait_for_transaction_receipt( tx_hash )
    print( receipt[ 'logs' ][ -1 ][ 'data' ][ -1 ] )
    if receipt[ 'logs' ][ -1 ][ 'data' ][ -1 ] == '1':
[...]

PR #4119, opened 2022-04-05
increase block gas limit to 50m

rlan35:
Lift the temporary gas limit from 20m to 50m now that https://github.com/harmony-one/harmony/pull/4118 is deployed.


sophoah (approved):
LGTM

PR #4126, opened 2022-04-10
[core]: fix: Cap suggested gas price to 500 gwei

MaxMustermann2:
...and set suggested gas percentile to 40 instead of 60.

Recently, the recommended gas price requested via `eth_gasPrice`,
`hmy_gasPrice` or `hmyv2_gasPrice` has been fairly high (think 100k
gwei). This patch changes the maximum suggested gas price via these APIs
to be capped at 500 gwei. Users are still allowed to change the gas
price manually to be higher than this, for faster execution.

The occurence of these extremely high gas prices can be attributed to
the self reinforcing mechanism of this implementation. If some 20 blocks
are produced with extremely high gas prices for all transactions, the
API queries the transactions and returns the (erstwhile) 60th percentile
from them. This then results in a feedback loop for the following
blocks. This loop continues until transactions which do not use the
suggested gas price become part of the following blocks - which in
itself is unlikely, since transactions are sorted by gas price for
inclusion.

## Issue
Screenshot from a Discord user
unknown

### Unit Test Coverage

Before:

```
?   	github.com/harmony-one/harmony/hmy	[no test files]
```

After:

```
?   	github.com/harmony-one/harmony/hmy	[no test files]
```

### Test/Run Logs
Not needed.

## Operational Checklist

No.

No.

No.

### DIFF
--- a/hmy/gasprice.go
+++ b/hmy/gasprice.go
@@ -34,7 +34,7 @@ import (
 
 const sampleNumber = 3 // Number of transactions sampled in a block
 
-var DefaultMaxPrice = big.NewInt(1 * params.Ether)
+var DefaultMaxPrice = big.NewInt(5e11) // 500 gwei is the max suggested limit
 
 type GasPriceConfig struct {
 	Blocks     int
--- a/hmy/hmy.go
+++ b/hmy/hmy.go
@@ -151,9 +151,9 @@ func New(
 
 	// Setup gas price oracle
 	gpoParams := GasPriceConfig{
-		Blocks:     20,
-		Percentile: 60,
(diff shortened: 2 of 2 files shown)


rlan35 (approved):

MaxMustermann2:
https://github.com/harmony-one/harmony/commit/5acf652a6d20b0f81e50c750e4d957d90460d5e2 updates the commit message over https://github.com/harmony-one/harmony/commit/23fa4cf46c8cd1c0f46830327b726e2c577eac0e and squashes the commits into one. There are no code changes.

LeoHChen (approved):

PR #4141, opened 2022-04-20
Fix BaseStream buf io concurrent access.

Frozen:
## Issue

Bufio Reader/Writer is not thread safe, different goroutines can work with BaseStream at the same time.

## Operational Checklist

    **NO**

    **NO**

    **NO**

### DIFF
--- a/p2p/stream/types/stream.go
+++ b/p2p/stream/types/stream.go
@@ -25,8 +25,9 @@ type Stream interface {
 
 // BaseStream is the wrapper around
 type BaseStream struct {
-	raw libp2p_network.Stream
-	rw  *bufio.ReadWriter
+	raw      libp2p_network.Stream
+	reader   *bufio.Reader
+	readLock sync.Mutex
 
 	// parse protocol spec fields
 	spec     ProtoSpec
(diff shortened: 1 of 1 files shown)


LeoHChen (approved):

LeoHChen:
do you have a testcase that can trigger the concurrency issue?

p2p/stream/types/stream.go
```diff
@@ -89,18 +90,19 @@ func (st *BaseStream) WriteBytes(b []byte) (err error) {
...
-	}
-	bytesWriteCounter.Add(sizeBytes)
-	if _, err = st.rw.Write(b); err != nil {
-		return
+	size := sizeBytes + len(b)
+	message := make([]byte, size)
+	copy(message, intToBytes(len(b)))
+	copy(message[sizeBytes:], b)
+
+	if _, err = st.raw.Write(message); err != nil {
```
JackyWYX, line 98: Shall we also add the lock to Write method?

JackyWYX:
Thanks for pointing this out! Please also add a lock to write method to ensure the atomacity in

```
st.raw.Write
```

p2p/stream/types/stream.go
Frozen: Actually, we use write only once and no need for lock here, but libp2p seems so complicated, it's better just to add it and don't worry. Added.

Frozen:
> do you have a testcase that can trigger the concurrency issue?

If node is running with race detector -race, it will fail after some time with that issue.

rlan35 (approved):

PR #4146, opened 2022-04-23
[consensus] HIP-18: Allowlist for external leaders

peekpi:
## Issue
https://talk.harmony.one/t/on-hold-hip-18-allowlist-for-external-leaders/4329

## Test
TODO

## Operational Checklist

    **YES, the on-disk data structure is not modified, but the consensus algorithm is affected.**

    1. Pass the test in the localnet&testnet.
    2. Make sure all nodes contain this pr before HIP-18 epoch is reached.

    1. add some blskeys of external validators into the allowlist
    2. make sure all the keys are elected
    3. after HIP18 epoch, forces all hmy nodes not to propose new blocks(this may need to hacking some code or just turn off these nodes)
          - we can turn off a node that is the current leader, so the leader will change. then we restart the node and turn off the new leader node. repeat until an external node becomes the leader.
    5. waiting for the external validator to become the leader and propose new blocks
6. **How much minimum baking period after the last flag epoch should we allow on Pangaea before promotion onto mainnet?**
    \>=3 epochs

8. **What should node operators know about this planned change?**
    **All nodes must contain this pr before HIP-18 epoch is reached, otherwise, hard forks may occur when HIP-18 epoch is reached.**
    **node operators who have slot keys in the allowlist may need to strengthen the security of the key.**

13. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**
     NO


MaxMustermann2:
I was under the impression that this is on hold at this stage. Has that changed?

LeoHChen:
> I was under the impression that this is on hold at this stage. Has that changed?

It will be enabled after HIP-16

peekpi:
# Test On Localnet
## Config
- AllowlistEpoch: 2
- StakingEpoch: 1
- TwoSecondsEpoch: 0
- CrossLinkEpoch:1
- Modify the code let hmy don't propose new blocks after HIP-18.
specific config and code are here: https://gist.github.com/peekpi/f10d716ae304ecd6e5795f8faed91fe4
## External Validator Info
One Address: one109r0tns7av5sjew7a7fkekg4fs3pw0h76pp45e
BLS Keys:
   - 0522db7de624b49ac48c821234a56232495b3f3548acdbf0bc9381bdb83fc46913c341ef84a8b16d640a010428d93594
  - 16c574539574a27b493f62dd0d30e10dfc6942a78b4e0668d5470f678c8dac8ee497ae58802361c2ba028b4ab236a801
BLS keys above are in allowlist.
Validator and blskey created/staking on epoch 1, would be elected on epoch2.

## Block Header
### Shard0
First Block Number of HIP18 is 60.
Block 60, the first block of HIP18
```JSON
{
  "id": "1",
  "jsonrpc": "2.0",
  "result": {
    "difficulty": 0,
    "epoch": "0x2",
    "extraData": "0x",
    "gasLimit": "0x4c4b400",
    "gasUsed": "0x0",
    "hash": "0xb1cf7fcd152b3be989aed753caa7f51372057c93a78d9a3c4de57dcba806e7e6",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "miner": "one1pdv9lrdwl0rg5vglh4xtyrv3wjk3wsqket7zxy",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": 0,
    "number": "0x3c",
    "parentHash": "0xd44606d9185032deeec71c391bfbb32687c0d0ea7095e327b110269f35f81044",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "size": "0x446",
    "stakingTransactions": [],
    "stateRoot": "0x7d3adb754f58200ab1d48918c319a0849ea99183648298a0168ac531b956b397",
    "timestamp": "0x62791800",
    "transactions": [],
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "uncles": [],
    "viewID": "0x1ea4c6",
    "vrf": "0x9ac882334a43b57eea9dd9a1f5a67571a29a8209d13d7b1c37a2d429529ce4e3",
    "vrfProof": "0x3fa20c2d4ea473ec64845540c2708d9c20b9425c6a56f5ec333738e4441d95a8522ea222c0e640f0a9f77323c7981d1571e4dc7b6d0835b4bd8a08ac75ab203cd4e8ca6714d089e2497b5e31d5eea2afa30d77c1b424a78216a0b7e0051c1595"
  }
}
```
Block 61, the second Block of HIP18
```JSON
{
  "id": "1",
  "jsonrpc": "2.0",
  "result": {
    "difficulty": 0,
    "epoch": "0x2",
    "extraData": "0x",
    "gasLimit": "0x4c4b400",
    "gasUsed": "0x0",
    "hash": "0x48f322ffd2999b5e9215e4c69011e6bd84d5e8319953447d5996415e4d2b24ff",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "miner": "one109r0tns7av5sjew7a7fkekg4fs3pw0h76pp45e",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": 0,
    "number": "0x3d",
    "parentHash": "0xb1cf7fcd152b3be989aed753caa7f51372057c93a78d9a3c4de57dcba806e7e6",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "size": "0x322",
    "stakingTransactions": [],
    "stateRoot": "0x113b88e727df8270db265193d172065cc990f403eb450091f8cfbb98b2820c2f",
    "timestamp": "0x627918fd",
    "transactions": [],
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "uncles": [],
    "viewID": "0x1ea4cd",
    "vrf": "0xb116c54960659848efedb911fc97124b79f0ca6ae34177ec4c86f7601e701a3e",
    "vrfProof": "0x6e50fdb0fee155bb9c54cf52234d9e90e506cc324a846b102f3e2dc08c5da3c79892690f0fde974d886089d1a113ba165845d2efad350e037ee74e10cabbf34fa8d7df9cfb2e5792d191968253bb27a6e7b1b8e5e3135198702c9a02219c2611"
  }
}

```
### Shard1
First Block Number of HIP18 is 59.
Block 59, the first block of HIP18 
```JSON
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "difficulty": 0,
    "epoch": "0x2",
    "extraData": "0x",
    "gasLimit": "0x4c4b400",
    "gasUsed": "0x0",
    "hash": "0xf0b2d6428598969066baba5c3fba6aea2bb4440d2477b101223f82d2315aedd2",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "miner": "one1m6m0ll3q7ljdqgmth2t5j7dfe6stykucpj2nr5",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": 0,
    "number": "0x3b",
    "parentHash": "0x670632aede1911b12c039452729aea2d7a4f576b180c2cc5f0543fb84f3aea70",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "size": "0x323",
    "stakingTransactions": [],
    "stateRoot": "0x027f148eb3b7f6aaebdbe54dbcbcd36f46d7005efd12aa700f2f25cc2f16a26d",
    "timestamp": "0x62791804",
    "transactions": [],
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "uncles": [],
    "viewID": "0x1ea4c5",
    "vrf": "0x8b71cb928b0f90a8b9575e2c78297b6a89231b0849a5b230519f006ac6ea3722",
    "vrfProof": "0xfee34254e0afe747b08b69a759d9befdc3fd6ad53c96c9cbccff521950da4038c094fda227be77d40aa53654837d5918f20a5ceea6e501d4d29075c0fb317e26045a2cf327c003b39273c6ba93e398a6e3ddc1d6d3f9e977cf8097f8d4b55f0d"
  }
}
```
Block 60, the second block of HIP18
```JSON
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "difficulty": 0,
    "epoch": "0x2",
    "extraData": "0x",
    "gasLimit": "0x4c4b400",
    "gasUsed": "0x0",
    "hash": "0xd71e8b90b7d21a143857c9597adca192cb9ed25c6d83f91ebaf5667959870221",
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "miner": "one109r0tns7av5sjew7a7fkekg4fs3pw0h76pp45e",
    "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
    "nonce": 0,
    "number": "0x3c",
    "parentHash": "0xf0b2d6428598969066baba5c3fba6aea2bb4440d2477b101223f82d2315aedd2",
    "receiptsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "size": "0x322",
    "stakingTransactions": [],
    "stateRoot": "0x027f148eb3b7f6aaebdbe54dbcbcd36f46d7005efd12aa700f2f25cc2f16a26d",
    "timestamp": "0x627918fa",
    "transactions": [],
    "transactionsRoot": "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421",
    "uncles": [],
    "viewID": "0x1ea4cc",
    "vrf": "0xb37a46cdde9487a4b1e7e6d92a2bc8285af4cf845e1beca76d87a897cec2e395",
    "vrfProof": "0x6e3e81a92e3ea011945c02a6340dcaedff363a42a4e9a9f110e6d179bf83f60530f97b7e6af9973bf66c576b60c57f145acd5b6b51bd43799a98c071d12c170d6c87495a414ecb2d90bb57123061b2d5a06518de62497b7ca139a44bd79dcd8b"
  }
}

```

LeoHChen (approved):
LGTM, please post test logs on testnet.

sophoah:
@rlan35 @MaxMustermann2 can we your approval here as well ?

consensus/quorum/quorum.go
```diff
@@ -246,6 +250,41 @@ func (s *cIdentities) NthNextHmy(instance shardingconfig.Instance, pubKey *bls.P
...
+		found = true
+	}
+	numHmyNodes := instance.NumHarmonyOperatedNodesPerShard()
+	// sanity check to avoid out of bound access
+	if numHmyNodes <= 0 || numHmyNodes > len(s.publicKeys) {
+		numHmyNodes = len(s.publicKeys)
+	}
+	nth := idx
+	if idx >= numHmyNodes {
+		nth = sort.SearchInts(s.allowlistIndex, idx) + numHmyNodes
```
rlan35, line 268: Is allowlistIndex sorted?

consensus/quorum/quorum.go
```diff
@@ -246,6 +250,41 @@ func (s *cIdentities) NthNextHmy(instance shardingconfig.Instance, pubKey *bls.P
...
+	}
+
+	numExtNodes := instance.ExternalAllowlistLimit()
+	if numExtNodes > len(s.allowlistIndex) {
+		numExtNodes = len(s.allowlistIndex)
+	}
+
+	totalNodes := numHmyNodes + numExtNodes
+	// (totalNodes + next%totalNodes) can convert negitive 'next' to positive
```
rlan35, line 277: How can next be negative?

consensus/quorum/quorum.go
```diff
@@ -246,6 +250,41 @@ func (s *cIdentities) NthNextHmy(instance shardingconfig.Instance, pubKey *bls.P
...
+	nth = (nth + totalNodes + next%totalNodes) % totalNodes
```
rlan35, line 278: why not just = (nth + next) % totalNodes?

internal/configs/sharding/allowlist.go
```diff
@@ -0,0 +1,47 @@
...
+		if key, err := bls_cosi.WrapperPublicKeyFromString(pubkeys[i]); err != nil {
+			panic(fmt.Sprintf("invalid bls key: %d:%s error:%s", i, pubkeys[i], err.Error()))
+		} else {
+			blsPubkeys[i] = *key
+		}
+	}
+	return blsPubkeys
+}
+
+// each time to update the allowlist, it requires a hardfork.
```
MaxMustermann2, line 26: Is there a better way out of this? How frequently do you think we'd be updating the allow list?

internal/configs/sharding/allowlist.go
```diff
@@ -0,0 +1,47 @@
...
+var mainnetAllowlist = Allowlist{
+	MaxLimitPerShard: 0,
+	BLSPublicKeys:    _BLS([]string{}),
+}
+
+var testnetAllowlist = Allowlist{
+	MaxLimitPerShard: 4,
+	BLSPublicKeys: _BLS([]string{
+		"7915b9cbae9d675af510cb252362b80ae6d68a3684bbea203bc30d2f5fda25ffcedfa3cf2a6c1d3051469379920a418d",
```
MaxMustermann2, line 35: Please add comments of validator names and addresses.

How are different shard ids in the BLS keys handled?

internal/configs/sharding/allowlist.go
```diff
@@ -0,0 +1,47 @@
...
+
+	bls_cosi "github.com/harmony-one/harmony/crypto/bls"
+)
+
+type Allowlist struct {
+	MaxLimitPerShard int
+	BLSPublicKeys    []bls_cosi.PublicKeyWrapper
+}
+
+func _BLS(pubkeys []string) []bls_cosi.PublicKeyWrapper {
```
MaxMustermann2, line 14: Pardon my ignorance but why does this start with an underscore?

MaxMustermann2 (commented):
Since there is always a possibility of view change during epoch transition, does that case need to be handled separately here?

consensus/quorum/quorum.go
peekpi: yeah, sorted in `UpdateParticipants`. it's around line 307: `sort.Ints(s.allowlistIndex)`

consensus/quorum/quorum.go
peekpi: The comment for the function 'NthNextHmy()' says that `next` can be negative number.

consensus/quorum/quorum.go
peekpi: just for negative `next`. if 'next' can be negative, `(nth + next) % totalNodes` will be wrong result.

internal/configs/sharding/allowlist.go
peekpi: I don't think we should update the allow list frequently, maybe once a month?

internal/configs/sharding/allowlist.go
peekpi: It takes the intersection of the committees of a shard and AllowList.

peekpi: we can add the comments of the validator, but blskeys in the allowlist don't bind to validators. a validator who owns the key can remove it, and then add it to another validator.

internal/configs/sharding/allowlist.go
peekpi: removed

internal/configs/sharding/allowlist.go
MaxMustermann2: @sophoah Can you comment if this is a feasible strategy: a hard fork every month until the leader allow list is in effect?

internal/configs/sharding/allowlist.go
MaxMustermann2: > we can add the comments of the validator, but blskeys in the allowlist don't bind to validators. a validator who owns the key can remove it, and then add it to another validator.

Sure, but I don't think most validators are sharing BLS keys so it makes sense to at least have a record in the code.

> It takes the intersection of the committees of a shard and AllowList.
To confirm, does that mean the shardID of the allow list BLS public key is inferred from the committee members of each shard?

MaxMustermann2:
Reposting because it seems to have been missed:

Since there is always a possibility of view change during epoch transition, does that case need to be handled separately here?

peekpi:
> Reposting because it seems to have been missed:
> 
> 
> 
> Since there is always a possibility of view change during epoch transition, does that case need to be handled separately here?
> 
> 

I don’t think so. This PR just extends the list of leader candidates, and all the blskey in the allowlist is appended to the hmy nodes. the view change logic still keeps the same. If a view change occurs during epoch transition, the first key would be the new leader(same as before).

internal/configs/sharding/allowlist.go
peekpi: > To confirm, does that mean the shardID of the allow list BLS public key is inferred from the committee members of each shard?

yeah, the allowlist just records blskeys and does not care which shard it is. if a key is in the allowlist but not elected, it will not be in the committees, so it can’t be the leader.

peekpi: > To confirm, does that mean the shardID of the allow list BLS public key is inferred from the committee members of each shard?

and each shard has its committee, keys in the committee are in the same shard.

internal/configs/sharding/allowlist.go
peekpi: > Is there a better way out of this? How frequently do you think we'd be updating the allow list?

hardfork is required each time the list is updated. unless we put it into the stateDB and update it with transactions(via precompiled contract).

peekpi: Although we could write allowlist and its epoch to an external config file, it would still cause hardfork if it is updated. nodes use different config file may pick different leaders. and nodes start at early block must know the allowlist at that block.

sophoah: > s is a feasible strategy: a hard fork

to be honest I really don't like the idea of a new hardfork every time we update the list or doing that every month ... best is something that can allow the update of the list without doing a hardfork. 

ie maybe just skip or ignore a bls key if one that is the list doesn't answer. I think we are already doing that, 

The challenge based on my understand here is to have the same list in all validator node so things can flow in order .. that's the main reason of the hardfork for me

sophoah: once a month hardfork is not something we can really do ... hardfork is really difficult to coordinate.

MaxMustermann2: @rlan35 What do you think? Updating the allow list via a precompile should be less complicated to manage compared to hard forking for each allow list addition.

rlan35: The allowlist feature is only a temporary experiment to test out having external validators becoming leaders. The list shouldn't change frequently. Maybe at most 2-3 times before the true decentralized leader rotation is launched and this feature gets deprecated.

internal/configs/sharding/allowlist.go
```diff
@@ -0,0 +1,47 @@
...
+			panic(fmt.Sprintf("invalid bls key: %d:%s error:%s", i, pubkeys[i], err.Error()))
+		} else {
+			blsPubkeys[i] = *key
+		}
+	}
+	return blsPubkeys
+}
+
+// each time to update the allowlist, it requires a hardfork.
+var mainnetAllowlist = Allowlist{
```
rlan35, line 27: AllowList should have a clear versioning too just like the sharding config instance. Used allowList can be changed anymore. So put like mainnetAllowlistV1 for example.

rlan35 (approved):

rlan35:
Looks good. Please fix the travis, and have enough local test before pushing to testnet.

PR #4150, opened 2022-04-25
[downloader] fix explorer node get stuck when doing short range sync

JackyWYX:
## Issue

Explorer testnet nodes get stuck when doing stream sync. 

Disable the subscription sending feed when the node is explorer node to avoid a sync process stuck observed on testnet.

The issue is caused by sending an event to a subscription that is not subscribed. When the subscription is not subscribed by any handler, the notification service will be blocked at:

```
// hmy/downloader/downloader.go:254
d.evtDownloadStarted.Send(struct{}{})
```

Adding a boolean value to tell whether the subscription is subscribed before will resolve the issue (not not send the event when the subscription is not subscribed).

## Test

Tested on testnet node.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -811,7 +811,7 @@ func setupSyncService(node *node.Node, host p2p.Host, hc harmonyconfig.HarmonyCo
 	node.RegisterService(service.Synchronize, s)
 
 	d := s.Downloaders.GetShardDownloader(node.Blockchain().ShardID())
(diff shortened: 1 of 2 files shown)


rlan35:
Can you describe more about the problem? Not very clear on the description.

rlan35 (approved):

JackyWYX:
Hey @rlan35 , can we merge this PR? Or is there anything I should do before the merge?

PR #4165, opened 2022-05-08
[core] feat: Add cross shard transfer precompile

### LINKED ISSUE #4132: precompile for cross-shard token transfer
We'd like to have a new precompile to support cross-shard transaction, so that users can use metamask to call smart contract for cross-shard token transfer.

MaxMustermann2:
...for native tokens only, and not for smart contracts (which will be
added later). Resolves #4132 and requires a hard fork, currently
scheduled at EpochTBD on main and test nets, but epoch 1 (after
AcceptsCrossTx) on other nets.

The precompile is at address 249 and the method has the signature
`crossShardTransfer(uint256 amount, address to, uint32 toShardID)`. It
requires a transfer of the `amount` to the precompile address first, in
the form of `msg.value`, and cascades a cross shard receipt to the
network when executed. At this stage, it is blocked for use by smart
contracts by checking that there is no code at the address, and that the
address is not a validator. Documentation for Python integration tests
to follow.

## Issue
#4132

### Unit Test Coverage

Before:

```
?   	github.com/harmony-one/harmony/internal/params	[no test files]
ok  	github.com/harmony-one/harmony/accounts/abi	(cached)	coverage: 90.2% of statements
ok  	github.com/harmony-one/harmony/core	(cached)	coverage: 30.8% of statements
ok  	github.com/harmony-one/harmony/core/vm	(cached)	coverage: 42.4% of statements
ok  	github.com/harmony-one/harmony/staking	(cached)	coverage: 89.1% of statements
```

After:

```
?   	github.com/harmony-one/harmony/internal/params	[no test files]
ok  	github.com/harmony-one/harmony/accounts/abi	(cached)	coverage: 90.0% of statements
ok  	github.com/harmony-one/harmony/core	(cached)	coverage: 30.7% of statements
ok  	github.com/harmony-one/harmony/core/vm	(cached)	coverage: 43.6% of statements
ok  	github.com/harmony-one/harmony/staking	(cached)	coverage: 87.5% of statements
```

### Test/Run Logs
```
Transaction sent: 0xf05827de42a3df6081b372f37a6d53e7b6de2ef0922e8b0096f4112de88603bd
  Gas price: 30.0 gwei   Gas limit: 50000   Nonce: 0
  CrossShardXfer.crossShardTransfer confirmed   Block: 12   Gas used: 42788 (85.58%)
```

### Integration Tests Code
```python
from brownie import network, Contract, accounts
from pyhmy import account, blockchain, transaction
from web3 import Web3
import time

abi = [
	  {
	    "inputs": [
		  {
	        "internalType": "uint256",
	        "name": "value",
	        "type": "uint256"
	      },
		  {
	        "internalType": "address",
	        "name": "to",
	        "type": "address"
	      },
		  {
	        "internalType": "uint64",
	        "name": "toShardID",
	        "type": "uint32"
	      }
	    ],
	    "name": "crossShardTransfer",
	    "outputs": [],
	    "stateMutability": "payable",
	    "type": "function"
	  }
]
timeout = 20

network.connect( "hmyLocal" )
precompile = Contract.from_abi( "CrossShardXfer", "0x00000000000000000000000000000000000000F9", abi )
pk = "1f84c95ac16e6a50f08d44c7bde7aff8742212fda6e4321fde48bf83bef266dc"
accounts.add( pk )
endpoints = [ 'http://localhost:9500', 'http://localhost:9598' ]
web3 = Web3( Web3.HTTPProvider( endpoints[ 0 ] ) )

while blockchain.get_current_epoch( endpoints[ 0 ] ) < 1:
    time.sleep( 5 )
for i, endpoint in enumerate( endpoints ):
    if blockchain.get_shard( endpoint ) != i:
        raise Exception( f'Shard {i} endpoint wrong' )
before = [ accounts[ 0 ].balance(), account.get_balance( accounts[ 0 ].address, endpoints[ 1 ] ) ]
value = web3.toWei( 100, 'ether' )
tx = precompile.crossShardTransfer( 
    value,
    accounts[ 0 ].address,
    1,
    {
        'from': accounts[ 0 ],
        'nonce': account.get_account_nonce( accounts[ 0 ].address, 'latest', endpoints[ 0 ] ),
        'gas_limit': 50000,
        'value': value,
    }
)
receipt = web3.eth.wait_for_transaction_receipt( tx.txid )
# get the cx receipt (use timeout so it can reach shard 1)
cx_receipt = None
start = time.time()
while time.time() - start <= timeout and cx_receipt is None:
    cx_receipt = transaction.get_cx_receipt_by_hash( tx.txid, endpoints[ 1 ] )
# wait upto 20s for it to be processed
start = time.time()
while time.time() - start <= timeout and \
    account.get_balance( accounts[ 0 ].address, endpoints[ 1 ] ) == before[ 1 ]:
        transaction.resend_cx_receipt( cx_receipt, endpoints[ 0 ] )
        time.sleep( 1 )
after = [ accounts[ 0 ].balance(), account.get_balance( accounts[ 0 ].address, endpoints[ 1 ] ) ]
assert before[ 0 ] - value - tx.gas_used * tx.gas_price == after[ 0 ]
assert before[ 1 ] + value == after[ 1 ]
```

## Operational Checklist

No.

No.

No.

### DIFF
--- a/accounts/abi/abi.go
+++ b/accounts/abi/abi.go
@@ -19,9 +19,11 @@ package abi
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
+	"math/big"
+
+	"github.com/pkg/errors"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
@@ -276,4 +278,36 @@ func UnpackRevert(data []byte) (string, error) {
 		return "", err
 	}
 	return unpacked[0].(string), nil
+}
+
+// ParseAddressFromKey pulls out the address value from a map with provided key,
+// andv alidates the data type for it
+func ParseAddressFromKey(args map[string]interface{}, key string) (common.Address, error) {
+	if address, ok := args[key].(common.Address); ok {
+		return address, nil
+	} else {
+		return common.Address{}, errors.Errorf("Cannot parse address from %v", args[key])
+	}
+}
+
+// ParseBigIntFromKey pulls out the *big.Int value from a map with provided key,
+// and validates the data type for it
+func ParseBigIntFromKey(args map[string]interface{}, key string) (*big.Int, error) {
+	bigInt, ok := args[key].(*big.Int)
+	if !ok {
+		return nil, errors.Errorf(
+			"Cannot parse BigInt from %v", args[key])
+	} else {
+		return bigInt, nil
+	}
+}
+
+// ParseUint32FromKey pulls out the uint64 value from a map with provided key,
+// and validates the data type for it
+func ParseUint32FromKey(args map[string]interface{}, key string) (uint32, error) {
+	if val, ok := args[key].(uint32); ok {
+		return val, nil
+	} else {
+		return 0, errors.Errorf("Cannot parse uint32 from %v", args[key])
+	}
 }
--- a/core/state_processor.go
+++ b/core/state_processor.go
@@ -299,9 +299,30 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
 	var cxReceipt *types.CXReceipt
 	// Do not create cxReceipt if EVM call failed
 	if txType == types.SubtractionOnly && !failedExe {
-		cxReceipt = &types.CXReceipt{TxHash: tx.Hash(), From: msg.From(), To: msg.To(), ShardID: tx.ShardID(), ToShardID: tx.ToShardID(), Amount: msg.Value()}
+		cxReceipt = &types.CXReceipt{
+			TxHash:    tx.Hash(),
+			From:      msg.From(),
+			To:        msg.To(),
+			ShardID:   tx.ShardID(),
+			ToShardID: tx.ToShardID(),
+			Amount:    msg.Value(),
+		}
+		if vmenv.CXReceipt != nil {
+			return nil, nil, nil, 0, errors.New("cannot have cross shard receipt via precompile and directly")
+		}
 	} else {
-		cxReceipt = nil
+		if !failedExe {
+			if vmenv.CXReceipt != nil {
+				cxReceipt = vmenv.CXReceipt
+				// this tx.Hash needs to be the "original" tx.Hash
+				// since, in effect, we have added
+				// support for cross shard txs
+				// to eth txs
+				cxReceipt.TxHash = tx.HashByType()
+			}
+		} else {
+			cxReceipt = nil
+		}
 	}
 
 	return receipt, cxReceipt, vmenv.StakeMsgs, result.UsedGas, err
--- a/core/vm/contracts_write.go
+++ b/core/vm/contracts_write.go
@@ -2,9 +2,13 @@ package vm
 
 import (
 	"errors"
+	"math/big"
+	"strings"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/harmony-one/harmony/accounts/abi"
+	"github.com/harmony-one/harmony/core/types"
 	"github.com/harmony-one/harmony/shard"
 	"github.com/harmony-one/harmony/staking"
 	stakingTypes "github.com/harmony-one/harmony/staking/types"
@@ -17,6 +21,19 @@ var WriteCapablePrecompiledContractsStaking = map[common.Address]WriteCapablePre
 	common.BytesToAddress([]byte{252}): &stakingPrecompile{},
 }
 
+var (
+	// reserve 250 for read only staking precompile and 251 for epoch
+	crossShardXferPrecompileAddress = common.BytesToAddress([]byte{249})
+)
+
+// WriteCapablePrecompiledContractsCrossXfer lists out the write capable precompiled contracts
+// which are available after the CrossShardXferPrecompileEpoch
+// It includes the staking precompile and the cross-shard transfer precompile
+var WriteCapablePrecompiledContractsCrossXfer = map[common.Address]WriteCapablePrecompiledContract{
+	crossShardXferPrecompileAddress:    &crossShardXferPrecompile{address: crossShardXferPrecompileAddress},
+	common.BytesToAddress([]byte{252}): &stakingPrecompile{},
+}
+
 // WriteCapablePrecompiledContract represents the interface for Native Go contracts
 // which are available as a precompile in the EVM
 // As with (read-only) PrecompiledContracts, these need a RequiredGas function
@@ -26,7 +43,7 @@ type WriteCapablePrecompiledContract interface {
 	// RequiredGas calculates the contract gas use
 	RequiredGas(evm *EVM, contract *Contract, input []byte) (uint64, error)
 	// use a different name from read-only contracts to be safe
-	RunWriteCapable(evm *EVM, contract *Contract, input []byte) ([]byte, error)
+	RunWriteCapable(evm *EVM, contract *Contract, input []byte, value *big.Int) ([]byte, error)
 }
 
 // RunWriteCapablePrecompiledContract runs and evaluates the output of a write capable precompiled contract.
@@ -36,6 +53,7 @@ func RunWriteCapablePrecompiledContract(
 	contract *Contract,
 	input []byte,
 	readOnly bool,
+	value *big.Int,
 ) ([]byte, error) {
 	// immediately error out if readOnly
 	if readOnly {
@@ -48,7 +66,7 @@ func RunWriteCapablePrecompiledContract(
 	if !contract.UseGas(gas) {
 		return nil, ErrOutOfGas
 	}
-	return p.RunWriteCapable(evm, contract, input)
+	return p.RunWriteCapable(evm, contract, input, value)
 }
 
 type stakingPrecompile struct{}
@@ -104,6 +122,7 @@ func (c *stakingPrecompile) RunWriteCapable(
 	evm *EVM,
 	contract *Contract,
 	input []byte,
+	value *big.Int,
 ) ([]byte, error) {
 	if evm.Context.ShardID != shard.BeaconChainShardID {
 		return nil, errors.New("Staking not supported on this shard")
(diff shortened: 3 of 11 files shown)


LeoHChen (commented):
LGTM, can you attach a testing log in localnet?

accounts/abi/abi.go
```diff
@@ -277,3 +279,35 @@ func UnpackRevert(data []byte) (string, error) {
 	}
 	return unpacked[0].(string), nil
 }
+
+// ParseAddressFromKey pulls out the address value from a map with provided key,
+// andv alidates the data type for it
```
rlan35, line 284: fix comment typo

MaxMustermann2:
> LGTM, how to support cross-shard resend if the current cross-shard receipt wasn't processed, or lost? There is a resendcx rpc call designed to handle this scenario. How do we support that in precompile?

The integration test in my message above includes resending the CX Receipt. 
```python
cx_receipt = transaction.get_cx_receipt_by_hash( tx.txid, endpoints[ 1 ] )
```

The CX Receipt is returned by `state_processor` similar to any other receipt.
https://github.com/harmony-one/harmony/blob/ec8d197f00aa51ce47c6403609373d2bef14f7e1/core/state_processor.go#L314-L322

It is generated in `contracts_write`.
https://github.com/harmony-one/harmony/blob/ec8d197f00aa51ce47c6403609373d2bef14f7e1/core/vm/contracts_write.go#L301-L307

core/state_processor.go
```diff
@@ -299,9 +299,30 @@ func ApplyTransaction(config *params.ChainConfig, bc ChainContext, author *commo
...
-		cxReceipt = &types.CXReceipt{TxHash: tx.Hash(), From: msg.From(), To: msg.To(), ShardID: tx.ShardID(), ToShardID: tx.ToShardID(), Amount: msg.Value()}
+		cxReceipt = &types.CXReceipt{
+			TxHash:    tx.Hash(),
+			From:      msg.From(),
+			To:        msg.To(),
+			ShardID:   tx.ShardID(),
+			ToShardID: tx.ToShardID(),
+			Amount:    msg.Value(),
+		}
+		if vmenv.CXReceipt != nil {
```
rlan35, line 310: nit, can move this above initialization of cxReceipt.

core/vm/contracts_write.go
```diff
@@ -150,3 +169,168 @@ func (c *stakingPrecompile) RunWriteCapable(
...
+// This method does not require any overflow checking as the input size gas costs
+// required for anything significant is so high it's impossible to pay for.
+func (c *crossShardXferPrecompile) RequiredGas(
+	evm *EVM,
+	contract *Contract,
+	input []byte,
+) (uint64, error) {
+	// we just charge the intrinsic gas here
+	// using data that includes: fromAddress, toAddress, fromShardID, toShardID, value
+	var payload []byte = make([]byte, 0)
```
rlan35, line 230: The gas charged on normal cross-shard transaction is just the gas for a normal token transfer. I think we can stick with it.

core/vm/contracts_write.go
```diff
@@ -150,3 +169,168 @@ func (c *stakingPrecompile) RunWriteCapable(
...
+	fromAddress, toAddress, fromShardID, toShardID, value, err := parseCrossShardXferData(evm, contract, input)
+	if err != nil {
+		return nil, err
+	}
+	// validate not a contract
+	if len(evm.StateDB.GetCode(fromAddress)) > 0 && !evm.IsValidator(evm.StateDB, fromAddress) {
+		return nil, errors.New("cross shard xfer not yet implemented for contracts")
+	}
+	// validate not a contract
+	if len(evm.StateDB.GetCode(toAddress)) > 0 && !evm.IsValidator(evm.StateDB, toAddress) {
```
rlan35, line 279: We don't have the restriction on normal cross-shard txn that the destination address can't be a contract. it can be any address for toAddress. fromAddress has to be EOA address since it's sent from EOA address.

core/vm/evm.go
```diff
@@ -314,8 +320,10 @@ func (evm *EVM) Call(caller ContractRef, addr common.Address, input []byte, gas
 			precompiles = PrecompiledContractsStaking
 			writeCapablePrecompiles = WriteCapablePrecompiledContractsStaking
 		}
-
-		if writeCapablePrecompiles[addr] == nil && precompiles[addr] == nil && evm.ChainConfig().IsS3(evm.EpochNumber) && value.Sign() == 0 {
+		if evm.chainRules.IsCrossShardXferPrecompile {
+			writeCapablePrecompiles = WriteCapablePrecompiledContractsCrossXfer
+		}
+		if (len(writeCapablePrecompiles) == 0 || writeCapablePrecompiles[addr] == nil) && precompiles[addr] == nil && evm.ChainConfig().IsS3(evm.EpochNumber) && value.Sign() == 0 {
```
rlan35, line 326: why add a len() check in the condition?

rlan35:
So the caller (EOA or a smart contract) of the precompiled contract must send the same amount of ONE token to the contract as in the parameter. Then the token is removed from the precompiled contract, right?

core/vm/evm.go
MaxMustermann2: The code prior to this change was initializing `writeCapablePrecompiles` with `make`. Now it is no longer initialized, so this check is needed to avoid a `nil` error.
https://github.com/harmony-one/harmony/blob/ec8d197f00aa51ce47c6403609373d2bef14f7e1/core/vm/evm.go#L306

MaxMustermann2:
> So the caller (EOA or a smart contract) of the precompiled contract must send the same amount of ONE token to the contract as in the parameter. Then the token is removed from the precompiled contract, right?
Yes, but note that the caller is not allowed to be a smart contract in this case.

core/vm/contracts_write.go
MaxMustermann2: The extra gas is being charged to prevent smart contracts from allowing EOAs to subsidize these transfers via `delegatecall`. The same was tested out in the staking precompile.
https://github.com/MaxMustermann2/harmony-staking-precompiles/blob/28c9f0761c3bd3cda5ee8dc9e42d4dd3ecd75e92/contracts/example/TrySubsidizeContract.sol#L29

A smart contract which delegate calls 0xF9 multiple times on behalf of an EOA will be able to save the EOA the base gas of 21000 per call and the system will just charge them for the overall data + 21000 only once.

rlan35: sounds good.

core/vm/contracts_write_test.go
```diff
@@ -211,3 +199,92 @@ var StakingPrecompileTests = []writeCapablePrecompileTest{
...
+
+func TestCrossShardXferPrecompile(t *testing.T) {
+	for _, test := range CrossShardXferPrecompileTests {
+		testCrossShardXferPrecompile(test, t)
+	}
+}
+
+var CrossShardXferPrecompileTests = []writeCapablePrecompileTest{
+	{
+		input:    []byte{40, 72, 8, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 5, 107, 199, 94, 45, 99, 16, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 165, 36, 21, 19, 218, 159, 68, 99, 241, 212, 135, 75, 84, 141, 251, 172, 41, 217, 31, 52, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1},
```
rlan35, line 259: does the test cases cover invalid values that could cause errors or failures? We need to make sure the new code won't cause evm or system fault.

rlan35 (approved):
Overall, looks good. Please add enough tests to make sure no corner cases exist that causes any failure.

core/vm/contracts_write.go
```diff
@@ -150,3 +163,163 @@ func (c *stakingPrecompile) RunWriteCapable(
...
+		return nil, errors.New("argument value and msg.value not equal")
+	}
+	// now do the actual transfer
+	// step 1 -> remove funds from the precompile address
+	evm.Transfer(evm.StateDB, contract.Address(), toAddress, value, types.SubtractionOnly)
+	// step 2 -> make a cross link
+	// note that the transaction hash is added by state_processor.go to this receipt
+	// and that the receiving shard does not care about the `From` but we use the original
+	// instead of the precompile address for consistency
+	evm.CXReceipt = &types.CXReceipt{
```
rlan35, line 290: what happens if the code calls the precompile multiple times in the same transaction? We probably need a list of CXReceipt, right? Or is there anyway to prevent repeated call to the precompile in the same txn?

MaxMustermann2: > Or is there anyway to prevent repeated call to the precompile in the same txn?

Aren't we already doing that by checking that the `From` address is not a contract?

MaxMustermann2:
Results of two integration tests are below. The first is for ensuring that the precompile is inaccessible to contracts, and the second is to ensure that the precompile cannot be used multiple times in one transaction. The second one is impossible due to Solidity's `delegatecall` not accepting a `value` parameter _and_ thus not allowing a split of `msg.value` across two different precompile calls.

```Solidity
//SPDX-License-Identifier: Unlicense
pragma solidity ^0.8.0;

abstract contract CrossShardTransferPrecompile {
  function crossShardTransfer(uint256 value, address to, uint32 toShardID) public virtual;
}

contract SampleContract {
  event UsefulEvent(bool success);

  function acceptMoney() public payable {
  }

  function contractCrossShardTransfer(uint256 value) public returns (uint256 result) {
    bytes memory encodedInput = abi.encodeWithSelector(
      CrossShardTransferPrecompile.crossShardTransfer.selector,
      value,
      address(this),
      1);
    assembly {
      result := call(
        25000,
        0xF9,
        value,
        add(encodedInput, 32),
        mload(encodedInput),
        mload(0x40),
        0x20
      )
    }
    emit UsefulEvent(result != 0);
  }

  // first case where we make 2 msg.value requests
  // the former goes through but the latter fails
  // due to insufficient balance
  function delegatedCrossShardTransfer1() public payable returns (bool success) {
    bytes memory encodedInput = abi.encodeWithSelector(
      CrossShardTransferPrecompile.crossShardTransfer.selector,
      msg.value,
      msg.sender,
      1);
    success = true;
    uint256 result;
    for(uint256 i = 0; i < 2; i++) {
      assembly {
        result := delegatecall(
          25000,
          0xF9,
          add(encodedInput, 32),
          mload(encodedInput),
          mload(0x40),
          0x20
        )
      }
      success = success && (result != 0);
    }
    emit UsefulEvent(success);
  }

  // second case where we make 2 (msg.value / 2) requests
  // neither goes through, since the precompile receives
  // 2 calls of msg.value (which we cannot change)
  function delegatedCrossShardTransfer2() public payable returns (bool success) {
    bytes memory encodedInput = abi.encodeWithSelector(
      CrossShardTransferPrecompile.crossShardTransfer.selector,
      msg.value / 2,
      msg.sender,
      1);
    success = true;
    uint256 result;
    for(uint256 i = 0; i < 2; i++) {
      assembly {
        result := delegatecall(
          25000,
          0xF9,
          add(encodedInput, 32),
          mload(encodedInput),
          mload(0x40),
          0x20
        )
      }
      success = success && (result != 0);
    }
    emit UsefulEvent(success);
  }
  
}
``` 
#### Contract deployment and environment set up
```python
$ brownie console --network hmyLocal
>>> from pyhmy import account, blockchain, transaction
>>> import time
>>> timeout = 20
>>> pk = "1f84c95ac16e6a50f08d44c7bde7aff8742212fda6e4321fde48bf83bef266dc"
>>> accounts.add( pk )
<LocalAccount '0xA5241513DA9F4463F1d4874b548dFBAC29D91f34'>
>>> endpoints = [ 'http://localhost:9500', 'http://localhost:9598' ]
>>> while blockchain.get_current_epoch( endpoints[ 0 ] ) < 1:
...     time.sleep( 1 )
...
>>> for i, endpoint in enumerate( endpoints ):
...     if blockchain.get_shard( endpoint ) != i:
...         raise Exception( f'Shard {i} endpoint wrong' )
...
>>> value = web3.toWei( 5, 'ether' )
>>> contract = SampleContract.deploy( { 'from': accounts[ 0 ] } )
Transaction sent: 0x049f75895424138ad97e3310ab924804e8d47bed582970df55980ba3cea3bb59
  Gas price: 30.0 gwei   Gas limit: 279360   Nonce: 9
  SampleContract.constructor confirmed   Block: 804   Gas used: 253964 (90.91%)
  SampleContract deployed at: 0x65A4191e03E681a35F0B524fbf3ac9323A17F545
```

#### Test for contract cross shard transfer
This transaction always fails, even when the `require` is removed.
```python
>>> contract.acceptMoney( { 'from': accounts[ 0 ], 'value': value } )
Transaction sent: 0x19534a31adb82fb97f318bcd717d5566c1fb6dfc82f1de495b60fabb7724265e
  Gas price: 30.0 gwei   Gas limit: 23300   Nonce: 10
  SampleContract.acceptMoney confirmed   Block: 813   Gas used: 21182 (90.91%)

<Transaction '0x19534a31adb82fb97f318bcd717d5566c1fb6dfc82f1de495b60fabb7724265e'>
>>> contract.balance()
5000000000000000000
>>> account.get_balance( contract.address, endpoints[ 1 ] )
0
>>> contract.contractCrossShardTransfer( contract.balance(), { 'from': accounts[ 0 ] } )
  File "<console>", line 1, in <module>
  File "brownie/network/contract.py", line 1693, in __call__
    return self.transact(*args)
  File "brownie/network/contract.py", line 1566, in transact
    return tx["from"].transfer(
  File "brownie/network/account.py", line 642, in transfer
    receipt, exc = self._make_transaction(
  File "brownie/network/account.py", line 725, in _make_transaction
    raise VirtualMachineError(e) from None
  File "brownie/exceptions.py", line 121, in __init__
    raise ValueError(str(exc)) from None
ValueError: Gas estimation failed: 'execution reverted: Failure'. This transaction will likely revert. If you wish to broadcast, you must set the gas limit manually.
>>> contract.contractCrossShardTransfer( contract.balance(), { 'from': accounts[ 0 ], 'gas_limit': 50000, 'allow_revert': True } )
Transaction sent: 0xa00e05e87e9e457429a3ac2b194a7014b0546676506416aa9fe96b15ff3c014f
  Gas price: 30.0 gwei   Gas limit: 50000   Nonce: 11
  SampleContract.contractCrossShardTransfer confirmed (reverted)   Block: 844   Gas used: 50000 (100.00%)

<Transaction '0xa00e05e87e9e457429a3ac2b194a7014b0546676506416aa9fe96b15ff3c014f'>
>>> contract.balance()
5000000000000000000
>>> account.get_balance( contract.address, endpoints[ 1 ] )
0
```

Within Solidity, it is impossible to change the cascaded value of `delegatecall` since it does not take `value` as a parameter. The below two tests demonstrate its use regardless.
### Test for multiple calls (via `delegatecall` on an EOA) using `msg.value`
In this case, _if_ the `require` statement is removed from Solidity the first cross shard transfer goes through with `msg.value`. The second one fails, since all of `msg.value` is exhausted by that point. With the `require` statement, however, everything is reverted to the original state as demonstrated below.
```python
$ brownie console --network hmyLocal
>>> accounts.add()
>>> accounts[ 0 ].transfer( accounts[ 1 ], value * 25 )
Transaction sent: 0x7718eaee65c977897af74937948c015b51e0945834c078a37c42e3b5cd43bd96
  Gas price: 30.0 gwei   Gas limit: 21000   Nonce: 14
  Transaction confirmed   Block: 1040   Gas used: 21000 (100.00%)

<Transaction '0x7718eaee65c977897af74937948c015b51e0945834c078a37c42e3b5cd43bd96'>
>>> accounts[ 1 ].balance()
125000000000000000000
>>> account.get_balance( accounts[ 1 ].address, endpoints[ 1 ] )
0
>>> tx = contract.delegatedCrossShardTransfer1( { 'from': accounts[ 1 ], 'value': value } )
  File "<console>", line 1, in <module>
  File "brownie/network/contract.py", line 1693, in __call__
    return self.transact(*args)
  File "brownie/network/contract.py", line 1566, in transact
    return tx["from"].transfer(
  File "brownie/network/account.py", line 642, in transfer
    receipt, exc = self._make_transaction(
  File "brownie/network/account.py", line 725, in _make_transaction
    raise VirtualMachineError(e) from None
  File "brownie/exceptions.py", line 121, in __init__
    raise ValueError(str(exc)) from None
ValueError: Gas estimation failed: 'execution reverted: Failure'. This transaction will likely revert. If you wish to broadcast, you must set the gas limit manually.
>>> tx = contract.delegatedCrossShardTransfer1( { 'from': accounts[ 1 ], 'value': value, 'allow_revert': True, 'gas_limit': 75000 } )
Transaction sent: 0x4efa5a263314e9307033539e5ec7e22b4029e6bb50b073e849b31ded2bba845f
  Gas price: 30.0 gwei   Gas limit: 75000   Nonce: 0
  SampleContract.delegatedCrossShardTransfer1 confirmed (reverted)   Block: 1045   Gas used: 49352 (65.80%)

>>> accounts[ 1 ].balance() + tx.gas_used * tx.gas_price
125000000000000000000
>>> account.get_balance( accounts[ 1 ].address, endpoints[ 1 ] )
0
```

### Test for multiple calls (via `delegatecall` on an EOA) using `msg.value / 2`
By encoding `msg.value / 2` in the method call, we can demonstrate that `msg.value` is directly cascaded to the precompile and cannot be changed. In this case, the method fails complaining about the `msg.value` and the `value` parameter being different.
```python
>>> accounts.add()
>>> accounts[ 0 ].transfer( accounts[ 2 ], value * 25 )
Transaction sent: 0x49fc37778353b0387fff18e8112ef7e20dd34cbd58a6216debe5cde517c9a087
  Gas price: 30.0 gwei   Gas limit: 21000   Nonce: 15
  Transaction confirmed   Block: 1195   Gas used: 21000 (100.00%)

<Transaction '0x49fc37778353b0387fff18e8112ef7e20dd34cbd58a6216debe5cde517c9a087'>
>>> accounts[ 2 ].balance()
125000000000000000000
>>> account.get_balance( accounts[ 2 ].address, endpoints[ 1 ] )
0
>>> tx = contract.delegatedCrossShardTransfer2( { 'from': accounts[ 2 ], 'value': value * 2 } )
  File "<console>", line 1, in <module>
  File "brownie/network/contract.py", line 1693, in __call__
    return self.transact(*args)
  File "brownie/network/contract.py", line 1566, in transact
    return tx["from"].transfer(
  File "brownie/network/account.py", line 642, in transfer
    receipt, exc = self._make_transaction(
  File "brownie/network/account.py", line 725, in _make_transaction
    raise VirtualMachineError(e) from None
  File "brownie/exceptions.py", line 121, in __init__
    raise ValueError(str(exc)) from None
ValueError: Gas estimation failed: 'execution reverted: Failure'. This transaction will likely revert. If you wish to broadcast, you must set the gas limit manually.
>>> tx = contract.delegatedCrossShardTransfer2( { 'from': accounts[ 2 ], 'value': value * 2, 'gas_limit': 75000, 'allow_revert': True } )
Transaction sent: 0x28a59e1b867737373390f862f7e9ec4f6fba41883c277d53de9c58fb41d81ded
  Gas price: 30.0 gwei   Gas limit: 75000   Nonce: 0
  SampleContract.delegatedCrossShardTransfer2 confirmed (reverted)   Block: 1347   Gas used: 74428 (99.24%)

>>> accounts[ 2 ].balance() + tx.gas_used * tx.gas_price
125000000000000000000
>>> account.get_balance( accounts[ 2 ].address, endpoints[ 1 ] )
0
```

internal/params/config.go
```diff
@@ -36,217 +36,227 @@ var once sync.Once
...
+		DataCopyFixEpoch:              big.NewInt(0),
+		IstanbulEpoch:                 big.NewInt(0),
+		ReceiptLogEpoch:               big.NewInt(0),
+		SHA3Epoch:                     big.NewInt(0),
+		HIP6And8Epoch:                 big.NewInt(0),
+		StakingPrecompileEpoch:        big.NewInt(2), // same as staking
+		ChainIdFixEpoch:               big.NewInt(0),
+		SlotsLimitedEpoch:             EpochTBD, // epoch to enable HIP-16
+		CrossShardXferPrecompileEpoch: big.NewInt(1),
+		AllowlistEpoch:                EpochTBD,
```
MaxMustermann2, line 146: @peekpi In `PangaeaChainConfig` and others below, `AllowlistEpoch` was unset. Was that intentional? I have added it here for uniformity. Let me know if we need to remove.

peekpi: yeah, setting it to EpochTBD is right.

PR #4100, opened 2022-03-19
[consensus] HIP-16: Enforce a 6% max keys per shard limit for each validator

peekpi:
## Issue
#4045 
the rules of #4045 only applied to external validators. 

## Test

Localnet: SUCCESS https://github.com/harmony-one/harmony/pull/4100#issuecomment-1107415219

## Operational Checklist

1. **Does this PR introduce backward-incompatible changes to the on-disk data structure and/or the over-the-wire protocol?**.

    **YES, the on-disk data structure is not modified, but the consensus algorithm is affected.**

2. **Describe the migration plan.**. For each flag epoch, describe what changes take place at the flag epoch, the anticipated interactions between upgraded/non-upgraded nodes, and any special operational considerations for the migration.
     1. Pass the test in the testnet.
     2. Make sure all nodes contain this pr before HIP-16 epoch is reached.

3. **Describe how the plan was tested.**
There are two main tests, one is that before HIP16-Epoch, upgraded nodes can work properly with the nodes that are not upgraded. The other is that all upgraded nodes can work properly after HIP16-Epoch.
     1. Assume that the maximum number of effective slots for each validator on the same shard is 15.
     2. **Test1:** Before HIP16-Epoch, assign more than 15 slots keys to a shard for an external validator.  Make sure all slots are elected.
     3.  Upgrade a node to include this pr.
     **Test2:** Confirm node is able to to synchronize and produce BINGO. Test at least 2 epochs before HIP16-Epoch.
     4. Then upgrade all nodes to include this pr before HIP16-Epoch.
     **Test3:** When reaching HIP16-Epoch, confirm the slots elected for the validator must be reduced to 15 (exceeded keys should have an effective stake of 0). All nodes must be synchronized properly. test at least 2 epochs after HIP16-Epoch.
     **Test4:** try to add more keys to the validators, should have error message

7. **How much minimum baking period after the last flag epoch should we allow on Pangaea before promotion onto mainnet?**
     **\>= 3 epochs**

8. **What should node operators know about this planned change?**
     ** All nodes must contain this pr before HIP-16 epoch is reached, otherwise, hard forks may occur when HIP-16 epoch is reached.**

9. **Does the existing `node.sh` continue to work with this change?**
    **no impact on this feature.**

10. **What should node operators know about this change?**
    **Avoid assigning more than 15(=250*6% Mainnet) slots BLS keys to the same shard. The slots beyond that will not participate in the election.**

[...]

PR #4168, opened 2022-05-13
set HIP16-epoch of mainnet to 999

peekpi:
Launch HIP16 on mainnet at epoch 999. it's around '27 May 2022 09:41:02 UTC' with 2s block time. 
previous pr: #4100


rlan35 (approved):

sophoah (approved):
LGTM

PR #4170, opened 2022-05-13
Feature 4052 epoch sync with crosslinks.

Frozen:
### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**


api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+	beaconChain        *core.BlockChain
+	selfip             string
+	selfport           string
+	selfPeerHash       [20]byte // hash of ip and address combination
+	commonBlocks       map[int]*types.Block
+	lastMileBlocks     []*types.Block // last mile blocks to catch up with the consensus
+	syncConfig         *SyncConfig
+	isExplorer         bool
+	stateSyncTaskQueue *queue.Queue
+	syncMux            sync.Mutex
```
rlan35, line 30: Is syncMux and lastMileMux needed for epoch sync?

rlan35, line 30: Is syncMux and lastMileMux needed for epoch sync?

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+}
+
+// isInSync query the remote DNS node for the latest height to check what is the current
+// sync status
+func (ss *EpochSync) isInSync(doubleCheck bool) SyncCheckResult {
+	if ss.syncConfig == nil {
+		return SyncCheckResult{} // If syncConfig is not instantiated, return not in sync
+	}
+	otherHeight1 := getMaxPeerHeight(ss.syncConfig)
+	curBlock := ss.beaconChain.CurrentBlock()
```
rlan35, line 52: From the beaconChain, you only have epoch blocks, right? The EpochLastBlock of a epoch block is the epoch block itself.

rlan35, line 52: From the beaconChain, you only have epoch blocks, right? The EpochLastBlock of a epoch block is the epoch block itself.

rlan35: What you need to check is whether the otherHeight is within the epoch that's immediately the next epoch of your latest epoch block. Say you have the epoch block of epoch N. If the otherHeight1 belongs to epoch N+1, then you are in sync, if it's > N+1, then you are out of sync (need to download the epoch block for epoch N+1 and so on).

rlan35: What you need to check is whether the otherHeight is within the epoch that's immediately the next epoch of your latest epoch block. Say you have the epoch block of epoch N. If the otherHeight1 belongs to epoch N+1, then you are in sync, if it's > N+1, then you are out of sync (need to download the epoch block for epoch N+1 and so on).

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+		curEpoch := block.Epoch()
+		for i := 0; i < int(SyncLoopBatchSize); i++ {
+			epochLastBlock := shard.Schedule.EpochLastBlock(curEpoch.Uint64())
+			if epochLastBlock > maxHeight {
+				break
+			}
+			if height != epochLastBlock {
+				heights = append(heights, epochLastBlock)
+			}
+			curEpoch = curEpoch.Add(curEpoch, common.Big1)
```
rlan35, line 134: This loop should stop after it processed the epoch before the epoch of the block at maxHeight, right? Say the epoch of block at maxHeight is N, then the loop should only process up to N-1.

api/service/legacysync/epoch_syncing.go
```diff
@@ -0,0 +1,240 @@
...
+		}
+
+		// Signature validation.
+		err = bc.Engine().VerifyHeaderSignature(bc, block.Header(), sig, bitmap)
+		if err != nil {
+			return errors.Wrap(err, "failed signature validation")
+		}
+
+		ep := block.Header().Epoch()
+		next := ep.Add(ep, common.Big1)
```
rlan35, line 213: Prefer to use the epoch within the shardState. If it doesn't exist, then use epoch + 1.

api/service/legacysync/helpers.go
```diff
@@ -0,0 +1,99 @@
...
+			}
+			lock.Unlock()
+		}()
+		return
+	})
+	wg.Wait()
+	return maxHeight
+}
+
+func createSyncConfig(syncConfig *SyncConfig, peers []p2p.Peer, isBeacon bool) (*SyncConfig, error) {
```
rlan35, line 53: Please refactor this so the code can be reused and not just a copy of the original code.

node/node_cross_link.go
```diff
@@ -120,6 +121,52 @@ func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
 		Len, _ := node.Blockchain().AddPendingCrossLinks(candidates)
 		utils.Logger().Debug().
 			Msgf("[ProcessingCrossLink] Add pending crosslinks,  total pending: %d", Len)
+	} else {
+		var crosslinks []types.CrossLink
```
rlan35, line 125: so the crosslink signal has the exactly the same data structure as cross links?

node/node_handler.go
```diff
@@ -189,27 +249,44 @@ func (node *Node) BroadcastCrossLink() {
 		"Construct and Broadcasting new crosslink to beacon chain groupID %s",
 		nodeconfig.NewGroupIDByShardID(shard.BeaconChainShardID),
 	)
-	headers := []*block.Header{}
-	lastLink, err := node.Beaconchain().ReadShardLastCrossLink(curBlock.ShardID())
+
+	for _, shardID := range []uint32{1, 2, 3} {
+		lastLink, err := node.Blockchain().ReadShardLastCrossLink(shardID)
```
rlan35, line 254: We probably don't need to send the whole crosslink data from beacon to shard, just need to send a message containing the number of the last continuous crosslink. Of course, the sender (leader or validators) can sign the message to prove the legitimacy of the data to other shards.

PR #4173, opened 2022-05-19
Feature/rpc filter

GheisMohammadi:
## Issue

We need to have a few flags to be able to enable/disable Eth and Staking RPCs.

## Test

Set the any of the flags below in configuration (under **RPCOpt**) can enable/disable related RPCs. These flags utilizes _**namespace**_ to differentiate between RPCs. All of these flags by default enable the RPCs.
  **EthRPCsEnabled:** to enable/disable Ethereum RPCs. (default: true)
  **StakingRPCsEnabled:** to enable/disable Staking RPCs. (default: true)
  **LegacyRPCsEnabled:** to enable/disable Legacy RPCs. (default: true)
In addition to these flags there is another option which you can set a RPC filter file in config file (under **RPCOpt**).
**RpcFilterFile** the path to the RPC filter toml file
 This file is a toml file which contains two fields: Allow and Deny. An example of this file is shown below:
```
 Allow = [ 
	"hmy_method1",
	"hmyv2_method?",
	"eth*",
	"hmy_getNetworkInfo",
	"^hmy_send[a-zA-Z]+"
]

Deny = [ 
	"*staking*",
	"eth_get*",
	"hmy_getStakingNetworkInfo",
	"exact:hmy_sendTx"
]
```
As it shown in example, this file accepts wildcard, regex, exact matching(case-sensitive) and case-insensitive matching. Any of the RPC name which pass the Allow and Deny Filters will be exposed. 
After add the Method filter file simply use a curl to see whether API is exposed or not:
```
curl -d '{
    "jsonrpc":"2.0",
    "method":"hmy_getStakingNetworkInfo",
    "params":[],
    "id":1
}' -H 'Content-Type:application/json' -X POST 'http://localhost:9599'
```

Before:

```
{"jsonrpc":"2.0","id":1,"result":{"circulating-supply":"40000000624.000000000000000000","epoch-last-block":9,"median-raw-stake":"0.000000000000000000","total-staking":0,"total-supply":"40000000624.000000000000000000"}}
```

After:

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32601,"message":"the method hmy_getStakingNetworkInfo does not exist/is not available"}}
```


rlan35 (approved):

rlan35:
generally looks good!

eth/rpc/method_filter.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	for _, filter := range filters {
+		if Match(filter, value) {
+			return true
+		}
+	}
+	return false
+}
+
+// MatchRegex -  finds whether the text matches/satisfies the regex pattern string.
```
MaxMustermann2, line 85: Please update the comment to say `Match`.

eth/rpc/method_filter.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+
+	if pattern == "*" {
+		return true
+	}
+	// Does only wildcard '*' match.
+	return deepMatchRune([]rune(name), []rune(pattern), true)
+}
+
+// Match -  finds whether the text matches/satisfies the pattern string.
```
MaxMustermann2, line 134: Please update the comment

rpc/rpc.go
```diff
@@ -71,30 +72,40 @@ func (n Version) Namespace() string {
...
+	var rmf rpc.RpcMethodFilter
+	rpcFilterFilePath := strings.Trim(rpcOpt.RpcFilterFile, " ")
+	if len(rpcFilterFilePath) > 0 {
+		if err := rmf.LoadRpcMethodFiltersFromFile(rpcFilterFilePath); err != nil {
+			return err
+		}
+	} else {
+		rmf.ExposeAll()
+	}
+	fmt.Println("start servers: ", rmf)
```
MaxMustermann2, line 88: Can this debugging line be removed?

test/deploy.sh
```diff
@@ -81,7 +81,7 @@ function launch_localnet() {
     i=$((i + 1))
 
     # Read config for i-th node form config file
-    IFS=' ' read -r ip port mode bls_key shard <<<"${line}"
+    IFS=' ' read -r ip port mode bls_key shard node_config <<<"${line}"
```
MaxMustermann2, line 84: What is this change and why is it part of this PR?

MaxMustermann2 (commented):
Generally looks good with a few specific comments.

test/deploy.sh
GheisMohammadi: This one allows to set a custom config file for each single node in local test net. So, we are able to enable/disable any RPC and test the RPC filter. By default it uses default configuration which all RPCs are enabled. Actually this change could be very helpful for other tests as well. 
If the config file is not set in local test net config file, then it will use the default values like how it works so far.
The test net config file is in this path: 
`github.com/harmony-one/harmony/test/configs/local-resharding.txt`
It includes a single line for each single node like this:
`127.0.0.1 9000 validator .hmy/65f55eb3052f9e9f632b2923be594ba77c55543f5c58ee1454b9cfd658d25e06373b0f7d42a19c84768139ea294f6204.key
127.0.0.1 9001 validator .hmy/40379eed79ed82bebfb4310894fd33b6a3f8413a78dc4d43b98d0adc9ef69f3285df05eaab9f2ce5f7227f8cb920e809.key
127.0.0.1 9002 validator .hmy/02c8ff0b88f313717bc3a627d2f8bb172ba3ad3bb9ba3ecb8eed4b7c878653d3d4faf769876c528b73f343967f74a917.key
...
`
So, if we add a config file path to the end of any line, that node will use that custom config file. An example is shown below:
`127.0.0.1 9000 validator .hmy/65f55eb3052f9e9f632b2923be594ba77c55543f5c58ee1454b9cfd658d25e06373b0f7d42a19c84768139ea294f6204.key null ./bin/node1_config.toml`
It makes local test net more flexible and we are able to feed different configs to each single node.

Frozen (approved):

Frozen:
Do we really need multiple ways of matching(exact/simple/wildcard)  when regex is enough for everything?

GheisMohammadi:
Actually, that's true, but I see it as extra options to make it easier to configure and test. Actually, sometimes using regex takes a little time to write the expression and analyze it. Just imagine a file with a couple of regex lines which you need to analyze which method could pass all of those filters. I think using a wildcard or exact match makes it simpler and  more handy and can improve the testing process and reduce the common configuration mistakes. An example of using wildcards can be seen in **.gitignore** files. To be honest, the idea of using filters is mostly about wildcards and I guess almost every scenario can be handled using wildcards or exact match and my suggestion is regex should be used only in complex filtering scenarios.

MaxMustermann2 (approved):

PR #4177, opened 2022-05-24
add new config to set local accounts from harmony config

GheisMohammadi:
## Issue

Need to be able to set the local accounts through harmony configuration.

## Test

We added a new option called **LocalAccountsFile** under **TxPool** to set a file path which includes the local account. The default path for local accounts file is: **_./.hmy/locals.txt_**. In order to test, just set the local accounts file path in configuration and run a local test net by custom configuration. Send a transaction using one of the local accounts. If the transaction is under priced, it should still go through.  

The local accounts file supports comment lines using "#". The comment helps us to make it more readable and it helps maintenance and debugging. An example of this file is shown below:

```
# Harmony Local Account
one1m6m0ll3q7ljdqgmth2t5j7dfe6stykucpj2nr5

# Binanace Withdrawal Account
one1fdv7u7rll9epgcqv9xxh9lhwq427nsqlr5wca5
```

### DIFF
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -130,6 +130,7 @@ var (
 		rosettaFixFileFlag,
 		tpBlacklistFileFlag,
 		legacyTPBlacklistFileFlag,
+		localAccountsFileFlag,
 	}
 
 	pprofFlags = []cli.Flag{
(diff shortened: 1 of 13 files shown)


cmd/harmony/flags.go
```diff
@@ -1048,6 +1054,11 @@ func applyTxPoolFlags(cmd *cobra.Command, config *harmonyconfig.HarmonyConfig) {
 	} else if cli.IsFlagChanged(cmd, legacyTPBlacklistFileFlag) {
 		config.TxPool.BlacklistFile = cli.GetStringFlagValue(cmd, legacyTPBlacklistFileFlag)
 	}
+	if cli.IsFlagChanged(cmd, localAccountsFileFlag) {
+		config.TxPool.LocalAccountsFile = cli.GetStringFlagValue(cmd, localAccountsFileFlag)
+	} else if cli.IsFlagChanged(cmd, localAccountsFileFlag) {
```
rlan35, line 1059: You don't need the "else if" part right? There is no legacy flag.

rlan35 (approved):

PR #4184, opened 2022-05-30
Fix race errors.

Frozen:
## Issue

Concurrent access problems that was caught by race detector.

    **NO**

    **NO**

    **NO**


consensus/consensus_fbft.go
```diff
@@ -0,0 +1,30 @@
+package consensus
+
+import "sync"
+
+type LockedFBFTPhase struct {
+	mu     sync.Mutex
+	phrase FBFTPhase
```
GheisMohammadi, line 7: - It might be a good idea to use RWMutex which allows to other threads also be able to read the value
- do you mean phrase or phase? is it typo?

consensus/consensus_fbft.go
```diff
@@ -0,0 +1,30 @@
...
+	return &LockedFBFTPhase{
+		phrase: initialPhrase,
+	}
+}
+
+func (a *LockedFBFTPhase) Set(phrase FBFTPhase) {
+	a.mu.Lock()
+	a.phrase = phrase
+	a.mu.Unlock()
+}
```
GheisMohammadi, line 20: you might want to consider using "defer"

consensus/consensus_fbft.go
```diff
@@ -0,0 +1,30 @@
...
+
+func (a *LockedFBFTPhase) Get() FBFTPhase {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	return a.phrase
+}
```
GheisMohammadi, line 26: It could be more efficient if we use RWMutex and then using RLock rather than Lock to allow other threads also use Get to read the value .

consensus/consensus_fbft_test.go
```diff
@@ -0,0 +1,15 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestLockedFBFTPhase(t *testing.T) {
+	s := NewLockedFBFTPhase(FBFTAnnounce)
+	require.Equal(t, FBFTAnnounce, s.Get())
+
+	s.Set(FBFTCommit)
+	require.Equal(t, FBFTCommit, s.Get())
+}
```
GheisMohammadi, line 15: It might be good if we test it by multiple threads rather than a normal Set and Get.

p2p/utils.go
```diff
@@ -0,0 +1,41 @@
+package p2p
+
+import "sync"
+
+type ConnectCallbacks struct {
+	cbs []ConnectCallback
+	mu  sync.Mutex
+}
```
GheisMohammadi, line 8: You might want to consider RWMutex here as well.

p2p/utils.go
```diff
@@ -0,0 +1,41 @@
...
+func (a *ConnectCallbacks) Add(cb ConnectCallback) {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	a.cbs = append(a.cbs, cb)
+}
+
+func (a *ConnectCallbacks) GetAll() []ConnectCallback {
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	out := make([]ConnectCallback, len(a.cbs))
```
GheisMohammadi, line 19: If you changed it to RWMutexm, then here we use RLock and RUnlock.

p2p/utils_test.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil
+	}
+
+	require.Equal(t, 0, len(cbs.GetAll()))
+
+	cbs.Add(fn)
+
+	require.Equal(t, 1, len(cbs.GetAll()))
+	require.Equal(t, reflect.ValueOf(fn).Pointer(), reflect.ValueOf(cbs.GetAll()[0]).Pointer())
+}
```
GheisMohammadi, line 23: have you considered to add multi-thread test for ConnectCallbacks?

MaxMustermann2 (commented):
Great work! I echo the opinion of @GheisMohammadi that multi-thread test cases should be added.

consensus/consensus_fbft.go
Frozen: FBFTPhase is a string, immutable struct, so only copy is possible to it. Copy takes less than 0.5ns, lock takes about  15ns. No reason to grant a shared access to smth that lasts less than lock getting time. It's a good idea to reduce lock getting time, and `sync.Mutex` is the most simplest and fastest.

consensus/consensus_fbft.go
Frozen: Be aware, that `defer` is not free, it has overhead. It is useful when we want to reduce code complexity or add panic catch possibility.  We are not interested in catching panics, and there is no complexity to reduce. I don't think that adding `defer` just in case is a good idea. But if code gets somehow more complex, i will add defer for sure.

consensus/consensus_fbft.go
Frozen: Mostly yes, but not for such simple case. Using rwlock here is over-engineering.

consensus/consensus_fbft_test.go
Frozen: We have
`
LockAcquire
`
`
Assign/Retrieve
`
`
LockRelease
`
What should be tested here by concurrent test? How we should prove that it was really concurrent?

Frozen: Besides this concurrent test, we have working code with no race errors. Is it not enough?

p2p/utils.go
Frozen: changed

consensus/consensus_fbft.go
rlan35: "phrase FBFTPhase"   phrase should be phase, right?

consensus/consensus_fbft.go
rlan35: If there is no downside or more complexity but only performance benefits for using RWMutex, we should use it.

Frozen: No benefits.
RWMutex is much more complex struct than Mutex, so it takes more time to lock. 
Benefits of shared read will be eaten by getting lock time.

consensus/consensus_fbft.go
Frozen: you mean `FBFTphase` or what?

MaxMustermann2 (approved):

rlan35 (approved):

PR #4193, opened 2022-06-08
[params] testnet hard fork for cross shard xfer

MaxMustermann2:
Set testnet epoch for cross shard transfer precompile (#4165) hard fork to be the same as that of HIP18 and `chainId` fix.


sophoah (approved):
LGTM

PR #4194, opened 2022-06-09
[rpc] use latest block as default params of EstimateGas

peekpi:
because harmony does not support the pending block, it should use the latest block as default.

### DIFF
--- a/rpc/transaction.go
+++ b/rpc/transaction.go
@@ -167,7 +167,7 @@ func (s *PublicTransactionService) EstimateGas(
 ) (hexutil.Uint64, error) {
 	timer := DoMetricRPCRequest(RpcEstimateGas)
 	defer DoRPCRequestDuration(RpcEstimateGas, timer)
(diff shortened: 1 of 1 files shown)


MaxMustermann2:
What do you mean by 'harmony does not support the pending block' ? Would it be possible (or better) to add support for pending block instead ?

peekpi:
> What do you mean by 'harmony does not support the pending block' ? Would it be possible (or better) to add support for pending block instead ?

I think pending block is only used in POW. in harmony chain, the block interval is 2s and only the leader can propose new blocks. so we can use the latest block instead of the pending block.

peekpi:
> What do you mean by 'harmony does not support the pending block' ? Would it be possible (or better) to add support for pending block instead ?

In pow, every node can mine a new block. I think pending block means the block the node is mining, the block has txs, state, but the difficulty requirement is not satisfied.

LeoHChen (approved):

PR #3872, opened 2021-09-07
[WIP] MMR-HardFork: add go-merklemountainrange and modified merkle proof logic

gupadhyaya:
This PR is a MMR hard fork PR that adds mmrRoot to block header, adds logic to compute the MMR tree every epoch, RPCs to fetch the MMR proof for any given transaction.

Features
* add receipt proof
* adding MMRRoot field to block header & cross-chain epoch
* add memdb and filedb mmr processing logic
* add GetProof rpc
* relayer rpcs for fetching full header
* adding block signers for rpc response, debug-only
* minor testing bls
* fix merge conflicts
* github.com/zmitton/go-merklemountainrange dependency
* minor fix
* moving mmr root compute/update logic to after the shard state is computed
* fix getting siblings bug
* adding index to mmr-proof and GetProof with respect to a block number

PR #4198, opened 2022-06-13
Mmr hardfork

peekpi:
## Issue

#3872 

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**


peekpi:
Other optimization points:
1. In the current scheme, MMR can only prove block hashes. To prove a receipt, it also needs to provide the block header, then gets receiptsRoot . If we add receiptsRoot to the MMR, we can omit the block header.
2. The verification of MMR proof is complicated. To optimize it, we can sort the child nodes when calculating the parent hash, so we don't need to know whether it's a left child or a right child when verifying the proof.

gupadhyaya:
@rlan35 @LeoHChen @MaxMustermann2 could you guys help with review of this PR?

internal/params/config.go
```diff
@@ -583,6 +594,11 @@ func (c *ChainConfig) IsPreStaking(epoch *big.Int) bool {
 	return isForked(c.PreStakingEpoch, epoch)
 }
 
+// IsCrossChain determines whether it is the epoch to support MMR root in the block header
+func (c *ChainConfig) IsCrossChain(epoch *big.Int) bool {
```
rlan35, line 598: Change to IsMMR

consensus/engine/consensus_engine.go
```diff
@@ -69,6 +69,8 @@ type ChainReader interface {
 
 	// ReadCommitSig read the commit sig of a given block number
 	ReadCommitSig(blockNum uint64) ([]byte, error)
+
```
rlan35, line 72: Add a comment here explaining this method.

consensus/engine/errors.go
```diff
@@ -40,4 +40,6 @@ var (
 
 	// ErrInvalidConsensusMessage is returned is the consensus message received is invalid
 	ErrInvalidConsensusMessage = errors.New("invalid consensus message")
+	// ErrInvalidConsensusMessage is returned is the consensus message received is invalid
```
rlan35, line 43: update comment. also  please fix to "is returned if".

core/blockchain_mmr.go
```diff
@@ -0,0 +1,242 @@
...
+	cacheLock     sync.RWMutex
+	MemMmrDB      *mmr.Mmr
+	MmrDBLock     sync.RWMutex
+	MmrCommitChan chan *mmr.Mmr
+	quit          chan struct{}
+	wait          sync.WaitGroup
+}
+
+func newMemMmrDB(epoch *big.Int) *mmr.Mmr {
+	leafLength := shard.Schedule.InstanceForEpoch(epoch).BlocksPerEpoch()
```
rlan35, line 32: does our MMR logic support the change of leafLength? what happens if the blocksPerEpoch change in the future?

hmy/hmy.go
```diff
@@ -116,6 +116,9 @@ type NodeAPI interface {
 	GetConfig() commonRPC.Config
 	ShutDown()
 	GetLastSigningPower() (float64, error)
+
+	// mmr APIs
+	// GetProof(txHash common.Hash, blockHash common.Hash, blockNumber, withRespectTo uint64) mmr.MmrProof
```
rlan35, line 121: why is this commented?

internal/configs/sharding/localnet.go
```diff
@@ -19,11 +19,11 @@ type localnetSchedule struct{}
 const (
 	localnetV1Epoch = 1
 
-	localnetEpochBlock1      = 10
-	localnetBlocksPerEpoch   = 5
-	localnetBlocksPerEpochV2 = 10
+	localnetEpochBlock1      = 5
```
rlan35, line 22: why change the localnet settings?

internal/shardchain/dbfactory.go
```diff
@@ -67,3 +122,8 @@ func (f *LDBShardFactory) NewChainDB(shardID uint32) (ethdb.Database, error) {
 		CacheSize: f.CacheSize,
 	})), nil
 }
+
+// MmrDB returns OpenMmrDB for the blockchain for given shard and epoch.
+func (f *LDBShardFactory) MmrDB(shardID uint32) (core.OpenMmrDB, error) {
```
rlan35, line 127: Is the reason to have a separate db for mmr data that it have potential conflict with the chain id entries?

rlan35:
> Other optimization points:
> 
> 1. In the current scheme, MMR can only prove block hashes. To prove a receipt, it also needs to provide the block header, then get receiptsRoot from the header. If we add receiptsRoot to the MMR, we can omit the block header.
> 2. The verification of MMR proof is complicated. To optimize it, we can sort the child nodes when calculating the parent hash, so we don't need to know whether it's a left child or a right child when verifying the proof. This is also what OpenZeppelin does.

Do these optimization need hard fork? If yes, we better do it now.

cmd/harmony/config_test.go
```diff
@@ -107,7 +107,8 @@ Version = "1.0.4"
 [WS]
   Enabled = true
   IP = "127.0.0.1"
-  Port = 9800`
+  Port = 9800
```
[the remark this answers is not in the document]
peekpi: the original pr(#3872) added some configuration here that is not needed now. will be modified to keep it the same as the main branch.

internal/shardchain/dbfactory.go
peekpi: no, the main reason is to use the database implementation of the MMR library directly, it simply stores/loads all nodes to/from a file. and also MMR file DB is optional, we can rebuild it from blocks of an epoch.

internal/configs/sharding/localnet.go
peekpi: commit by mistake. this is my local test configuration, to speed up the test.

hmy/hmy.go
peekpi: will remove it. it's the old MMR-proof RPC function.

core/blockchain_mmr.go
peekpi: yeah, MMR supports arbitrary leaf length. here `leafLength` is just used to optimize map memory allocation.

internal/params/config.go
peekpi: also rename `CrossChainEpoch` to `MMRHeaderEpoch`.

peekpi:
> > Other optimization points:
> > 
> > 1. In the current scheme, MMR can only prove block hashes. To prove a receipt, it also needs to provide the block header, then get receiptsRoot from the header. If we add receiptsRoot to the MMR, we can omit the block header.
> > 2. The verification of MMR proof is complicated. To optimize it, we can sort the child nodes when calculating the parent hash, so we don't need to know whether it's a left child or a right child when verifying the proof. This is also what OpenZeppelin does.
> 
> Do these optimization need hard fork? If yes, we better do it now.

yeah, it needs hard fork. this optimization also needs to do more tests, both go and smart contract. do you think we can create another pr to do it? we can just merge this pr, but do not enable the feature until the optimization is done.

gupadhyaya:
> Other optimization points:
> 
> 1. In the current scheme, MMR can only prove block hashes. To prove a receipt, it also needs to provide the block header, then get receiptsRoot from the header. If we add receiptsRoot to the MMR, we can omit the block header.
> 2. The verification of MMR proof is complicated. To optimize it, we can sort the child nodes when calculating the parent hash, so we don't need to know whether it's a left child or a right child when verifying the proof. This is also what OpenZeppelin does.

@peekpi since these optimizations break the MMRVerifier smart contract in the horizon repo, could you also make a PR there to correctly verify these changes? https://github.com/harmony-one/horizon/blob/main/contracts/lib/MMRVerifier.sol

internal/params/config.go
gupadhyaya: > also rename `CrossChainEpoch` to `MMRHeaderEpoch`.

recommend changing the naming to `MMRHardForkEpoch`

peekpi:
> @peekpi since these optimizations break the MMRVerifier smart contract in the horizon repo, could you also make a PR there to correctly verify these changes? https://github.com/harmony-one/horizon/blob/main/contracts/lib/MMRVerifier.sol

these optimizations are only my ideas and not have been done in this PR, but I think it can significantly reduce the gas cost. of course, I’ll also change the horizon contract to match the optimizations.

core/blockchain_mmr.go
MaxMustermann2: Will the (disk) database need to be re-created in the event the `leafLength` changes?

cmd/harmony/default.go
```diff
@@ -22,6 +22,7 @@ var defaultConfig = harmonyconfig.HarmonyConfig{
 		IsOffline:        false,
 		DataDir:          "./",
 		TraceEnable:      false,
+		MmrDBEnable:      false,
```
MaxMustermann2, line 25: Do we need a config migration for this? When would (external) validators want to use this flag?

core/blockchain_mmr.go
```diff
@@ -0,0 +1,242 @@
+package core
```
MaxMustermann2, line 1: Can I ask you to please add a few more comments above function definitions in this file?

cmd/harmony/default.go
peekpi: may not. most nodes do not need to use this flag, only 1 or 2 nodes need to enable this flag to provide MMR proof queries.

core/blockchain_mmr.go
peekpi: > Will the (disk) database need to be re-created in the event the `leafLength` changes?

don't need.

PR #4200, opened 2022-06-14
Increase voting power of harmony nodes in testnet to 0.7

peekpi:
## Issue
Testnet failed to propose the last block of epoch 75893. It stopped at 26481646, and the last block of epoch 75893 is 26481647. 
## Change
Increase voting power of harmony nodes in testnet to 0.7


MaxMustermann2 (approved):
`UpdateValidatorVotingPower` is called with next epoch's voting power so this makes sense.

rlan35 (approved):

MaxMustermann2 (approved):

PR #4203, opened 2022-06-16
[node] fix: always broadcast cross links

MaxMustermann2:
Continuation of #4199 to always broadcast cross links even when beacon blocks channel is empty.


node/node_syncing.go
```diff
@@ -232,15 +232,15 @@ func (node *Node) doBeaconSyncing() {
...
-						} else if node.Consensus.IsLeader() || rand.Intn(100) == 0 {
-							// Only leader or 1% of validators broadcast crosslink to avoid spamming p2p
-							// if beaconBlock.NumberU64() == node.Beaconchain().CurrentBlock().NumberU64() {
-							node.BroadcastCrossLinkFromShardsToBeacon()
-							// }
 						}
 					}
 				}
 			}
+			// regardless of whether we received any blocks in the beacon block channel
```
rlan35, line 239: This will never be executed right?

rlan35 (approved):

PR #4214, opened 2022-06-29
Blockchain interface.

Frozen:
## Motivation

So, we are going to have different implementations for shard and epoch chains. Current implementation for blockchain requires some hacks for epoch chain, which can be avoided by using it under interface. 

## Operational Checklist

    **NO**

    **NO**

    **NO**


MaxMustermann2 (approved):
Great, looking forward to the full implementation for shard chains!

rlan35 (approved):

PR #4222, opened 2022-07-04
Increase suggested gas and reduce block gas limit

Frozen:
Increase suggested gas to be 70 percentile and max suggested gas to 1000 gwei.

Reduce block gas limit to be 30M.

### DIFF
--- a/node/worker/worker.go
+++ b/node/worker/worker.go
@@ -69,9 +69,9 @@ func (w *Worker) CommitSortedTransactions(
 	coinbase common.Address,
 ) {
 	for {
-		if w.current.gasPool.Gas() < 30000000 {
-			// Temporary solution to reduce the fullness of the block. Break here when the available gas left hit 30M.
-			// Effectively making the gas limit 50M (since 80M is the default gas limit)
+		if w.current.gasPool.Gas() < 50000000 {
+			// Temporary solution to reduce the fullness of the block. Break here when the available gas left hit 50M.
+			// Effectively making the gas limit 30M (since 80M is the default gas limit)
 			utils.Logger().Info().Uint64("have", w.current.gasPool.Gas()).Uint64("want", params.TxGas).Msg("[Temp Gas Limit] Not enough gas for further transactions")
 			break
 		}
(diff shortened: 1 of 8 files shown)


MaxMustermann2 (approved):

rlan35:
Thanks for fixing the tests.

rlan35 (approved):

node/worker/worker.go
```diff
@@ -69,9 +69,9 @@ func (w *Worker) CommitSortedTransactions(
 	coinbase common.Address,
 ) {
 	for {
-		if w.current.gasPool.Gas() < 30000000 {
-			// Temporary solution to reduce the fullness of the block. Break here when the available gas left hit 30M.
-			// Effectively making the gas limit 50M (since 80M is the default gas limit)
+		if w.current.gasPool.Gas() < 50000000 {
```
snawaz, line 72: Why not define some constant in a single file and use that instead of using magic numbers? If such numbers are used at multiple places and their values are not in sync, then we're in trouble, right?

mhd999: Better use environment variables in this case. Doing so you'll have names for those magic number and also can change them without producing code changes.

snawaz: @mhd999 Environment variables are "code" as well, just different kinda code. But then, for such values, environment variables are not _probably_ needed and sounds overkill. Defining named constants in a single file should be enough!

mhd999: Read the comment again :) if those are changing values having them as environment variables means no code changes (golang) the changes will be on the infrastructure level especially if those env vars are passed at run time.

snawaz: I'd just define them as constants! I see no value in environment variables! Currently, the values are hardcoded, that implies all environments use the same values and so there is no need to add "infrastructure level" configurations!

BTW, Golang compiles quickly.... and constants often makes programs performant as the compilers have more information and thus they can optimise further!

rlan35: Thanks for the discussion, but this is an emergency update, we need to roll out asap to fix the network issues. Eventually the lower gas limit will be implemented as a formal hard fork.

PR #4239, opened 2022-07-19
feat: update dockerfile with a few enhancement 

jasonyic:
Configuration file should be used as the single source of truth (ssot), and provide an simple and clean way for operation. a few enhancements have been included in this PR:

 - refactor `dockerfile` to use multi stage build 
 - use non-root user & `tini` for better process handling
 - provide docker-compose file for local testing
 - add README.md

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,7 +40,7 @@ jobs:
     if: needs.check.outputs.tag_annotated == 'true'
     strategy:
       matrix:
-        os: [ubuntu-18.04, macos-10.15, [self-hosted, linux, ARM64]]
+        os: [ubuntu-18.04, macos-12, [self-hosted, linux, ARM64]]
 
     steps:
       - name: Import GPG key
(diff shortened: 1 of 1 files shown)


scripts/docker/README.md
```diff
@@ -0,0 +1,63 @@
...
+make docker
+```
+
+If your build machine has an ARM-based chip, like Apple silicon (M1), the image is built for `linux/arm64` by default. To build for `x86_64`, apply the --platform arg:
+
+```bash
+docker build --platform linux/amd64 -t harmonyone/harmony -f Dockerfile .
+```
+
+Before start the docker, dump the default config `harmony.conf`, make your customization. `harmony.conf`  should be mounted into default `HOME` directory `/harmony` inside the container. Assume `harmony.conf`, `blskeys` and `hmykey` are under `./config` in your current working directory, you can start your docker container with the following command:
```
MaxMustermann2, line 16: Can you add a one liner instruction that shows how to dump the `harmony.conf` out from the docker image `harmonyone/harmony` ?

jasonyic: done

MaxMustermann2 (approved):

PR #4246, opened 2022-07-27
[node] nitpick: move heartbeat validation logic

MaxMustermann2:


rlan35 (approved):
thanks for putting it in a better position.

PR #4258, opened 2022-08-07
Fix staking bug in elastic rpc mode

LuttyYang:
## Operational Checklist

   NO

   NO

   NO

### DIFF
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -3186,7 +3186,13 @@ func (bc *BlockChainImpl) isInitTiKV() bool {
 func (bc *BlockChainImpl) tikvPreemptMaster(fromBlock, toBlock uint64) bool {
 	for {
 		// preempt master
-		if ok, _ := bc.redisPreempt.TryLock(60); ok {
+		lockType, _ := bc.redisPreempt.TryLock(60)
+		switch lockType {
+		case redis_helper.LockResultRenewalSuccess:
+			return true
+		case redis_helper.LockResultSuccess:
+			// first to master
+			bc.validatorListByDelegatorCache.Purge()
 			return true
 		}
 
--- a/internal/tikv/redis_helper/lock.go
+++ b/internal/tikv/redis_helper/lock.go
@@ -5,6 +5,7 @@ import (
 	"crypto/rand"
 	"encoding/base64"
 	"errors"
+
 	"github.com/go-redis/redis/v8"
 )
 
--- a/internal/tikv/redis_helper/lock.go
+++ b/internal/tikv/redis_helper/lock.go
@@ -4,8 +4,17 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
-
+	"errors"
 	"github.com/go-redis/redis/v8"
+)
+
+type LockResult int
+
+const (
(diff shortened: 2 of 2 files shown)


core/blockchain_impl.go
```diff
@@ -3186,7 +3186,13 @@ func (bc *BlockChainImpl) isInitTiKV() bool {
 func (bc *BlockChainImpl) tikvPreemptMaster(fromBlock, toBlock uint64) bool {
 	for {
 		// preempt master
-		if ok, _ := bc.redisPreempt.TryLock(60); ok {
+		lockType, _ := bc.redisPreempt.TryLock(60)
```
LeoHChen, line 3189: why 60?  can it be a random number in a certain range?

core/blockchain_impl.go
```diff
@@ -3186,7 +3186,13 @@ func (bc *BlockChainImpl) isInitTiKV() bool {
...
+		switch lockType {
+		case redis_helper.LockResultRenewalSuccess:
+			return true
+		case redis_helper.LockResultSuccess:
+			// first to master
+			bc.validatorListByDelegatorCache.Purge()
```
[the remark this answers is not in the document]
LeoHChen: agree, other cache needs to be purged?  Please add a more detailed comment here about why this is important.

internal/tikv/redis_helper/lock.go
```diff
@@ -31,10 +40,15 @@ func (p *RedisPreempt) init() {
...
 	p.password = base64.StdEncoding.EncodeToString(byt)
 	p.lockScript = redis.NewScript(`
-		if redis.call('get',KEYS[1]) == ARGV[1] then
-			return redis.call('expire', KEYS[1], ARGV[2])
-		else
-			return redis.call('set', KEYS[1], ARGV[1], 'ex', ARGV[2], 'nx')
+		local val = redis.call('get', KEYS[1])
+		if (val==nil or (type(val) == "boolean" and not val)) then
+			redis.call('set', KEYS[1], ARGV[1], 'ex', ARGV[2])
+			return 'LockResultSuccess'
```
LeoHChen, line 46: why return text and later return 0 here?  can you return the same kind of value for consistency?

LeoHChen (commented):
how to reproduce the error and prove the fix?

core/blockchain_impl.go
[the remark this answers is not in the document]
LuttyYang: Other caches are cached based on blockhash, blockid or txid, no such problems

core/blockchain_impl.go
LuttyYang: This is the upper limit time for writing a block normally. The definition of 60 is because I think it will be more reasonable, writing a block usually takes less than 60 seconds, in case of accident, other writers will take over writing the block after 60 seconds

internal/tikv/redis_helper/lock.go
LuttyYang: This is a two-part script. There are 3 states returned when lock, and only 2 states are returned by unlock, indicating success or failure.

LuttyYang:
> how to reproduce the error and prove the fix?

reproduce the error can follow the process below:
Bug

PR #4263, opened 2022-08-15
fix: tikv graceful shutdown on close

diego1q2w:
## Issue

Sometimes when the writer is shutdown, panics because the channel **l2ExpiredRefresh** is closed before and there might still be some data being pushed to it, preventing the tikv and redis connection to close. In this PR I'm defering the close to make sure the connections are closed even if it panics, also moving the wait before the channel is closed to give some time to the cache to stop pushing messages to the channel 

## Operational Checklist

    **NO**

2. **Does this PR introduce backward-incompatible changes *NOT* related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)

    **NO**


LuttyYang (approved):
LGTM

Issue #4261, opened 2022-08-12
[hardfork] transaction fee goes to a multisig/foundation account


LeoHChen:
## Summary

As part of the recovery plan, https://talk.harmony.one/t/recovery-one-community-governance-for-depegged-tokens/21104, we can add a hark fork in the protocol to route all the transaction fees to a community-managed account.

## Current Design

Since the very beginning, all the transaction fees are burnt in the protocol, not going to any validator or account.

## Problems

The Recovery plan would like to assemble all the transaction fees to support either ongoing protocol development or reimburse the off-peg wrapped tokens.

## Proposal

A dedicated community-owned multisig account can be set up, and route all the transaction fees in each block to this account.
The account address has to be publically visible and hard-coded in the protocol.

This requires a hard fork upgrade of the protocol.

PR #4265, opened 2022-08-15
[hardfork] collect txns fees to a specific account

peekpi:
## Issue
#4261

After FeeCollectEpoch, all the transactions fees will be collected into the feeCollector account instead of be burned.
If the feeCollector account is a contract, it will not trigger any of its functions(fallback()/receive() in solidity).
feeCollector’s balance increases after each transaction is executed.


rlan35:
looks good. How is this change tested? We need to at least test on localnet and testnet before launching on mainnet since it's token-related critical change.

rlan35 (approved):

core/state_transition.go
```diff
@@ -282,6 +280,18 @@ func (st *StateTransition) refundGas() {
 	st.gp.AddGas(st.gas)
 }
 
+func (st *StateTransition) collectGas() {
+	// Burn Txn Fees after staking epoch
+	if config := st.evm.ChainConfig(); !config.IsStaking(st.evm.EpochNumber) {
+		txFee := new(big.Int).Mul(new(big.Int).SetUint64(st.gasUsed()), st.gasPrice)
+		st.state.AddBalance(st.evm.Coinbase, txFee)
+	} else if config.IsFeeCollectEpoch(st.evm.EpochNumber) { // collect Txn Fees to community-managed account.
```
gupadhyaya, line 288: should this be the first condition that is checked? otherwise after this fork, `IsStaking` will always be true and you may never hit `IsFeeCollectEpoch` condition? I think the `IsFeeCollectEpoch` condition should come first and fallback `else` should check `IsStaking` for burning?

rlan35: The first condition is "!config.IsStaking(st.evm.EpochNumber)"  so right now, it's always false. The catch is fee collection only works after staking epoch. If it's set before staking epoch, the original logic of fee going to the miner take precedence.

@peekpi please add the catch as a comment to the FeeCollectionEpoch definition.

sophoah:
@here, my ask here is to have a new test written in https://github.com/harmony-one/harmony-test so this feature can be tested at every new PR and be part of our CI/CD thanks.

internal/configs/sharding/mainnet.go
```diff
@@ -44,15 +46,29 @@ var (
...
+
+	emptyAddress = ethCommon.Address{}
+	// TODO: set a valid address
+	feeCollector ethCommon.Address // = mustAddress("0xXXX or onexxx")
 )
 
+func mustAddress(addrStr string) ethCommon.Address {
+	addr, err := common.ParseAddr(addrStr)
+	if err != nil {
+		panic("inlivad feeCollector")
```
MaxMustermann2, line 58: Nitpick: Typo.

MaxMustermann2 (approved):
Looks good. Please add a test as Soph described; let me know if you need help.

core/state_transition.go
peekpi: > ...
> @peekpi please add the catch as a comment to the FeeCollectionEpoch definition.

Also add function `mustValid()` to internal/params/config.go to ensure `ChainConfig` is valid.

peekpi:
> looks good. How is this change tested? We need to at least test on localnet and testnet before launching on mainnet since it's token-related critical change.

because testnet has not been fixed, we can only test it in localnet.

peekpi:
test js script:
```js
// fee.js
const ethers = require('ethers')

const provider = new ethers.providers.JsonRpcProvider('http://127.0.0.1:9500')
const feeCollector = '0x19E7E376E7C213B7E7e7e46cc70A5dD086DAff2A'
const faucetPrivateKey = '7472616374206b65e45ffeb29e933944f5027ef139f124f430641487e70ea9a1'
const faucetWallet = new ethers.Wallet(faucetPrivateKey, provider)

async function main() {
    const getBalance = async ()=>({
        feeCollector: await provider.getBalance(feeCollector).then(ethers.utils.formatEther),
        faucet: await faucetWallet.getBalance().then(ethers.utils.formatEther)
    })

    const beforeBalance = await getBalance()
    const gasPrice = ethers.utils.parseEther('0.001')
    await faucetWallet.sendTransaction({to:faucetWallet.address, gasPrice}).then(tx=>tx.wait())
    const afterBalance = await getBalance()
    console.log('beforeBalance:', beforeBalance)
    console.log('afterBalance :', afterBalance)
}

main()
```
run `node fee.js` after FeeCollectEpoch, we can see the balance of feeCollector increases.

log:
```shell
beforeBalance: { feeCollector: '0.0', faucet: '10000000000.0' }
afterBalance : { feeCollector: '21.0'', faucet: '9999999979.0' }
```

MaxMustermann2:
Can you move the checks here to be inside `mustValid` ?
https://github.com/harmony-one/harmony/blob/7c1c43ecbd00db4481ae0c19871c170761b6aa0c/internal/params/config.go#L706-L715

Secondly, can you ensure the `FeeCollectEpoch` variable is set in all `ChainConfig` objects in the file ? For example, it is currently missing in `StressnetChainConfig`.

peekpi:
> 



> Can you move the checks here to be inside `mustValid` ?
> 
> https://github.com/harmony-one/harmony/blob/7c1c43ecbd00db4481ae0c19871c170761b6aa0c/internal/params/config.go#L706-L715
> 
> Secondly, can you ensure the `FeeCollectEpoch` variable is set in all `ChainConfig` objects in the file ? For example, it is currently missing in `StressnetChainConfig`.

fixed

MaxMustermann2 (approved):

sophoah:
@peekpi will that work for shard 1/2/3 fees ? cc @MaxMustermann2

peekpi:
> 



> @peekpi will that work for shard 1/2/3 fees ? cc @MaxMustermann2
> 
> Also let's assume one day we need a 5th shard, will that still work ?
> 
> here are some test case scenario : fee on a tx between s0 to s0 fees on a tx between s1 to s1 fees on a tx between s0 to s1
> 
> What about future cross shard contract transaction fee ?

fee are always collected in src shard.
s0->s0: s0
s1->s1: s1
s0->s1: s0
s1->s0: s1

for future cross shard contract transaction, we need to handle this scenario carefully.

MaxMustermann2:
If the `FeeCollector` is going to be a multisig, #4165 must be activated at the same time (or before) this change. This will allow movement of funds across shards through the smart contract, something that is not currently possible.

peekpi:
> If the `FeeCollector` is going to be a multisig, #4165 must be activated at the same time (or before) this change. This will allow movement of funds across shards through the smart contract, something that is not currently possible.

does pr #4165 only support cross shard transfer? or also support cross shard contract call?

MaxMustermann2:
> does pr #4165 only support cross shard transfer? or also support cross shard contract call?

It supports only transfers. And it's actually disabled for smart contract (although that needs some more hardening in my opinion). Please ignore my original comment then.

sophoah (approved):

PR #4266, opened 2022-08-15
Resharding: leader rotation.

Frozen:
Rotate leader. Currently we choose one leader for all epoch lifetime and change only if leader goes down. PR changes this, so new leader changes every 64 blocks. It calculates blocks count produced by leader, and move to next leader when  enough blocks. It doesn't conflict with current view change behaviour.


consensus/consensus_service.go
```diff
@@ -77,24 +78,30 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys, allowlist []bls_cosi.Publi
...
-			Int("index", i).
-			Str("BLSPubKey", pubKeys[i].Bytes.Hex()).
-			Msg("Member")
-	}
-
 	allKeys := consensus.Decider.Participants()
+	consensus.pubKeyLock.Unlock()
 	if len(allKeys) != 0 {
-		consensus.LeaderPubKey = &allKeys[0]
+		first := consensus.Decider.FirstParticipant(
```
rlan35, line 84: why do such change? and deviate from simply getting the first key in the committee list and the first leader?

consensus/consensus_service.go
```diff
@@ -77,24 +78,30 @@ func (consensus *Consensus) UpdatePublicKeys(pubKeys, allowlist []bls_cosi.Publi
...
 	}
-	consensus.pubKeyLock.Unlock()
+	for i := range pubKeys {
+		consensus.getLogger().Info().
+			Int("index", i).
+			Str("BLSPubKey", pubKeys[i].Bytes.Hex()).
+			Msg("Member")
+	}
+	if consensus.Blockchain.Config().IsLeaderRotation(consensus.GetCurEpoch()) {
+		consensus.updateLeader()
```
rlan35, line 102: when the committee (publicKeys) are updated, that means the epoch is switching, we should just default back to the first key as the leader. Why call updateLeader() here?

consensus/consensus_v2.go
```diff
@@ -679,13 +683,104 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
...
+			}()
+		}
+	}
+}
+
 // SetupForNewConsensus sets the state for new consensus
 func (consensus *Consensus) SetupForNewConsensus(blk *types.Block, committedMsg *FBFTMessage) {
 	atomic.StoreUint64(&consensus.blockNum, blk.NumberU64()+1)
 	consensus.SetCurBlockViewID(committedMsg.ViewID + 1)
+	if blk.IsLastBlockInEpoch() {
```
rlan35, line 742: why do this change? why not directly fetch the epoch from the proposed block? that should be already correct.

consensus/consensus_service.go
Frozen: rolled back

consensus/consensus_v2.go
```diff
@@ -679,13 +681,75 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
...
+			header := consensus.Blockchain.GetHeaderByNumber(curNumber - uint64(i))
+			if header == nil {
+				return
+			}
+			// Previous epoch, we should not change leader.
+			if header.Epoch().Uint64() != epoch.Uint64() {
+				return
+			}
+			// Check if the same leader.
+			pub, err := chain.GetLeaderPubKeyFromCoinbase(consensus.Blockchain, header)
```
rlan35, line 700: We should keep a local cache of the leader situation for past blocks, instead of reading from the DB and parse the coinbase for every block it's pretty heavy. This DB reading and parsing should only happen when there is not enough cache data in memory to use (e.g. node first startup and for the first N blocks)

consensus/consensus_v2.go
```diff
@@ -679,13 +681,75 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
...
+				return
+			}
+		}
+		// Passed all checks, we can change leader.
+		var (
+			wasFound bool
+			next     *bls.PublicKeyWrapper
+		)
+		// The same leader for N blocks.
+		if consensus.Blockchain.Config().IsAllowlistEpoch(epoch) {
```
rlan35, line 716: Let's remove the allowlist logic for now. We may not need to use it eventually.

consensus/consensus_v2.go
Frozen: cached

consensus/consensus_v2.go
Frozen: Removed. I can remove `IsAllowlistEpoch` method and logic if we don't need it anymore in another pr.

internal/params/config.go
```diff
@@ -342,7 +348,10 @@ var (
 		big.NewInt(0),        // SlotsLimitedEpoch
 		big.NewInt(1),        // CrossShardXferPrecompileEpoch
 		big.NewInt(0),        // AllowlistEpoch
-		big.NewInt(0),        // TesnetNinetyPercentEpoch
+		// TODO place correct epoch number
```
MaxMustermann2, line 351: In addition to this `TODO`, please also add `LeaderRotationEpoch` to be `EpochTBD` for all non-testnet chains.

consensus/consensus_v2.go
```diff
@@ -679,13 +680,66 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
 	return nil
 }
 
+func (consensus *Consensus) rotateLeader(epoch *big.Int) {
+	prev := consensus.GetLeaderPubKey()
+	curNumber := consensus.Blockchain.CurrentHeader().Number().Uint64()
+	utils.Logger().Info().Msgf("[Rotating leader] epoch: %v rotation:%v numblocks:%d", epoch.Uint64(), consensus.Blockchain.Config().IsLeaderRotation(epoch), consensus.Blockchain.Config().LeaderRotationBlocksCount)
+	if consensus.Blockchain.Config().IsLeaderRotation(epoch) {
```
MaxMustermann2, line 687: Why are we checking this again?

consensus/consensus_v2.go
```diff
@@ -679,13 +680,66 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
...
+		leader := consensus.GetLeaderPubKey()
+		for i := 0; i < consensus.Blockchain.Config().LeaderRotationBlocksCount; i++ {
+			header := consensus.Blockchain.GetHeaderByNumber(curNumber - uint64(i))
+			if header == nil {
+				return
+			}
+			// Previous epoch, we should not change leader.
```
MaxMustermann2, line 694: Nit: Previous `block was epoch block`, we should not change leader.

node/api.go
```diff
@@ -18,6 +19,11 @@ func (node *Node) IsCurrentlyLeader() bool {
 	return node.Consensus.IsLeader()
 }
 
+// GetPublicKeys exposes if node is currently the leader node
+func (node *Node) GetPublicKeys() multibls.PublicKeys {
```
MaxMustermann2, line 23: Where is this used?

node/api.go
```diff
@@ -18,6 +19,11 @@ func (node *Node) IsCurrentlyLeader() bool {
...
+	return node.Consensus.GetPrivateKeys().GetPublicKeys()
```
MaxMustermann2, line 24: Why not `node.Consensus.GetPublicKeys()` ?

internal/utils/singleton.go
```diff
@@ -195,3 +196,18 @@ func updateZeroLogLevel(level int) {
 	childLogger := Logger().Level(zeroLoggerLevel)
 	zeroLogger = &childLogger
 }
+
+func GetPort() int {
```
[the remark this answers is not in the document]
Frozen: It's useful for debugging. I prefer keep it, it's used often.

consensus/consensus_service.go
```diff
@@ -451,6 +446,18 @@ func (consensus *Consensus) IsLeader() bool {
 	return false
 }
 
+// isLeader check if the node is a leader or not by comparing the public key of
+// the node with the leader public key. This function assume it runs under lock.
+func (consensus *Consensus) isLeader() bool {
```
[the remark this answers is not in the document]
Frozen: No use currently, but better keep it, it will be overwritten in next PR.

consensus/consensus_v2.go
Frozen: This check is cheap, we can keep it. Previously this function was called from different places, so we needed it.

node/api.go
Frozen: This function will be overwritten in next PR https://github.com/harmony-one/harmony/pull/4300/files#diff-b1f5bb00d842600c8142ae84ffba5888cbcc451590063e43198771d4d0b6a1b4R167 . We can keep it now, i'll change it in next pr.

consensus/consensus_v2.go
Frozen: fixed

internal/params/config.go
Frozen: fixed

internal/utils/singleton.go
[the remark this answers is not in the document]
MaxMustermann2: Agreed; it's very practical. But please add a comment indicating its unused in the code and only used during development.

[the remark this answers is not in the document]
Frozen: Added comment.

consensus/consensus_service.go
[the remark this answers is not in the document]
Frozen: Actually, this function does what described. I don't see any problems here.

ONECasey (changes requested):
@Frozen , can you please fix the conflicts?

ONECasey:
Fixed conflicts.

ONECasey (approved):
Approved. Looks good to go into `dev` once state sync has been merged into `main`

PR #4268, opened 2022-08-17
Reduced cache size.

Frozen:
Measured on explorer node, it saves about 1.5G of memory, and doesn't impact somehow on LA, CPU, DISC IO or reads count.

### DIFF
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -93,10 +93,10 @@ var (
 )
 
 const (
-	bodyCacheLimit                     = 2048
-	blockCacheLimit                    = 2048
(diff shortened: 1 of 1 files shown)


rlan35:
Great job! Excited to see it's deployed on some mainnet rpc nodes @sophoah

sophoah:
> Measured on explorer node, it saves about 1.5G of memory, and doesn't impact somehow on LA, CPU, DISC IO or reads count.

hey Konstantin, I guess you haven't tested on whether it impacts :
- sync speed
- RPC performance call ?
- validator node using the staking data ?

PR #4273, opened 2022-08-25
[Reimbursement] TBD

peekpi:
## Issue
Issue additional 2.391 billion ONE tokens over 3 years to do the reimbursement.


core/state_processor.go
```diff
@@ -402,6 +404,37 @@ func ApplyIncomingReceipt(
...
+		startNumber := shard.Schedule.EpochLastBlock(config.ReimbursementEpoch.Uint64()-1) + 1
+		startHeader := bc.GetHeaderByNumber(startNumber) // first block to reimburse, use it as the starting time
+		// endTime = startTime + duration
+		endTime := startHeader.Time().Uint64() + reimbursement.Duration.Uint64()
+		parentTime := parentHeader.Time().Uint64()
+		if parentTime >= endTime { // reimburse over
+			return
+		}
+		currentTime := header.Time().Uint64()
+		deltaTime := currentTime - parentTime
```
rlan35, line 420: actually don't need to compute time and reimburse based on block time. We just assume the block time is 2 seconds and issue the same fixed number of tokens (hard-coded) for 3 years based on 2-second block time.

internal/configs/sharding/mainnet.go
```diff
@@ -204,22 +204,22 @@ func (ms mainnetSchedule) IsSkippedEpoch(shardID uint32, epoch *big.Int) bool {
...
-	mainnetV3   = MustNewInstance(4, 250, 90, 0, numeric.MustNewDecFromStr("0.68"), genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV1_5, emptyAllowlist, mainnetReshardingEpoch, MainnetSchedule.BlocksPerEpoch())
-	mainnetV3_1 = MustNewInstance(4, 250, 50, 0, numeric.MustNewDecFromStr("0.60"), genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV1_5, emptyAllowlist, mainnetReshardingEpoch, MainnetSchedule.BlocksPerEpoch())
-	mainnetV3_2 = MustNewInstance(4, 250, 25, 0, numeric.MustNewDecFromStr("0.49"), genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV1_5, emptyAllowlist, mainnetReshardingEpoch, MainnetSchedule.BlocksPerEpoch())
-	mainnetV3_3 = MustNewInstance(4, 250, 25, 0.06, numeric.MustNewDecFromStr("0.49"), genesis.HarmonyAccounts, genesis.FoundationalNodeAccountsV1_5, emptyAllowlist, mainnetReshardingEpoch, MainnetSchedule.BlocksPerEpoch())
+	mainnetV0   = MustNewInstance(4, 150, 112, 0, numeric.OneDec(), genesis.HarmonyAccounts, genesis.FoundationalNodeAccounts, emptyAllowlist, mainnetReshardingEpoch, MainnetSchedule.BlocksPerEpochOld(), nil)
```
rlan35, line 207: For this change, we don't have to modify the config in sharding/*.go. It's too much. We don't expect this logic to apply to other networks or other cases, so it's a one-off hardfork, we can just add it as a one-off logic directly like:

if reimburseEpoch and mainnet:
.....

rlan35:
Closing it for now since the reimbursement plan likely won't need this.

PR #4275, opened 2022-08-30
Fix/filter new tikv

diego1q2w:
## Issue

Two issues were found:

First: when creating a new filter, the filters are stored in memory so you would create the filter in one reader but then when you would like to pull the result of that filter and your request was routed to a different node then you would get "filter not found error". This PR makes it so other readers are notified about new filters created so they can create one with the same filter ID. This is done via redis channels

Second: the writer sends the newst blocks and logs to the readers, so they can keep them in memory and run in-memory feeds. The issue was that when serializing the data some fields where cut off from the [serialized output ](https://github.com/harmony-one/harmony/blob/da3c93fb58371b5355923ede044ab0ad398d15e6/core/types/log.go#L83).  So we needed to wrap the Log struct with LogForStorage [which includes all the fields](https://github.com/harmony-one/harmony/blob/da3c93fb58371b5355923ede044ab0ad398d15e6/core/types/log.go#L102).

Update: tested it on a non-erpc node and it works fine too.

### DIFF
--- a/internal/tikv/redis_helper/pubsub.go
+++ b/internal/tikv/redis_helper/pubsub.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum"
@@ -22,7 +21,7 @@ var big0 = big.NewInt(0)
 // BlockUpdate block update event
 type BlockUpdate struct {
 	BlkNum uint64
-	Logs   []*types.LogForStorage
+	Logs   []*types.LogForStorage // types.Log will cut some fields when rpl encoding/decoding look at core/types/log.go:83
 }
 
 // NewFilterUpdated new filter update event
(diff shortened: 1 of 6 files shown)


internal/tikv/redis_helper/pubsub.go
```diff
@@ -12,28 +17,50 @@ import (
 	"github.com/pkg/errors"
 )
 
+var big0 = big.NewInt(0)
```
MaxMustermann2, line 20: nit: `common.Big0` please

internal/tikv/redis_helper/pubsub.go
```diff
@@ -44,6 +71,47 @@ func PublishShardUpdate(shardID uint32, blkNum uint64, logs []*types.Log) error
 	return redisInstance.Publish(context.Background(), fmt.Sprintf("shard_update_%d", shardID), msg).Err()
 }
 
+// SubscribeNewFilterLogEvent subscribe new filter log event from other readers
+func SubscribeNewFilterLogEvent(shardID uint32, namespace string, cb func(id string, crit ethereum.FilterQuery)) {
+	if redisInstance == nil {
+		return
+	}
```
MaxMustermann2, line 78: This check, is it supposed to be everywhere? And should it return an error?

LuttyYang (approved):
LGTM

MaxMustermann2 (approved):

PR #4276, opened 2022-09-01
add description for closing client and change randomize process to ma…

GheisMohammadi:
In order to analyse the node logs and debug the connectivity issues, we added a few descriptions for connection closure, which help us figure out the reason for peer disconnections. 
PR changes the peers' randomize function as well to make sure we get a list of online nodes for sync. 
**This PR has no effect on other functions and is completely backward compatible.**


api/service/legacysync/helpers.go
```diff
@@ -58,39 +58,39 @@ func createSyncConfig(syncConfig *SyncConfig, peers []p2p.Peer, shardID uint32)
...
 
 	utils.Logger().Debug().
-		Int("len", len(peers)).
+		Int("peers count", len(peers)).
+		Int("target size", targetSize).
 		Uint32("shardID", shardID).
 		Msg("[SYNC] CreateSyncConfig: len of peers")
 
-	if len(peers) == 0 {
+	if targetSize == 0 {
```
MaxMustermann2, line 69: Why this change?

GheisMohammadi: Basically, what we were doing in this function was getting a list of peers, then randomly choosing a set of peers from that list, and then connecting to the peers in the selected set. If the connection to any peer in that set is broken and the client has failed to build, we remove that peer from the list. So, imagine the set contains 3 peers. If two of them are broken, it leaves only one peer connected to the node. What I did was change a bit of this flow. So, we first shuffle the peers, but we still have all the peers in our set. Then we try to connect to N peers, where N is targetSize. After we connect to those N peers, we remove the rest from the peer set. By this way we can always be sure that the peers list has maximum online nodes and we add maximum available random online peers to list. targetSize is a random number between low and high bounds, and the final number of peers will be set by that number. This part is the same as before. But we remove extra peers from the list later after initializing connections; that's why we use this variable.

api/service/legacysync/helpers.go
```diff
@@ -58,39 +58,39 @@ func createSyncConfig(syncConfig *SyncConfig, peers []p2p.Peer, shardID uint32)
...
+			continue
+		}
+		peerConfig := &SyncPeerConfig{
+			ip:     peer.IP,
+			port:   peer.Port,
+			client: client,
+		}
+		syncConfig.AddPeer(peerConfig)
+		connectedPeers++
+		if connectedPeers >= targetSize {
```
Frozen, line 90: Do we really need new variable when we can just call smth like 
`syncConfig.connectedPeersCount()` ?

api/service/legacysync/helpers.go
```diff
@@ -58,39 +58,39 @@ func createSyncConfig(syncConfig *SyncConfig, peers []p2p.Peer, shardID uint32)
...
+	if targetSize == 0 {
 		return syncConfig, errors.New("[SYNC] no peers to connect to")
 	}
 	if syncConfig != nil {
 		syncConfig.CloseConnections()
 	}
 	syncConfig = NewSyncConfig(shardID, nil)
 
-	var wg sync.WaitGroup
+	var connectedPeers int
```
Frozen, line 77: What is the reason to remove async code?

api/service/legacysync/helpers.go
GheisMohammadi: I think you mean using syncConfig.PeersCount(). right? I should say it also uses mutex and for sure is slower, so, in my opinion in terms of memory usage and performance this variable makes more sense. Any concern you see about this variable?

api/service/legacysync/helpers.go
GheisMohammadi: Because I need to count the connected peers one by one till they reach a certain number of connected peers. If I do parallel, first I have to wait for all client routines to finish and then disconnect the extra ones. Just imagine if we have a few hundred peers in our peer list and we need a few connected nodes. Does it make sense to connect to all of them (just to get the benefit of parallel routines) and then remove tenths of extra connections? And also, this function is called a few times, mostly at the beginning, and technically it won't affect performance that much, and in some certain scenarios it improves performance as well.

Frozen (approved):

ONECasey (approved):

PR #4279, opened 2022-09-12
Fix getting epoch number.

Frozen:
### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -1187,9 +1187,10 @@ func (node *Node) InitConsensusWithValidators() (err error) {
 		)
 	}
 	shardID := node.Consensus.ShardID
-	blockNum := node.Blockchain().CurrentBlock().NumberU64()
+	currentBlock := node.Blockchain().CurrentBlock()
+	blockNum := currentBlock.NumberU64()
 	node.Consensus.SetMode(consensus.Listening)
-	epoch := shard.Schedule.CalcEpochNumber(blockNum)
+	epoch := currentBlock.Epoch()
 	utils.Logger().Info().
 		Uint64("blockNum", blockNum).
 		Uint32("shardID", shardID).


MaxMustermann2 (approved):

rlan35 (approved):

sophoah:
hey Konstantin, the code you are pushing was working fine before, it only recently it showed the weird issue we had at node start on testnet. Do you have an explanation for the issue (ie why we just have that issue now) ? I am worried `shard.Schedule.CalcEpochNumber(blockNum)` might be called somewhere else and the same issue may happened in another edge case scenario

Frozen:
It's difficult to say, that all calls of that method are correct, but exactly this line is really weird, because calculate epoch when we already have epoch doesn't make any sense. And this why i want to test it on testnet before.

MaxMustermann2:
Some more color on this PR:

Within `InitConsensusWithValidators` a node is supposed to determine the BLS keys with which it will sign the blocks. This is done once at startup, and **again** at epoch change as part of the regular syncing process (via `UpdateConsensusInformation`). The inaccurate epoch number calculation, fixed in this PR, does not directly impact the epoch number of the block produced. It is calculated by the below code after each block of all shards.
https://github.com/harmony-one/harmony/blob/b1ec028031d7cb607f293f42f4fb9172e0f9773d/node/worker/worker.go#L353-L371

Within our testing, this bug was discovered as a potential cause to the following issue (but it was not). A specific external validator's BLS key was retained in committee, even though it was not voting. Another key was added to that validator, which never got elected. This is because the `shardState` is repeated at the next epoch if there are no auction winners - which happened because it was the singular validator.
https://github.com/harmony-one/harmony/blob/b1ec028031d7cb607f293f42f4fb9172e0f9773d/shard/committee/assignment.go#L383-L386

ONECasey (approved):

ONECasey:
This will be merged to `dev` once state sync has been merged to `main`.

PR #4280, opened 2022-09-13
Feature peers storage.

Frozen:
`Storage` is struct where we keep 2 types of peers: valid peers and invalid. When we ask storage to give us peers to connect, first we get them from valid peers, and if it's not enough then from invalid in hope, that somehow we will succeed to connect to them. If it happens, we move peer from invalid to valid and vice versa for valid peers. 
The difference is that currently we get all peers, shuffle them and get N (probably 3) of them. We have no clue are those peers have ever been valid, and we have chance not to connect to even 1 peer from that list. With `storage` we have same chance too, but it's less probably. 

Actually, it contains more cleanup than logic, `storage` is really simple.

### DIFF
--- a/api/service/legacysync/storage.go
+++ b/api/service/legacysync/storage.go
@@ -3,6 +3,7 @@ package legacysync
 import (
 	"sync"
 
+	"github.com/harmony-one/harmony/internal/linkedhashmap"
 	"github.com/harmony-one/harmony/p2p"
 )
 
@@ -14,8 +15,8 @@ type peerDupID struct {
 // Storage keeps successful and failed peers.
 type Storage struct {
 	mu   sync.Mutex
-	succ map[peerDupID]p2p.Peer
-	fail map[peerDupID]p2p.Peer
+	succ *linkedhashmap.HashMap[peerDupID, p2p.Peer]
+	fail *linkedhashmap.HashMap[peerDupID, p2p.Peer]
 }
 
 func (s *Storage) Contains(p p2p.Peer) bool {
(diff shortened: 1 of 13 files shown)


GheisMohammadi:
Could you please add a description and explain what this PR does and why we need that?

api/service/legacysync/storage.go
```diff
@@ -0,0 +1,111 @@
...
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	d := peerDupID{p.IP, p.Port}
+	if s.contains(d) {
+		return
+	}
+	s.succ[d] = p
+}
+
+func (s *Storage) Fail(p p2p.Peer) {
```
rlan35, line 68: => SetFailure

api/service/legacysync/storage.go
```diff
@@ -0,0 +1,111 @@
...
+	defer s.mu.Unlock()
+	return s.isFail(peerDupID{p.IP, p.Port})
+}
+
+func (s *Storage) isFail(d peerDupID) bool {
+	_, ok := s.fail[d]
+	return ok
+}
+
+func (s *Storage) Success(p p2p.Peer) {
```
rlan35, line 87: => SetSuccess

rlan35, line 87: Seems like Success() is never really used in sync logic?

Frozen: renamed

Frozen: True. I had some doubts about understanding when successful is successful. Added.

Frozen:
I've added linked hash map for better peers rotations, because even hashmap is random, it doesn't mean it shuffles, and can produce same result every time. It has 99% code coverage, but using it is still tricky, because it requires huge amount of pointers, and small mistake can lead to UB. 
I don't recommend merge it now, it needs proper testing.

ONECasey:
@Frozen, can you please fix the conflicts on this PR?

Frozen:
There are changes from Gheis for this part, and i have already cherry-picked some commits from it, so i think it's outdated.

PR #4286, opened 2022-10-04
[p2p] prevent dialing of private ips

MaxMustermann2:
The original feature (erroneously) prevents only querying of the private IPs. This change prevents dialing private IPs altogether when the flag is activated.

## Issue
One of our (newly added) internal nodes received a netscan abuse email even though the flag `DisablePrivateIPScan` was turned on.

### Unit Test Coverage

Before:
```
p2p: 15.7%
```

After:

```
p2p: 17.2%
```

## Operational Checklist

No.

No.

No.

### DIFF
--- a/p2p/gater.go
+++ b/p2p/gater.go
@@ -1,0 +1,44 @@
+package p2p
+
+import (
+	"github.com/libp2p/go-libp2p-core/connmgr"
+	"github.com/libp2p/go-libp2p-core/control"
+	"github.com/libp2p/go-libp2p-core/network"
+	"github.com/libp2p/go-libp2p-core/peer"
+	libp2p_dht "github.com/libp2p/go-libp2p-kad-dht"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+type Gater struct{}
+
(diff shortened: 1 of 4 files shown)


rlan35 (approved):

PkayJava:
Tested this PR (Harmony (C) 2020. harmony, version v7593-v4.3.12-35-gee52bc80)

restarted harmony service for many times. hetzner does not block that node.

p2p/gater.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/libp2p/go-libp2p-core/peer"
+	libp2p_dht "github.com/libp2p/go-libp2p-kad-dht"
+	ma "github.com/multiformats/go-multiaddr"
+)
+
+type Gater struct{}
+
+func NewGater(disablePrivateIPScan bool) connmgr.ConnectionGater {
+	if !disablePrivateIPScan {
+		return nil
```
Frozen, line 16: Returning nil from constructor seems like bad design for me. Maybe exactly here it's ok, but such style leads to runtime panics.

PR #4287, opened 2022-10-10
fix: erpc raw transaction error handling

diego1q2w:
## Issue

Initially the erpc reader was thought to only listen to existing tx from the reader, so any new transaction coming from a client would get broadcasted to the network by the reader without any validation, then the client would get a valid tx hash for every transaction even if that transaction could be discarded by the validator afterwards. 
We did this because we didn't want the reader to get tx from the p2p network and only from the writer. But we then decided to run the reader in offline mode which discarded that issue but we never enabled the readers to process and validate pending transactions. In this PR im removing the lines of code that prevented just that, now the reader can validate a transaction and only if valid then it will broadcasting it, giving timely feedback to the client, the writer will still catch up with that transaction eventually and notify the other readers about it.

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
-	"log"
 	"math/big"
 	"os"
 	"runtime/pprof"
@@ -235,12 +234,6 @@ func (node *Node) addPendingTransactions(newTxs types.Transactions) []error {
 	// 		Msg("[addPendingTransactions] Node out of sync, ignoring transactions")
 	// 	return nil
 	// }
-
-	// in tikv mode, reader only accept the pending transaction from writer node, ignore the p2p message
-	if node.HarmonyConfig.General.RunElasticMode && node.HarmonyConfig.TiKV.Role == tikv.RoleReader {
-		log.Printf("skip reader addPendingTransactions: %#v", newTxs)
-		return nil
-	}
 
 	poolTxs := types.PoolTransactions{}
 	errs := []error{}


MaxMustermann2 (approved):

MaxMustermann2:
Before this change
```bash
curl -H "Content-Type: application/json" -d '{"method":"hmyv2_sendRawTransaction","params":["0xf86f80830186a085e8d4a51000808094966a2fba069b863eced2686839b56796e53dcbc1880de0b6b3a76400008026a0c0607a814ebb5ac3de55272ba4b97008e7264b65a37f4fa7c193af510dbcd6a6a04c0e30f45785d67117a24f8dc84982aedcb348c24dc9b5a2b0539c8470bac942"],"id":1}' https://api.harmony.one
{"jsonrpc":"2.0","id":1,"result":"0x6d6fcbb89aa64610f8a8497295e36c06740c8a573d828b2ac75658a6605bfc44"}
```

After this change
```bash
curl -H "Content-Type: application/json" -d '{"method":"hmyv2_sendRawTransaction","params":["0xf86f80830186a085e8d4a51000808094966a2fba069b863eced2686839b56796e53dcbc1880de0b6b3a76400008026a0c0607a814ebb5ac3de55272ba4b97008e7264b65a37f4fa7c193af510dbcd6a6a04c0e30f45785d67117a24f8dc84982aedcb348c24dc9b5a2b0539c8470bac942"],"id":1}' https://api.harmony.one
{"jsonrpc":"2.0","id":1,"error":{"code":-32000,"message":"exceeds block gas limit"}}
```

PR #4289, opened 2022-10-18
Decider no longer requires public keys as a dependency.

Frozen:
It was required by method that no one uses, removing this method allows us simplify Decider dependencies.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/harmony-one/harmony/consensus/quorum"
 	"github.com/harmony-one/harmony/internal/shardchain/tikv_manage"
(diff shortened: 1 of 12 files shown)


GheisMohammadi:
It seems safe to remove, do you have any comments on why even these were there?

Frozen:
> It seems safe to remove, do you have any comments on why even these were there?
No idea

MaxMustermann2 (approved):

sophoah:
`AmIMemberOfCommitee` could have been something used by an RPC that would determine if the a validator or bls key is actually elected or not, but seems it was never used or implemented afterward.

sophoah:
looks to be safe to remove but as suggested by @GheisMohammadi understanding why it was there at the first place would have been nice

sophoah (approved):

PR #4291, opened 2022-10-22
slow down the growth of statedb disk usage

peekpi:
## Issue
1. removed writing stateDB to disk once per epoch and now it only writes to disk when memory garbage collection happens.
2. fix minor bug of tikvCleanCache


MaxMustermann2 (approved):

MaxMustermann2:
Supersed by #4295

PR #4293, opened 2022-10-24
Consensus doesn't require anymore `Node` as a circular dependency.

Frozen:
Resolved circular dependency 
`currentConsensus.SetBlockVerifier(currentNode.VerifyNewBlock)` 
where `*Node` method was provided as validator for new block. Checks from node were moved inside blockschain_impl.


MaxMustermann2:
Can this be closed in favour of #4296 ?

PR #4295, opened 2022-10-29
Prune state trie in run-time

peekpi:
This pr is designed to prune the state-tree in run-time to slow down the growth of disk usage for non-archival nodes.
When a new state-tree is written to disk, it compares the old state-tree with the new state-tree, and deletes data exists in old state-tree but not in the new state-tree.

### DIFF
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -3145,11 +3145,16 @@ func (bc *BlockChainImpl) tikvCleanCache() {
 			if err != nil {
 				continue
 			}
-
-			// build current block statedb
-			toBlock := bc.GetBlockByNumber(i + 1)
-			toTrie, err := state.New(toBlock.Root(), bc.stateCache)
-			if err != nil {
+			var toTrie *state.DB
+			// find next state trie
+			for j := i + 1; j <= to+1; j++ {
+				toBlock := bc.GetBlockByNumber(j)
+				toTrie, err = state.New(toBlock.Root(), bc.stateCache)
+				if err != nil {
+					continue
+				}
+			}
+			if toTrie == nil {
 				continue
 			}
 
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -219,7 +219,7 @@ func newBlockChainWithOptions(
 	engine consensus_engine.Engine, vmConfig vm.Config, options Options) (*BlockChainImpl, error) {
 	if cacheConfig == nil {
 		cacheConfig = &CacheConfig{
-			TrieNodeLimit: 256 * 1024 * 1024,
+			TrieNodeLimit: 256,
 			TrieTimeLimit: 2 * time.Minute,
 		}
 	}
@@ -1214,7 +1214,7 @@ func (bc *BlockChainImpl) WriteBlockWithState(
 
 	// Flush trie state into disk if it's archival node or the block is epoch block
 	triedb := bc.stateCache.TrieDB()
-	if bc.cacheConfig.Disabled || block.IsLastBlockInEpoch() {
+	if bc.cacheConfig.Disabled {
 		if err := triedb.Commit(root, false); err != nil {
 			if isUnrecoverableErr(err) {
 				fmt.Printf("Unrecoverable error when committing triedb: %v\nExitting\n", err)
@@ -3282,6 +3282,7 @@ var (
 //  3. Corrupted db data. (leveldb.errors.ErrCorrupted)
 //  4. OS error when open file (too many open files, ...)
 //  5. OS error when write file (read-only, not enough disk space, ...)
+//
 // Among all the above leveldb errors, only `too many open files` error is known to be recoverable,
 // thus the unrecoverable errors refers to error that is
 //  1. The error is from the lower storage level (from module leveldb)
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -1262,6 +1262,7 @@ func (bc *BlockChainImpl) WriteBlockWithState(
 					}
 					// Flush an entire trie and restart the counters
 					triedb.Commit(header.Root(), true)
+					bc.pruneOldTrie(lastWrite, chosen)
 					lastWrite = chosen
 					bc.gcproc = 0
 				}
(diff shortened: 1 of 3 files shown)


MaxMustermann2 (commented):
I like this PR, but we need more production level tests. It would also help to get a potential estimate of the space saved.

In addition, could you add some more comments to `state_prune.go` for clarity and of course make this configurable by the node runner?

CC @PkayJava @sophoah

MaxMustermann2:
@peekpi do we know why state trie was intentionally dumped to disk at the end of each epoch?

peekpi:
> @peekpi do we know why state trie was intentionally dumped to disk at the end of each epoch?

I don't know, but I don't see where this feature is needed. 
do you know about this? @rlan35

MaxMustermann2:
> I don't know, but I don't see where this feature is needed.
Let's retain it for now. According to RJ, it is "convenient for fast state sync where only state is synced".

peekpi:
> > I don't know, but I don't see where this feature is needed.
> 
> Let's retain it for now. According to RJ, it is "convenient for fast state sync where only state is synced".

How about make it to optional? And the default is to keep it.

MaxMustermann2:
> How about make it to optional? And the default is to keep it.

Let's keep the last block in epoch please.

cmd/harmony/flags.go
```diff
@@ -319,6 +320,12 @@ var (
 		Usage:    "indicates if full transaction tracing should be enabled",
 		DefValue: defaultConfig.General.TraceEnable,
 	}
+
+	statePruneFlag = cli.BoolFlag{
```
MaxMustermann2, line 324: Add a test for this flag to to `flags_test.go`

core/blockchain_impl.go
```diff
@@ -139,7 +139,10 @@ type BlockChainImpl struct {
 	// This can improve the cache hit rate of redis
 	latestCleanCacheNum uint64      // The most recently cleaned cache of block num
 	cleanCacheChan      chan uint64 // Used to notify blocks that will be cleaned up
-
+	lastWrite           uint64
+	statePruneEnable    bool
+	statePruneRuning    sync.WaitGroup
```
MaxMustermann2, line 144: nit: `statePruneRunning`

core/blockchain_impl.go
```diff
@@ -899,7 +904,8 @@ func (bc *BlockChainImpl) Stop() {
 	bc.scope.Close()
 	close(bc.quit)
 	atomic.StoreInt32(&bc.procInterrupt, 1)
-
+	// wait pruning completed
+	bc.waitPruning(false)
```
MaxMustermann2, line 908: Should this be done after the `HEAD`, `HEAD-1` and `HEAD-127` ?

core/blockchain_impl.go
```diff
@@ -1214,7 +1218,7 @@ func (bc *BlockChainImpl) WriteBlockWithState(
 
 	// Flush trie state into disk if it's archival node or the block is epoch block
 	triedb := bc.stateCache.TrieDB()
-	if bc.cacheConfig.Disabled || block.IsLastBlockInEpoch() {
+	if bc.cacheConfig.Disabled {
```
MaxMustermann2, line 1221: Add back `|| block.IsLastBlockInEpoch()`

core/blockchain_impl.go
```diff
@@ -1253,16 +1257,21 @@ func (bc *BlockChainImpl) WriteBlockWithState(
...
-					if chosen < lastWrite+triesInMemory && bc.gcproc >= 2*bc.cacheConfig.TrieTimeLimit {
+					if chosen < bc.lastWrite+triesInMemory && bc.gcproc >= 2*bc.cacheConfig.TrieTimeLimit {
 						utils.Logger().Info().
 							Dur("time", bc.gcproc).
 							Dur("allowance", bc.cacheConfig.TrieTimeLimit).
-							Float64("optimum", float64(chosen-lastWrite)/triesInMemory).
+							Float64("optimum", float64(chosen-bc.lastWrite)/triesInMemory).
 							Msg("State in memory for too long, committing")
 					}
+					// before flush next state tire, must ensure pre-pruning finished
```
MaxMustermann2, line 1267: nit: `tire` -> `trie`

core/blockchain_impl.go
```diff
@@ -1341,6 +1350,45 @@ func (bc *BlockChainImpl) WriteBlockWithState(
 	return CanonStatTy, nil
 }
 
+// pruneOldTrieAsync prune old trie in async.
+// The task must be completed before the next call, or it will be aborted.
+// Example of why:
+// At State-1, A=1.
+// At State-2, A=0. This means A will be removed from the trie.
+// At State-3, A=1.
+// Assuming that the pruning of (State-1, State-2) is performed after State-3 is flushed to disk, the A will be removed. State-3 will have the wrong value for A.
```
MaxMustermann2, line 1359: Interesting. Do we have a test case for this?

core/state/state_prune.go
```diff
@@ -0,0 +1,81 @@
...
+			// if old state is empty, skip
+			if oldAccount.Root == emptyRoot || oldAccount.Root == (common.Hash{}) {
+				continue
+			}
+
+			// create account difference iterator
+			oldAccountTrie := newObject(oldDB, addr, oldAccount).getTrie(oldDB.db)
+			newAccountTrie := newObject(newDB, addr, newAccount).getTrie(newDB.db)
+
+			leafKeyCache, _ := lru.New(32 * 1024)
```
MaxMustermann2, line 57: How is this size determined?

core/state/state_prune.go
```diff
@@ -0,0 +1,81 @@
...
+					return count, errors.New("aborted")
+				default:
+				}
+				if newTrieIt.Leaf() {
+					leafKeyCache.Add(string(newTrieIt.LeafKey()), struct{}{})
+				}
+				if !accountDifferIt.Leaf() {
+					count++
+					batch.Delete(accountDifferIt.Hash().Bytes())
+				} else if !leafKeyCache.Contains(string(accountDifferIt.LeafKey())) {
```
MaxMustermann2, line 73: Please add a comment explaining this.

MaxMustermann2 (commented):
Looks good overall; needs minor changes.

core/blockchain_impl.go
peekpi: no, here is waiting for the previous pruning to complete, we must ensure the pruning is finished before next state writing.

core/blockchain_impl.go
peekpi: yeah, in `TestDiffAndPrune`, It writes dataset1, then dataset2, then dataset1 again.

core/state/state_prune.go
peekpi: I didn't calculate it. but I think 32k is enough.

ONECasey:
@peekpi , is there anyway we can get an update on this PR and possibly fix the conflicts on the following files:

 cmd/harmony/config_migrations.go
cmd/harmony/default.go
core/blockchain_impl.go
internal/shardchain/shardchains.go

peekpi:
> @peekpi , is there anyway we can get an update on this PR and possibly fix the conflicts on the following files:
> 
> 
> 
>  cmd/harmony/config_migrations.go
> 
> cmd/harmony/default.go
> 
> core/blockchain_impl.go
> 
> internal/shardchain/shardchains.go 

yeah, I can fix it.
But there is a problem with this pr that is difficult to solve... If two contracts have the same state, they share the same state tire. If the data of one contract changes, we cannot prune it, otherwise the data of the other contract will be missing.

MaxMustermann2:
> If two contracts have the same state, they share the same state tire. If the data of one contract changes, we cannot prune it, otherwise the data of the other contract will be missing.

Is there no reference counting for these states that can be used to assess whether it can be deleted? In other words, if there is a state with two references, it should not be deleted.

PR #4300, opened 2022-11-01
Feature thread-safe consensus. 

Frozen:
## Issue

Current consensus uses both mutex and channel synchronization, but doesn't do it properly. Race detector complains about concurrent access to properties. 

## Solution
Pr removes channel synchronization, only mutex available. Consensus divided on public methods with locks, and private without locks. No race detector complains.


ONECasey (changes requested):
Please fix merge conflicts.

ONECasey (approved):

consensus/checks.go
```diff
@@ -56,9 +56,10 @@ func (consensus *Consensus) senderKeySanityChecks(msg *msg_pb.Message, senderKey
 }
 
 func (consensus *Consensus) isRightBlockNumAndViewID(recvMsg *FBFTMessage) bool {
-	if recvMsg.ViewID != consensus.GetCurBlockViewID() || recvMsg.BlockNum != consensus.BlockNum() {
+	blockNum := consensus.getBlockNum()
```
MaxMustermann2, line 59: What is the difference between `consensus.getBlockNum()` and `consensus.BlockNum()` ?
https://github.com/harmony-one/harmony/blob/258aff0630b9029b35278cfa7ec71a9e232da77e/consensus/consensus.go#L223-L229

consensus/consensus.go
```diff
@@ -200,7 +196,7 @@ func (consensus *Consensus) GetPrivateKeys() multibls.PrivateKeys {
 }
 
 // GetLeaderPrivateKey returns leader private key if node is the leader
-func (consensus *Consensus) GetLeaderPrivateKey(leaderKey *bls_core.PublicKey) (*bls.PrivateKeyWrapper, error) {
+func (consensus *Consensus) getLeaderPrivateKey(leaderKey *bls_core.PublicKey) (*bls.PrivateKeyWrapper, error) {
```
MaxMustermann2, line 199: Maybe change the name of this function to `getPrivateKeyFromPublic`

consensus/checks.go
Frozen: Naming. Just to be sure no public calls from private methods.

consensus/consensus.go
Frozen: I don't see any advantages or disadvantages from this renaming, i will leave it as is.

sophoah:
there are some stagesync code @GheisMohammadi could you help review this as well ?

sophoah:
travis-ci already worked so this is already a good sign that consensus is fine however the PR is also touching on view change. So here a test proposal : 
- use diego's (active testnet validator)  node and confirm consensus is still working in testnet for his node
- trigger a view change in testnet and confirm it everything still works
- monitor for 1 day (ie at least 2 epoch happened) with no issue

if no issue for the above test / comment addressed / @GheisMohammadi reviewed, let's merge this PR and test it on devnet

consensus/consensus_fbft.go
```diff
@@ -1,9 +1,6 @@
 package consensus
 
-import "sync"
-
 type LockedFBFTPhase struct {
-	mu    sync.Mutex
 	phase FBFTPhase
 }
```
GheisMohammadi, line 5: if there is no lock here, then how this struct helps? Also, The name like **LockedFBFTPhase** doesn't match with what it is doing. Basically we don't need this file and could be deleted and LockedFBFTPhase could be replaced with a simple **FBFTPhase** variable everywhere in codes.

consensus/consensus_fbft.go
```diff
@@ -14,14 +11,10 @@ func NewLockedFBFTPhase(initialPhrase FBFTPhase) *LockedFBFTPhase {
...
 
 func (a *LockedFBFTPhase) Set(phrase FBFTPhase) {
```
GheisMohammadi, line 13: I think the parameter name should be  **phase** , please spell check

consensus/consensus_v2.go
```diff
@@ -290,168 +298,168 @@ func (consensus *Consensus) BlockCommitSigs(blockNum uint64) ([]byte, error) {
...
+		go func() {
+			ticker := time.NewTicker(250 * time.Millisecond)
+			defer ticker.Stop()
+			for {
+				select {
+				case <-stopChan:
+					return
+				case <-ticker.C:
+					consensus.mutex.Lock()
+					consensus.tick()
```
GheisMohammadi, line 314: I think if you move mutex inside tick() function would be better. Because there you can use defer to unlock mutex. So, for any reason if tick() function fails, mutex would be unlocked. In current implementation if tick() fails, mutex will remain locked.

consensus/consensus_v2.go
```diff
@@ -468,24 +476,29 @@ type LastMileBlockIter struct {
 }
 
 // GetLastMileBlockIter get the iterator of the last mile blocks starting from number bnStart
-func (consensus *Consensus) GetLastMileBlockIter(bnStart uint64) (*LastMileBlockIter, error) {
+func (consensus *Consensus) GetLastMileBlockIter(bnStart uint64, cb func(iter *LastMileBlockIter) error) error {
 	consensus.mutex.Lock()
```
GheisMohammadi, line 480: can we use RLock?

consensus/consensus_v2.go
```diff
@@ -290,168 +298,168 @@ func (consensus *Consensus) BlockCommitSigs(blockNum uint64) ([]byte, error) {
...
 	if consensus.dHelper != nil {
-		consensus.dHelper.close()
+		consensus.dHelper.start()
 	}
-	consensus.waitForCommit()
-	return nil
 }
 
 // waitForCommit wait extra 2 seconds for commit phase to finish
 func (consensus *Consensus) waitForCommit() {
```
GheisMohammadi, line 454: where we use this function?

consensus/consensus.go
```diff
@@ -157,12 +153,14 @@ func (consensus *Consensus) VerifyBlock(block *types.Block) error {
 // BlocksSynchronized lets the main loop know that block synchronization finished
 // thus the blockchain is likely to be up to date.
 func (consensus *Consensus) BlocksSynchronized() {
-	consensus.syncReadyChan <- struct{}{}
+	consensus.mutex.Lock()
+	consensus.syncReadyChan()
+	consensus.mutex.Unlock()
```
GheisMohammadi, line 158: using defer to unlock would be safer

GheisMohammadi (changes requested):

consensus/consensus_fbft.go
Frozen: Removed

consensus/consensus_fbft.go
Frozen: Removed

consensus/consensus_v2.go
Frozen: No, it inserts into blockchain inside.

consensus/consensus.go
Frozen: fixed

consensus/consensus_v2.go
Frozen: Dead code, removed.

consensus/consensus_v2.go
Frozen: fixed

GheisMohammadi (approved):

internal/params/config.go
```diff
@@ -271,9 +271,9 @@ var (
 		SlotsLimitedEpoch:             EpochTBD, // epoch to enable HIP-16
 		CrossShardXferPrecompileEpoch: big.NewInt(1),
 		AllowlistEpoch:                EpochTBD,
-		FeeCollectEpoch:               big.NewInt(5),
-		LeaderRotationEpoch:           EpochTBD,
+		LeaderRotationEpoch:           big.NewInt(4),
```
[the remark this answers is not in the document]
Frozen: i disabled this feature temporary.

Frozen:
I've checked view change, it switched successfully.

MaxMustermann2 (approved):

PR #4302, opened 2022-11-11
add openssl compatibility on m2 chips using darwin

ONECasey:
Adds support for OpenSSL on macOS Ventura using m2 chips. This PR requires a commit on the `mcl` that includes SSL header files necessary to compile on m1/m2 systems.


ONECasey:
Here is the PR on the `mcl` repository. 

https://github.com/harmony-one/mcl/pull/6

sophoah (approved):

PR #4307, opened 2022-11-17
Fix for consensus stuck.

Frozen:
We can see that error happened on `onNewView`:

`{"level":"warn","myBlock":33980416,"myViewID":33984024,"phase":"Commit","mode":"ViewChanging","error":"Block verification failed: crosslink already exist","caller":"/home/sp/harmony/harmony/consen
sus/view_change.go:496","time":"2022-11-15T05:44:05.852088781Z","message":"[onNewView] Verify New View Msg Failed"}`

`onNewView` is final step of view change mechanism, and block is already invalid. Previous step is `startNewView`, so i added check there for no unchecked blocks sending.

### DIFF
--- a/consensus/validator.go
+++ b/consensus/validator.go
@@ -62,15 +62,12 @@ func (consensus *Consensus) onAnnounce(msg *msg_pb.Message) {
 	consensus.switchPhase("Announce", FBFTPrepare)
 
 	if len(recvMsg.Block) > 0 {
-		go func() {
-			// Best effort check, no need to error out.
-			_, err := consensus.validateNewBlock(recvMsg)
-
-			if err == nil {
-				consensus.getLogger().Info().
-					Msg("[Announce] Block verified")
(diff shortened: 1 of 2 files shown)


sophoah:
@Frozen could you update the PR description explaining the why, how, this PR help.

ONECasey:
I agree with @sophoah. All my tests and local deployments work well with this change but still a little confused about how this fixes the issue.

MaxMustermann2:
Can you perhaps provide an example to reproduce the issue in localnet? You suggestion on Notion that the bug can be replicated by creating "empty blocks to send".

Frozen:
I can't reproduce same error on localnet.

MaxMustermann2 (approved):

Frozen:
Starts new view change if block invalid.

MaxMustermann2 (approved):

PR #3906, opened 2021-10-22
Resolve harmony-one/bounties#77: Staking precompiles

MaxMustermann2:
This PR adds staking precompiled contracts to the Harmony EVM, at address 0xFC after a hard fork at `EpochTBD`. The precompiles can be called from Solidity by using `abi.encodeWithSelector` on the required parameters to produce the arguments for assembly `call`.  This will allow any contract to participate in staking. 3/5 directives (`Delegate`, `Undelegate` and `CollectRewards`  are accessible from the precompiled contract, which follows exactly the same logic as the original`ApplyStakingMessage`.

A sample Solidity implementation is outlined [here](https://github.com/MaxMustermann2/harmony-staking-precompiles). The repository also contains integration tests for the (modified) node; the test report is available below. Note that a contract may only use the precompile for its own address; otherwise, the node will return an error indicating address mismatch. One instance of such a test case is covered in the sample implementation as `test_delegate_fail_malicious`.  Additionally, I have added a test case to ensure that the precompile can be called multiple times in one transaction (`test_multiple_calls_success`). This can be used, for example, by a staking contract, to delegate 50% of its ONE to one validator, 25% to another, and 25% to a third validator - all in a single transaction hash.

#### Commit message:
Create write capable precompiles that can perform staking transactions
Add hard fork logic (EpochTBD) for these precompiles
Tests for new code with at least 80% unit test coverage
Staking library + tests in [MaxMustermann2/harmony-staking-precompiles](https://github.com/MaxMustermann2/harmony-staking-precompiles)

## Issue
[harmony-one/bounties#77](https://github.com/harmony-one/bounties/issues/77) | [Gitcoin](https://gitcoin.co/issue/harmony-one/bounties/77/100026734)

### 1.1 Core protocol
- [x] Add staking logic to precompiles.
    - [x] Precompiles can be directly called with staking data as arguments.
    - [x] The execution of the precompiles will return result and error as necessary.
   - [x]  The execution logic will be exactly the same as ApplyStakingTransaction.
    - [x] Staking directive:
        - [ ] ~~Create Validator~~
        - [ ] ~~Edit Validator~~
        - [x] Delegate
        - [x] Undelegate
        - [x] Collect Rewards
- [x] Add hard fork logic for the new precompiles.
    - [x] Aggregate the new precompile contracts to a new set.
    - [x] Add the fork logic in evm run and Call.
    - [x] Add the fork epoch number to localnet at LocalnetChainConfig.
    - [x] Misc
[...]

PR #4308, opened 2022-11-21
staking: enable migration precompile

MaxMustermann2:
This change [re-](https://github.com/harmony-one/harmony/pull/3906#issuecomment-1014158304)adds support for staking migration to the staking precompile located at `0xFC`. Staking migration is defined as the transfer of all delegated balance, undelegated balance, and pending rewards from address A to address B. In the event that a wallet's private key is compromised, a single transaction can transfer all the staked balance (+ rewards) to a new, clean address. 

We have seen a few events where attackers undelegate the balances of compromised users, and wait for the 7 epoch period to end. A user who notices this transaction on time may redelegate their balance to save it, but no other transaction is possible because the malicious attacker can steal any unlocked funds. This ends up becoming a cat and mouse game where the attacker undelegates and the owner delegates, on repeat. This change is built to work around that, by allowing users to transfer their (locked) funds to a new address.

### DIFF
--- a/core/evm.go
+++ b/core/evm.go
@@ -97,6 +97,7 @@ func NewEVMContext(msg Message, header *block.Header, chain ChainContext, author
 		Delegate:              DelegateFn(header, chain),
 		Undelegate:            UndelegateFn(header, chain),
 		CollectRewards:        CollectRewardsFn(header, chain),
+		MigrateDelegations:    MigrateDelegationsFn(header, chain),
 		CalculateMigrationGas: CalculateMigrationGasFn(chain),
 		ShardID:               chain.ShardID(),
 		NumShards:             shard.Schedule.InstanceForEpoch(header.Epoch()).NumShards(),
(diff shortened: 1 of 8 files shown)


MaxMustermann2:
The migration of delegated / locked balances can be transferred by anyone (original owner or malicious attacker, whoever has access to the private key), to a new address. Unless the original owner immediately, upon compromise of the key, makes such a transaction, the attacker can create the same transaction to an address controlled by them. 

We can work around the above hole by making use of the denylist / allowlist of transactions, but that can be done without needing a migration precompile in the first place. My suggestion, therefore, is to not merge it, unless we plan to add another level of (possibly centralized) filtering to such a transaction.

GheisMohammadi:
This is very interesting challenge. The concern is this feature is a double-edged sword. It can be used by attacker as well to drain the account as fast as possible. If a user is not aware of such a feature and attacker do some phishing attack to get him sign the transaction, account owner will lose everything. Do you think this concern is right?

sophoah:
the same non precompile feature was suggested into our protocol at one point, we decided to not implement for the reason mentioned above.

I am also against to implement this. 7 days locked staking is also a safety feature against private key leak

MaxMustermann2:
BREAKING CHANGES: Following the hard fork, (1) the staking precompile
will not accept any transactions with a `msg.value` attached to them;
and (2) the event topic hash of CollectRewards will change to include
the return type; and lastly, (3) the event data will be correctly padded
to 32 bytes, or `uint256`.

ONECasey:
After careful review with the rest of the team, we agreed not to push this feature forward at this time.

PR #4309, opened 2022-11-22
[rpc] query node for syncing calls

### LINKED ISSUE #3966: hmy_syncing & hmyv2_syncing return wrong payload
**Describe the bug**
hmy_syncing & hmyv2_syncing (v1 & v2) based on the documentation should work like eth_syncing, when the node is syncing, it should return a result key with an object, when the node is sync it will return a false.

In both case here this rpc method return false even when the node is a lot behind of the shard height. In my case my node was 100k behind and it return "false" that mean "sync is done"

**To Reproduce**
1. Start a node
2. Run a curl request to it
```bash
curl -X POST --location "http://localhost:9500" \
    -H "Content-Type: application/json" \
    -d "{
        	\"jsonrpc\":\"2.0\",
        	\"method\":\"hmyv2_syncing\",
        	\"params\":[],
        	\"id\":0
        }" 
```
**Expected behavior**
Base on the documentation when the node is syncing it should return an object on the result key.
When the node sync is done, it should return false.

**Environment (please complete the following information):**
 - OS: Ubuntu 20.04
 - Docker - `docker pull pocketfoundation/harmony` try with 4.3.1 and testnet

MaxMustermann2:
Resolves #3966

### DIFF
--- a/hmy/downloader/downloader.go
+++ b/hmy/downloader/downloader.go
@@ -125,7 +125,8 @@ func (d *Downloader) SyncStatus() (bool, uint64, uint64) {
 	if !syncing { // means synced
 		target = current
 	}
-	return syncing, target, current - target
+	// isSyncing, target, blocks to target
+	return syncing, target, target - current
 }
 
 // SubscribeDownloadStarted subscribe download started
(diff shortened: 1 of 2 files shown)


ONECasey (approved):
Fixes issues and returns proper payload.

PR #4310, opened 2022-11-23
Remove blockchain dependency from engine.

Frozen:
`Engine` is used like a storage for beaconChain, but doesn't use beaconChain internally.


MaxMustermann2 (approved):

ONECasey (approved):
Looks good! Nice work.

PR #4311, opened 2022-11-28
[dumpdb] ensure each cross link is dumped

MaxMustermann2:
## Issue
This pull request fixes the root cause behind the recent consensus stuck issue.
 - `dumpdb` does not write the crosslink to disk.
 - a new node is set up with `snapdb` (obtained from `dumpdb` command).
 - it receives an old crosslink, likely over the wire, from a syncing s1/2/3 node.
 - it adds this crosslink to the store of pending crosslinks, while a valid node would have dropped such a crosslink. 
 - the pending crosslinks are saved to disk and reloaded each time the node restarts.
 - with time more invalid crosslinks are stored in the list, since the list is capped at 1000 of them.
 - when this node becomes a leader, it proposes a block with invalid crosslink.
 - other nodes, which already have this crosslink, reject the block - but do not trigger a view change.
 - some nodes, which were loaded from the same `snapdb` accept the block and return a `prepare` message. however, it does not have enough voting power for the block to be considered valid.
 - consensus times out, and the next leader broadcasts a view change message.
 - some nodes respond to the view change message but others still consider the previous block as valid and can ignore the view change.
 - view change times out as well.
 - repeat until a valid node has its view change proposal accepted. it then proposes a valid block, which restores consensus.

### DIFF
--- a/cmd/harmony/dumpdb.go
+++ b/cmd/harmony/dumpdb.go
@@ -15,12 +15,14 @@ import (
 	ethRawDB "github.com/ethereum/go-ethereum/core/rawdb"
 	"github.com/ethereum/go-ethereum/ethdb"
 	"github.com/ethereum/go-ethereum/params"
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/harmony-one/harmony/block"
 	"github.com/harmony-one/harmony/core/rawdb"
 	"github.com/harmony-one/harmony/core/state"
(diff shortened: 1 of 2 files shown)


sophoah (approved):

ONECasey (approved):

ONECasey:
@MaxMustermann2 can you please verify all commits are signed to push this PR forward?

PR #4324, opened 2022-12-29
Feature registry

Frozen:
Registry for services. First moved `core.Blockchain`, because node can change it's shard during epoch change.


ONECasey (changes requested):
Seems to be a few issues with the tests. Able to check it out and make necessary changes?

ONECasey (approved):
Tests passed. Thanks @Frozen

PR #4356, opened 2023-02-03
fix max height issue

GheisMohammadi:
## Issue
This PR fixes the eth_syncing rpc, which sometimes returns the wrong maximum height, and makes the rpc result exactly the same as Ethereum's. The issue was that some nodes were not able to query other nodes for block height. It's mostly because of the instability of the network connection or not getting proper responses on time from other peers. This PR tries to keep the last sync status in cache for 6s, and it only updates it if the returned height is valid; otherwise, it uses the last valid sync status.

## Test
Call the rpc and check the max block height:
`curl -X POST --data '{"jsonrpc":"2.0","method":"eth_syncing","params":[],"id":1}'`

### DIFF
--- a/api/service/legacysync/helpers.go
+++ b/api/service/legacysync/helpers.go
@@ -14,7 +14,7 @@ import (
 )
 
 // getMaxPeerHeight gets the maximum blockchain heights from peers
-func getMaxPeerHeight(syncConfig *SyncConfig) uint64 {
+func getMaxPeerHeight(syncConfig *SyncConfig) (uint64, error) {
 	maxHeight := uint64(math.MaxUint64)
 	var (
 		wg   sync.WaitGroup
@@ -49,7 +49,12 @@ func getMaxPeerHeight(syncConfig *SyncConfig) uint64 {
 		return
 	})
 	wg.Wait()
-	return maxHeight
+
+	if maxHeight == uint64(math.MaxUint64) {
+		return 0, fmt.Errorf("[SYNC] get max peer height failed")
+	}
+
+	return maxHeight, nil
 }
 
 func createSyncConfig(syncConfig *SyncConfig, peers []p2p.Peer, shardID uint32, selfPeerID libp2p_peer.ID, waitForEachPeerToConnect bool) (*SyncConfig, error) {
--- a/api/service/legacysync/syncing.go
+++ b/api/service/legacysync/syncing.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"fmt"
+	"math"
 	"math/rand"
 	"reflect"
 	"sort"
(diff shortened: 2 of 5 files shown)


api/service/legacysync/syncing.go
```diff
@@ -1079,9 +1083,9 @@ func (ss *StateSync) SyncLoop(bc core.BlockChain, worker *worker.Worker, isBeaco
 
 	for {
 		start := time.Now()
-		otherHeight := getMaxPeerHeight(ss.syncConfig)
 		currentHeight := bc.CurrentBlock().NumberU64()
-		if currentHeight >= otherHeight {
+		otherHeight, errMaxHeight := getMaxPeerHeight(ss.syncConfig)
+		if errMaxHeight != nil || currentHeight >= otherHeight {
```
MaxMustermann2, line 1088: Add a separate log here if `errMaxHeight != nil` so that we can easily `grep` the occurrence of a syncing failure. It happens if the node forks, or if the DNS is pointed to nodes with snap database.

MaxMustermann2 (commented):
Just one edit needed, thanks.

api/service/legacysync/syncing.go
GheisMohammadi: added

MaxMustermann2 (approved):

MaxMustermann2:
Please wait for @Frozen's review before merging. Thank you.

ONECasey (approved):

api/service/legacysync/helpers.go
```diff
@@ -49,7 +49,12 @@ func getMaxPeerHeight(syncConfig *SyncConfig) uint64 {
 		return
 	})
 	wg.Wait()
-	return maxHeight
+
+	if maxHeight == uint64(math.MaxUint64) {
+		return 0, fmt.Errorf("[SYNC] get max peer height failed")
```
Frozen, line 54:
```suggestion
		return 0, fmt.Errorf("failed to get max peer height")
```

api/service/stagedsync/sync_status.go
```diff
@@ -75,7 +76,9 @@ func (status *syncStatus) Get(fallback func() SyncCheckResult) SyncCheckResult {
 	defer status.lock.Unlock()
 	if status.expired() {
 		result := fallback()
-		status.update(result)
+		if result.OtherHeight > 0 && result.OtherHeight < uint64(math.MaxUint64) {
```
Frozen, line 79: if you expect this is possible, please log otherwise case.

Frozen:
Getting height is really tricky, i got several bugs here too. LGTM

Frozen (approved):

api/service/stagedsync/sync_status.go
GheisMohammadi: getMaxHeight returns either **0**  or **math.MaxUint64**, in case it is not able to query other peers. So, here, first we validate the return value, and if it is valid, we update the status. The last code was updating the status regardless of the validity of the returned value.

PR #4359, opened 2023-02-09
Leader rotation. Check next leader aliveness.

Frozen:
Checks last 10 blocks. If no sigs from future leader - skip.


ONECasey (approved):

ONECasey:
@Frozen Please fix these conflicts so we can merge for the next release.

PR #4363, opened 2023-02-13
Fix p2p host discovery for legacy sync

GheisMohammadi:
## Issue

This PR fixes the discovery issue for legacy sync. It also adds two new options to adjust the configurations of host connection manager.


rosetta/infra/harmony-mainnet.conf
```diff
@@ -68,6 +68,8 @@ Version = "2.5.11"
   MaxConnsPerIP = 10
   MaxPeers = 0
   Port = 9000
+  ConnManagerLowWatermark = 160
+	ConnManagerHighWatermark = 192
```
MaxMustermann2, line 72: Dedent please

rosetta/infra/harmony-pstn.conf
```diff
@@ -68,6 +68,8 @@ Version = "2.5.11"
   MaxConnsPerIP = 10
   MaxPeers = 0
   Port = 9000
+  ConnManagerLowWatermark = 160
+	ConnManagerHighWatermark = 192
```
MaxMustermann2, line 72: Dedent please

p2p/host.go
```diff
@@ -286,6 +279,33 @@ func NewHost(cfg HostConfig) (Host, error) {
 	return h, nil
 }
 
+// connectionManager creates a new connection manager and configures libp2p to use the
+// given connection manager.
+// lo and hi are watermarks governing the number of connections that'll be maintained.
+// When the peer count exceeds the 'high watermark', as many peers will be pruned (and
+// their connections terminated) until 'low watermark' peers remain.
+func connectionManager(low int, high int) libp2p_config.Option {
+	if low > 0 && high > low {
```
MaxMustermann2, line 288: Isn't high and low already checked above ?

Frozen (approved):

p2p/host.go
GheisMohammadi: We're looking to see if high>low at the top. but here we are making sure they are set to be used by connection manager; otherwise, we don't initiate connection manager. In this way, using Connection Manager will be an optional feature. If we don't want to use it, just set both high and low to zero.

MaxMustermann2 (approved):

sophoah:
I am curious here, is that connManager redundant with :
```
p2p.MaxConnsPerIP
p2p.MaxPeers
```

sophoah:
following up the previous comment, any risk of conflict ?

internal/configs/node/network.go
```diff
@@ -63,7 +63,13 @@ const (
 	DefaultMaxConnPerIP = 10
 	// DefaultMaxPeers is the maximum number of remote peers, with 0 representing no limit
 	DefaultMaxPeers = 0
-	// DefaultWaitForEachPeerToConnect sets the sync configs to connect to neighbor peers one by one and waits for each peer to connect
+	// DefaultConnManagerLowWatermark is the lowest number of connections that'll be maintained in connection manager
+	DefaultConnManagerLowWatermark = 160
+	// DefaultConnManagerHighWatermark is the highest number of connections that'll be maintained in connection manager
+	// When the peer count exceeds the 'high watermark', as many peers will be pruned (and
+	// their connections terminated) until 'low watermark' peers remain.
+	DefaultConnManagerHighWatermark = 192
```
sophoah, line 71: so in here, it kind of conflict with the `DefaultMaxPeers = 0` above

p2p/host.go
```diff
@@ -141,16 +130,20 @@ func NewHost(cfg HostConfig) (Host, error) {
 	)
 
 	ctx, cancel := context.WithCancel(context.Background())
-	// TODO: move low and high to configs
-	connmgr, err := connmgr.NewConnManager(
-		int(cfg.MaxConnPerIP),      // LowWater
```
sophoah, line 146: is LowWater/HighWater, differentiate a host IP that has multiple instance of the harmony binary with p2p ID

GheisMohammadi:
> I am curious here, is that connManager redundant with :
> 
> ```
> p2p.MaxConnsPerIP
> p2p.MaxPeers
> ```

Actually those configs are remained from old codes and we are not using them. So, I think we can remove those dead codes later.

GheisMohammadi:
> following up the previous comment, any risk of conflict ?

Actually these two p2p options (MaxConnsPerIP, MaxPeers) are remained from old codes. We are connection manager to handle connections and so, we can remove dead codes later.

internal/configs/node/network.go
GheisMohammadi: > I am curious here, is that connManager redundant with :
> 
> ```
> p2p.MaxConnsPerIP
> p2p.MaxPeers
> ```

these two options are for security manager which is able to control the number of connections per each IP. So, any IP can't have more than a certain number of connections to node. 
ibp2p peer ID is based only on RSA public key. This allows an attacker to create unlimited number of nodes on a single machine (same IP address) and use these nodes to monopolize the incoming/outgoing connections of victim nodes. That's why we use security manager.

GheisMohammadi:
> following up the previous comment, any risk of conflict ?

Connection manager cares about total number of connections regardless of IP. connection manager has a range of connections and if connections get more than that, it will close extra connections. Security even won't allow to peer get connected if same IP has more connections than threshold. Technically, they act as two different security layer

sophoah (approved):

sophoah:
Considering this PR has been successfully tested on a testnet node by @GheisMohammadi. all reviews were completed, checked all passed, I'll merge this one

PR #4368, opened 2023-02-19
Fixed Bitmap race error.

Frozen:
Wrapped Bitmap in thread safe struct.

### DIFF
--- a/api/service/explorer/storage.go
+++ b/api/service/explorer/storage.go
@@ -34,11 +34,54 @@ const (
 // explorer db is doing migration and unavailable
 var ErrExplorerNotReady = errors.New("explorer db not ready")
 
+type Bitmap interface {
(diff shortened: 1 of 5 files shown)


api/service/explorer/storage.go
```diff
@@ -89,11 +132,12 @@ func newStorage(hc *harmonyconfig.HarmonyConfig, bc core.BlockChain, dbPath stri
 		return nil, err
 	}
 
+	safeBitmap := NewThreadSafeBitmap(bitmap)
 	return &storage{
 		db:        db,
 		bc:        bc,
-		rb:        bitmap,
-		tm:        newTaskManager(bitmap),
```
Frozen, line 96: Here we pass bitmap to another thread.

ONECasey (approved):

PR #4370, opened 2023-02-21
Removed unused method.

Frozen:
Removed unused method.


ONECasey:
What was the original purpose of this method?

ONECasey (approved):

PR #4380, opened 2023-03-16
Feature disable internal/external validators.

Frozen:


Frozen:
This is one of alternative branches how leader rotation could be, not relevant.

PR #4381, opened 2023-03-16
evm: don't return extcode for validators

MaxMustermann2:
Due to technical debt, validator information is stored in the code field of the address. The code field can be accessed in Solidity for an arbitrary address using `extcodesize`, `extcodehash`, and `extcodecopy` or helper commands (such as `address.code.Length`). The presence of this field is used by contract developers to (erroneously) deny smart contract access to other smart contracts (and therefore, validators). This PR fixes that oversight by returning the same values as other EOAs for known validator addresses. Obviously, it needs a hard fork that will be scheduled separately.

### DIFF
--- a/core/vm/instructions.go
+++ b/core/vm/instructions.go
@@ -24,6 +24,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/math"
 	"github.com/harmony-one/harmony/core/types"
 	"github.com/harmony-one/harmony/internal/params"
+	"github.com/harmony-one/harmony/shard"
 	"golang.org/x/crypto/sha3"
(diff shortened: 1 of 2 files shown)


ONECasey:
Awesome fix @MaxMustermann2 .

Let's hold off on merging this until the next hard fork. We have a temporary fix for the meantime but this will allow future validators to access smart contracts.

ONECasey:
Not activated. Merging

ONECasey (approved):

sophoah:
testing : https://github.com/harmony-one/harmony/pull/4456#issuecomment-1623159350

PR #4392, opened 2023-03-28
Revert "Removed engine dependency."

ONECasey:
Reverts harmony-one/harmony#4385


GheisMohammadi:
what's the reason for the revert?

PR #4402, opened 2023-04-06
Add prefix for validator wrapper

GheisMohammadi:
## Issue
The validator wrapper is encoded to bytes and then is saved in the code field of the account. Following latest changes, this information will be prefixed with c. This PR sets a different prefix for validator code. Using a separate prefix helps to distinguish the data on some level if it's needed later. There is old validator wrapper information that's not really useful right now but it might still be stored in there. This PR helps it to be identified via the prefix and then pruned.

## Operational Checklist

    **NO**


ONECasey (approved):

core/rawdb/accessors_state.go
```diff
@@ -93,3 +93,57 @@ func DeleteCode(db ethdb.KeyValueWriter, hash common.Hash) {
...
+	data, _ := db.Get(validatorCodeKey(hash))
+	return data
+}
+
+// HasValidatorCode checks if the validator code corresponding to the
+// provided code hash is present in the db.
+func HasValidatorCode(db ethdb.KeyValueReader, hash common.Hash) bool {
+	// Try with the prefixed code scheme first, if not then try with legacy
+	// scheme.
+	if ok := HasValidatorCodeWithPrefix(db, hash); ok {
```
MaxMustermann2, line 122: Why not simply use golang's in-built short circuiting? `return HasValidatorCodeWithPrefix(db, hash) || db.Has(Hash.Bytes())`

PR #4404, opened 2023-04-08
Consensus: removed double initialization.

Frozen:
A good example of double initialization and recursive dependency is the `Node` and `Consensus` structs. The `Node` struct has a method called `SetConsensus(Consensus)`, and the `Consensus` struct has a method called `SetNode(Node)`. These structs cannot be fully initialized in the constructor and require additional initialization steps through methods, so they can serve requests even if not initialized properly. Another issue is that the `Consensus` can internally call `Node` with calls `Consensus` back, and it can stuck on mutex double lock. This sort of bugs called `reentrant lock bugs`, and we have already faced with several of them. 

My goal is to fully initialize Consensus first and then pass it to Node in the constructor without requiring any additional actions.

### DIFF
--- a/api/service/explorer/service.go
+++ b/api/service/explorer/service.go
@@ -190,21 +190,20 @@ type HeightResponse struct {
 func (s *Service) GetHeight(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 
-	blockchain := s.backend.Blockchain()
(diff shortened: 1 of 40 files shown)


GheisMohammadi:
could you please add a description for the PR? What you mean by double initialization and how this PR removes it? Was there an issue or this is an improvement for initialization?

ONECasey (approved):

PR #4410, opened 2023-04-21
core, internal/configs: HIP28-v2 fee collection

MaxMustermann2:
Based on the Snapshot vote that has passed, collect 50% of the fee to a community maintained account and the remainder to an account used to pay for infrastructure costs. Note that these accounts need to be decided and set in the code at this moment, and the feature needs to be activated by setting the `FeeCollectEpoch` of the `ChainConfig` object.

The implementation for devnet is a bit different than compared to others because the feature was activated on devnet with 100% collection to an account. I have handled this case separately in `devnet.go`.

### DIFF
--- a/internal/params/config.go
+++ b/internal/params/config.go
@@ -193,10 +193,10 @@ var (
 		SlotsLimitedEpoch:                      EpochTBD, // epoch to enable HIP-16
 		CrossShardXferPrecompileEpoch:          big.NewInt(1),
 		AllowlistEpoch:                         EpochTBD,
-		FeeCollectEpoch:                        EpochTBD,
+		FeeCollectEpoch:                        big.NewInt(848), // 2023-04-28 04:33:33+00:00
 		LeaderRotationExternalNonBeaconLeaders: EpochTBD,
 		LeaderRotationExternalBeaconLeaders:    EpochTBD,
-		ValidatorCodeFixEpoch:                  EpochTBD,
+		ValidatorCodeFixEpoch:                  big.NewInt(848),
 	}
 
 	// StressnetChainConfig contains the chain parameters for the Stress test network.


internal/configs/sharding/localnet.go
```diff
@@ -14,8 +14,12 @@ import (
...
 var LocalnetSchedule localnetSchedule
 
-// privatekey: 0x1111111111111111111111111111111111111111111111111111111111111111
-var feeCollectorLocalnet = mustAddress("0x19E7E376E7C213B7E7e7e46cc70A5dD086DAff2A")
+var feeCollectorsLocalnet = FeeCollectors{
+	// 0x1111111111111111111111111111111111111111111111111111111111111111
+	mustAddress("0x19E7E376E7C213B7E7e7e46cc70A5dD086DAff2A"): numeric.MustNewDecFromStr("0.5"),
+	// 0x2222222222222222222222222222222222222222222222222222222222222222
+	mustAddress("0x1563915e194D8CfBA1943570603F7606A3115508"): numeric.MustNewDecFromStr("0.5"),
+}
```
sophoah, line 22: are the private key or mnemonic of the collector in localnet somewhere available in the code ?

internal/configs/sharding/partner.go
```diff
@@ -13,7 +13,17 @@ import (
...
-var feeCollectorDevnet = mustAddress("0xb728AEaBF60fD01816ee9e756c18bc01dC91ba5D")
+var feeCollectEpochV1 = big.NewInt(574)
+
+var feeCollectorsDevnet = []FeeCollectors{
+	FeeCollectors{
+		mustAddress("0xb728AEaBF60fD01816ee9e756c18bc01dC91ba5D"): numeric.OneDec(),
+	},
+	FeeCollectors{
+		mustAddress("0xb728AEaBF60fD01816ee9e756c18bc01dC91ba5D"): numeric.MustNewDecFromStr("0.5"),
+		mustAddress("0xb41B6B8d9e68fD44caC8342BC2EEf4D59531d7d7"): numeric.MustNewDecFromStr("0.5"),
```
sophoah, line 24: we should have the private keys/mnemonic of those account available to us

sophoah (commented):
is there a way to actually develop unit test here for the fee collection ? or it can only be done with a live network (ie localnet)

ONECasey (approved):

sophoah (approved):

sophoah (approved):

PR #4418, opened 2023-05-02
Remove dockerfile hardcoded version numbers

ONECasey:
This is required for tests


Frozen:
I would recommend to create base image with already installed apk dependencies, so, even we don't know exact version, it will be same always. And it will be faster, we will skip downloading stage. I can do by my own, but i don't have access to harmony dockerhub.

Frozen (approved):

PR #4419, opened 2023-05-02
Snapshot integration and add cache to statedb

GheisMohammadi:
## Issue
It is pointless to check and re-verify every state item every time we retrieve it from disk. This PR makes advantage of the Ethereum snapshot functionality. When a new block is processed using snapshot, we do not merge the writes directly into the disk layer, but rather generate a new in-memory diff layer containing the modifications. When enough in-memory diff layers are piled on top, the bottom ones begin to merge and are eventually pushed to disk. When we need to read a state item, we start at the topmost diff layer and work our way backwards until we discover it or reach the disk.
This PR has the potential to dramatically enhance the performance of the state database. The snapshot reduces read times from O(log n) to O(1).

## Operational Checklist

    **NO**

### DIFF
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -39,6 +39,7 @@ import (
 	"github.com/ethereum/go-ethereum/event"
 	"github.com/ethereum/go-ethereum/metrics"
 	"github.com/ethereum/go-ethereum/rlp"
+	"github.com/ethereum/go-ethereum/trie"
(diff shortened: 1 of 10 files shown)


GheisMohammadi:
> Interesting! Do you have any results to see the improvements?

This PR is tested on local node and it is supposed to reduce disk accesses for state writing/retrieving by a few times which is depends on the state path and type of access. It has been already deployed to devnet to do the functionality tests. If it were stable enough, we can do some exact benchmarks.

PR #4422, opened 2023-05-05
state pruning (offline version)

GheisMohammadi:
## State Pruning
The initial version of state pruning. It's an offline tool which uses a snapshot to prune the stale state. Pruner's workflow is pretty straightforward. The pruner iterates the snapshot and reconstructs the relevant state and removes all state entries that they are not neither belongs to the target state and the genesis state. State pruning takes several hours to be completed. It can reduce the disk usage and improve the disk read performance.


adsorptionenthalpy (approved):

core/state/pruner/pruner.go
```diff
@@ -0,0 +1,516 @@
...
+	// If the target state root is not specified, use the HEAD-127 as the
+	// target. The reason for picking it is:
+	// - in most of the normal cases, the related state is available
+	// - the probability of this layer being reorg is very low
+	var layers []snapshot.Snapshot
+	if root == (common.Hash{}) {
+		// Retrieve all snapshot layers from the current HEAD.
+		// In theory there are 128 difflayers + 1 disk layer present,
+		// so 128 diff layers are expected to be returned.
+		layers = p.snaptree.Snapshots(p.chainHeader.Root(), 128, true)
```
sophoah, line 258: what about on an archival node ? I assume in this case there will be more than 128 difflayers right ?

core/state/pruner/pruner.go
```diff
@@ -0,0 +1,516 @@
...
+			return fmt.Errorf("snapshot not old enough yet: need %d more blocks", 128-len(layers))
+		}
+		// Use the bottom-most diff layer as the target
+		root = layers[len(layers)-1].Root()
+	}
+	// Ensure the root is really present. The weak assumption
+	// is the presence of root can indicate the presence of the
+	// entire trie.
+	if !rawdb.HasLegacyTrieNode(p.db, root) {
+		// The special case is for clique based networks(rinkeby, goerli
```
sophoah, line 272: is this something that can happen on the harmony network ? i don't think we have clique poa https://eips.ethereum.org/EIPS/eip-225 enabled on harmony right ?

core/state/pruner/pruner.go
GheisMohammadi: We shouldn't execute state pruning on archival node if we need to have all of those states. The harmony nodes store all the intermediary states. State pruning removes those intermediary states to reduce disk usage.

core/state/pruner/pruner.go
GheisMohammadi: True, we are using Ethereum DB Pruner module. Those features are not used by harmony.

sophoah:
@GheisMohammadi have you tested it on devnet/testnet or even on a test mainnet node ? ie once the pruning is completed is the harmony process still without issue ?

is the pruning only for the ONE accounts and their balance ? or is there any other state ?

GheisMohammadi:
> @GheisMohammadi have you tested it on devnet/testnet or even on a test mainnet node ? ie once the pruning is completed is the harmony process still without issue ?
> 
> is the pruning only for the ONE accounts and their balance ? or is there any other state ?

Tests are in progress. Haven't tested yet on testnet/mainnet. First I should validate things work after pruning.

PR #4427, opened 2023-05-11
Feature: generating protoflles with docker.

Frozen:
Protofiles generated by developers with different versions of proto, gen-go-grpc and protoc-gen-go. This PR aims to unify process of generating by using same version saved in the docker image. 
I contains test that checks generated output and fails if it's not generate by image. 
Only 1 concern here - i use my own repo for image frozen621/harmony-proto:latest, because i don't have access to harmony docker hub.


sophoah:
i guess two things we can do here :
- build the docker image and use it locally
- for the upload of the image to the harmony docker account, do you think we need to upload it often ? Questions is whether we need to build the whole CICD around this or this should be a one time thing ?

Frozen:
1) It's already built and ready for use. 
2) I guess update once per year is enough, no additional actions required.

ONECasey (approved):

PR #4429, opened 2023-05-11
Removed concurrent block SetCurrentCommitSig.

Frozen:
2 goroutines where block.SetCurrentCommitSig can be set.


ONECasey (approved):

ONECasey:
If removing the concurrent block SetCurrentCommitSig operation introduces blocking behavior and hampers the intended gains, alternative approaches might be worth considering.

Using a channel or an atomic number as a check before starting the goroutine can be a viable solution to prevent unnecessary execution and premature setting of the commit signature. Implementing such a check ensures that the goroutine proceeds only when the conditions are met, thus preserving the intended behavior and avoiding potential conflicts.

PR #4430, opened 2023-05-12
Fix code hash conflicts

GheisMohammadi:
## Issue

The validator wrapper code is prefixed differently than contract codes. Prior to the current release (2023.2.0), neither of them used a prefix and both used a code hash as the key to store the code in the database. We used a flag called **isValidatorCode** to tell state object how to fetch code and which prefix to use, but this flag cannot be identified for old data because validator and contract codes do not have prefixes. So, it was using **false** by default in most parts of the code, which can cause problems if the code belongs to a validator wrapper and is stored by prefix. This PR removes that flag and tries to retrieve the code without and with all possible code prefixes.

### DIFF
--- a/core/rawdb/database.go
+++ b/core/rawdb/database.go
@@ -412,7 +412,7 @@ func InspectDatabase(db ethdb.Database, keyPrefix, keyStart []byte) error {
 		}
 		count++
 		if count%1000 == 0 && time.Since(logged) > 8*time.Second {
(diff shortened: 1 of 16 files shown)


ONECasey:
Do we still need this PR @GheisMohammadi if were planning on just rebasing?

ONECasey (dismissed):

GheisMohammadi:
> Do we still need this PR @GheisMohammadi if we're planning on just rebasing?

This PR is just to have the same changes as main. If we want to merge main into dev, then we don't need this PR anymore,

PR #4431, opened 2023-05-12
[hotfix] fix code hash conflicts

GheisMohammadi:
## Issue

The validator wrapper code is prefixed differently than contract codes. Prior to the current release (2023.2.0), neither of them used a prefix and both used a code hash as the key to store the code in the database. We used a flag called isValidatorCode to tell state object how to fetch code and which prefix to use, but this flag cannot be identified for old data because validator and contract codes do not have prefixes and there is no way to distinguish between the keys. So, it was using false by default in most parts of the code, which can cause problems if the code belongs to a validator wrapper and is stored by validator prefix. This PR removes that flag and tries to retrieve the code without and with all possible code prefixes.


sophoah (approved):
no major comment here except that we'll need proper testing as usual

ONECasey (approved):

PR #4362, opened 2023-02-10
[Enhancement]  Add Gas Price Oracle User Configuration

Aishlia:
## Issue
#4357 Add feature to enable users to specify gas price parameters (blocks, percentile, defaultPriceGwei, maxPriceGwei, and numberTxsnSampled) in config and as flags. 

### Unit Test Coverage

Before:
```
ok      github.com/harmony-one/harmony/cmd/harmony    0.756s    coverage: 44.8% of statements
```

After:
```
PASS
ok  	github.com/harmony-one/harmony/cmd/harmony	0.568s
```

### Test/Run Logs
```
julianai@Harmonys-MacBook-Air harmony % go test ./cmd/harmony
# gopkg.in/olebedev/go-duktape.v3
In file included from _cgo_export.c:4:
debugger.go:23:13: warning: unused function '_duk_debugger_attach' [-Wunused-function]
ok  	github.com/harmony-one/harmony/cmd/harmony	0.495s
julianai@Harmonys-MacBook-Air harmony % go test -run TestHarmonyFlags
# gopkg.in/olebedev/go-duktape.v3
In file included from _cgo_export.c:4:
debugger.go:23:13: warning: unused function '_duk_debugger_attach' [-Wunused-function]
PASS
ok  	github.com/harmony-one/harmony/cmd/harmony	0.497s
```

## Operational Checklist

1. **Does this PR introduce backward-incompatible changes to the on-disk data structure and/or the over-the-wire protocol?**. (If no, skip to question 8.)

    **NO**

2. **Describe the migration plan.**. For each flag epoch, describe what changes take place at the flag epoch, the anticipated interactions between upgraded/non-upgraded nodes, and any special operational considerations for the migration.

3. **Describe how the plan was tested.**

4. **How much minimum baking period after the last flag epoch should we allow on Pangaea before promotion onto mainnet?**

5. **What are the planned flag epoch numbers and their ETAs on Pangaea?**

6. **What are the planned flag epoch numbers and their ETAs on mainnet?**

    Note that this must be enough to cover baking period on Pangaea.

7. **What should node operators know about this planned change?**

8. **Does this PR introduce backward-incompatible changes *NOT* related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)

    **NO**

9. **Does the existing `node.sh` continue to work with this change?**

10. **What should node operators know about this change?**

11. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**

[...]

PR #4436, opened 2023-05-22
hmy: Gas Price Oracle improvements

### LINKED ISSUE #4357: Gas Price Oracle Configuration and Redesign
The `gpoParams` are hardcoded
https://github.com/harmony-one/harmony/blob/b2e50e694ca7b7cd2adde45724711f9eafff049b/hmy/hmy.go#L160-L164.
https://github.com/harmony-one/harmony/blob/b2e50e694ca7b7cd2adde45724711f9eafff049b/hmy/gasprice.go#L37
Make them customizable similar to Ethereum.
```
GAS PRICE ORACLE OPTIONS:
  --gpo.blocks value                  Number of recent blocks to check for gas prices (default: 20)
  --gpo.percentile value              Suggested gas price is the given percentile of a set of recent transaction gas prices (default: 60)
  --gpo.maxprice value                Maximum transaction priority fee (or gasprice before London fork) to be recommended by gpo (default: 500000000000)
  --gpo.ignoreprice value             Gas price below which gpo will ignore transactions (default: 2)
```

In a separate pull request, add a better alternative to the gas price oracle's percentile mechanism. For example, as opposed to a constant percentile of the last `N` transactions, define low percentile and high percentile which are used when the last `N` blocks are at least `X%` full where low percentile, high percentile and X are user configurable.

MaxMustermann2:
This change makes the gas price oracle options available as options in the configuration file / command line. In addition, the gas price oracle's suggestion mechanism has been modified to return the default gas price when block utilization is low. In other words, the oracle can be configured to return the 60th percentile gas price from the last 5 blocks with 3 transactions each, or return the default gas price if those 5 blocks were utilized less than 50% of their capacity

Fixes #4357 and supersedes #4362

### DIFF
--- a/cmd/harmony/config_migrations.go
+++ b/cmd/harmony/config_migrations.go
@@ -378,6 +378,33 @@ func init() {
 		return confTree
 	}
 
+	migrations["2.5.14"] = func(confTree *toml.Tree) *toml.Tree {
+		if confTree.Get("GPO.Blocks") == nil {
+			confTree.Set("GPO.Blocks", defaultConfig.GPO.Blocks)
+		}
+		if confTree.Get("GPO.Transactions") == nil {
+			confTree.Set("GPO.Transactions", defaultConfig.GPO.Transactions)
+		}
+		if confTree.Get("GPO.Percentile") == nil {
+			confTree.Set("GPO.Percentile", defaultConfig.GPO.Percentile)
+		}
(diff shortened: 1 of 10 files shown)


ONECasey (approved):

sophoah (approved):
LGTM

PR #4445, opened 2023-06-09
Fixed bug with storing already cancelled context.

Frozen:
Firstly, according to this article https://go.dev/blog/context-and-structs
**Contexts should not be stored inside a struct type, but instead passed to each function that needs it.**

Secondly, there is a bug with storing context at https://github.com/harmony-one/harmony/blob/main/api/service/stagedstreamsync/syncing.go#L190

```
		ctx, cancel := context.WithCancel(downloaderContext)
		s.ctx = ctx
		s.SetNewContext(ctx)
```
not sure how exaclty it happenss, but this fix solved issue. 

When something happens, such as a network error, the already canceled context keeps getting passed repeatedly, thereby preventing subsequent synchronization.


ONECasey (approved):

GheisMohammadi:
The staged sync supports two approaches for syncing. It can either go through multiple cycles of syncing or it can do all in one sync cycle. That's why we use two context. one for each cycle and one for single cycle syncing. 
Have you tested your code for both of these scenarios? Please share how you tested and please share the results here. Thanks

PR #4448, opened 2023-06-19
update the hardcoded settings in stream sync

GheisMohammadi:
## Issue

We used hard-coded values for a few of the sync settings in the present version of staged stream sync so that we could test them in dev-net. The plan was to make them dynamic after all of the testing were completed successfully. This PR updates those options and get the values from config file.


ONECasey (approved):

ONECasey:
@GheisMohammadi Can you please fix these conflicts for the next release?

PR #4463, opened 2023-07-11
Fix kill_node script

GheisMohammadi:
## Issue

The script to kill harmony process doesn't work. This is the error given from Travis logs:
```
pkill: pattern that searches for process name longer than 15 characters will result in zero matches
Try `pkill -f' option to match against the complete command line.
```
The issue is the pattern for **pkill** command can't be longer than 15 characters. This PR uses an array rather than a long pattern to address the issue.


sophoah:
was it tested locally ?

ONECasey:
> was it tested locally ?

I ran it locally, and it worked on my end.

PR #4467, opened 2023-07-17
Improve staged stream sync, fix the devnet node stuck issue

GheisMohammadi:
## Issue
This PR improves the stream sync on the items below:
- improves the short range stage
- improves the epoch sync stage
- adds the last mile stage to staged stream sync
- adds Debug logs to  staged stream sync
- removes redundant stage executions
- addresses the issue for one of the dev-net nodes.

### DIFF
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -195,16 +195,19 @@ func (d *Downloader) loop() {
 	// for epoch chain we do only need to go through epoch sync process
 	initSync := d.isBeaconNode || d.bc.ShardID() != shard.BeaconChainShardID
 
-	done := make(chan struct{})
+	doneC := make(chan struct{})
 
-	select {
-	case done <- struct{}{}:
-	default:
+	done := func() {
+		select {
+		case doneC <- struct{}{}:
+		default:
+		}
 	}
+	done()
 
 	trigger := func() {
 		select {
-		case <-done:
+		case <-doneC:
 			select {
 			case d.downloadC <- struct{}{}:
 			default:
@@ -212,7 +215,6 @@ func (d *Downloader) loop() {
 		case <-time.After(100 * time.Millisecond):
 		}
 	}
-
 	go trigger()
 
 	for {
@@ -221,7 +223,11 @@ func (d *Downloader) loop() {
 			go trigger()
 
 		case <-d.downloadC:
+			d.stagedSyncInstance.Debug("downloader_loop/downloadC", "received signal from downloadC channel")
+			d.stagedSyncInstance.Debug("downloader_loop/downloadC/shardID", d.bc.ShardID())
 			addedBN, err := d.stagedSyncInstance.doSync(d.ctx, initSync)
+			d.stagedSyncInstance.Debug("downloader_loop/downloadC/addedBN", addedBN)
+			d.stagedSyncInstance.Debug("downloader_loop/downloadC/error", err)
 			if err != nil {
 				//TODO: if there is a bad block which can't be resolved
 				if d.stagedSyncInstance.invalidBlock.Active {
@@ -255,6 +261,7 @@ func (d *Downloader) loop() {
 			}
 			if addedBN != 0 {
 				// If block number has been changed, trigger another sync
+				done()
 				go trigger()
 				// try to add last mile from pub-sub (blocking)
 				if d.bh != nil {
@@ -266,7 +273,7 @@ func (d *Downloader) loop() {
 		case <-d.closeC:
 			return
 		}
-
-		done <- struct{}{}
+		d.stagedSyncInstance.Debug("downloader_loop/downloadC", "done")
+		done()
 	}
 }
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -194,14 +194,20 @@ func (d *Downloader) loop() {
 	// make sure it goes through the long range sync first.
 	// for epoch chain we do only need to go through epoch sync process
 	initSync := d.isBeaconNode || d.bc.ShardID() != shard.BeaconChainShardID
+	
+	done := make(chan struct{})
+	d.downloadC <- struct{}{}
 
 	trigger := func() {
 		select {
-		case d.downloadC <- struct{}{}:
+		case <-done:
+			select {
+			case d.downloadC <- struct{}{}:
+			default:
+			}
 		case <-time.After(100 * time.Millisecond):
 		}
 	}
-	go trigger()
 
 	for {
 		select {
@@ -223,7 +229,6 @@ func (d *Downloader) loop() {
 						//TODO: if we don't have any new or untried stream in the list, sleep or panic
 					}
 				}
-
 				// If any error happens, sleep 5 seconds and retry
 				d.logger.Error().
 					Err(err).
@@ -233,8 +238,7 @@ func (d *Downloader) loop() {
 					time.Sleep(5 * time.Second)
 					trigger()
 				}()
-				time.Sleep(1 * time.Second)
-				continue
+				break
 			}
 			if initSync {
 				d.logger.Info().Int("block added", addedBN).
@@ -243,7 +247,6 @@ func (d *Downloader) loop() {
 					Uint32("shard", d.bc.ShardID()).
 					Msg(WrapStagedSyncMsg("sync finished"))
 			}
-
 			if addedBN != 0 {
 				// If block number has been changed, trigger another sync
 				go trigger()
@@ -257,5 +260,7 @@ func (d *Downloader) loop() {
 		case <-d.closeC:
 			return
 		}
+
+		done <- struct{}{}
 	}
 }
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -195,15 +195,15 @@ func (d *Downloader) loop() {
 	// for epoch chain we do only need to go through epoch sync process
 	initSync := d.isBeaconNode || d.bc.ShardID() != shard.BeaconChainShardID
 
-	doneC := make(chan struct{}, 1)
+	doneC := make(chan struct{})
 
 	done := func() {
 		select {
 		case doneC <- struct{}{}:
-		default:
+		case <-time.After(100 * time.Millisecond):
 		}
 	}
-	done()
+	go done()
 
 	trigger := func() {
 		select {
@@ -263,15 +263,6 @@ func (d *Downloader) loop() {
 					Uint32("shard", d.bc.ShardID()).
 					Msg(WrapStagedSyncMsg("sync finished"))
 			}
-			if addedBN != 0 {
-				// If block number has been changed, trigger another sync
-				done()
-				go trigger()
-				// try to add last mile from pub-sub (blocking)
-				if d.bh != nil {
-					d.bh.insertSync()
-				}
-			}
 			// if last doSync needed only to add a few blocks less than LastMileBlocksThreshold and
 			// the node is fully synced now, then switch to short range
 			// the reason why we need to check distanceBeforeSync is because, if it was long distance,
@@ -285,11 +276,21 @@ func (d *Downloader) loop() {
 				distanceAfterSync <= uint64(LastMileBlocksThreshold) {
 				initSync = false
 			}
+			// If block number has been changed, trigger another sync
+			if addedBN != 0 {
+				// try to add last mile from pub-sub (blocking)
+				if d.bh != nil {
+					d.bh.insertSync()
+				}
+
+				go done()
+				go trigger()
+			}
 
 		case <-d.closeC:
 			return
 		}
 		d.stagedSyncInstance.Debug("downloader_loop/downloadC", "done")
-		done()
+		go done()
 	}
 }
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -194,7 +194,7 @@ func (d *Downloader) loop() {
 	// make sure it goes through the long range sync first.
 	// for epoch chain we do only need to go through epoch sync process
 	initSync := d.isBeaconNode || d.bc.ShardID() != shard.BeaconChainShardID
-	
+
 	done := make(chan struct{})
 	d.downloadC <- struct{}{}
 
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -225,7 +225,8 @@ func (d *Downloader) loop() {
 		case <-d.downloadC:
 			d.stagedSyncInstance.Debug("downloader_loop/downloadC", "received signal from downloadC channel")
 			d.stagedSyncInstance.Debug("downloader_loop/downloadC/shardID", d.bc.ShardID())
-			addedBN, err := d.stagedSyncInstance.doSync(d.ctx, initSync)
+			bnBeforeSync := d.bc.CurrentBlock().NumberU64()
+			estimatedHeight, addedBN, err := d.stagedSyncInstance.doSync(d.ctx, initSync)
 			d.stagedSyncInstance.Debug("downloader_loop/downloadC/addedBN", addedBN)
 			d.stagedSyncInstance.Debug("downloader_loop/downloadC/error", err)
 			if err != nil {
@@ -268,7 +269,19 @@ func (d *Downloader) loop() {
 					d.bh.insertSync()
 				}
 			}
-			initSync = false
+			// if last doSync needed only to add a few blocks less than LastMileBlocksThreshold and
+			// the node is fully synced now, then switch to short range
+			// the reason why we need to check distanceBeforeSync is because, if it was long distance,
+			// very likely, there are a couple of new blocks have been added to the other nodes which
+			// we should still stay in long range and check them.
+			bnAfterSync := d.bc.CurrentBlock().NumberU64()
+			distanceBeforeSync := estimatedHeight - bnBeforeSync
+			distanceAfterSync := estimatedHeight - bnAfterSync
+			if estimatedHeight > 0 && addedBN != 0 &&
+				distanceBeforeSync <= uint64(LastMileBlocksThreshold) &&
+				distanceAfterSync <= uint64(LastMileBlocksThreshold) {
+				initSync = false
+			}
 
 		case <-d.closeC:
 			return
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -68,8 +68,8 @@ func NewDownloader(host p2p.Host, bc core.BlockChain, consensus *consensus.Conse
 
 	ctx, cancel := context.WithCancel(context.Background())
 
-	//TODO: use mem db should be in config file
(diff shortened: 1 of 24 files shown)


ONECasey (changes requested):
Do we need all these debug logic?

GheisMohammadi:
> Do we need all these debug logic?

I agree that it makes the code a bit messy. But they are very useful to trace the issues and every single process. By default, it does nothing and only if the **DebugMode** is enabled, the logs will be printed out.

sophoah:
instead of having a specific debug flag for sync, can't we use simply use `utils.Logger().Debug()` I am trying to avoid having more hidden/unknown config parameter that make our harmony binary UX less friendly

api/service/stagedstreamsync/stage_bodies.go
```diff
@@ -60,6 +61,11 @@ func (b *StageBodies) Exec(ctx context.Context, firstCycle bool, invalidBlockRev
 		return nil
 	}
 
+	// shouldn't execute for epoch chain
+	if b.configs.bc.ShardID() == shard.BeaconChainShardID && !s.state.isBeaconNode {
+		return nil
```
sophoah, line 66: can you add a info message here ?

api/service/stagedstreamsync/stage_epoch.go
```diff
@@ -34,6 +35,9 @@ func NewStageEpochCfg(bc core.BlockChain, db kv.RwDB) StageEpochCfg {
 
 func (sr *StageEpoch) Exec(ctx context.Context, firstCycle bool, invalidBlockRevert bool, s *StageState, reverter Reverter, tx kv.RwTx) error {
 
+	s.state.Debug("shard", sr.configs.bc.ShardID())
```
sophoah, line 38: Any issue updating 
```golang
func (fi *fastIterator) Debug() 
```
to use `utils.Logger().Debug()` ?

api/service/stagedstreamsync/stage_lastmile.go
```diff
@@ -0,0 +1,109 @@
...
+		bc:  bc,
+		db:  db,
+	}
+}
+
+func (lm *StageLastMile) Exec(ctx context.Context, firstCycle bool, invalidBlockRevert bool, s *StageState, reverter Reverter, tx kv.RwTx) (err error) {
+
+	// no need to download the last mile blocks if we are redoing the stages because of bad block
+	if invalidBlockRevert {
+		return nil
```
sophoah, line 40: add a log message ?

api/service/stagedstreamsync/stage_lastmile.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	// shouldn't execute for epoch chain
+	if lm.configs.bc.ShardID() == shard.BeaconChainShardID && !s.state.isBeaconNode {
+		return nil
```
sophoah, line 45: add a log message ?

api/service/stagedstreamsync/stage_lastmile.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (lm *StageLastMile) Revert(ctx context.Context, firstCycle bool, u *RevertState, s *StageState, tx kv.RwTx) (err error) {
+	useInternalTx := tx == nil
+	if useInternalTx {
+		tx, err = lm.configs.db.BeginRw(lm.configs.ctx)
+		if err != nil {
+			return err
+		}
+		defer tx.Rollback()
```
sophoah, line 78: @MaxMustermann2 @GheisMohammadi @Frozen please properly review this rollback method to avoid any unwanted security risk

api/service/stagedstreamsync/stage_state.go
```diff
@@ -57,6 +58,11 @@ func (stg *StageStates) Exec(ctx context.Context, firstCycle bool, invalidBlockR
 		return nil
 	}
 
+	// shouldn't execute for epoch chain
+	if stg.configs.bc.ShardID() == shard.BeaconChainShardID && !s.state.isBeaconNode {
+		return nil
```
sophoah, line 63: add message log

api/service/stagedstreamsync/staged_stream_sync.go
```diff
@@ -583,3 +595,133 @@ func (s *StagedStreamSync) EnableStages(ids ...SyncStageID) {
...
+}
+
+func (ss *StagedStreamSync) RollbackLastMileBlocks(ctx context.Context, hashes []common.Hash) error {
+	if len(hashes) == 0 {
+		return nil
+	}
+	utils.Logger().Info().
+		Interface("block", ss.bc.CurrentBlock()).
+		Msg("[STAGED_STREAM_SYNC] Rolling back last mile blocks")
+	if err := ss.bc.Rollback(hashes); err != nil {
```
sophoah, line 657: same as previous rollback comment @GheisMohammadi @MaxMustermann2 @Frozen

sophoah (changes requested):

api/service/stagedstreamsync/stage_bodies.go
GheisMohammadi: Each stage is executed once for each sync cycle, and the sync cycles will run in a loop. Therefore, this line will continuously run, and if the chain is an epoch chain, it will log the information repeatedly within a cycle, resulting in many entries in the log file. In some cases, this could lead to thousands of logs being generated per hour.

api/service/stagedstreamsync/syncing.go
```diff
@@ -38,41 +40,48 @@ var Buckets = []string{
...
 	protocol syncProtocol,
 	config Config,
 	logger zerolog.Logger,
-	logProgress bool,
 ) (*StagedStreamSync, error) {
 
-	isBeacon := bc.ShardID() == shard.BeaconChainShardID
-
+	if config.DebugMode {
+		fmt.Printf("[SSSD:CreateStagedSync] creating staged sync for shard:%d beaconNode:%t memdb:%t dbDir:%s serverOnly:%t minStreams:%d\n",
```
sophoah, line 52: seems to be a good log for utils.Logger().info() ? not sure if there are going to be a lot of more logs but it can't be worse than the p2p connected/disconnected logs right ?

api/service/stagedstreamsync/stage_epoch.go
GheisMohammadi: Agreed. I have been working on sync issues for the last few weeks, and it was quite challenging to trace and replicate the errors locally. To address this, I added a debug mode, which was extremely useful.
The debug mode is a hidden option located under sync options in the config file. It is optional and, by default, set to false, meaning it doesn't affect the normal execution. Even if you mistakenly set this flag incorrectly, it won't cause any issues; it exists only for display purposes. The sync debug mode allows printing out the variables and each individual stage's process, helping to trace the issues and implementation correctness.
I'm considering removing all the debug lines and retaining this function for future use, but I'm not entirely sure about it yet.

api/service/stagedstreamsync/stage_lastmile.go
GheisMohammadi: Same thing here.

api/service/stagedstreamsync/stage_lastmile.go
GheisMohammadi: Same thing here

api/service/stagedstreamsync/stage_lastmile.go
GheisMohammadi: good idea

api/service/stagedstreamsync/stage_state.go
GheisMohammadi: same here

api/service/stagedstreamsync/syncing.go
GheisMohammadi: it will create much more logs than the p2p connection logs

GheisMohammadi: I think so

api/service/stagedstreamsync/syncing.go
```diff
@@ -156,67 +166,119 @@ func getBlockDbPath(beacon bool, loopID int, dbDir string) string {
 		}
 	} else {
 		if loopID >= 0 {
-			return fmt.Sprintf("%s_%d", filepath.Join(dbDir, "cache/blocks_db"), loopID)
+			return fmt.Sprintf("%s_%d_%d", filepath.Join(dbDir, "cache/blocks_db"), shardID, loopID)
```
sophoah, line 187: use utils.logger() ?

api/service/stagedstreamsync/syncing.go
```diff
@@ -156,67 +166,119 @@ func getBlockDbPath(beacon bool, loopID int, dbDir string) string {
...
+	if len(callerParts) > 0 {
+		caller = callerParts[len(callerParts)-1]
+	}
+	src := source
+	if src == "" {
+		src = "message"
+	}
+	// SSSD: STAGED STREAM SYNC DEBUG
+	if msg == nil {
+		fmt.Printf("[SSSD:%s] %s: nil or no error\n", caller, src)
```
sophoah, line 211: use utils.logger() ?

api/service/stagedstreamsync/syncing.go
```diff
@@ -156,67 +166,119 @@ func getBlockDbPath(beacon bool, loopID int, dbDir string) string {
...
 	}
 
+	s.startSyncing()
+	defer s.finishSyncing()
+
 	for {
 		ctx, cancel := context.WithCancel(downloaderContext)
 
 		n, err := s.doSyncCycle(ctx, initSync)
+		s.Debug("doSyncCycle/n", n)
```
sophoah, line 260: better log description? or let's remove if no longer required ?

sophoah:
@GheisMohammadi questions:
1) can the explorer node be as a stream peer ?
2) many s0 node in the network (mainnet) are using snapDB (DB with only recent block information), this will be a risk for node willing to sync from 0. In mainnet, mainly explorer node will have all the blocks from genesis

cmd/harmony/default.go
```diff
@@ -178,6 +178,7 @@ var defaultStagedSyncConfig = harmonyconfig.StagedSyncConfig{
 	MaxMemSyncCycleSize:    1024,  // max number of blocks to use a single transaction for staged sync
 	UseMemDB:               true,  // it uses memory by default. set it to false to use disk
```
sophoah, line 179: is there a limit for `UseMemDB` on the memory usage to avoid a possible OOM ?

p2p/stream/protocols/sync/protocol.go
```diff
@@ -180,7 +180,8 @@ func (p *Protocol) HandleStream(raw libp2p_network.Stream) {
 			Msg("failed to add new stream")
 		return
 	}
-	fmt.Println("Connected to", raw.Conn().RemotePeer().String(), "(", st.ProtoID(), ")", "my ID: ", raw.Conn().LocalPeer().String())
+	//to get my ID use raw.Conn().LocalPeer().String()
+	fmt.Println("Connected to", raw.Conn().RemotePeer().String(), "(", st.ProtoID(), ")")
```
sophoah, line 184: change it to utils.logging() ?

Frozen: better remove it

sophoah: why is that ? and different it is between the p2p connected/disconnected message ? not saying we should add it, but we should at least be consistent.

api/service/stagedstreamsync/syncing.go
GheisMohammadi: This function returns a path in string format and it does not perform any action. Sprintf doesn't print anything, it formats the string

api/service/stagedstreamsync/syncing.go
GheisMohammadi: This is the Debug function designed to print information in the console instead of the log file. As mentioned earlier, the Debug function prints each process very frequently, leading to an excessive amount of log lines. The underlying concept of the debug feature is to monitor data in real-time.

api/service/stagedstreamsync/syncing.go
GheisMohammadi: I am going to remove it.

GheisMohammadi:
> @GheisMohammadi questions:
> 
> 1. can the explorer node be as a stream peer ?
> 2. many s0 node in the network (mainnet) are using snapDB (DB with only recent block information), this will be a risk for node willing to sync from 0. In mainnet, mainly explorer node will have all the blocks from genesis

1- Based on the codes they are stream peer and can talk to other nodes through streams.
2- That's something definitely should be considered. Thanks for sharing it. Let me check first how the logic is and then we need proper sync strategy for that scenario.

cmd/harmony/default.go
GheisMohammadi: Exactly, That's the reason I added a parameter for it. It depends on the available memory and system resources. I used the small Cycle Size to make sure it works on almost every node with minimum required spec.

p2p/stream/protocols/sync/protocol.go
GheisMohammadi: Actually this line was one of the most useful lines to debug and trace the connectivity issues. That's how I could check the node connections. The idea of logging connection is common on many existed blockchains. The reason why we haven't have this connection lines is because we were using the DNS nodes which are trusted nodes. I can move this line to log file if preferred.

GheisMohammadi:
I removed all extra debug logs

p2p/stream/protocols/sync/protocol.go
sophoah: yeah agree i'd move this to the log

cmd/harmony/default.go
sophoah: what's the parameter and can this be somewhat dynamic based on the machine memory ?

api/service/stagedstreamsync/stage_lastmile.go
sophoah: any update for this ?

cmd/harmony/default.go
GheisMohammadi: it can be enabled on every single node regardless of the specs. But MaxMemSyncCycleSize should be set appropriately.

api/service/stagedstreamsync/staged_stream_sync.go
GheisMohammadi: As we discussed, the Rollback function is being used on DNS sync, stream sync, staged sync, command line and so on. So, if any changes is needed for the Rollback function, we will do it on separate PR.

api/service/stagedstreamsync/stage_lastmile.go
GheisMohammadi: this rollback is for transaction and not doing any thing for db. It's one of the Erigon db functionalities.

cmd/harmony/default.go
sophoah: so it's manual. Do you have a recommended settings that match the expected memory for `MaxMemSyncCycleSize`

GheisMohammadi: The default value of MaxMemSyncCycleSize is the recommended one. That's the same value used by erigon as well. It is on 1024.

sophoah (approved):

api/service/stagedstreamsync/downloader.go
```diff
@@ -242,16 +246,27 @@ func (d *Downloader) loop() {
...
+				}
 			}
-			// try to add last mile from pub-sub (blocking)
-			if d.bh != nil {
-				d.bh.insertSync()
+			// if last doSync needed only to add a few blocks less than LastMileBlocksThreshold and
+			// the node is fully synced now, then switch to short range
+			// the reason why we need to check distanceBeforeSync is because, if it was long distance,
+			// very likely, there are a couple of new blocks have been added to the other nodes which
+			// we should still stay in long range and check them.
```
MaxMustermann2, line 261: Great explanation.

api/service/stagedstreamsync/staged_stream_sync.go
```diff
@@ -583,3 +594,133 @@ func (s *StagedStreamSync) EnableStages(ids ...SyncStageID) {
...
+		ss.lastMileBlocks = append(ss.lastMileBlocks, block)
+	}
+}
+
+func (ss *StagedStreamSync) getBlockFromLastMileBlocksByParentHash(parentHash common.Hash) *types.Block {
+	ss.lastMileMux.Lock()
+	defer ss.lastMileMux.Unlock()
+	for _, block := range ss.lastMileBlocks {
+		ph := block.ParentHash()
+		if bytes.Equal(ph[:], parentHash[:]) {
```
MaxMustermann2, line 622: nit: you don't need bytes.Equal for arrays of equal length. Just `==` works fine and is marginally faster.

api/service/stagedstreamsync/staged_stream_sync.go
```diff
@@ -583,3 +594,133 @@ func (s *StagedStreamSync) EnableStages(ids ...SyncStageID) {
...
+	if err := ss.bc.Rollback(hashes); err != nil {
+		utils.Logger().Error().Err(err).
+			Msg("[STAGED_STREAM_SYNC] failed to rollback last mile blocks")
+		return err
+	}
+	return nil
+}
+
+// UpdateBlockAndStatus updates block and its status in db
+func (ss *StagedStreamSync) UpdateBlockAndStatus(block *types.Block, bc core.BlockChain, verifyAllSig bool) error {
```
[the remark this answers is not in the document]
GheisMohammadi: This function is a member of **ISync** interface. In stream sync the logs are different and a few small changes compared to others. But for sure all can be implemented into one common function. It needs a bit change in sync structure. I will do it in a separate PR.

cmd/harmony/default.go
```diff
@@ -178,6 +178,7 @@ var defaultStagedSyncConfig = harmonyconfig.StagedSyncConfig{
...
 	LogProgress:            false, // log the full sync progress in console
+	DebugMode:              false, // log every single process and error to help to debug the syncing
```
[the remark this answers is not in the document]
GheisMohammadi: done

api/service/stagedstreamsync/staged_stream_sync.go
GheisMohammadi: done

MaxMustermann2:
> That's something definitely should be considered. Thanks for sharing it. Let me check first how the logic is and then we need proper sync strategy for that scenario. if any logic needs to be added about it, will do it in separate PR.

@GheisMohammadi - in regards to the above, what happens based on the current code? If a node requests for a block from another node that does not have said block, does the stream simply error out?

GheisMohammadi:
> > That's something definitely should be considered. Thanks for sharing it. Let me check first how the logic is and then we need proper sync strategy for that scenario. if any logic needs to be added about it, will do it in separate PR.
> 
> 
> 
> @GheisMohammadi - in regards to the above, what happens based on the current code? If a node requests for a block from another node that does not have said block, does the stream simply error out?

Current version of stream sync will receive empty response from other peers if they don't have that block.
We had a few discussions with the team on this matter. We need to add a new logic flow for that. This will be done on a separate PR.

MaxMustermann2 (approved):

ONECasey (approved):

Issue #4468, opened 2023-07-17
FBFTLog: Сoncurrent map iteration and map write


Frozen:
fatal error: concurrent map iteration and map write

goroutine 28578 [running]:
github.com/harmony-one/harmony/consensus.(*FBFTLog).GetMessagesByTypeSeqHash(...)
	/home/runner/work/harmony/harmony/harmony/consensus/fbft_log.go:228
github.com/harmony-one/harmony/node.(*Node).explorerMessageHandler(0xc001780300, {0xc0067addd0?, 0xc007a4cf58?}, 0xc007a4cf68?)
	/home/runner/work/harmony/harmony/harmony/node/node_explorer.go:99 +0x24e
github.com/harmony-one/harmony/node.(*Node).StartPubSub.func2.1()
	/home/runner/work/harmony/harmony/harmony/node/node.go:881 +0x115
created by github.com/harmony-one/harmony/node.(*Node).StartPubSub.func2
	/home/runner/work/harmony/harmony/harmony/node/node.go:874 +0x70

goroutine 1 [select (no cases), 1 minutes]:
main.setupNodeAndRun({{0xc000b7c530, 0x5}, {{0xc000b7ca70, 0x8}, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...}, ...})
	/home/runner/work/harmony/harmony/harmony/cmd/harmony/main.go:477 +0x105c
main.runHarmonyNode(0x3f0b560?, {0x26ba187?, 0x2?, 0x2?})
	/home/runner/work/harmony/harmony/harmony/cmd/harmony/main.go:157 +0x1f3
github.com/spf13/cobra.(*Command).execute(0x3f0b560, {0xc00011c160, 0x2, 0x2})
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:876 +0x67b
github.com/spf13/cobra.(*Command).ExecuteC(0x3f0b560)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:990 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.5.0/command.go:918
main.main()
	/home/runner/work/harmony/harmony/harmony/cmd/harmony/main.go:131 +0x25

MaxMustermann2:
Closed by #4469

PR #4469, opened 2023-07-17
Fixed concurrent map access in FBFTLog

Frozen:
Fixed concurrent map access in FBFTLog https://github.com/harmony-one/harmony/issues/4468


ONECasey (approved):
Very interesting issue and bugfix. All seems good from my point of view. Nice work @Frozen

PR #4476, opened 2023-08-02
fix Travic CI build enviroment, update docker daemon to be the latest  stable version

mur-me:


sophoah:
doesn't seems required after recent testing as docker version is now more recent

PR #4477, opened 2023-08-03
Dead code and cleanup.

Frozen:


core/blockchain_impl.go
```diff
@@ -447,7 +447,7 @@ func VerifyIncomingReceipts(blockchain BlockChain, block *types.Block) error {
 			}
 		}
 
-		if err := blockchain.Validator().ValidateCXReceiptsProof(cxp); err != nil {
+		if err := NewBlockValidator(blockchain).ValidateCXReceiptsProof(cxp); err != nil {
```
MaxMustermann2, line 450: Would it be better to keep `bc.Validator()` without mutex as a function for cases like this?

core/blockchain_impl.go
```diff
@@ -1764,7 +1746,7 @@ func (bc *BlockChainImpl) insertChain(chain types.Blocks, verifyHeaders bool) (i
 			err = <-verifyHeadersResults
 		}
 		if err == nil {
-			err = bc.Validator().ValidateBody(block)
+			err = NewBlockValidator(bc).ValidateBody(block)
```
MaxMustermann2, line 1749:
```suggestion
			err = bc.validator.ValidateBody(block)
```

core/staking_verifier.go
```diff
@@ -30,14 +30,9 @@ var (
 )
 
 func checkDuplicateFields(
-	bc ChainContext, state vm.StateDB,
+	addrs []common.Address, state vm.StateDB,
```
MaxMustermann2, line 33: For clarity, can you rename `addrs` to be `validatorList` ?

MaxMustermann2 (changes requested):

core/blockchain_impl.go
sophoah: @MaxMustermann2 just curious, could you explain your rationale of doing so ?

ONECasey (approved):

PR #4484, opened 2023-08-05
Fixed debug run for mac.

Frozen:
### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**

### DIFF
--- a/test/kill_node.sh
+++ b/test/kill_node.sh
@@ -10,7 +10,7 @@ do
   if [ $NAME == "Darwin" ]; then
     pkill -9 "^(${pname})$"
   else
-    sudo pkill -9 -e "^(${pname})$"
+    pkill -9 -e "^(${pname})$"
   fi
 done
 
--- a/test/kill_node.sh
+++ b/test/kill_node.sh
@@ -3,9 +3,15 @@
 # list of process name to be killed
 targetProcess=( "harmony" "bootnode" "soldier" "commander" "profiler" )
 
+NAME="$(uname -s)"
+
 for pname in "${targetProcess[@]}"
 do
-  sudo pkill -9 -e "^(${pname})$"
+  if [ $NAME == "Darwin" ]; then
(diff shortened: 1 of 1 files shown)


ONECasey (approved):

sophoah:
hey @Frozen what was the error message you had so you don't need to use sudo on your mac with that PR

test/kill_node.sh
```diff
@@ -3,9 +3,15 @@
 # list of process name to be killed
 targetProcess=( "harmony" "bootnode" "soldier" "commander" "profiler" )
 
+NAME="$(uname -s)"
+
 for pname in "${targetProcess[@]}"
 do
-  sudo pkill -9 -e "^(${pname})$"
+  if [ $NAME == "Darwin" ]; then
+    pkill -9 "^(${pname})$"
```
sophoah, line 11: so you removed the sudo here, what was the error you had before that ? i assume on your mac your user account has root privilege ? is that the same for all mac OS version ?

Frozen: sudo requires password, but pkill works without it. 

> is that the same for all mac OS version ?
I don't really know, but more likely i have default settings.

test/kill_node.sh
```diff
@@ -3,9 +3,15 @@
...
+  else
+    sudo pkill -9 -e "^(${pname})$"
```
MaxMustermann2, line 13: I think this was introduced by @GheisMohammadi but it should be removed. Rather, the user should call `sudo ./test/kill_node.sh` instead. 

```suggestion
    pkill -9 -e "^(${pname})$"
```

MaxMustermann2 (approved):

test/kill_node.sh
ONECasey: @Frozen can you please change before we merge?

Frozen: fixed

PR #4492, opened 2023-08-20
Next validator in view change. 

Frozen:
Multiple consecutive keys within a list of validators can be assigned to a single validator. In the worst-case scenario, these consecutive keys could be grouped together. If this validator becomes unavailable, the system experiences prolonged downtime as each key associated with the validator is sequentially checked. The current approach involves iterating through the list of keys, and if the next key belongs to the same validator, it is skipped to optimize the process.

In the default implementation, keys are retrieved one by one from a list, and in the current implementation, each key is checked sequentially. If the next key in line belongs to the same validator as the previous one, it is skipped during the validation process.

go test -cover github.com/harmony-one/harmony/consensus/quorum
ok      github.com/harmony-one/harmony/consensus/quorum 1.813s  coverage: 52.9% of statements

### DIFF
--- a/consensus/quorum/quorum.go
+++ b/consensus/quorum/quorum.go
@@ -217,7 +217,35 @@ func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bl
 	return found, &s.publicKeys[idx]
 }
 
-// NthNextHmy return the Nth next pubkey of Harmony nodes, next can be negative number
+// NthNextValidatorHmy return the Nth next pubkey nodes, but from another validator.
+func (s *cIdentities) NthNextValidatorHmy(instance shardingconfig.Instance, slotList shard.SlotList, pubKey *bls.PublicKeyWrapper, next int) (bool, *bls.PublicKeyWrapper) {
+	found := false
+
+	publicToAddress := make(map[bls.SerializedPublicKey]common.Address)
+	for _, slot := range slotList {
+		publicToAddress[slot.BLSPublicKey] = slot.EcdsaAddress
+	}
+
+	idx := s.IndexOf(pubKey.Bytes)
+	if idx != -1 {
+		found = true
+	} else {
+		utils.Logger().Error().
+			Str("key", pubKey.Bytes.Hex()).
+			Msg("[NthNextHmy] pubKey not found")
+	}
+	for {
+		numNodes := len(s.publicKeys)
+		idx = (idx + next) % numNodes
+		if publicToAddress[s.publicKeys[idx].Bytes] == publicToAddress[pubKey.Bytes] {
+			// same validator, go next
+			idx++
+			continue
+		}
+		return found, &s.publicKeys[idx]
+	}
+}
+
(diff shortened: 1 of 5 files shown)


sophoah:
Can you add test similar to TestNthNextHmyExt ?

sophoah:
@GheisMohammadi can you help review this one ?

GheisMohammadi:
> @GheisMohammadi can you help review this one ?

@Frozen could you please add a proper description for this PR. It helps us to understand how this PR works.

consensus/quorum/quorum.go
```diff
@@ -217,7 +217,35 @@ func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bl
 	return found, &s.publicKeys[idx]
 }
 
-// NthNextHmy return the Nth next pubkey of Harmony nodes, next can be negative number
+// NthNextValidatorHmy return the Nth next pubkey nodes, but from another validator.
+func (s *cIdentities) NthNextValidatorHmy(instance shardingconfig.Instance, slotList shard.SlotList, pubKey *bls.PublicKeyWrapper, next int) (bool, *bls.PublicKeyWrapper) {
```
GheisMohammadi, line 221: it would be good if you validate publicKeys here:

```golang
if len(s.publicKeys)==0 {
   return false, nil, fmt.Errorf("...")
}
```
maybe you can check it with a minimum number of publicKeys rather than 0???

consensus/quorum/quorum.go
```diff
@@ -217,7 +217,35 @@ func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bl
...
+	found := false
+
```
GheisMohammadi, line 223: it would be good if you validate next here:
```golang
if next <= 0 {
      return false, nil, fmt.Errorf("invalid ...")
}
```

consensus/quorum/quorum.go
```diff
@@ -217,7 +217,35 @@ func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bl
...
+	if idx != -1 {
+		found = true
+	} else {
+		utils.Logger().Error().
+			Str("key", pubKey.Bytes.Hex()).
+			Msg("[NthNextHmy] pubKey not found")
+	}
+	for {
+		numNodes := len(s.publicKeys)
+		idx = (idx + next) % numNodes
```
GheisMohammadi, line 239: suggested checking above, helps to ensure here we always have idx>=0 and also numNodes>0

consensus/quorum/quorum.go
GheisMohammadi, line 221: where we are using NthNextValidatorHmy?

consensus/quorum/quorum.go
Frozen: added check

consensus/quorum/quorum.go
Frozen: added check

consensus/quorum/quorum.go
Frozen: pushed new commits with usage

Frozen: added check

GheisMohammadi (approved):

ONECasey (approved):

PR #4493, opened 2023-08-22
Improvements of streamsync to deploy on mainnet

GheisMohammadi:
## Issue

The current stream sync codes find the neighboring peers and try to connect to them, completing the discovery through the bootstrap process. Once nodes have connected to the minimum number of peers adjusted in the config file, they consider the bootstrap complete and move on to the next step: syncing with the connected peers.
In the main-net, most validators do not have a full database and only have the latest blocks. If a node wants to ask them for old blocks, they most likely won't have them. In this scenario, new nodes won't be able to use neighbors for sync from scratch and will have to use explorer or archive nodes for syncing, which is more centralized.
After the bootstrap process is completed and the node is connected to its neighbors, during the syncing phase, if the remote peer doesn't have the required blocks, the node removes it from the list. In the mainnet, after a few seconds, the stream list will become empty because most validators don't have the full database. Consequently, the node becomes stuck and cannot proceed with the syncing process.
This PR improves discovery process and tries to let streams be longer in list. Also, it ensures that for each sync cycle it has enough streams available.

### DIFF
--- a/api/service/stagedstreamsync/stage_bodies.go
+++ b/api/service/stagedstreamsync/stage_bodies.go
@@ -167,13 +167,22 @@ func (b *StageBodies) runBlockWorkerLoop(ctx context.Context, gbm *blockDownload
 				Msg(WrapStagedSyncMsg("downloadRawBlocks failed"))
 			err = errors.Wrap(err, "request error")
 			gbm.HandleRequestError(batch, err, stid)
-		} else if blockBytes == nil || len(blockBytes) == 0 {
+		} else if blockBytes == nil {
 			utils.Logger().Warn().
 				Str("stream", string(stid)).
 				Interface("block numbers", batch).
-				Msg(WrapStagedSyncMsg("downloadRawBlocks failed, received empty blockBytes"))
+				Msg(WrapStagedSyncMsg("downloadRawBlocks failed, received invalid (nil) blockBytes"))
+			err := errors.New("downloadRawBlocks received invalid (nil) blockBytes")
+			gbm.HandleRequestError(batch, err, stid)
+			b.configs.protocol.StreamFailed(stid, "downloadRawBlocks failed")
+		} else if len(blockBytes) == 0 {
+			utils.Logger().Warn().
+				Str("stream", string(stid)).
+				Interface("block numbers", batch).
+				Msg(WrapStagedSyncMsg("downloadRawBlocks failed, received empty blockBytes, remote peer is not fully synced"))
 			err := errors.New("downloadRawBlocks received empty blockBytes")
 			gbm.HandleRequestError(batch, err, stid)
+			b.configs.protocol.RemoveStream(stid)
 		} else {
 			if err = b.saveBlocks(ctx, gbm.tx, batch, blockBytes, sigBytes, loopID, stid); err != nil {
 				panic(ErrSaveBlocksToDbFailed)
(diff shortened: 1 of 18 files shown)


p2p/stream/common/requestmanager/interface_test.go
```diff
@@ -118,7 +119,7 @@ func (st *testStream) FailedTimes() int {
 	return 0
 }
 
-func (st *testStream) AddFailedTimes() {
+func (st *testStream) AddFailedTimes(faultRecoveryThreshold time.Duration) {
```
sophoah, line 122: test will be developed in a separate PR ?

p2p/stream/types/stream.go
```diff
@@ -82,7 +84,11 @@ func (st *BaseStream) FailedTimes() int {
 	return st.failedTimes
 }
 
-func (st *BaseStream) AddFailedTimes() {
+func (st *BaseStream) AddFailedTimes(faultRecoveryThreshold time.Duration) {
+	durationSinceLastFailure := time.Now().Sub(st.lastFailureTime)
+	if durationSinceLastFailure >= faultRecoveryThreshold {
```
sophoah: since this function is only called by `StreamFailed`, it looks to me that a failed stream will never be back to 0, and a previously failed stream will always be at 1 at the minimum until the next node restart

api/service/stagedstreamsync/stage_bodies.go
```diff
@@ -167,13 +167,22 @@ func (b *StageBodies) runBlockWorkerLoop(ctx context.Context, gbm *blockDownload
...
-				Msg(WrapStagedSyncMsg("downloadRawBlocks failed, received empty blockBytes"))
+				Msg(WrapStagedSyncMsg("downloadRawBlocks failed, received invalid (nil) blockBytes"))
+			err := errors.New("downloadRawBlocks received invalid (nil) blockBytes")
+			gbm.HandleRequestError(batch, err, stid)
+			b.configs.protocol.StreamFailed(stid, "downloadRawBlocks failed")
+		} else if len(blockBytes) == 0 {
+			utils.Logger().Warn().
+				Str("stream", string(stid)).
+				Interface("block numbers", batch).
+				Msg(WrapStagedSyncMsg("downloadRawBlocks failed, received empty blockBytes, remote peer is not fully synced"))
```
sophoah, line 182: if the stream peer is offline, blockbyte will also be empty right ? if so not all the time it means the peer is not fully synced

sophoah (commented):
I just have a few comments / questions but overall looks ok. We still need to test that in devnet cc @diego1q2w

p2p/stream/types/stream.go
GheisMohammadi: They never back to 0 except if they can work without issue for 30 minutes. Otherwise once they reach max failures, they will be removed

api/service/stagedstreamsync/stage_bodies.go
GheisMohammadi: Based on my tests, only if stream is active, but it is unsynced, they return empty blocks, otherwise they return nil or error or timeout

p2p/stream/common/requestmanager/interface_test.go
GheisMohammadi: can be another PR

p2p/stream/types/stream.go
sophoah: thanks yeah this is what was also expected as a behavior here.

api/service/stagedstreamsync/stage_bodies.go
sophoah: i see noted thanks for the clarification

ONECasey (approved):

PR #4495, opened 2023-08-26
HIP-30 Boilerplate

MaxMustermann2:
This pull request contains
- Addition of HIP-30 parameters to the sharding configuration: emission split and emission destination address
- A hard fork scheduled sometime TBD in the future for HIP-30, at which point `genesis.HarmonyAccountsPostHIP30` goes into effect. It contains our existing keys except those which belonged to shards 2 and 3 per the original formula.
- Minor comment edits in `core/state/statedb.go`
- Largely cosmetic changes to the many `MustNewInstance` lines. Earlier, they were extending across the screen's viewport but now they have been broken down into multiple lines for improved readability.


ONECasey (approved):

ONECasey (approved):

internal/configs/sharding/instance.go
```diff
@@ -37,18 +37,26 @@ type instance struct {
...
+	numNodesPerShard,
+	numHarmonyOperatedNodesPerShard,
+	slotsLimit int,
+	harmonyVotePercent numeric.Dec,
 	hmyAccounts []genesis.DeployAccount,
 	fnAccounts []genesis.DeployAccount,
 	allowlist Allowlist,
 	feeCollectors FeeCollectors,
+	emissionFractionToRecovery numeric.Dec,
+	recoveryAddress ethCommon.Address,
```
sophoah, line 59: why is recovery address not a configuration as part of sharding config like here https://github.com/harmony-one/harmony/blob/1b9614ba240de1fb431c8f800bb378f2358ae7eb/internal/configs/sharding/mainnet.go#L52

sophoah: let's leave this here as I am thinking  we may in the future change to another address eventually.

sophoah (approved):
LGTM

sophoah:
note something important was the HarmonyAccountsPostHIP30 which I've confirmed is correct.

sophoah:
@ONECasey if we do two hardfork instead of 1, this one will definitely require some updates, however we need to see those two hardforks globally with all the PRs submitted by Max

ONECasey:
> @ONECasey if we do two hardfork instead of 1, this one will definitely require some updates, however we need to see those two hardforks globally with all the PRs submitted by Max

I say we try and put all HIP-30 updates into 1 fork

PR #4499, opened 2023-08-26
HIP-30: Balance migration

MaxMustermann2:
Depends on #4498 and #4494. 

Needs significant testing, so I am leaving it as draft with edits by maintainers allowed (you can simply push to my remote). I did add the ability to test this on devnet, so maybe that could bring some clarity about its efficacy. 

Lastly, the feature to generate a pre-migration report needs to be built. It should be fairly trivial to dump everything in a CSV at the last block of epoch T - 2.

PR #4494, opened 2023-08-26
HIP-30: Set up pre-image generation, recording, export and import

MaxMustermann2:
Leaving here what's a preimage : https://river.com/learn/terms/p/preimage

This PR implement the preimage function which will allow us to retrieve all the addresses bases on their hash.

PR #4503, opened 2023-09-11
HIP-30 Shard Reduction Process

ONECasey:
Combination of #4494, #4498, #4499, #4507, #4508, #4510, #4513, #4514 and #4502


ONECasey:
@diego1q2w Final one! This should be a combination of the entire migration and shard reduction process.

adsorptionenthalpy (approved):

sophoah:
we need 4508 also in which was tested locally and with the account migration working.

sophoah (approved):

PR #4517, opened 2023-10-03
reset devnet and set 30M epoch for all network except mainnet/testnet

sophoah:
This PR is preparing for the devnet reset by :
- updating the different epoch feature activation
- updating the devnet instance configuration

and also update the epoch activation for 30M block gas for all network except mainnet and testnet

### DIFF
--- a/internal/configs/sharding/partner.go
+++ b/internal/configs/sharding/partner.go
@@ -14,8 +14,6 @@ import (
 // configuration schedule.
 var PartnerSchedule partnerSchedule
 
-var feeCollectEpochV1 = big.NewInt(574)
-
 var feeCollectorsDevnet = []FeeCollectors{
 	FeeCollectors{
 		mustAddress("0xb728AEaBF60fD01816ee9e756c18bc01dC91ba5D"): numeric.OneDec(),
--- a/internal/configs/sharding/partner.go
+++ b/internal/configs/sharding/partner.go
@@ -44,10 +44,6 @@ func (ps partnerSchedule) InstanceForEpoch(epoch *big.Int) Instance {
 	switch {
 	case params.PartnerChainConfig.IsHIP30(epoch):
 		return partnerV4
-	case params.PartnerChainConfig.IsFeeCollectEpoch(epoch):
-		return partnerV3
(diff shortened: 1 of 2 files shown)


internal/configs/sharding/partner.go
```diff
@@ -44,10 +44,6 @@ func (ps partnerSchedule) InstanceForEpoch(epoch *big.Int) Instance {
 	switch {
 	case params.PartnerChainConfig.IsHIP30(epoch):
 		return partnerV4
-	case params.PartnerChainConfig.IsFeeCollectEpoch(epoch):
-		return partnerV3
-	case epoch.Cmp(feeCollectEpochV1) >= 0:
-		return partnerV2
```
GheisMohammadi, line 50: do we need to remove the checkings for V2 and V3? Jumping from V1 to V4?
I think keeping these lines helps to see other versions and users could see the path from V1 to V4.

GheisMohammadi:
The Travis tests are failed in this PR because the **checkGetReceiptsResult** redeclared in same package. I fixed this in PR #4518

internal/configs/sharding/partner.go
sophoah: this section doesn't check anything it just return the instance configuration ie `InstanceForEpoch` to be used
v2 basically implemented hip28 partially when we only had one multisig to send the fee collection.
v3 implemented the full hip29 that contained both multisig (infra and community)
i kept partnerV2 and partnerV3 for historical reason so we can go back on it eventually.

GheisMohammadi: developers look at this part and can't see any historical reason for jumping from V1 to V4. This is the reason we should keep that part. I don't see any benefits of removing it.

Frozen (approved):

internal/configs/sharding/partner.go
sophoah: but there is also no business benefits of keeping instance configuration. Ok you know what, I'll simply delete and replace V4 with V2 :)

GheisMohammadi (approved):

sunwavesun (approved):

sophoah:
rebased the PR with the latest dev

PR #4518, opened 2023-10-03
fix duplicate function def

GheisMohammadi:
## Issue

function **checkGetReceiptsResult** is defined two times in same package. This issue could be because of merge conflicts resolution. This PR fixes that.


sophoah (approved):

sophoah:
this PR also fixes the latest CI failure

PR #4519, opened 2023-10-05
add GetNodeData tests for stream client, increase nodes and receipts cap

GheisMohammadi:
## Issue

This PR adds a test for **GetNodeData** in stream client and increases the client cap for **GetNodedata** and **GetReceipts**. So, this PR increases the capacity for receipt and state requests. This enhancement enables nodes to request more data per query. Although the numbers are still lower than Ethereum's capacity, they are now substantial enough to handle large batches of receipts and states per request. This has the potential to significantly boost the performance of state synchronization.

### DIFF
--- a/p2p/stream/protocols/sync/const.go
+++ b/p2p/stream/protocols/sync/const.go
@@ -19,11 +19,11 @@ const (
 
 	// GetNodeDataCap is the cap of request of single GetNodeData request
 	// This number has an effect on maxMsgBytes as 20MB defined in github.com/harmony-one/harmony/p2p/stream/types.
-	GetNodeDataCap = 10
+	GetNodeDataCap = 256
 
 	// GetReceiptsCap is the cap of request of single GetReceipts request
 	// This number has an effect on maxMsgBytes as 20MB defined in github.com/harmony-one/harmony/p2p/stream/types.
-	GetReceiptsCap = 10
+	GetReceiptsCap = 128
 
 	// MaxStreamFailures is the maximum allowed failures before stream gets removed
 	MaxStreamFailures = 5
--- a/p2p/stream/protocols/sync/chain_test.go
+++ b/p2p/stream/protocols/sync/chain_test.go
@@ -75,6 +75,21 @@ func checkGetReceiptsResult(b []byte, hs []common.Hash) error {
 	return nil
 }
 
+func checkGetNodeDataResult(b []byte, hs []common.Hash) error {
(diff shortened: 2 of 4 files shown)


p2p/stream/protocols/sync/const.go
```diff
@@ -19,11 +19,11 @@ const (
 
 	// GetNodeDataCap is the cap of request of single GetNodeData request
 	// This number has an effect on maxMsgBytes as 20MB defined in github.com/harmony-one/harmony/p2p/stream/types.
-	GetNodeDataCap = 10
+	GetNodeDataCap = 256
```
sophoah, line 22: were there any analysis done on the change of this cap ? is the comment `// This number has an effect on maxMsgBytes as 20MB defined in github.com/harmony-one/harmony/p2p/stream/types.` still applicable ?

p2p/stream/protocols/sync/const.go
```diff
@@ -19,11 +19,11 @@ const (
...
 
 	// GetReceiptsCap is the cap of request of single GetReceipts request
 	// This number has an effect on maxMsgBytes as 20MB defined in github.com/harmony-one/harmony/p2p/stream/types.
-	GetReceiptsCap = 10
+	GetReceiptsCap = 128
```
sophoah, line 26: same question as above

p2p/stream/protocols/sync/const.go
GheisMohammadi: I chose a maximum of 256 for the number of states data per request because the states themselves are relatively small in size. Assuming an average state size of 1 kilobyte, when we request 256 states, the total data size amounts to 256 kilobytes. By keeping the number of states per request at 256, we ensure that the data size remains well below 20 megabytes (20MB), which is a reasonable and manageable size for data transfer and processing. This number for other chain like Ethereum is larger. Later we can increase this number again.

p2p/stream/protocols/sync/const.go
GheisMohammadi: same answer as above

GheisMohammadi:
> Could update the PR description to explain what's the intent of this PR

updated

sophoah (approved):

p2p/stream/protocols/sync/const.go
sophoah: thanks

ONECasey (approved):

PR #4521, opened 2023-10-06
fix: make the comission rate increasse come into effect one epoch before

diego1q2w:


ONECasey (approved):

sophoah:
@diego1q2w before approving that PR, we should be able to test that in localnet

sophoah:
@diego1q2w please fix the travis-ci build as well

PR #4526, opened 2023-10-10
enable hip30 epoch for testnet

diego1q2w:
## Issue

Setting the epoch activation for hip30 in testnet

### DIFF
--- a/internal/params/config.go
+++ b/internal/params/config.go
@@ -117,7 +117,7 @@ var (
 		FeeCollectEpoch:                        big.NewInt(1296), // 2023-04-28 07:14:20+00:00
 		ValidatorCodeFixEpoch:                  big.NewInt(1296), // 2023-04-28 07:14:20+00:00
 		HIP30Epoch:                             big.NewInt(2176), // 2023-10-12 10:00:00+00:00
-		BlockGas30MEpoch:                       EpochTBD,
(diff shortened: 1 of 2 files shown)


internal/params/config.go
```diff
@@ -116,7 +116,7 @@ var (
 		LeaderRotationExternalBeaconLeaders:    EpochTBD,
 		FeeCollectEpoch:                        big.NewInt(1296), // 2023-04-28 07:14:20+00:00
 		ValidatorCodeFixEpoch:                  big.NewInt(1296), // 2023-04-28 07:14:20+00:00
-		HIP30Epoch:                             EpochTBD,
+		HIP30Epoch:                             big.NewInt(2176), // 2023-10-10 10:00:00+00:00
```
sophoah, line 119: big.NewInt(2176), // 2023-10-10 10:00:00+00:00 too early ? remember we need to sometime for the pre hip30 epoch for account migration to happen

sophoah:
could you add activation epoch 30m gas as well

sophoah:
yes same as hip30

internal/params/config.go
sophoah: can you update the comment with the correct date and time

sophoah (approved):

GheisMohammadi (approved):

PR #4529, opened 2023-10-11
fix preimage import bugs

diego1q2w:
## Issue

Fixing two things here:  
1. If the import haven't run in the past then we need to ignore the not found error when trying to fetch the last successful import (that is used for metrics only)
2. If the ending state is nil (meaning the generation is up to date) then we skip the final commitment

### DIFF
--- a/core/preimages.go
+++ b/core/preimages.go
@@ -233,7 +233,7 @@ func GeneratePreimages(chain BlockChain, start, end uint64) error {
 
 		if stateAt != nil {
 			if root, err := endingState.Commit(false); err != nil {
-				return fmt.Errorf("unabe to commit state for block '%d': %w", i, err)
+				return fmt.Errorf("unable to commit state for block '%d': %w", i, err)
 			} else if root.Hex() != block.Root().Hex() {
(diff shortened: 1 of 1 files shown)


ONECasey (approved):

core/preimages.go
```diff
@@ -245,10 +248,11 @@ func GeneratePreimages(chain BlockChain, start, end uint64) error {
 	}
 	// force any pre-images in memory so far to go to disk, if they haven't already
 	fmt.Println("committing images")
-	if _, err := endingState.Commit(false); err != nil {
-		return fmt.Errorf("unabe to commit state for block: %w", err)
+	if endingState != nil {
+		if _, err := endingState.Commit(false); err != nil {
+			return fmt.Errorf("unabe to commit state for block: %w", err)
```
sophoah, line 253: there is a small typo here

GheisMohammadi (approved):

core/preimages.go
GheisMohammadi: unabe > unable

sophoah (approved):

PR #4539, opened 2023-10-18
Fixed memory leak and deadcode.

Frozen:


consensus/downloader.go
```diff
@@ -39,24 +39,20 @@ func newDownloadHelper(c *Consensus, d downloader) *downloadHelper {
...
 		finishedSub: finishedSub,
 	}
+	go out.downloadStartedLoop()
+	go out.downloadFinishedLoop()
+	return out
 }
 
 func (dh *downloadHelper) start() {
-	go dh.downloadStartedLoop()
-	go dh.downloadFinishedLoop()
```
Frozen, line 54: The function `start` initiates an internal loop and awaits the occurrence of `close`, which never happens.

adsorptionenthalpy (approved):

Issue #4549, opened 2023-11-02
HIP30 - min 7% fee not working for validator with max fee below 7%


sophoah:
There are validators with max fees at 5% and the HIP30 protocol automatic increase didn't work:
- https://staking.harmony.one/validators/mainnet/one1xrtkrcpx7edw40zxpp26up939gc68u8hwepvnx
- https://staking.harmony.one/validators/mainnet/one132ggm0mgxdt4l8etud9e5gzpv8psgx30p7hxsq

```
hmy --node="https://api.s0.t.hmny.io" blockchain validator information one1xrtkrcpx7edw40zxpp26up939gc68u8hwepvnx | jq '.result.validator["max-rate"]'
"0.050000000000000000"

hmy --node="https://api.s0.t.hmny.io" blockchain validator information one132ggm0mgxdt4l8etud9e5gzpv8psgx30p7hxsq | jq '.result.validator["max-rate"]'
"0.050000000000000000"
```

sophoah:
@diego1q2w  when you have some time could you help check whether your fix will also fix the validator information.

The below output is for a validator currently hitting the 5% fee bug:
```
hmy --node="https://api.s0.t.hmny.io" blockchain validator information one1rxdtt9fvgdcy3n5sy6e4zxg0s4hn4shaqf4qtp
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "active-status": "active",
    "booted-status": "lost epos auction",
    "current-epoch-performance": null,
    "currently-in-committee": false,
    "epos-status": "eligible to be elected next epoch",
    "epos-winning-stake": null,
```
the issue here is that currently this validator is active and signing and some of the field above are not right, see here for another validator also active but not hitting the 5% fee bug

```
hmy --node="https://api.s0.t.hmny.io" blockchain validator information one1upv5r5y3n5zce7a3wdj4z7pqzussqtdm35rcl5 | more
{
  "id": "0",
  "jsonrpc": "2.0",
  "result": {
    "active-status": "active",
    "booted-status": null,
    "current-epoch-performance": {
      "block": 49638807,
      "current-epoch-signing-percent": {
        "current-epoch-signed": 561830,
        "current-epoch-signing-percentage": "1.000000000000000000",
        "current-epoch-to-sign": 561830
      },
      "epoch": 1687
    },
    "currently-in-committee": true,
    "epos-status": "currently elected",
    "epos-winning-stake": "296673299565500747935060671.600000000000000000",
 ```

[...]

PR #4552, opened 2023-11-03
fix: max-rate bellow the era min-rate

diego1q2w:
### Issue 
Following the discovery in [issue #4549](https://github.com/harmony-one/harmony/issues/4549) on the Harmony repository, there is an operational inconsistency that occurs before a block is finalized. During the process where the validator data stored in memory, are written to disk as defined in [statedb.go lines 896-898](https://github.com/harmony-one/harmony/blob/258d5de4661d5b4b5ca911640400c47753c8159d/core/state/statedb.go#L896-L898), a sanity check is performed. This check executes a series of rules to ensure the integrity of the data before it's committed to disk. One particular rule, which can be found in [validator.go lines 315-320](https://github.com/harmony-one/harmony/blob/258d5de4661d5b4b5ca911640400c47753c8159d/staking/types/validator.go#L315-L320), verifies that the rate is not exceeding the maximum allowed rate. If the rate is greater than the maximum, the write operation fails.

This issue became prominent after the HIP30 hard fork, which increased the minimum rate to 7%. Subsequently, validators with a 5% maximum rate were unable to adjust their fees accordingly due to the aforementioned validation rule.

In this PR, we address the issue by adjusting the max-rate to be at least equal to the sum of the minRate and the max-increase-rate. This change requires a hard fork, as it necessitates consensus across the majority of the nodes to ensure that all participants agree on the same value.

### DIFF
--- a/staking/availability/measure.go
+++ b/staking/availability/measure.go
@@ -267,3 +267,27 @@ func UpdateMinimumCommissionFee(
 	}
 	return false, nil
 }
+
+// UpdateMaxCommissionFee makes sure the max-rate is at least higher than the rate + max-rate-change.
+func UpdateMaxCommissionFee(state *state.DB, addr common.Address, minRate numeric.Dec) (bool, error) {
+	utils.Logger().Info().Msg("begin update max commission fee")
+
+	wrapper, err := state.ValidatorWrapper(addr, true, false)
+	if err != nil {
+		return false, err
+	}
+
+	minMaxRate := minRate.Add(wrapper.MaxChangeRate)
+
+	if wrapper.MaxRate.LT(minMaxRate) {
+		utils.Logger().Info().
+			Str("addr", addr.Hex()).
+			Str("old max-rate", wrapper.MaxRate.String()).
+			Str("new max-rate", minMaxRate.String()).
+			Msg("updating max commission rate")
+		wrapper.MaxRate.SetBytes(minMaxRate.Bytes())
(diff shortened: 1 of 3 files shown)


sophoah:
can you test it on localnet with external validator created with a max rat of 5% ?

staking/availability/measure.go
```diff
@@ -267,3 +267,27 @@ func UpdateMinimumCommissionFee(
 	}
 	return false, nil
 }
+
+// UpdateMaxCommissionFee makes sure the max-rate is at least higher than the rate + max-rate-change.
+func UpdateMaxCommissionFee(state *state.DB, addr common.Address, minRate numeric.Dec) (bool, error) {
```
Frozen, line 272: bool value returns from function, but never used. Maybe use it in tests?

diego1q2w: yeah good catch! i wanted to keep the same function signature as the `UpdateMinCommissionFee` but as you said if not used then we can remove it

diego1q2w:
> > can you test it on localnet with external validator created with a max rat of 5% ?
> 
> yup! will do

And it works! tested it two validators as you can see in the PR test script, one had the max-rate bellow the min and the other had the max-rate above, the latter got updated as soon as hip30 hit, while the former didn't (as expected) however when the MaxRate epoch hit the former got the update as well.

Frozen (approved):

PR #4553, opened 2023-11-07
Timeout for block proposal.

Frozen:


consensus/consensus_v2.go
```diff
@@ -688,37 +687,36 @@ func (consensus *Consensus) commitBlock(blk *types.Block, committedMsg *FBFTMess
...
 	var (
 		bc        = consensus.Blockchain()
-		prev      = consensus.getLeaderPubKey()
 		leader    = consensus.getLeaderPubKey()
 		curBlock  = bc.CurrentBlock()
 		curNumber = curBlock.NumberU64()
 		curEpoch  = curBlock.Epoch().Uint64()
 	)
-	const blocksCountAliveness = 10
+	const blocksCountAliveness = 4
```
GheisMohammadi, line 698: could you please explain this?

Frozen: 4 is more comfortable for localnet. If no blocks from validator for 4 blocks means validator is inactive, and we can skip him from future rotating.

PR #4554, opened 2023-11-08
Refactored verify block.

Frozen:
Complicated logic with closures and abstraction level for no reason.


GheisMohammadi:
Why do even we need to remove **BlockVerifier**? It doesn't seems complicated. How that helps? **BlockVerifier** lets you attach a verifier dynamically based on the structure. This PR makes it static and rigid.

Issue #4558, opened 2023-11-09
Data race


Frozen:
WARNING: DATA RACE
Write at 0x00c000940e60 by goroutine 1903:
  github.com/harmony-one/harmony/consensus/quorum.(*cIdentities).UpdateParticipants()
      /go/src/github.com/harmony-one/harmony/consensus/quorum/quorum.go:332 +0xf0
  github.com/harmony-one/harmony/consensus/quorum.(*uniformVoteWeight).UpdateParticipants()
      <autogenerated>:1 +0x7c
  github.com/harmony-one/harmony/consensus.(*Consensus).updatePublicKeys()
      /go/src/github.com/harmony-one/harmony/consensus/consensus_service.go:85 +0x80
  github.com/harmony-one/harmony/consensus.(*Consensus).updateConsensusInformation()
      /go/src/github.com/harmony-one/harmony/consensus/consensus_service.go:412 +0x94c
  github.com/harmony-one/harmony/consensus.(*Consensus).syncReadyChan()
   

Previous read at 0x00c000940e60 by goroutine 2851:
  github.com/harmony-one/harmony/consensus/quorum.(*cIdentities).Participants()
      /go/src/github.com/harmony-one/harmony/consensus/quorum/quorum.go:318 +0x2c
  github.com/harmony-one/harmony/consensus/quorum.(*uniformVoteWeight).Participants()
      <autogenerated>:1 +0x48
  github.com/harmony-one/harmony/consensus.(*Consensus).NumSignaturesIncludedInBlock()
      /go/src/github.com/harmony-one/harmony/consensus/consensus_service.go:630 +0x4c
  github.com/harmony-one/harmony/node.(*Node).PostConsensusProcessing()

Issue #4557, opened 2023-11-08
Data Race


Frozen:
WARNING: DATA RACE
Read at 0x00c00a8da7b0 by goroutine 5551:
  github.com/harmony-one/harmony/p2p/stream/common/streammanager.(*streamManager).loop.func1()
      /Users/frozen/go/src/github.com/harmony-one/harmony/p2p/stream/common/streammanager/streammanager.go:143 +0x34

Previous write at 0x00c00a8da7b0 by goroutine 1435:
  github.com/harmony-one/harmony/p2p/stream/common/streammanager.(*streamManager).loop()
      /Users/frozen/go/src/github.com/harmony-one/harmony/p2p/stream/common/streammanager/streammanager.go:141 +0x290
  github.com/harmony-one/harmony/p2p/stream/common/streammanager.(*streamManager).Start.func1()
      /Users/frozen/go/src/github.com/harmony-one/harmony/p2p/stream/common/streammanager/streammanager.go:105 +0x34

Issue #4556, opened 2023-11-08
DATA RACE


Frozen:
WARNING: DATA RACE
Read at 0x00c0097b3ce8 by goroutine 3363:
  github.com/harmony-one/harmony/core.(*BlockChainImpl).saveLeaderRotationMeta()
      /Users/frozen/go/src/github.com/harmony-one/harmony/core/blockchain_leader_rotation.go:101 +0xd0
  github.com/harmony-one/harmony/core.(*BlockChainImpl).InsertChain()
      /Users/frozen/go/src/github.com/harmony-one/harmony/core/blockchain_impl.go:1658 +0x3a4
  github.com/harmony-one/harmony/consensus.(*Consensus).commitBlock()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:665 +0x198
  github.com/harmony-one/harmony/consensus.(*Consensus).finalCommit()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:222 +0xbe4
  github.com/harmony-one/harmony/consensus.(*Consensus).onCommit()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/leader.go:287 +0xe18
  github.com/harmony-one/harmony/consensus.(*Consensus).HandleMessageUpdate()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:129 +0x428
  github.com/harmony-one/harmony/consensus.(*Consensus).HandleMessageUpdate-fm()
      <autogenerated>:1 +0x64
  github.com/harmony-one/harmony/node.(*Node).StartPubSub.func2.1()
      /Users/frozen/go/src/github.com/harmony-one/harmony/node/node.go:908 +0x19c

Previous write at 0x00c0097b3ce8 by goroutine 3358:
  github.com/harmony-one/harmony/core.(*BlockChainImpl).saveLeaderRotationMeta()
      /Users/frozen/go/src/github.com/harmony-one/harmony/core/blockchain_leader_rotation.go:101 +0x250
  github.com/harmony-one/harmony/core.(*BlockChainImpl).InsertChain()
      /Users/frozen/go/src/github.com/harmony-one/harmony/core/blockchain_impl.go:1658 +0x3a4
  github.com/harmony-one/harmony/consensus.(*Consensus).preCommitAndPropose.func1()
[...]

Issue #4555, opened 2023-11-08
DATA RACE


Frozen:
WARNING: DATA RACE
Write at 0x00c000b327b0 by goroutine 3905:
  github.com/harmony-one/harmony/internal/utils.(*Timeout).Start()
      /Users/frozen/go/src/github.com/harmony-one/harmony/internal/utils/timer.go:32 +0x4f4
  github.com/harmony-one/harmony/consensus.(*Consensus).preCommitAndPropose.func1()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:583 +0x4b4

Previous read at 0x00c000b327b0 by goroutine 747:
  github.com/harmony-one/harmony/internal/utils.(*Timeout).Expired()
      /Users/frozen/go/src/github.com/harmony-one/harmony/internal/utils/timer.go:44 +0x3c
  github.com/harmony-one/harmony/consensus.(*Consensus).tick()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:405 +0x384
  github.com/harmony-one/harmony/consensus.(*Consensus).Tick()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:376 +0x84
  github.com/harmony-one/harmony/consensus.(*Consensus).Start.func1.1()
      /Users/frozen/go/src/github.com/harmony-one/harmony/consensus/consensus_v2.go:314 +0x8c

PR #4559, opened 2023-11-09
Fixed data race.

Frozen:
https://github.com/harmony-one/harmony/issues/4558
https://github.com/harmony-one/harmony/issues/4557
https://github.com/harmony-one/harmony/issues/4556
https://github.com/harmony-one/harmony/issues/4555

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -575,19 +575,19 @@ func (consensus *Consensus) preCommitAndPropose(blk *types.Block) error {
 		if _, err := consensus.Blockchain().InsertChain([]*types.Block{blk}, !consensus.FBFTLog().IsBlockVerified(blk.Hash())); err != nil {
 			switch {
 			case errors.Is(err, core.ErrKnownBlock):
-				consensus.getLogger().Info().Msg("[preCommitAndPropose] Block already known")
+				consensus.GetLogger().Info().Msg("[preCommitAndPropose] Block already known")
 			default:
-				consensus.getLogger().Error().Err(err).Msg("[preCommitAndPropose] Failed to add block to chain")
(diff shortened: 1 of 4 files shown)


consensus/consensus_v2.go
```diff
@@ -575,19 +575,19 @@ func (consensus *Consensus) preCommitAndPropose(blk *types.Block) error {
...
 				return
 			}
 		}
-
+		consensus.mutex.Lock()
 		consensus.getLogger().Info().Msg("[preCommitAndPropose] Start consensus timer")
 		consensus.consensusTimeout[timeoutConsensus].Start()
 
 		// Send signal to Node to propose the new block for consensus
 		consensus.getLogger().Info().Msg("[preCommitAndPropose] sending block proposal signal")
```
GheisMohammadi, line 589: GetLogger() ?

consensus/consensus_v2.go
```diff
@@ -575,19 +575,19 @@ func (consensus *Consensus) preCommitAndPropose(blk *types.Block) error {
...
-
+		consensus.mutex.Unlock()
```
GheisMohammadi, line 590: defer?

consensus/consensus_v2.go
Frozen: No, we obtained lock above.

Frozen: GetLogger() has lock inside, so it will be deadlock.

GheisMohammadi: GetLogger is a public function, it should be improved to not have such a constraints. We can't make a function public and ask users to not use it in a certain conditions, right?

consensus/consensus_v2.go
GheisMohammadi: to excludes line  `consensus.ReadySignal(AsyncProposal)` ? any reason for that?

PR #4562, opened 2023-11-14
Fixed panic with dHelper.

Frozen:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1a973f5]
goroutine 5441 [running]:
github.com/harmony-one/harmony/consensus.(*Consensus).spinUpStateSync(0xc00017a1e0)
        /home/sp/harmony/harmony/consensus/downloader.go:107 +0x35
github.com/harmony-one/harmony/consensus.(*Consensus).onCommitted(0xc00017a1e0, 0xc008e3cfc0)

### DIFF
--- a/consensus/consensus_v2.go
+++ b/consensus/consensus_v2.go
@@ -277,6 +277,8 @@ func (consensus *Consensus) BlockCommitSigs(blockNum uint64) ([]byte, error) {
 		return nil, nil
 	}
 	lastCommits, err := consensus.Blockchain().ReadCommitSig(blockNum)
+	consensus.mutex.Lock()
+	defer consensus.mutex.Unlock()
 	if err != nil ||
 		len(lastCommits) < bls.BLSSignatureSizeInBytes {
 		msgs := consensus.FBFTLog().GetMessagesByTypeSeq(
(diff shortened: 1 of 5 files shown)


GheisMohammadi:
it would be good if you can explain the issue and how this PR fixes it

consensus/downloader.go
```diff
@@ -19,12 +19,13 @@ type downloader interface {
 // Set downloader set the downloader of the shard to consensus
 // TODO: It will be better to move this to consensus.New and register consensus as a service
 func (consensus *Consensus) SetDownloader(d downloader) {
+	consensus.mutex.Lock()
+	defer consensus.mutex.Unlock()
 	consensus.dHelper = newDownloadHelper(consensus, d)
```
Frozen, line 24: This is where we initialize value

consensus/downloader.go
```diff
@@ -104,7 +101,7 @@ func (consensus *Consensus) AddConsensusLastMile() error {
 }
 
 func (consensus *Consensus) spinUpStateSync() {
-	consensus.dHelper.d.DownloadAsync()
```
Frozen, line 107: Here we use dHelper

Frozen:
Use of dHelper can happen before it initialization. 
New version initializes dHelper in constructor with default value in order to avoid panic. 
Other changes are fixes for a small bugs, i've found while debugging.

consensus/consensus_v2.go
```diff
@@ -277,6 +277,8 @@ func (consensus *Consensus) BlockCommitSigs(blockNum uint64) ([]byte, error) {
 		return nil, nil
 	}
 	lastCommits, err := consensus.Blockchain().ReadCommitSig(blockNum)
+	consensus.mutex.Lock()
+	defer consensus.mutex.Unlock()
```
sophoah, line 281: could you explain why we need this here ?

sophoah:
> Use of dHelper can happen before it initialization.
> New version initializes dHelper in constructor with default value in order to avoid panic.
> Other changes are fixes for a small bugs, i've found while debugging.

Could you explain how at this issue happen. You have a fix here which is great, but i don't get how this issue happens at the first place

GheisMohammadi (approved):

consensus/consensus_v2.go
Frozen: this code is called from function `BlockCommitSigs` which is running in different goroutine. Without this lock it's going to be a race condition, because FBFTLog, which is called later, is not thread safe.

Frozen:
> > Use of dHelper can happen before it initialization.
> > New version initializes dHelper in constructor with default value in order to avoid panic.
> > Other changes are fixes for a small bugs, i've found while debugging.
> 
> Could you explain how at this issue happen. You have a fix here which is great, but i don't get how this issue happens at the first place

We had checks like `if dSync != nil` previously, so it didn't fail. Value is initialized with nil in constructor, later it's called or reinitialized, depending what happens faster

sophoah (approved):

PR #4565, opened 2023-11-17
Fix the block insertion in legacy sync for existed block

GheisMohammadi:
## Issue

In the legacy sync, if a block insertion is attempted and the block already exists, it breaks the entire sync process. This PR addresses this issue by adding an exception for known blocks.


Frozen (approved):

Frozen:
Thanks!

Issue #4564, opened 2023-11-16
ShardIDFromKey still return 4 shards when starting from 0


sophoah:
**Describe the bug**

When the node start from 0, HIP30 epoch hasn't been reached yet and the function below will still return shard 2 or 3 instead of 0 or 1
```golang
// ShardIDFromKey returns the shard ID statically determined from the input key
func (conf *ConfigType) ShardIDFromKey(key *bls_core.PublicKey) (uint32, error) {
	var pubKey bls.SerializedPublicKey
	if err := pubKey.FromLibBLSPublicKey(key); err != nil {
		return 0, errors.Wrapf(err,
			"cannot convert libbls public key %s to internal form",
			key.SerializeToHexStr())
	}
	epoch := conf.networkType.ChainConfig().StakingEpoch
	numShards := conf.shardingSchedule.InstanceForEpoch(epoch).NumShards()
	shardID := new(big.Int).Mod(pubKey.Big(), big.NewInt(int64(numShards)))
	return uint32(shardID.Uint64()), nil
}
```

**What to expect**
The shard belonging should be dependent on the latest epoch activation despite where the node has synced so far

mur-me:
Hey @sophoah and @Frozen, users are complaining about this on the Oct 21 and today 14 Nov, please bump the priority for this one, because validators user experience is suffering here :pray:

mur-me:
Double checked that this fix is working, methodology:
1. Generate keys for both shards - around 50 for each via `./hmy keys generate-bls-keys --count 50 --shard 1 --passphrase`
2. Filter only previously problematic one via - `ls -1 | grep key | cut -d '.' -f 1 | grep -E "[2,3,6,7,a,b,e,f]$")`
3. Run `./hmy utility shard-for-bls $i` against each one`

Expected result - only shard 0 and shard 1 returned
Result in format last chard - shard, no more shards 2 and 3 : 
```
7 - {"shard-id":1}
b - {"shard-id":1}
7 - {"shard-id":1}
6 - {"shard-id":0}
3 - {"shard-id":1}
2 - {"shard-id":0}
7 - {"shard-id":1}
f - {"shard-id":1}
7 - {"shard-id":1}
2 - {"shard-id":0}
3 - {"shard-id":1}
a - {"shard-id":0}
3 - {"shard-id":1}
e - {"shard-id":0}
f - {"shard-id":1}
6 - {"shard-id":0}
3 - {"shard-id":1}
3 - {"shard-id":1}
[...]

PR #4566, opened 2023-11-17
Fix for `ShardIDFromKey`

Frozen:
https://github.com/harmony-one/harmony/issues/4564

### DIFF
--- a/internal/configs/node/config.go
+++ b/internal/configs/node/config.go
@@ -344,7 +344,7 @@ func (conf *ConfigType) ShardIDFromKey(key *bls_core.PublicKey) (uint32, error) 
 			"cannot convert libbls public key %s to internal form",
 			key.SerializeToHexStr())
 	}
-	epoch := conf.networkType.ChainConfig().StakingEpoch
+	epoch := conf.networkType.ChainConfig().HIP30Epoch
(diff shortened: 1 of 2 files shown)


sophoah:
@Frozen i think that one is good. Can you add two tests dealing with a network that has HIP30 epoch defined as TBD, and another one set with a fixed number. First one should use the default / genesis sharding config while the other one should use the sharding config of HIP30.

Things to note, at the end what we want is to know the latest sharding configuration (4 shard vs 2 shards) when we start the node from height 0, be it with HIP30 or with future HIP, so maybe we could define a function that does just that.

@diego1q2w @GheisMohammadi feel free to chime in any other ideas.

sophoah:
@Frozen not sure why but something weird here. I was working on testing that PR and 
```golang
numShards := conf.shardingSchedule.InstanceForEpoch(epoch).NumShards()
```
always give me numShards = 2 whether I use HIP30Epoch or stakingEpoch in `epoch := conf.networkType.ChainConfig().StakingEpoch`

Can you double confirm ?

sophoah:
nevermind I found the issue

GheisMohammadi (approved):

mur-me (approved):

PR #4567, opened 2023-11-20
Fix for last mile block

Frozen:
`addConsensusLastMile` should be used only on `0` shard

### DIFF
--- a/api/service/stagedstreamsync/syncing.go
+++ b/api/service/stagedstreamsync/syncing.go
@@ -285,7 +285,7 @@ func (s *StagedStreamSync) doSync(downloaderContext context.Context, initSync bo
 	}
 
 	// add consensus last mile blocks
-	if s.consensus != nil {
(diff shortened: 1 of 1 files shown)


sophoah:
@Frozen can you explain why you say so ? isn't there as well consensus running on shard 1 ?

GheisMohammadi (approved):

api/service/stagedstreamsync/syncing.go
```diff
@@ -285,7 +285,7 @@ func (s *StagedStreamSync) doSync(downloaderContext context.Context, initSync bo
 	}
 
 	// add consensus last mile blocks
-	if s.consensus != nil {
+	if s.consensus != nil && s.consensus.ShardID == 0 {
```
GheisMohammadi, line 288: if you need to have this only for beacon nodes, please use   `... && s.isBeaconNode` because the epoch chain shard number is also zero.

PR #4568, opened 2023-11-21
Rollback removes physically block data from db. 

Frozen:
Rollback removes physically block data from db.


sophoah:
I believe merely deleting the block physically may not suffice for a rollback. We need to consider the rollback of states, the explorer database, and for s0, the rollback of stored crosslinks. There may be more DBs to consider. f we merge this PR, what would be the impact or doing a rollback without reverting the previously suggested DBs

As discussed in our recent meeting, the current state of the rollback should only be used in cases where the node has forked or gotten stuck, and a rollback of just one block is sufficient to restore the node.

PR #4569, opened 2023-11-21
Removed future blocks from blockchain_impl.go

Frozen:
Redundant code: It is imperative that we exercise control over block insertion. Running the specified goroutine may lead to unintended side effects.


sophoah (approved):

sophoah:
Note : the code we are removing is not being used in the go-ethereum code. While i am not sure why we started to use it, i think for now we can remove it.

PR #4570, opened 2023-11-21
Elk stack for logs.

Frozen:


sophoah:
hey @Frozen let's move this to https://github.com/harmony-one/harmony-ops

there is another similar solution actually https://github.com/harmony-one/harmony-ops/tree/master/devops/harmony-node-log but i like what you did with the direct log view of the localnet

PR #4575, opened 2023-11-30
Fixed infinity loop sync.

Frozen:
Fixed issue and added additional logs.

### DIFF
--- a/api/service/stagedstreamsync/staged_stream_sync.go
+++ b/api/service/stagedstreamsync/staged_stream_sync.go
@@ -642,8 +642,9 @@ func (ss *StagedStreamSync) addConsensusLastMile(bc core.BlockChain, cs *consens
 			case errors.Is(err, core.ErrNotLastBlockInEpoch):
 			case err != nil:
 				return errors.Wrap(err, "failed to InsertChain")
+			default:
+				hashes = append(hashes, block.Header().Hash())
 			}
-			hashes = append(hashes, block.Header().Hash())
 		}
 		return nil
 	})
--- a/node/node.go
+++ b/node/node.go
@@ -499,10 +499,6 @@ func (node *Node) validateNodeMessage(ctx context.Context, payload []byte) (
 					utils.Logger().Debug().Uint64("receivedNum", block.NumberU64()).
 						Uint64("currentNum", curBeaconHeight).Msg("beacon block sync message rejected")
 					return nil, 0, errors.New("beacon block height smaller than current height beyond tolerance")
-				} else if block.NumberU64()-beaconBlockHeightTolerance > curBeaconHeight {
-					utils.Logger().Debug().Uint64("receivedNum", block.NumberU64()).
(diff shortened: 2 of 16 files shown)


api/service/stagedstreamsync/staged_stream_sync.go
```diff
@@ -642,8 +642,9 @@ func (ss *StagedStreamSync) addConsensusLastMile(bc core.BlockChain, cs *consens
 			case errors.Is(err, core.ErrNotLastBlockInEpoch):
 			case err != nil:
 				return errors.Wrap(err, "failed to InsertChain")
+			default:
```
Frozen, line 645: This line prevents the following code from executing, so no new hashes added to slice. Later we check count of hashes, and if it's not zero, we start sync again and again, which cause infinity loop.

sophoah: but if `err != nil` is true then we'll also never hit the `default` case

sophoah: is this links to the `Block exists` issue ?

Frozen: if `err != nil` we don't need default case and should return error.

node/node.go
```diff
@@ -499,10 +499,6 @@ func (node *Node) validateNodeMessage(ctx context.Context, payload []byte) (
 					utils.Logger().Debug().Uint64("receivedNum", block.NumberU64()).
 						Uint64("currentNum", curBeaconHeight).Msg("beacon block sync message rejected")
 					return nil, 0, errors.New("beacon block height smaller than current height beyond tolerance")
-				} else if block.NumberU64()-beaconBlockHeightTolerance > curBeaconHeight {
-					utils.Logger().Debug().Uint64("receivedNum", block.NumberU64()).
-						Uint64("currentNum", curBeaconHeight).Msg("beacon block sync message rejected")
-					return nil, 0, errors.Errorf("beacon block height too much higher than current height beyond tolerance, block %d, current %d, epoch %d , current %d", block.NumberU64(), curBeaconHeight, block.Epoch().Uint64(), curBeaconBlock.Epoch().Uint64())
```
sophoah, line 505: why did you remove this?

Frozen: https://github.com/harmony-one/harmony/blob/dev/node/node.go#L502
This code part doesn't exist in dev branch, it was previously deleted, looks like bug in github

Frozen: https://github.com/harmony-one/harmony/pull/4574

api/service/stagedstreamsync/staged_stream_sync.go
GheisMohammadi: @sophoah current code considers existed blocks and counts them as new added blocks. adding default helps to not count them as new added blocks because the block hash won't be added to slice. So, the sync loop checks number of hashes and it is zero, so it breaks.

GheisMohammadi (approved):

sophoah (approved):

PR #4580, opened 2023-12-04
Fix: max rate issue

diego1q2w:
### Issue 
Following the discovery in [issue #4549](https://github.com/harmony-one/harmony/issues/4549) on the Harmony repository, there is an operational inconsistency that occurs before a block is finalized. During the process where the validator data stored in memory, are written to disk as defined in [statedb.go lines 896-898](https://github.com/harmony-one/harmony/blob/258d5de4661d5b4b5ca911640400c47753c8159d/core/state/statedb.go#L896-L898), a sanity check is performed. This check executes a series of rules to ensure the integrity of the data before it's committed to disk. One particular rule, which can be found in [validator.go lines 315-320](https://github.com/harmony-one/harmony/blob/258d5de4661d5b4b5ca911640400c47753c8159d/staking/types/validator.go#L315-L320), verifies that the rate is not exceeding the maximum allowed rate. If the rate is greater than the maximum, the write operation fails.

This issue became prominent after the HIP30 hard fork, which increased the minimum rate to 7%. Subsequently, validators with a 5% maximum rate were unable to adjust their fees accordingly due to the aforementioned validation rule.

In this PR, we address the issue by adjusting the max-rate to be at least equal to the sum of the minRate and the max-increase-rate. This change requires a hard fork, as it necessitates consensus across the majority of the nodes to ensure that all participants agree on the same value.

### DIFF
--- a/internal/params/config.go
+++ b/internal/params/config.go
@@ -75,6 +75,7 @@ var (
 		ValidatorCodeFixEpoch:                 big.NewInt(1535), // 2023-07-20 05:51:07+00:00
 		HIP30Epoch:                            big.NewInt(1673), // 2023-11-02 17:30:00+00:00
 		BlockGas30MEpoch:                      big.NewInt(1673), // 2023-11-02 17:30:00+00:00
+		MaxRateEpoch:                          EpochTBD,
 	}
 
 	// TestnetChainConfig contains the chain parameters to run a node on the harmony test network.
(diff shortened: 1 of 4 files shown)


diego1q2w:
Re-opening this so the tests run

internal/params/config.go
```diff
@@ -290,8 +295,9 @@ var (
 		LeaderRotationExternalValidatorsEpoch: big.NewInt(6),
 		FeeCollectEpoch:                       big.NewInt(2),
 		ValidatorCodeFixEpoch:                 big.NewInt(2),
-		HIP30Epoch:                            EpochTBD,
+		HIP30Epoch:                            big.NewInt(3),
```
Frozen, line 298: It will never pass tests with such condition. We need to remove it from localnet, or create additional command in Makefile to start localnet with different options.

Frozen (approved):

PR #4584, opened 2023-12-11
Feature: last signing power.

Frozen:
Calculation on last inserted block, no leader required.


ONECasey (approved):

sophoah:
the last signing power is a good to have feature, however I am interested in the current signing power. During the multiple outages we had in the past, knowing how much voting power trying to sign the latest block proposed help for troubleshooting. Only the leader has that information currently.

with that PR, all validator could query for the last block total vote but not the current

PR #4586, opened 2023-12-13
Enable 1 second finality blocks 

ONECasey:


Frozen:
We can't just update the current code; we need new code for compatibility with previously stored data. I'm going to close this pr

PR #4608, opened 2024-01-14
Recreate trie after revert.

Frozen:


core/blockchain_impl.go
```diff
@@ -781,13 +781,37 @@ func (bc *BlockChainImpl) resetWithGenesisBlock(genesis *types.Block) error {
 	return nil
 }
 
+func (bc *BlockChainImpl) repairRecreateStateTries(head **types.Block) error {
```
Frozen, line 784: This should be ok, because blocks we previously checked, so we can recreate state.

sophoah:
let's test it we can kill the harmony process `sudo kill -9 harmony` and restart

PR #4609, opened 2024-01-16
fix concurrent map read and write issue for stakedVoteWeight

GheisMohammadi:
## Issue
The **stakedVoteWeight** structure doesn't support concurrent read and write. This could break the main process if an RPC triggers any write or read from that. This PR addresses this issue by adding a mutex to it and make it thread-safe.


ONECasey (approved):

Frozen:
Bug fixed in https://github.com/harmony-one/harmony/pull/4610

PR #4613, opened 2024-01-19
Fix fast sync null snapshot and null response issue

GheisMohammadi:
## Issue
If a snapshot is not available, the client receives null for snapshots and attempts to call methods on a null instance. This pull request addresses the issue of null snapshots and null responses in fast sync by checking and validating values.


ONECasey (approved):

api/service/stagedstreamsync/stage_statesync_full.go
```diff
@@ -190,7 +190,10 @@ func (sss *StageFullStateSync) runStateWorkerLoop(ctx context.Context, sdm *Full
 		default:
 		}
 		accountTasks, codes, storages, healtask, codetask, nTasks, err := sdm.GetNextBatch()
-		if nTasks == 0 || err != nil {
+		if nTasks == 0 {
+			return
```
sophoah, line 194: should we log a message here ?

sophoah (approved):

PR #4618, opened 2024-01-25
Fix snapshot creation, add snapshot configurations

GheisMohammadi:
## Issue
The snapshot creation was broken due to an incorrect root hash. This PR addresses that issue and also adds a new configuration section for cache settings.

### DIFF
--- a/cmd/harmony/flags.go
+++ b/cmd/harmony/flags.go
@@ -32,7 +32,6 @@ var (
 		legacyDataDirFlag,
 
 		taraceFlag,
-		triesInMemoryFlag,
 	}
 
 	dnsSyncFlags = []cli.Flag{
@@ -268,6 +267,16 @@ var (
 		gpoBlockGasLimitFlag,
 	}
 
+	cacheConfigFlags = []cli.Flag{
+		cacheDisabled,
+		cacheTrieNodeLimit,
+		cacheTriesInMemory,
+		cachePreimages,
+		cacheSnapshotLimit,
+		cacheSnapshotNoBuild,
+		cacheSnapshotWait,
+	}
+
 	metricsFlags = []cli.Flag{
 		metricsETHFlag,
 		metricsExpensiveETHFlag,
@@ -352,11 +361,6 @@ var (
 		Usage:    "indicates if full transaction tracing should be enabled",
 		DefValue: defaultConfig.General.TraceEnable,
 	}
-	triesInMemoryFlag = cli.IntFlag{
-		Name:     "blockchain.tries_in_memory",
-		Usage:    "number of blocks from header stored in disk before exiting",
-		DefValue: defaultConfig.General.TriesInMemory,
-	}
 )
 
 func getRootFlags() []cli.Flag {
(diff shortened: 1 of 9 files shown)


ONECasey (approved):

cmd/harmony/flags.go
```diff
@@ -2115,3 +2111,70 @@ func applyGPOFlags(cmd *cobra.Command, cfg *harmonyconfig.HarmonyConfig) {
 		cfg.GPO.BlockGasLimit = cli.GetIntFlagValue(cmd, gpoBlockGasLimitFlag)
 	}
 }
+
+// cache config flags
+var (
+	cacheDisabled = cli.BoolFlag{
+		Name:     "cache.disabled",
+		Usage:    "Whether to disable trie write caching (archive node)",
+		DefValue: defaultCacheConfig.Disabled,
```
sophoah, line 2120: so default will be enabled for archival ?

core/blockchain_impl.go
```diff
@@ -354,7 +354,8 @@ func newBlockChainWithOptions(
 			NoBuild:    bc.cacheConfig.SnapshotNoBuild,
 			AsyncBuild: !bc.cacheConfig.SnapshotWait,
 		}
-		bc.snaps, _ = snapshot.New(snapconfig, bc.db, bc.triedb, head.Hash())
+		fmt.Println("loading/generating snapshot...")
```
sophoah, line 357: should we have this log as part of the harmony system log ?

core/blockchain_impl.go
```diff
@@ -354,7 +354,8 @@ func newBlockChainWithOptions(
 			NoBuild:    bc.cacheConfig.SnapshotNoBuild,
 			AsyncBuild: !bc.cacheConfig.SnapshotWait,
 		}
-		bc.snaps, _ = snapshot.New(snapconfig, bc.db, bc.triedb, head.Hash())
```
sophoah, line 357: just out of curiosity, why is this fixing the hash ?

cmd/harmony/flags.go
sophoah: ok. that cache configuration doesn't determine whether the node will act as archive node or not right ?

core/blockchain_impl.go
sophoah: not all user are looking at the console logs .. since it may take a long time, let's also add a log in the system log.

core/blockchain_impl.go
sophoah: thanks

cmd/harmony/flags.go
GheisMohammadi: > ok. that cache configuration doesn't determine whether the node will act as archive node or not right ?

Correct 👍

core/blockchain_impl.go
GheisMohammadi: > not all user are looking at the console logs .. since it may take a long time, let's also add a log in the system log. 

True, I will add a log for system log as well

ONECasey (approved):

core/blockchain_impl.go
GheisMohammadi: > not all user are looking at the console logs .. since it may take a long time, let's also add a log in the system log.

done

PR #4621, opened 2024-01-26
Removed outdated flag, additional checks and simplified logic.

Frozen:


sophoah:
since the logic has changed, the proper way to upgrade the node with this PR would be :
- stop all of them 
- upgrade
- start all of them

there might be some node with a weird state after that if something goes wrong, so if that happened we can still reset devnet cc @diego1q2w

sophoah:
@Frozen would be great to explain the new logic  when you implement one. So from what i could read can you confirm if the new logic is as follow, from multiple set of error checking, if any of them happen, you are returning the leader that committed the last block.

https://github.com/harmony-one/harmony/blob/603d48a59d5e9f6885ec78cbb63d5066dd9eccc0/consensus/consensus_v2.go#L672

I think it's a good approach and worst that can happen, we'll always have the same leader until next election. In the other hand, a malicious leader could try to fail condition right after he become leader, luckily that bad leader should not be able to do more than being the leader since bad block / tx would be rejected by other validator in the network

GheisMohammadi:
> @Frozen would be great to explain the new logic when you implement one. So from what i could read can you confirm if the new logic is as follow, from multiple set of error checking, if any of them happen, you are returning the leader that committed the last block.
> 
> https://github.com/harmony-one/harmony/blob/603d48a59d5e9f6885ec78cbb63d5066dd9eccc0/consensus/consensus_v2.go#L672
> 
> I think it's a good approach and worst that can happen, we'll always have the same leader until next election. In the other hand, a malicious leader could try to fail condition right after he become leader, luckily that bad leader should not be able to do more than being the leader since bad block / tx would be rejected by other validator in the network
> 
> what do you think @ONECasey @GheisMohammadi @diego1q2w

What do we do with the malicious leader then? Can it be present in the next election again? Can the same situation occur again in subsequent elections?

sophoah:
> What do we do with the malicious leader then? Can it be present in the next election again? Can the same situation occur again in subsequent elections?

i think first is to be able to detect it. However, without moving further into those detail now are we ok with @Frozen proposed changes.

PR #4629, opened 2024-02-12
[HOTFIX] fix leader crosslink issue to not include old cross link in the propo…

GheisMohammadi:
## Issue

The leader is adding pending crosslinks without verifying them. This could break the consensus if other validators don't have the old shard state needed to verify crosslinks. This PR addresses that issue by ignoring very old crosslinks and not adding them to proposing blocks.

### DIFF
--- a/consensus/view_change_construct.go
+++ b/consensus/view_change_construct.go
@@ -464,7 +464,7 @@ func (vc *viewChange) InitPayload(
 	if !inited {
 		viewIDBytes := make([]byte, 8)
 		binary.LittleEndian.PutUint64(viewIDBytes, viewID)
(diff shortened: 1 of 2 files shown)


sophoah:
Just add on, afaik, crosslink are today only used to calculate the reward to be distributed every 64 blocks, hence discarding those pending crosslink should be safe and shouldn't affect the protocol

sophoah (approved):

ONECasey (approved):

sophoah (approved):

Frozen (approved):

sophoah (approved):

PR #4630, opened 2024-02-12
fix crosslink processing, add logs for crosslinks

GheisMohammadi:
## Issue
The leader is adding pending crosslinks without verifying them. This could break the consensus if other validators don't have the old shard state needed to verify crosslinks. This PR addresses that issue by ignoring very old crosslinks and not adding them to proposing blocks. crosslinks are today used only to calculate the reward to be distributed every 64 blocks, hence discarding those old pending crosslinks should be safe and shouldn't affect the protocol.


sophoah (approved):

ONECasey (approved):

adsorptionenthalpy (approved):

sophoah:
closing this one as not required

PR #4645, opened 2024-03-20
Signers count metric.

Frozen:
Signers count metric for prometheus.

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -131,7 +131,7 @@ func (consensus *Consensus) onPrepare(recvMsg *FBFTMessage) {
 	signerCount := consensus.decider.SignersCount(quorum.Prepare)
 	//// Read - End
 
-	consensus.UpdateLeaderMetrics(float64(signerCount), float64(consensus.getBlockNum()))
+	consensus.SignersCountMetrics(float64(signerCount), float64(consensus.getBlockNum()))
 
 	// Check BLS signature for the multi-sig
 	prepareSig := recvMsg.Payload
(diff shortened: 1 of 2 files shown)


ONECasey (approved):

consensus/leader.go
```diff
@@ -131,7 +131,7 @@ func (consensus *Consensus) onPrepare(recvMsg *FBFTMessage) {
 	signerCount := consensus.decider.SignersCount(quorum.Prepare)
 	//// Read - End
 
-	consensus.UpdateLeaderMetrics(float64(signerCount), float64(consensus.getBlockNum()))
```
sophoah, line 134: for the leader lets keep UpdateLeaderMetrics

consensus/metrics.go
```diff
@@ -149,6 +149,12 @@ func (consensus *Consensus) UpdateLeaderMetrics(numCommits float64, blockNum flo
 	consensusCounterVec.With(prometheus.Labels{"consensus": "num_commits"}).Add(numCommits)
 	consensusGaugeVec.With(prometheus.Labels{"consensus": "num_commits"}).Set(numCommits)
 }
+
+func (consensus *Consensus) SignersCountMetrics(signersCount float64, blockNum float64) {
+	consensusCounterVec.With(prometheus.Labels{"consensus": "signers_count"}).Add(signersCount)
+	consensusGaugeVec.With(prometheus.Labels{"consensus": "signers_count"}).Set(signersCount)
+}
```
sophoah, line 156: i don't think we need SignersCountMetrics since we already have `UpdateValidatorMetrics`

Frozen:
docker run -d -p 9091:9091 prom/pushgateway
if start node with flags "--prometheus.push=true" "--prometheus.pushgateway=http://127.0.0.1:9091"
you can get data at http://127.0.0.1
image

sophoah:
i don't think we need that PR anymore, i will close it for now since we now have the signing power broadcasted in the network

Issue #4632, opened 2024-02-15
pending undelegation not release with inactive validator


sophoah:
**Describe the bug**
https://staking.harmony.one/validators/mainnet/one1nef0c8kgy30delcckf87jaegdp7s7ntrknvajj has pending delegation that are not released at the end of epoch as it should

the list can be found with this curl command:
```
curl --location 'https://rpc.s0.t.hmny.io' --header 'Content-Type: application/json' --data '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "hmyv2_getDelegationsByValidator",
    "params": [
        "one1nef0c8kgy30delcckf87jaegdp7s7ntrknvajj"
    ]
}' | jq '.result | map(select(.Undelegations | length > 0))'
```

there are many pending delegation with old epoch (>7 epoch) and fund are not released

**To Reproduce**
NA

**Expected behavior**

When a validator is not online for more than 7 epoch, the delegator has initiated the staking undelegation, fund should be released at the end of the epoch undelegation + 7.

**Screenshots**

**Environment (please complete the following information):**
 - OS: [Linux, MacOS]
- Go environment [ ```go env``` ]

**Additional context**

validator has been inactive for a while now. see result of
```
hmy blockchain validator information one1nef0c8kgy30delcckf87jaegdp7s7ntrknvajj -n https://api.harmony.one
```

i checked a few others validators that are also not active but they don't have any pending undelegation not cleared:
ie : one1aagrkq9vg49jm5xx22fvm4leg0lt48qeupexvu, one1s269hx6ch3u6hxuag5sp5ggwc68vem6d8du6ut,one1kh9xzevupc60jyw3t0r7wgsgkml8kw8w32z3h3

sophoah:
fixed with v2024.1.0

Issue #4605, opened 2024-01-11
Buggy validator state


sophoah:
**Describe the bug**

A mainnet validator https://staking.harmony.one/validators/mainnet/one1dgn6cnhueyg0vqvq6g84hljwvl0nruwlt4huhy is currently elected and have `BINGO` in their logs. However there are inconsistencies in its validator information

```
curl -s -d '{"jsonrpc":"2.0", "method":"hmyv2_getValidatorInformation", "params":["one1dgn6cnhueyg0vqvq6g84hljwvl0nruwlt4huhy"], "id":1}' -H 'Content-Type:application/json' -X POST 'https://api.s0.t.hmny.io' | jq .result
```
```json
{
  "active-status": "active",
  "booted-status": "lost epos auction",
  "current-epoch-performance": null,
  "currently-in-committee": false,
  "epos-status": "eligible to be elected next epoch",
  "epos-winning-stake": null,
  "last-epoch-in-committee": 1732,
```

last-epoch-in-committee should be the current epoch here 1765

**To Reproduce**
1. Check out code with "sha1" commit
2. Build
3. run local test using "...." commands

**Expected behavior**

A normal output for an active validator should look as follow :
```json
  "active-status": "active",
  "booted-status": null,
  "current-epoch-performance": {
    "block": 52193081,
    "current-epoch-signing-percent": {
      "current-epoch-signed": 263700,
      "current-epoch-signing-percentage": "0.998258631132646881",
      "current-epoch-to-sign": 264160
    },
    "epoch": 1765
  },
  "currently-in-committee": true,
  "epos-status": "currently elected",
  ```
**Screenshots**

![doc_2024-01-11_17-12-15](https://github.com/harmony-one/harmony/assets/35721420/e726a76d-56d5-4d0f-b7cf-6d2e8f50ad89)

**Environment (please complete the following information):**
 - OS: [Linux, MacOS]
- Go environment [ ```go env``` ]

**Additional context**

sophoah:
fixed with v2024.1.0

PR #4647, opened 2024-03-22
fix validator state when trying to update the max-rate

diego1q2w:
During the previous hardfork we mad sure to update the max-rate to be at least the `min-rate` + `max-rate-change` to comply with the the minimum 7% rate imposed in HIP30, for validators running under that threshold. However, with that we introduced a subtle bug under the assumption that `max-rate-change` was a small value but in practice is not, it can go up to a 100% which will sum to a number higher than a 100% when adding the min-rate. 
For a handful of validators that caused the sanity checks fails when persisting the state in the hard-drive. A issue that is hard to catch as it'll only happen once per epoch, this bug explains issues like the buggy validator or the validator with undelegated funds issues.
As this is a change in the way the data is stored in the state this will have to go with a hardfork, I guess we can inlcude it with the leader rotation.

### DIFF
--- a/core/state/statedb.go
+++ b/core/state/statedb.go
@@ -895,7 +895,10 @@ func (db *DB) Finalise(deleteEmptyObjects bool) {
 	// TODO: remove validator cache after commit
 	for addr, wrapper := range db.stateValidators {
 		if err := db.UpdateValidatorWrapper(addr, wrapper); err != nil {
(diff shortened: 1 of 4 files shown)


ONECasey (approved):

sophoah:
hey @diego1q2w PR looks good. Were you able to test in localnet whether all is good and whether that really fix #4632 and #4605

ONECasey:
Theres a few conflicts @diego1q2w

PR #4653, opened 2024-04-03
Additional logs for stagesync.

Frozen:
1. Additional logs
2. Refactored db initializing
3. Removed deadcode

### DIFF
--- a/api/service/stagedstreamsync/downloader.go
+++ b/api/service/stagedstreamsync/downloader.go
@@ -68,9 +68,17 @@ func NewDownloader(host p2p.Host, bc core.BlockChain, consensus *consensus.Conse
 
 	ctx, cancel := context.WithCancel(context.Background())
 
-	// create an instance of staged sync for the downloader
-	stagedSyncInstance, err := CreateStagedSync(ctx, bc, consensus, dbDir, isBeaconNode, sp, config, logger)
+	mainDB, dbs, err := InitDB(ctx, bc.ShardID(), config, isBeaconNode, dbDir)
 	if err != nil {
+		logger.Error().Err(err).Msg("failed to init db")
+		cancel()
+		return nil
+	}
+
+	// create an instance of staged sync for the downloader
+	stagedSyncInstance, err := CreateStagedSync(bc, consensus, mainDB, dbs, isBeaconNode, sp, config, logger)
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to create staged sync instance")
 		cancel()
 		return nil
 	}
--- a/api/service/stagedstreamsync/downloaders.go
+++ b/api/service/stagedstreamsync/downloaders.go
@@ -59,14 +59,6 @@ func (ds *Downloaders) Close() {
 	}
 }
 
-// DownloadAsync triggers a download
-func (ds *Downloaders) DownloadAsync(shardID uint32) {
-	d, ok := ds.ds[shardID]
-	if !ok && d != nil {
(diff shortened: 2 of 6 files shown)


api/service/stagedstreamsync/downloader.go
```diff
@@ -68,9 +68,17 @@ func NewDownloader(host p2p.Host, bc core.BlockChain, consensus *consensus.Conse
 
 	ctx, cancel := context.WithCancel(context.Background())
 
+	mainDB, dbs, err := InitDB(ctx, bc.ShardID(), config, isBeaconNode, dbDir)
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to init db")
+		cancel()
+		return nil
+	}
+
```
GheisMohammadi, line 77: How this helps? What's the reason to move staged sync db initiation to downloader?

api/service/stagedstreamsync/downloaders.go
```diff
@@ -59,14 +59,6 @@ func (ds *Downloaders) Close() {
 	}
 }
 
-// DownloadAsync triggers a download
-func (ds *Downloaders) DownloadAsync(shardID uint32) {
```
GheisMohammadi, line 63: Please keep this function. We may use it on state sync.

api/service/stagedstreamsync/stage_lastmile.go
```diff
@@ -25,9 +25,9 @@ func NewStageLastMile(cfg StageLastMileCfg) *StageLastMile {
 	}
 }
 
-func NewStageLastMileCfg(ctx context.Context, bc core.BlockChain, db kv.RwDB) StageLastMileCfg {
+func NewStageLastMileCfg(bc core.BlockChain, db kv.RwDB) StageLastMileCfg {
 	return StageLastMileCfg{
-		ctx: ctx,
+		ctx: context.TODO(),
```
GheisMohammadi, line 30: Touching context like this could possibly make same issues as we had for disk cache. Please take a look at pr #4450

Frozen: ```ctx``` is ```context.Background()```. No real changes here.

api/service/stagedstreamsync/downloaders.go
Frozen: You can add this function when you will use it. This function inherits interface, but does nothing.

api/service/stagedstreamsync/downloader.go
Frozen: No real need, but having db and struct initializations as different steps makes it more clear.

GheisMohammadi: Please keep it under staged stream sync. Downloader itself starts staged sync creation. There we initiate db.

api/service/stagedstreamsync/downloaders.go
GheisMohammadi: Please keep this function. I will remove it if it is not needed.

api/service/stagedstreamsync/stage_lastmile.go
GheisMohammadi: if no real changes, then why we should even do it? It would be good if you share how you test this part.

GheisMohammadi:
The title of this PR is "Additional logs ..." while it does a few unnecessary changes. It would be good to stick with goal and leave unnecessary changing in code structure for right time.

ONECasey (approved):

PR #4660, opened 2024-04-15
Improve staged stream sync

GheisMohammadi:
## Issue

This PR improves staged stream sync stages and refactors block storage.

### DIFF
--- a/api/service/stagedsync/syncing.go
+++ b/api/service/stagedsync/syncing.go
@@ -112,6 +112,24 @@ func CreateStagedSync(
 		finishCfg,
 	)
 
+	utils.Logger().Info().
+		Str("ip", ip).
+		Str("port", port).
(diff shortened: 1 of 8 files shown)


adsorptionenthalpy (approved):

api/service/stagedsync/syncing.go
```diff
@@ -112,6 +112,24 @@ func CreateStagedSync(
 		finishCfg,
 	)
 
+	utils.Logger().Info().
```
sophoah, line 115: put that into debug instead ? seems we might have a lot of those logs

GheisMohammadi: It is supposed to be only once after downloader creation.

sophoah (approved):

PR #4661, opened 2024-04-21
Force sync when view change and received message with higher height.

Frozen:
Force sync when view change and received message with higher height.


ONECasey (approved):

sophoah:
would that also exit the current view change process ?

Frozen:
`consensus.onAnnounceSanityChecks` called from `onAnnounce`, so it can't interrupt view change.

sophoah (approved):

PR #4669, opened 2024-05-15
Improve crosslink on devnet

Frozen:
Improved crosslink signing

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -510,6 +510,15 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 
 	go listenOSSigAndShutDown(currentNode)
 
+	if hc.Network.NetworkType == nodeconfig.Partner { // devnet
+		go func() {
+			for {
+				<-time.After(10 * time.Second)
+				go currentNode.BroadcastCrossLinkFromShardsToBeacon()
+			}
+		}()
+	}
+
 	if !hc.General.IsOffline {
 		if err := myHost.Start(); err != nil {
 			utils.Logger().Fatal().
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -510,7 +510,7 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 
 	go listenOSSigAndShutDown(currentNode)
 
-	if hc.Network.NetworkType == nodeconfig.Partner { // devnet
+	if hc.Network.NetworkType == nodeconfig.Partner || hc.Network.NetworkType == nodeconfig.Testnet { // devnet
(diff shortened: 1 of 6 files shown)


cmd/harmony/main.go
```diff
@@ -510,6 +510,15 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 
 	go listenOSSigAndShutDown(currentNode)
 
+	if hc.Network.NetworkType == nodeconfig.Partner { // devnet
+		go func() {
+			for {
+				<-time.After(10 * time.Second)
+				go currentNode.BroadcastCrossLinkFromShardsToBeacon()
```
sophoah, line 517: why do we need this here ?

Frozen: `BroadcastCrossLinkFromShardsToBeacon` happens only on a new block insertion. There is infinity view change, no new blocks.

sophoah: It's not always the case that we are in an infinite view change. Under normal conditions (i.e., no view change), could there be any race conditions here, or what happens if we broadcasting a crosslink twice?

Question is why would a shard need to broadcast a crosslink when it is not producing block?

node/node_handler.go
```diff
@@ -277,6 +304,27 @@ func (node *Node) BroadcastCrosslinkHeartbeatSignalFromBeaconToShards() { // lea
...
+	}
+	return d
+}
+
+func getPrivToSign(privs multibls.PrivateKeys, comm *shard.Committee) *bls.PrivateKeyWrapper {
+	var privToSing *bls.PrivateKeyWrapper
+	q := createVoters(comm)
+	for _, priv := range privs {
+		_, found := q[priv.Pub.Bytes]
+		if found { // if key is in committee
```
sophoah, line 320: depending on the use case of getPrivToSign, there might be instances where we need the leader key instead of any keys currently elected in the committee

cmd/harmony/main.go
Frozen: > what happens if we broadcasting a crosslink twice

crosslink broadcasted by leader and 1% of validators. It's duplicated by design. 

> why would a shard need to broadcast a crosslink when it is not producing block

Because shard 0 broadcasts struct `CrosslinkHeartbeat` with field `LatestContinuousBlockNum`. It knows about last blocks received by crosslink, but crosslink should be sent first. Actually, that is the second issue, where we can't verify `CrosslinkHeartbeat` message, because no key in committee, which is solved [here](https://github.com/harmony-one/harmony/pull/4669)

so, there is circular dependency, no heartbeat because no crosslinks, but no crosslinks, because no heartbeat.

ONECasey (approved):

PR #4671, opened 2024-05-16
Removed threshold 10 epochs

Frozen:
Removes threshold for 10 blocks for devnet crosslinks fix.

### DIFF
--- a/node/node_newblock.go
+++ b/node/node_newblock.go
@@ -1,7 +1,6 @@
 package node
 
 import (
-	"math/big"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


sophoah:
Hey @Frozen, instead of removing the crosslink outright, can we first check whether it has already been processed? If it has, then we can proceed with the removal. This particular piece of code was a legitimate removal due to a previous mainnet issue where old pending crosslinks, which had already been processed, were stuck in the database.

Frozen:
> Hey @Frozen, instead of removing the crosslink outright, can we first check whether it has already been processed? If it has, then we can proceed with the removal. This particular piece of code was a legitimate removal due to a previous mainnet issue where old pending crosslinks, which had already been processed, were stuck in the database.

As i can understand the next line of code does what you expect [link](https://github.com/harmony-one/harmony/pull/4671/files#diff-a7c611f40d3b2c6baa2977b87bce2219962e836042cb3f96175b18259609ff7cR238)

ONECasey (approved):

GheisMohammadi (approved):

PR #4673, opened 2024-05-21
Crosslink heartbeat verify through current committee

Frozen:
Crosslink heartbeat verify through current committee

### DIFF
--- a/node/node_cross_link.go
+++ b/node/node_cross_link.go
@@ -1,8 +1,6 @@
 package node
 
 import (
-	"math/big"
-
 	common2 "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
 	ffi_bls "github.com/harmony-one/bls/ffi/go/bls"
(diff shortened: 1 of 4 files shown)


Frozen:
https://github.com/harmony-one/harmony/pull/4673/files#diff-53cafc20d348cad861dbaad811b8e24eed9320f4382536da667b166aa641eb6eR71

ONECasey (approved):

sophoah:
PR Explanation:
In current dev shard 0 node are signing the crosslink heartbeat with the current leader private key, however upon processing the heartbeat shard 1 is verifying the public key by using the epoch number in the heartbeat hb.epoch(). In normal situation it usually works fine since hb.epoch() and cur.epoch() will be the same or more or less 1 epoch difference. However, in cases where the crosslink wasn't processed for a long time, the leader in hb.epoch() may not have been elected. It will be especially true in the future with external leader rotation when external validator can be kick out from committee due to low signature rate.

This PR is takes the assumption that the heartbeat has to be processed as soon as they are sent, and making sure it is sent by a valid/elected validator should be enough for now

node/node_cross_link.go
```diff
@@ -103,7 +103,12 @@ func (node *Node) processCrossLinkHeartbeatMessage(msgPayload []byte) error {
 func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
 	if node.IsRunningBeaconChain() {
 		pendingCLs, err := node.Blockchain().ReadPendingCrossLinks()
-		if err == nil && len(pendingCLs) >= maxPendingCrossLinkSize {
+		if err != nil {
+			utils.Logger().Debug().
```
GheisMohammadi, line 107: it would be good to have the error in logs as well
```golang
utils.Logger().Debug().Err(err).
```

sophoah:
@Frozen can you review the conflict and @GheisMohammadi comment

node/node_cross_link.go
Frozen: Added log message.

GheisMohammadi (approved):

PR #4674, opened 2024-05-21
[HOTFIX] fix boot node instability issue, refactor p2p host configurations and add new features

GheisMohammadi:
## Issue

After upgrading the boot nodes to the latest version of the main branch, which includes several changes compared to the previous boot node code, the boot node service process began restarting repeatedly. This was accompanied by numerous error messages related to the QUIC protocol, noise, sessions, muxers, and various other lower layer P2P issues.

This PR addresses these issues by making the following changes:
- Removes QUIC, an unnecessary feature, and switches to TCP transport.
- Makes TLS/Noise, relays, and NAT optional features that can be disabled via configuration.
- Adds a dial timeout for the connection manager.
- Allows the use of the MPlex muxer in addition to Yamux.

Additionally, this PR introduces several new configurations and flags, detailed below:
	**NoTransportSecurity:** to disable p2p security (tls and noise)
	**NAT:** enable p2p NAT. NAT Manager takes care of setting NAT port mappings, and discovering external addresses
	**UserAgent:** custom user agent; explicitly set the user-agent, so we can differentiate from other Go libp2p users
	**DialTimeout:** p2p dial timeout 
	**MuxerType:** P2P multiplexer type (default:Mplex, 0:Mplex, 1:Yamux)
	**NoRelay:** No relay services, direct connections between peers only


ONECasey (approved):

sophoah:
hey @GheisMohammadi those new settings are advanced p2p configuration, can you please hide them in the config file ?

GheisMohammadi:
> hey @GheisMohammadi those new settings are advanced p2p configuration, can you please hide them in the config file ?

I believe they are needed to be customized by user. They can be adjusted based on the environment
**NoTransportSecurity**:  User can decide to have TLS/Noise security. Sometimes that make instability, so user can disable it
**NAT**:   This is useful to differentiate between boot node and normal validator node. Also for those servers running behind router                             	
**UserAgent**:   A custom agent name which each user can set to differentiate from other libp2p users
**DialTimeout**:   A simple timeout, depends on how long make sense to wait for timeout
**MuxerType**:  useful for boot node and the debugging and instability issues.
**NoRelay**:  useful for boot node host and nodes can only talk directly to each other not tthrough the relays

adsorptionenthalpy (approved):

sophoah (approved):

PR #4675, opened 2024-05-22
Fix for crosslink snap db.

Frozen:
Fix for crosslink snap db.

### DIFF
--- a/node/node_newblock.go
+++ b/node/node_newblock.go
@@ -248,6 +248,16 @@ func (node *Node) ProposeNewBlock(commitSigs chan []byte) (*types.Block, error) 
 						AnErr("[ProposeNewBlock] pending crosslink is already committed onchain", err)
 					continue
 				}
+				last, err := node.Blockchain().ReadShardLastCrossLink(pending.ShardID())
+				if err != nil {
+					utils.Logger().Debug().
+						AnErr("[ProposeNewBlock] failed to read last crosslink", err)
+					// no return
+				}
+				// if pending crosslink is older than the last crosslink, delete it and continue
+				if exist == nil && last != nil && last.BlockNum() >= pending.BlockNum() {
+					invalidToDelete = append(invalidToDelete, pending)
+				}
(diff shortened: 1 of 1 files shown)


sophoah:
Here the reasoning for that PR :

Crosslink before an epoch or blocks that have been pruned will set `exist` to `nil` in
https://github.com/harmony-one/harmony/blob/main/node/node_newblock.go#L243 
```golang
exist, err := node.Blockchain().ReadCrossLink(pending.ShardID(), pending.BlockNum())
```
This is what happened last time in mainnet when we tried to add a crosslink for a block/epoch we didn't have. Our solution back then was to remove those pending crosslink because we knew they were already processed.

in theory, crosslink already added on s0 chain are all continuous, so we can create a function that check the lastcrosslink (added on chain) and delete the pending crosslink when the block number in the shard to be processed is older.

sophoah (approved):

node/node_newblock.go
```diff
@@ -248,6 +248,16 @@ func (node *Node) ProposeNewBlock(commitSigs chan []byte) (*types.Block, error)
...
 					continue
 				}
+				last, err := node.Blockchain().ReadShardLastCrossLink(pending.ShardID())
+				if err != nil {
+					utils.Logger().Debug().
+						AnErr("[ProposeNewBlock] failed to read last crosslink", err)
+					// no return
+				}
+				// if pending crosslink is older than the last crosslink, delete it and continue
+				if exist == nil && last != nil && last.BlockNum() >= pending.BlockNum() {
```
GheisMohammadi, line 258: there is no return even if ReadShardLastCrossLink fails. So, along with checking last!=nil, it would make more sense to check error as well
```golang
if err==nil  && exist == nil && last != nil && last.BlockNum() >= pending.BlockNum() {
```

Frozen: Fixed

GheisMohammadi (approved):

PR #4698, opened 2024-06-21
Block proposing moved from Node to Consensus.

Frozen:
1. Block proposing uses 99% consensus and 1% node code, but it is located within the node.
2. The chained protocol requires block proposing from consensus, causing a circular dependency error.


adsorptionenthalpy (approved):

internal/utils/utils.go
```diff
@@ -235,11 +235,18 @@ func IsPrivateIP(ip net.IP) bool {
...
+	shardID  uint32
+	blockNum uint64
+}
+
 // GetPendingCXKey creates pending CXReceiptsProof key given shardID and blockNum
 // it is to avoid adding duplicated CXReceiptsProof from the same source shard
-func GetPendingCXKey(shardID uint32, blockNum uint64) string {
-	key := strconv.FormatUint(uint64(shardID), 10) + "-" + strconv.FormatUint(blockNum, 10)
-	return key
+func GetPendingCXKey(shardID uint32, blockNum uint64) CXKey {
```
sophoah, line 245: can the difference of key format returned cause any issue in our codebase calling that function ?

Frozen: Cache keeped only in the memory, so, restart will erase everything. 
Exactly in this case, we have 3 memory allocations for no reason. My version has no allocations.

ONECasey (approved):

PR #4706, opened 2024-07-04
Fixed PeerStore usage.

Frozen:
PeerStore is storage for all known peers.
`host.Network().Peers()` is connected peer.

### DIFF
--- a/node/node_handler.go
+++ b/node/node_handler.go
@@ -421,16 +421,17 @@ func (node *Node) BootstrapConsensus() error {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 	min := node.Consensus.MinPeers
-	enoughMinPeers := make(chan struct{})
+	enoughMinPeers := make(chan struct{}, 1)
 	const checkEvery = 3 * time.Second
 	go func() {
 		for {
 			<-time.After(checkEvery)
 			numPeersNow := node.host.GetPeerCount()
-			if numPeersNow >= min {
+			connectedPeers := len(node.host.Network().Peers())
+			if connectedPeers >= min {
 				utils.Logger().Info().Msg("[bootstrap] StartConsensus")
 				enoughMinPeers <- struct{}{}
-				fmt.Println("Bootstrap consensus done.", numPeersNow, " peers are connected")
+				fmt.Printf("Bootstrap consensus done. Connected %d, known %d, shard: %d\n", connectedPeers, numPeersNow, node.Consensus.ShardID)
 				return
 			}
 			utils.Logger().Info().
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -57,6 +57,7 @@ type Host interface {
 	GetID() libp2p_peer.ID
 	GetP2PHost() libp2p_host.Host
 	GetDiscovery() discovery.Discovery
+	Network() libp2p_network.Network
 	GetPeerCount() int
 	ConnectHostPeer(Peer) error
 	// AddStreamProtocol add the given protocol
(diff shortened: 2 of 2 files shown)


p2p/host.go
```diff
@@ -443,17 +444,9 @@ func (host *HostV2) Close() error {
...
-	peers := host.h.Peerstore().Peers()
-	for _, peer := range peers {
-		result := host.h.Network().Connectedness(peer)
-		if result == libp2p_network.Connected {
-			connected++
-		} else if result == libp2p_network.NotConnected {
-			not++
-		}
-	}
-	return len(peers), connected, not
```
GheisMohammadi, line 456: For me old code looks safer approach. For each peer it checks Connectedness. If you are sure about 
```golang
connected := len(host.h.Network().Peers())
```
Then no issue.

node/node_handler.go
```diff
@@ -421,16 +421,17 @@ func (node *Node) BootstrapConsensus() error {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
 	defer cancel()
 	min := node.Consensus.MinPeers
-	enoughMinPeers := make(chan struct{})
+	enoughMinPeers := make(chan struct{}, 1)
```
GheisMohammadi, line 424: I am not sure if this change is necessary. The unbuffered channel ensures the consensus process only starts when the signal is received. For me unbuffered is safer in terms of synchronization. are you concerned about blocking here? Is there any other goroutine that this change taking care of?

p2p/host.go
Frozen: We have about 16000 knows peers, so it will make 16k checks. Only concern here, that it works as the previous one.

node/node_handler.go
Frozen: > The unbuffered channel ensures the consensus process only starts when the signal is received
How does the consensus start depends on the channel buffering?

Frozen: > are you concerned about blocking here

it pushes message to the channel and frees goroutine, because it doesn't need to wait, because the next line is `return`

Frozen: > Is there any other goroutine that this change taking care of?

no, only this one

GheisMohammadi: > How does the consensus start depends on the channel buffering?

The go routine counts the peers and once it reaches to the minimum required peers for consensus, it pushes message to channel.

The select part is waiting for the ```enoughMinPeers``` signal. **It blocks here** to receive the signal and once it receives enough peers, it calls ```node.Consensus.StartChannel()``` in a go-routine, otherwise it return ```ctx.Err()```. How buffered channel helps? 

My concern is we should be very careful about the logic behind a code that for several years is working.

> because it doesn't need to wait, because the next line is return

what you mean by this? It returns nil, but for sure it has received enough peers on this point. So, triggering start consensus itself doesn't need to wait.

p2p/host.go
GheisMohammadi: It seems this function is only used for public API (```GetNodeMetadata```). So, if you have properly tested the output, it should be fine.

node/node_handler.go
Frozen: > It returns nil

goroutine returns nothing

Frozen: > How buffered channel helps?

it frees goroutine. It does not affect the consensus in any way.

Frozen: > My concern is we should be very careful about the logic behind a code that for several years is working.

What can happen? It just bad designed code, which was written by someone who has no idea how does it work.

Frozen: > It blocks here

it blocks here independently channel buffered or no

Frozen:
image

Frozen:
image
this is for dev branch,

node/node_handler.go
GheisMohammadi: Its good the core logic of the function remains unchanged. The improvement here is that a buffered channel may help the sending go-routine to free up resources more promptly.
However, I have a few points I'd like to share with you:
 The fact that this code has been working for years shows the previous developer knew what they were doing. It's essential to maintain a respectful tone when discussing others' work. Mentioning original developer had "no idea how it works" can be demoralizing. Let's keep our feedback constructive to maintain a good team vibe.

GheisMohammadi (approved):

sophoah (approved):

PR #4722, opened 2024-07-29
Go1.22

Frozen:
Golang version 1.22


ONECasey (approved):

cmd/bootnode/main.go
```diff
@@ -108,7 +108,7 @@ func main() {
 	logConn := flag.Bool("log_conn", false, "log incoming/outgoing connections")
 	maxConnPerIP := flag.Int("max_conn_per_ip", 10, "max connections number for same ip")
 	forceReachabilityPublic := flag.Bool("force_public", false, "forcing the local node to believe it is reachable externally")
-	noTransportSecurity := flag.Bool("no_transport_security", false, "disable TLS encrypted transport")
+	noTransportSecurity := flag.Bool("no_transport_security", true, "disable TLS encrypted transport")
```
GheisMohammadi, line 111: It seems it disables transport security without proper discussion and a finalized decision from the team. Have we checked the compatibility of nodes with disabled security with the current mainnet nodes? We have to ensure they can communicate properly to avoid any network issues. Also it would be good to provide us with a few references that shows this change helps.

GheisMohammadi:
Thanks for the PR. It is a great work. There are a few points here to mention:

- The PR is quite extensive, with over thousands lines of changes, but the description is only three words long. I see a lot more changes rather than updating go version. Could you please provide a more detailed description? This will help us understand the purpose and scope of the changes better.

- It looks like you’re bringing in over 5000 lines of code from the `harmony-one/harmony-test` repository and placing them in a new folder called `tests`. We already have a `test` folder, and having both `test` and `tests` could be confusing. Additionally, duplicating code across repositories might make it very hard to maintain consistency. Could you explain the reasoning behind this approach? Perhaps we could consider using Docker to clone the `harmony-test` repository and use it directly, which might simplify maintenance.

- The commit titles are quite brief and a bit unclear (e.g., "test", "tests", "1.21.11"). Could you please provide more descriptive commit messages? This would greatly help in understanding the changes at a glance and maintaining the project history.

- There’s a commit titled "Fixed trailing zeroes" which seems to actually update dependencies instead. Clear and accurate commit messages are important to avoid confusion, so it would be helpful to correct this.

- I see that the tests for this PR are still ongoing. It might be best to wait until all tests are complete and verified to ensure there are no unexpected issues. We have to have this PR on test-net for awhile to ensure consistency and stability.

cmd/bootnode/main.go
mur-me: Small inputs:
* Default value for the release 2024.1.0 is `"NoTransportSecurity": false`, thus `"NoTransportSecurity": true` cases is out of scope
* If mixing all possible combinations together we will end up in 16(!) possible cases on the communication between bootnode and Harmony node, but if we remove out of scope cases the number is decreasing to 9.
* the next step will be to try enable security via  `"NoTransportSecurity": false` on the bootnode 1.22 and test 3 cases

  | Bootnode1.22 Node "NoTransportSecurity": true | Bootnode1.22 Node "NoTransportSecurity": false | Bootnode1.19 Node "NoTransportSecurity": true | Bootnode1.19 Node "NoTransportSecurity": false
-- | -- | -- | -- | --
Harmony 1.19 Node "NoTransportSecurity": true |  Out of scope  | Out of scope  | Out of scope  | Out of scope 
Harmony 1.19 Node "NoTransportSecurity": false | **OK** | NOT tested  | Out of scope | **OK**
Harmony 1.22 Node "NoTransportSecurity": true | **FAILED** | NOT tested  | Out of scope  | **FAILED**
Harmony 1.22 Node "NoTransportSecurity": false | **OK** | NOT tested  | Out of scope | **OK**

GheisMohammadi: Great tests, thanks

Frozen:
check this one https://github.com/harmony-one/harmony/pull/4736

cmd/bootnode/main.go
sophoah: > the same test on the devnet have given me inverted results:

just check two devnet normal node and they had  `"NoTransportSecurity": true,`

based on test above:
1) NoTransportSecurity is not playing a role for a normal node to get the list of other node in the network (from the bootnode).
2) the non bootnode node in the network need to match their  settings `NoTransportSecurity`

mur-me: Results of canary tests all nodes have `NoTransportSecurity: false`:
* RPC node 1.22 + Bootnode 1.22 ✅
* RPC node 1.22 + Bootnode 1.19 ✅
* RPC 1.19 + Bootnode 1.22 ✅
* RPC node 1.22(95.217.207.240 aka mhe-explorers1-02) + traffic  ✅
* s1 validator 1.22  ✅
* s0 validator 1.22  ✅

PR #4731, opened 2024-08-09
Fix Check for Existence of `allowedtxs.txt` File Before Applying Logic

GheisMohammadi:
## Issue

This PR resolves the issue where an error occurs if the `allowedtxs.txt` file is not present in the `.hmy` directory. Previously, the absence of this file would result in an error:
```
AllowedTxs setup error: open ./.hmy/allowedtxs.txt: no such file or directory
```
The fix involves adding a check to verify the existence of the `allowedtxs.txt` file before proceeding with the setup logic.

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -1112,9 +1112,11 @@ func setupAllowedTxs(hc harmonyconfig.HarmonyConfig) (map[ethCommon.Address][]co
 		return parseAllowedTxs(data)
 	} else if errors.Is(err, os.ErrNotExist) {
 		// file path does not exist
+		utils.Logger().Debug().Str("AllowedTxsFile", hc.TxPool.AllowedTxsFile).Msgf("AllowedTxs file doesn't exist")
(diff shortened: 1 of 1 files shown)


Frozen (approved):

cmd/harmony/main.go
```diff
@@ -1101,12 +1101,22 @@ func parseAllowedTxs(data []byte) (map[ethCommon.Address][]core.AllowedTxData, e
...
+		data, err := os.ReadFile(hc.TxPool.AllowedTxsFile)
+		if err != nil {
+			return nil, err
+		}
+		return parseAllowedTxs(data)
+	} else if errors.Is(err, os.ErrNotExist) {
+		// file path does not exist
+		return make(map[ethCommon.Address][]core.AllowedTxData), nil
+	} else {
+		// some other errors happened
```
sophoah, line 1117: add a debug logs here so we know some other error happened

GheisMohammadi: added

sophoah (approved):

PR #4739, opened 2024-08-19
Lock free consensus methods

Frozen:
1. Moved `PostConsensusProcessing` from node to consensus, because method uses consensus methods, but not node. 
2. Consensus uses `validateShardBoundMessage` indirectly, but `validateShardBoundMessage` uses consensus internally which causes deadlock. Changed method called by `validateShardBoundMessage` to be lock free.
3. Synchronious calls to `SendMessageToGroups` because it uses `validateShardBoundMessage` internally, but now it save to use without deadlocks.


cmd/harmony/main.go
```diff
@@ -884,7 +884,6 @@ func setupConsensusAndNode(hc harmonyconfig.HarmonyConfig, nodeConfig *nodeconfi
 		Msg("Init Blockchain")
 
 	currentNode.Consensus.Registry().SetNodeConfig(currentNode.NodeConfig)
-	currentConsensus.PostConsensusJob = currentNode.PostConsensusProcessing
```
sophoah, line 887: could you explain here why removing this line is safe ?

sophoah (approved):

mur-me (approved):

PR #4751, opened 2024-09-05
Fix for crosslink processing.

Frozen:
Fix for crosslink processing.

### DIFF
--- a/node/node_cross_link.go
+++ b/node/node_cross_link.go
@@ -103,13 +103,7 @@ func (node *Node) processCrossLinkHeartbeatMessage(msgPayload []byte) error {
 func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
 	if node.IsRunningBeaconChain() {
 		pendingCLs, err := node.Blockchain().ReadPendingCrossLinks()
-		if err != nil {
-			utils.Logger().Debug().
-				Err(err).
-				Msgf("[ProcessingCrossLink] Pending Crosslink reach maximum size: %d", len(pendingCLs))
-			return
-		}
-		if len(pendingCLs) >= maxPendingCrossLinkSize {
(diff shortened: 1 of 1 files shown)


node/node_cross_link.go
```diff
@@ -103,13 +103,7 @@ func (node *Node) processCrossLinkHeartbeatMessage(msgPayload []byte) error {
...
 	if node.IsRunningBeaconChain() {
 		pendingCLs, err := node.Blockchain().ReadPendingCrossLinks()
-		if err != nil {
-			utils.Logger().Debug().
-				Err(err).
-				Msgf("[ProcessingCrossLink] Pending Crosslink reach maximum size: %d", len(pendingCLs))
-			return
-		}
-		if len(pendingCLs) >= maxPendingCrossLinkSize {
+		if err == nil && len(pendingCLs) >= maxPendingCrossLinkSize {
```
GheisMohammadi, line 106: if err!=nil ?

sophoah: In the original code (see [here](https://github.com/harmony-one/harmony/blob/0b342d769e0223e0aad7453dc636209a8bb5ab50/node/node_cross_link.go#L106)), we also didn't handle the case where `err != nil`. @Frozen made an attempt to address this in PR #4673 (see [here](https://github.com/harmony-one/harmony/blob/a7560fed6ca312ab6b45bc3f47939a5d2515806d/node/node_cross_link.go#L106)), which was included in the v2024.2.0 release. However, we hadn't tested this with a fresh network using external validators until now.

@Frozen @GheisMohammadi, the error we're encountering is `level DB: not found`. Since this error only arises when the key in LevelDB doesn't exist, should we consider updating our code to simply continue execution in this case, and return if any other error is encountered?

sophoah:
hey @Frozen let's use #4754 since both have the same objective and 4754 is more complete. I will close this one for now.

PR #4755, opened 2024-09-11
feat(Makefile) - cover verbose mode for the local debug

mur-me:
## Issue

Our devs need a convenient way to run `debug` and `debug-ext` in the verbose mode, so I've created this PR to handle this.
VERBOSE variable can be enabled by:
1. uncomment in the Makefile
2. `export VERBOSE=true` in the current bash session
3. as a variable before the make run like `VERBOSE=true make debug` 

* Additionally, I've added explicit log saying that you are in verbose mode

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

5. **Describe the migration plan.**. For each flag epoch, describe what changes take place at the flag epoch, the anticipated interactions between upgraded/non-upgraded nodes, and any special operational considerations for the migration.

6. **Describe how the plan was tested.**

7. **How much minimum baking period after the last flag epoch should we allow on Pangaea before promotion onto mainnet?**

8. **What are the planned flag epoch numbers and their ETAs on Pangaea?**

9. **What are the planned flag epoch numbers and their ETAs on mainnet?**

10. **What should node operators know about this planned change?**

11. **Does this PR introduce backward-incompatible changes *NOT* related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)

    **YES|NO**

12. **Does the existing `node.sh` continue to work with this change?**

13. **What should node operators know about this change?**

14. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,8 @@ debug:
 	# uncomment the following lines to enable debug logging for libp2p, it produces a lot of logs, so disabled by default
 	#export GOLOG_LOG_LEVEL=debug
 	#export GOLOG_OUTPUT=stdout
+	# uncomment the following line to enable max logging
+	# export VERBOSE=true
 	bash ./test/debug.sh
 
 debug-kill:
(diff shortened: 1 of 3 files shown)


GheisMohammadi:
I feel like uncommenting is not much user friendly. Maybe If we can pass the verbose as a parameter would be more convenient, something like:

make  debug -v
make debug -verbose
make verbose=true debug

GheisMohammadi (approved):

PR #4757, opened 2024-09-12
Not Checking Pending Crosslinks for First ten Epoch

GheisMohammadi:
## Issue
The current code attempts to delete pending crosslinks that are older than 10 epochs. However, if the pending crosslinks were not added to the database within the first 10 epochs, this can lead to an `leveldb: not found`. This PR resolves the issue by adding a condition to skip checking pending crosslinks during the first 10 epochs, ensuring we avoid unnecessary database lookups and potential errors.

### DIFF
--- a/node/harmony/node.go
+++ b/node/harmony/node.go
@@ -1108,8 +1108,9 @@ func New(
 
 	// delete old pending crosslinks
 	ten := big.NewInt(10)
-	if node.Blockchain().ShardID() == shard.BeaconChainShardID && node.Blockchain().CurrentBlock().Epoch().Cmp(ten) > 0 {
-		crossLinkEpochThreshold := new(big.Int).Sub(node.Blockchain().CurrentHeader().Epoch(), ten)
+	epoch := node.Blockchain().CurrentHeader().Epoch()
+	if node.Blockchain().ShardID() == shard.BeaconChainShardID && node.Blockchain().Config().IsCrossLink(epoch) {
+		crossLinkEpochThreshold := new(big.Int).Sub(epoch, ten)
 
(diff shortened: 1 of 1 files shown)


node/harmony/node.go
```diff
@@ -1107,8 +1107,8 @@ func New(
 	node.serviceManager = service.NewManager()
 
 	// delete old pending crosslinks
-	if node.Blockchain().ShardID() == shard.BeaconChainShardID {
-		ten := big.NewInt(10)
+	ten := big.NewInt(10)
+	if node.Blockchain().ShardID() == shard.BeaconChainShardID && node.Blockchain().CurrentBlock().Epoch().Cmp(ten) > 0 {
```
sophoah, line 1111: in localnet the crosslink hardfork epoch happens before 10th epoch, you can use `func (c *ChainConfig) IsCrossLink(epoch *big.Int) bool` to improve your code

Frozen: image
no, it's not. What means 10 then?

sophoah: it is [2 for localnet](https://github.com/harmony-one/harmony/blob/main/internal/params/config.go#L281), and [186 for mainnet](https://github.com/harmony-one/harmony/blob/main/internal/params/config.go#L44)

GheisMohammadi: done

sophoah (approved):

sophoah:
@Frozen please review / approve. Thanks

node/harmony/node.go
Frozen: 10 is magic number, you should use `IsCrossLink ` method instead.

Frozen (approved):

Issue #4749, opened 2024-09-03
Stream sync: context deadline exceeded


Frozen:
**Describe the bug**
`{"level":"error","port":"9000","ip":"127.0.0.1","module":"staged` stream sync","ShardID":0,"error":"context deadline exceeded","streamID":"QmSAaTjhHWNF42A4xWLkmcwnmEJhatjbqMpv84HBxeNzkt","caller":"/Users/frozen/go/src/github.com/harmony-one/harmony/api/service/stagedstreamsync/syncing.go:520","time":"2024-09-03T13:45:10.258677-04:00","message":"[STAGED_STREAM_SYNC]: getCurrentNumber request `failed"}`

**To Reproduce**
1. Check out code with dev branch
2. Build
3. run `make debug`

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [Linux, MacOS]
- Go environment [ ```go env``` ]

**Additional context**

Frozen:
`{"level":"warn","port":"9002","ip":"127.0.0.1","mode":"epoch chain short range","error":"stream removed when doing request","stream":"QmTXwseacFvVFeHfSBSwL3xdAHMyTeesNUu2DJaGuYWZt5","caller":"/Users/frozen/go/src/github.com/harmony-one/harmony/api/service/stagedstreamsync/short_range_helper.go:185","time":"2024-09-03T22:43:13.597391-04:00","message":"[STAGED_STREAM_SYNC]: failed to doGetBlockHashesRequest"}`

GheisMohammadi:
The removal of the P2P stream is expected in this case, as the getCurrentNumber request failed multiple times. When the sync process is unable to retrieve data from the stream after several attempts, the stream is replaced. However, the key issue lies in understanding why the getCurrentNumber request is failing in the first place. This needs further investigation to determine the root cause behind the failed request.

GheisMohammadi:
The PR #4762 tries to address the issue. It is in progress.

GheisMohammadi:
The issue was introduced by the latest update to the muxers for the P2P host. Switching localnet to exclusively use Yamux (PR #4764 ) resolved the stream instability problems.

Frozen:
Resolved by #4764 and #4762: localnet moved from mplex to Yamux and stream lifecycle/error handling was improved. Closing as completed.

PR #4760, opened 2024-09-18
Disable stream sync in localnet

Frozen:
Disables stream sync for `make debug`

### DIFF
--- a/test/deploy.sh
+++ b/test/deploy.sh
@@ -72,7 +72,7 @@ function launch_localnet() {
     verbosity=3
   fi
 
-  base_args=(--log_folder "${log_folder}" --min_peers "${MIN}" --bootnodes "${BN_MA}" "--network_type=$NETWORK" --blspass file:"${ROOT}/.hmy/blspass.txt" "--dns=false" "--verbosity=${verbosity}" "--p2p.security.max-conn-per-ip=100")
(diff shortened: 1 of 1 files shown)


sophoah:
hey @Frozen could you explain  the reason for the PR ? what are you testing, and if a test do we need that change to be "permanent" ? Our direction is to use streamsync and disabling it doesn't make sense.

Frozen:
Because it has https://github.com/harmony-one/harmony/issues/4749, it unstable until it will not be fixed.

sophoah:
Considering :
1) the issue happens in localnet
2) streamsync is still experimental

I am ok with the change. However, let's have another Makefile target or an ENV so we can easily switch to and back that @mur-me can create with another PR

sophoah (approved):

sophoah:
@GheisMohammadi for your information and approval, see my rational above.

GheisMohammadi:
I don’t think this will be helpful in resolving any issue. The localnet is producing new blocks and functioning well, even with unstable connections. The proof is that all tests in the CI/CD pipeline, including the ones you've implemented, have passed. If you suspect connection issues and need to test something, I recommend trying these changes on your local system instead of submitting a PR. However, if you really need to create a PR, you could add a new command to the Makefile.

sophoah:
closing this one since is is done in #4761

PR #4762, opened 2024-09-24
Improve P2P Stream Handling & Stream Sync Enhancements

GheisMohammadi:
## Issue
This PR introduces several key improvements to the P2P stream handling and Stream Sync mechanisms, focusing on stability, error handling, and code optimization. Key changes include:

- Enhanced stream error handling for more visibility.
- Optimized pickAvailableStream to continue searching for other available streams when one fails.
- Refactored the stream request manager, including a Safe Map to ensure thread-safety.
- Improved logic for switching between short-range and long-range syncing.
- Optimized computeLongestHashChain and its whitelist functionality.
- Improved countHashMaxVote for accurate vote calculations on a specific hash.
- Added new configurations to better support staged stream sync.
- Fixed issues with Beacon Node and Epoch Chain detection.
- Introduced new metrics to monitor stream sync performance.
- Enhanced lock management and concurrency control.
- Passing the node roles throughout the stream protocol for better context awareness.
- Refined stream sync short-range helper functions.
- Improved the stream shutdown and cool-down processes, ensuring smooth terminations and halting unnecessary discoveries.
- Comprehensive stream manager refactor for improved readability and maintainability.
- Added more meaningful logs while removing unnecessary log entries or fields.


api/service/stagedstreamsync/metric.go
```diff
@@ -82,6 +94,18 @@ var (
 		[]string{"ShardID"},
 	)
 
+	numBlocksInsertedEpochSyncHistogramVec = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: "hmy",
+			Subsystem: "staged_stream_sync",
+			Name:      "num_blocks_inserted_epoch_sync",
+			Help:      "number of blocks inserted for each epoch sync",
```
sophoah, line 102: what do you mean here by  `each epoch sync`

GheisMohammadi: the number of blocks have been inserted by epoch sync

sophoah:
@Frozen please review/approve

api/service/stagedstreamsync/metric.go
sophoah: it's the word "each" that bothers me. Does it mean we have multiple epoch sync ? if yes wouldn't have number always be 1 ? as it would increment at every epoch ?

GheisMohammadi: Actually it is a cycle of **epoch synchronization**, where each cycle attempts to add any blocks that were missed on the **epoch chain**. The number of missed blocks can vary; it might be just one or multiple blocks, depending on factors such as node crashes or the node being stuck for a period of time.

GheisMohammadi: BTW, any suggestion for name? I can change it to whatever makes more sense

sophoah (approved):

Frozen (approved):

PR #4765, opened 2024-10-02
Improve Crosslink Retrieve from RawDB

GheisMohammadi:
## Issue

This PR introduces a check for the Crosslink Epoch before attempting to retrieve crosslinks from the database. By validating the epoch first, it prevents errors related to the retrieval of non-existent crosslinks.

### DIFF
--- a/core/blockchain_impl.go
+++ b/core/blockchain_impl.go
@@ -90,7 +90,10 @@ var (
 	blockValidationTimer = metrics.NewRegisteredTimer("chain/validation", nil)
 	blockExecutionTimer  = metrics.NewRegisteredTimer("chain/execution", nil)
 	blockWriteTimer      = metrics.NewRegisteredTimer("chain/write", nil)
-
+	// ErrCrosslinkNotFound is the error when no crosslink found
+	ErrCrosslinkNotFound = errors.New("crosslink not found")
+	// ErrZeroBytes is the error when it reads empty crosslink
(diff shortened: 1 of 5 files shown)


core/blockchain_impl.go
```diff
@@ -2367,8 +2370,15 @@ func (bc *BlockChainImpl) DeleteCrossLinks(cls []types.CrossLink) error {
...
 func (bc *BlockChainImpl) ReadCrossLink(shardID uint32, blockNum uint64) (*types.CrossLink, error) {
 	bytes, err := rawdb.ReadCrossLinkShardBlock(bc.db, shardID, blockNum)
-	if err != nil {
+	if err == goleveldb.ErrNotFound {
+		utils.Logger().Debug().Msgf("ReadCrossLink: not found")
+		return nil, ErrCrosslinkNotFound
+	} else if err != nil {
+		utils.Logger().Debug().Err(err).Msgf("ReadCrossLink failed")
 		return nil, err
+	} else if len(bytes) == 0 {
```
sophoah, line 2379: Q: if that case ever happened, what would be the fix ? delete shard 0 DB and resync ?

GheisMohammadi: This indicates that the database is not returning an error but is returning nil bytes for the crosslink. If this occurs, it suggests the crosslink is corrupted or not stored properly in the database. Resyncing the database should resolve the issue.

sophoah (approved):

Frozen (approved):

PR #4774, opened 2024-10-16
[ci] upgrade github actions and merge artifacts

sophoah:
Testing logs for ci.yaml can be seen here : https://github.com/sophoah/harmony/actions/runs/11362513771/job/31604875903

the error at the end is the docker image push which would work if the github action was running on the harmony github organization

upload-artifact:v4 doesn't accept an artefact with the same name, they all have to be unique and be merged at the end. More detail here https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#merging-multiple-artifacts

### DIFF
--- a/.github/workflows/build-docker-proto-image.yaml
+++ b/.github/workflows/build-docker-proto-image.yaml
@@ -9,14 +9,14 @@ jobs:
 
     steps:
       - name: Checkout harmony core code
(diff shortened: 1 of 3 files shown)


mur-me (approved):

nadem88:
Very strong project

PR #4778, opened 2024-10-22
Early return from `finalCommit` to prevent processing with empty hash.

Frozen:
`StagedStreamSync::doSync` calls `finishSyncing` which internally calls `reset` on consensus. `finalCommit`runs in goroutine, so it receives empty hash after `reset`.


sophoah:
We can't merged that commit. It will impact the validator signature received after the quorum is met

Issue #4796, opened 2024-11-11
Leader Rotation Corner Cases and Design Issues


GheisMohammadi:
**Detected Edge Cases**

- If a leader is not fully synced when taking over as leader after a view change, it may use an outdated blockHash or public key set. We should verify that, upon view change, the leader’s block and view state is fully updated before it starts processing `PREPARE` messages. Maybe checking for race conditions or delays in leader sync, especially after view changes.

- Another point here is: Since the leader processes all `PREPARE` messages from validators, it’s possible that only the leader encounters certain edge cases if `recvMsg.Payload` contains unexpected data.  Maybe we can put some logs to see if a certain payload or data is triggering some edge cases.

on the `rotateLeader` we find the committee like this:
```golang
    ss, err := bc.ReadShardState(epoch)
    if err != nil {
        utils.Logger().Error().Err(err).Msg("Failed to read shard state")
        return defaultKey
    }
    committee, err := ss.FindCommitteeByID(consensus.ShardID)
    if err != nil {
        utils.Logger().Error().Err(err).Msg("Failed to find committee")
        return defaultKey
    }
```
if the validator is not fully synced specially on view change (even 1 block difference), it detects the wrong epoch and it can cause the wrong leader.

- There is a variable called `IgnoreViewIDCheck`. This is the comment for this variable:
```
in a very rare case, when a M1 view change happened, the block contains coinbase for last leader but the new leader is actually recognized by most of the nodes. At this time, if a node sync to this exact block and set its leader, it will set with the failed leader as in the coinbase of the block.
This is a very rare case scenario and not likely to cause any issue in mainnet. But we need to think about a solution to take care of this case because the coinbase of the latest block doesn't really represent the the real current leader in case of M1 view change.
```
This indicates that, in rare instances, the coinbase of a block may not correctly represent the actual current leader. As a result, ensuring that the leader rotation mechanism accurately reflects the active leader in all network states is essential.

- The leader rotation mechanism includes a timeout for the next leader to produce and propagate a block. However, if the new leader has poor connectivity, it may delay the block’s propagation across the network. In this scenario, some nodes might receive the block and continue with consensus, while others that haven’t yet received it will trigger another leader rotation, waiting for the subsequent leader to produce a block. This can lead to network desynchronization, where part of the network moves forward while other nodes remain waiting, impacting overall consensus stability and efficiency.

**Unaddressed Edge Cases in Leader Rotation Calculations**

[...]

PR #4798, opened 2024-11-12
Enhance Leader Rotation Logic to Address Edge Cases in Leader Selection

GheisMohammadi:
## Issue
This PR refines the leader rotation logic in `NthNextValidator` to handle previously unaddressed edge cases that could lead to unexpected behavior or repeated selection of the same leader. In the refactored function for leader rotation, several corner cases were addressed to ensure robustness in the selection of the next validator.

One critical improvement was handling cases where `next` is greater than zero. In the original code, if the list wraps around from the last validator back to the beginning, there was a risk that the first validator could be skipped entirely. This is especially problematic when the validator list is circular, as it could lead to unfair distribution in leader selection. The refactored function avoids this by introducing attempts as a counter that methodically cycles through all nodes in `publicKeys`, ensuring no validator is skipped due to wrap-around.

Another addressed case involves scenarios where `next` is zero, yet the `pubKey` of the current leader isn't found within `publicKeys`. In the original code, if the `pubKey` index was -1 (indicating it wasn’t present), this situation could cause a crash when trying to access an invalid index. The refactored version improves safety by first logging this absence as an error and, if `next` is zero, returning the first public key in the list as a fallback. This default ensures stability, allowing the function to handle missing keys gracefully without interrupting leader selection.

Finally, the revised code prevents repeatedly selecting the same leader when a different validator isn’t readily available. In the original function, if the loop failed to find a unique validator, it would repeatedly return the same `pubKey`, potentially leading to repetitive selection cycles. The refactored version adds a check to ensure that only a distinct validator is returned. Additionally, a limit on attempts guarantees that the function terminates if no unique validator is found after a full cycle, logging a warning as an indication of this fallback behavior. This effectively prevents inefficient leader selection and ensures smoother rotations.

This PR introduces changes to the leader rotation logic that require a hard fork (HF) for full network-wide consistency. Because this update impacts how validators are selected, implementing it without a hard fork would lead to differing leader rotations among nodes running different versions of the code, potentially causing consensus issues.

For now, the hard fork epoch has been set to "TBD" (to be determined) to allow further discussion and coordination among stakeholders. This placeholder will be updated once the team agrees on an appropriate epoch for deployment.

This PR addresses 3 corner cases mentioned in issue #4796


sophoah:
@GheisMohammadi please check the travis failure

Frozen:
New functionality contains 0 tests coverage.

sophoah:
> New functionality contains 0 tests coverage.

Good point though it's not a new feature. Is it possible to write test cases here cc @GheisMohammadi

GheisMohammadi:
> > New functionality contains 0 tests coverage.
> 
> Good point though it's not a new feature. Is it possible to write test cases here cc @GheisMohammadi

I added tests for three edge cases to demonstrate how `Leader Rotation v1` fails to handle them. Additionally, I created the same tests for `Leader Rotation v2` to validate that it successfully addresses all three edge cases.

Frozen (approved):

PR #4799, opened 2024-11-12
Improve Consensus logic for Multi-BLS Validators with quorum

GheisMohammadi:
## Issue

This PR improves the consensus logic to handle cases where validators with multiple BLS keys achieve quorum on their initial signatures or commits. Currently, such scenarios can block phase transitions because the condition `if !quorumWasMet && quorumIsMet` doesn't account for quorum being achieved immediately.

To fix this, a new function, `checkFirstReceivedSignature`, is introduced. It checks whether quorum is reached on the first batch of signatures or commits. The PR also updates the logic for `Prepare` and `Commit` phases to properly handle these edge cases. The PR helps to better handling of multi-BLS key validators while maintaining compatibility with single-key setups.

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -216,6 +216,16 @@ func (consensus *Consensus) onCommit(recvMsg *FBFTMessage) {
 	quorumWasMet := consensus.decider.IsQuorumAchieved(quorum.Commit)
 
 	signerCount := consensus.decider.SignersCount(quorum.Commit)
+
+	// check if it is first received commit
+	// it may multi bls key validators can achieve quorum on first commit
+	var myPubkeys []bls.SerializedPublicKey
+	for _, key := range consensus.priKey {
+		myPubkeys = append(myPubkeys, key.Pub.Bytes)
+	}
+	mySignsCount := consensus.decider.GetBallotsCount(quorum.Commit, myPubkeys)
+	isFirstReceivedSignature := signerCount == mySignsCount
+
 	//// Read - End
 
 	// Verify the signature on commitPayload is correct
@@ -289,8 +299,10 @@ func (consensus *Consensus) onCommit(recvMsg *FBFTMessage) {
 
 	quorumIsMet := consensus.decider.IsQuorumAchieved(quorum.Commit)
 	//// Read - End
+	quorumAchievedByFirstCommit := isFirstReceivedSignature && quorumWasMet
+	quorumAchievedByThisCommit := !quorumWasMet && quorumIsMet
 
-	if !quorumWasMet && quorumIsMet {
+	if quorumAchievedByFirstCommit || quorumAchievedByThisCommit {
 		logger.Info().Msg("[OnCommit] 2/3 Enough commits received")
 		consensus.fBFTLog.MarkBlockVerified(blockObj)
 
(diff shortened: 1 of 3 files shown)


sophoah:
@GheisMohammadi i am good with that PR. Can you do run the same logic with `isFirstReceivedSignature` for onPrepare ? and use it here https://github.com/harmony-one/harmony/blob/3b6fde45185f136038d1d2ebfc1885248284c4a7/consensus/leader.go#L124 ? It prevent doing the `return` when the leader node already has quorum on the first onPrepare. Goal is to run `didReachPrepareQuorum` https://github.com/harmony-one/harmony/blob/3b6fde45185f136038d1d2ebfc1885248284c4a7/consensus/leader.go#L188

GheisMohammadi:
> @GheisMohammadi i am good with that PR. Can you do run the same logic with `isFirstReceivedSignature` for onPrepare ? and use it here
> 
> https://github.com/harmony-one/harmony/blob/3b6fde45185f136038d1d2ebfc1885248284c4a7/consensus/leader.go#L124
> 
> ? It prevent doing the `return` when the leader node already has quorum on the first onPrepare. Goal is to run `didReachPrepareQuorum`
> https://github.com/harmony-one/harmony/blob/3b6fde45185f136038d1d2ebfc1885248284c4a7/consensus/leader.go#L188

done

consensus/leader.go
```diff
@@ -121,14 +134,22 @@ func (consensus *Consensus) onPrepare(recvMsg *FBFTMessage) {
...
+
+	quorumPreExisting := consensus.decider.IsQuorumAchieved(quorum.Prepare)
+	//// Read - End
+
+	if quorumPreExisting {
 		// already have enough signatures
 		consensus.getLogger().Debug().
 			Interface("validatorPubKeys", recvMsg.SenderPubkeys).
 			Msg("[OnPrepare] Received Additional Prepare Message")
 		return
```
sophoah, line 151: when leader node with multi bls key has quorum, we hit that `return` code and never and will never hit `consensus.didReachPrepareQuorum()`

sophoah:
```
if !isFirstReceivedSignature && quorumPreExisting
```
first time with quorum we don't return so we can hit `consensus.didReachPrepareQuorum();` then subsequently we return

consensus/leader.go
```diff
@@ -183,7 +204,11 @@ func (consensus *Consensus) onPrepare(recvMsg *FBFTMessage) {
 	//// Write - End
 
 	//// Read - Start
-	if consensus.decider.IsQuorumAchieved(quorum.Prepare) {
+	quorumFromInitialSignature := hasMultiBlsKeys && isFirstReceivedSignature && quorumPreExisting
+	quorumPostNewSignatures := consensus.decider.IsQuorumAchieved(quorum.Prepare)
+	quorumFromNewSignatures := !quorumPreExisting && quorumPostNewSignatures
+
+	if quorumFromInitialSignature || quorumFromNewSignatures {
```
sophoah, line 211: old code was fine, as we just need `consensus.didReachPrepareQuorum()` to hit once

sophoah:
could you add PR description, and is this something we can write unit test (for onPrepare and onCommit) here as well ?

sophoah:
@Frozen please review/approve

GheisMohammadi:
> @GheisMohammadi please update PR description

done

sophoah:
updated this PR to "ready for review"
@Frozen please review/approve

Frozen (approved):

Frozen:
should i run it with  make debug-multi-bls or debug-ext or debug?

sophoah:
@Frozen you should run it with debug-multi-bls as it fixes a bug that only happens when using multi-bls and when the harmony node have quorum

Frozen:
rebase pls branch on dev

sophoah (approved):

PR #4802, opened 2024-11-14
View change: Support for checking if validators belongs to the same key.

Frozen:
Support for checking if validators belongs to the same key.


internal/params/config.go
```diff
@@ -596,6 +598,8 @@ type ChainConfig struct {
 	// vote power feature  https://github.com/harmony-one/harmony/pull/4683
 	// if crosslink are not sent for an entire epoch signed and toSign will be 0 and 0. when that happen, next epoch there will no shard 1 validator elected in the committee.
 	HIP32Epoch *big.Int `json:"hip32-epoch,omitempty"`
+
+	ViewChangeSkipValidatorsSameAddressEpoch *big.Int `json:"view-change-skip-validators-same-address-epoch,omitempty"`
```
sophoah, line 602: can we use the same epoch hardfork name as in https://github.com/harmony-one/harmony/pull/4798 ?

consensus/quorum/quorum.go
```diff
@@ -201,20 +202,18 @@ func (s *cIdentities) IndexOf(pubKey bls.SerializedPublicKey) int {
 }
 
 // NthNext return the Nth next pubkey, next can be negative number
-func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (bool, *bls.PublicKeyWrapper) {
-	found := false
-
+func (s *cIdentities) NthNext(pubKey *bls.PublicKeyWrapper, next int) (*bls.PublicKeyWrapper, error) {
```
sophoah, line 205: NthNext seems to be an unused function ?

sophoah: ok my bad i just saw you use for view change, do we have to use NthNext for view change and NthNextValidator seperately ? can we combine both ?

consensus/view_change.go
```diff
@@ -140,10 +147,36 @@ func (consensus *Consensus) getNextViewID() (uint64, time.Duration) {
...
+}
+
+func (a nextLeaderParams) Config() *params.ChainConfig {
+	return a.config
+}
+
 // getNextLeaderKey uniquely determine who is the leader for given viewID
 // It reads the current leader's pubkey based on the blockchain data and returns
 // the next leader based on the gap of the viewID of the view change and the last
 // know view id of the block.
```
sophoah, line 162: can you review the comments here and update accordingly ? Or move that to getNextLeaderKey and have a new one for getNextLeaderKeySkipSameAddress

consensus/view_change.go
```diff
@@ -140,10 +147,36 @@ func (consensus *Consensus) getNextViewID() (uint64, time.Duration) {
...
+func (consensus *Consensus) getNextLeaderKeySkipSameAddress(viewID uint64, committee *shard.Committee) *bls.PublicKeyWrapper {
```
sophoah, line 163: i would renamed getNextLeaderKeyskipSameValidator

internal/chain/engine.go
```diff
@@ -190,44 +206,30 @@ func (e *engineImpl) VerifyVRF(
...
-		return nil, errors.Wrapf(err, "cannot read shard state %v %s",
-			h.Epoch(),
-			h.Coinbase().Hash().Hex(),
-		)
-	}
-
-	committee, err := shardState.FindCommitteeByID(h.ShardID())
-	if err != nil {
-		return nil, err
-	}
```
sophoah, line 206: Can you explain why removing that code helps with anything ?

consensus/view_change.go
Frozen: updated

consensus/quorum/quorum.go
Frozen: no, doesn't look like we can use both right now

internal/chain/engine.go
Frozen: moved to another pr, those changes does not affect view change directly

consensus/view_change.go
Frozen: updated

sophoah:
@Frozen can you help rebase and fix conflict here.

sophoah:
@Frozen can you rebase this one with the latest dev ?

sophoah:
@Frozen I tested killing the external validator node to test view change in two scenarios: make multi-bls-key and make-multi-bls-key-multi-extl-node (https://github.com/harmony-one/harmony/pull/4819) 

Both fails to even recover from view change. Networks is basically down when view change kick off when the external leader fails to propose

for the test `ps aux | grep 9008` in both scenario will give you the external leader PID to kill (kill -9 pid)

cc @mur-me

sophoah:
@Frozen can you fix the conflict

PR #4806, opened 2024-11-28
Update Quorum Check to Use GTE for Consistency

GheisMohammadi:
## Issue

This PR updates the `IsQuorumAchievedByMask` function in the `stakedVoteWeight` struct to use GTE (greater than or equal to) instead of GT (greater than) for comparing the voting power with the quorum threshold. It ensures the consistency across all quorum checks in the codebase


sophoah:
can you explain 
```
It ensures the consistency across all quorum checks in the codebase
```
is there any other places in the code we use GTE and not GT ?

For me it's just a small detail, we are but in our consensus the threshold is actually 2/3 +1 not 2/3, might have been on purpose that we initially used GT and not GTE.

what do you think ?

GheisMohammadi:
> can you explain
> 
> ```
> It ensures the consistency across all quorum checks in the codebase
> ```
> 
> is there any other places in the code we use GTE and not GT ?
> 
> For me it's just a small detail , but in our consensus the threshold is actually 2/3 +1 not 2/3, might have been on purpose that we initially used GT and not GTE.
> 
> what do you think ?

there’s a mismatch in how the two `IsQuorumAchievedByMask` functions handle quorum checks. In `uniformVoteWeight`, quorum is considered achieved if the `currentTotalPower` is greater than or equal to the threshold (>=). But in `stakedVoteWeight`, quorum is only achieved if the `currentTotalPower` is strictly greater than the threshold (>). This means one allows quorum when the voting power exactly equals the threshold, while the other doesn’t.

We need to decide how quorum should work. should it include the threshold (>=) or only exceed it (>)? Usually, quorum is about meeting or surpassing a threshold (>=), so it makes sense to align both functions to that. If we agree, we just need to tweak `stakedVoteWeight` to match the logic in `uniformVoteWeight` for consistency.

sophoah:
ok if you ask me i would go with GT, not GTE.

GheisMohammadi:
> we just need to tweak `stakedVoteWeight` to match the logic in `uniformVoteWeight` for consistency.

Then we need to tweak `uniformVoteWeight` to match the logic in `stakedVoteWeight`.

sophoah:
Consensus/quorum/one-node-one-vote.go is no longer used in mainnet/testnet/devnet. The only env that uses it in localnet and happens before the staking epoch

PR #4811, opened 2024-12-09
Fix Multi-BLS Key Validator Leader Rotation to Prevent Multiple Block Proposals

GheisMohammadi:
## Issue
This PR resolves an issue in the consensus module where multi-BLS key validators, upon retaining leadership, could create multiple block proposals and potentially crash the network. It adds the logic to verify if the leader has changed or if the previous leader belongs to the same node (multi-BLS key validator scenario) and ensures that the consensus does not re-trigger leadership actions unnecessarily for the same node. Also, it updates logs to better describe leader transitions and ensure clarity in network operation debugging.


sophoah:
PR has been tested multiple last week and the only one edge case left :
- at epoch transition
- leader bls key is on the same node (last block and first block)

sophoah (approved):

Frozen (approved):

PR #4815, opened 2024-12-12
Add Resource Manager to P2P Host

GheisMohammadi:
This PR adds resource manager to p2p configurations. The new configs are as below:

```
ResourceMgrEnabled   (**FLAG:**  --p2p.resmgr-enabled)     --> default: false
ResourceMgrMemoryLimitBytes (**FLAG:** --p2p.resmgr-memory-limit-bytes)   --> 0: unlimited
ResourceMgrFileDescriptorsLimit (**FLAG:** --p2p.resmgr-file-descriptor-limit)  --> 0: unlimited
```
config v2.6.5

The resource manager is disabled by default. When enabled, the default configuration allows unlimited resource usage.
The resource manager is enabled by default for bootnode


sophoah (approved):

sophoah:
tested on bootnode1 out of 4 and hat confirmed resolving the p2p bootnode issue at boot.

internal/configs/node/network.go
```diff
@@ -77,6 +77,12 @@ const (
 	// When the peer count exceeds the 'high watermark', as many peers will be pruned (and
 	// their connections terminated) until 'low watermark' peers remain.
 	DefaultConnManagerHighWatermark = 192
+	// DefaultResourceMgrEnabled enables P2P host Resource Manager
+	DefaultResourceMgrEnabled = false
```
mur-me, line 81: here we have false because this settings are for the harmony binary, right?

GheisMohammadi: True, it is disabled by default for harmony node. For boot node, it's enabled by default.

p2p/resource_manager.go
```diff
@@ -0,0 +1,161 @@
...
+		utils.Logger().Info().Msg("go-libp2p Resource Manager DISABLED")
+		return rmgr, err
+	}
+
+	// Auto-scaled limits based on available memory/fds.
+	if maxMemory == 0 {
+		maxMemory = uint64((float64(memory.TotalMemory()) * 0.25))
+		if maxMemory < 1<<20 { // 1 GiB
+			maxMemory = 1 << 20
+		}
```
mur-me, line 35: So here we are reserving 25% of whole RAM, but if the RAM < 1GB - we will put here 1 GB as a limit, this one looks a bit tricky imho.

From one side:
* we will need to remember that we must use machines with > 1GB of RAM or linux OOM killer will find us...
From other side:
* I believe this is okay, rn we have 2 or 4 GB machines

cc: @sophoah lets just have this in your mind for now

mur-me (approved):

p2p/resource_manager.go
mur-me: Agreed, @GheisMohammadi let's put 0.5GB as minimal amount of RAM

Frozen (approved):

p2p/resource_manager.go
sophoah: by the way, i think that PR is good already.  1 << 20 = 1 MiB, not 1 GiB.

mur-me: yes, you are right, thanks for fixing comment @sophoah, let's merge it

PR #4816, opened 2024-12-14
Fix for excess propose signals.

Frozen:
Added block number for ReadySignal. Additional check for firing block proposing from `setupForNewConsensus`.


consensus/consensus_v2.go
```diff
@@ -606,12 +606,23 @@ func (consensus *Consensus) preCommitAndPropose(blk *types.Block) error {
...
+			if consensus.isMyKey(next) {
+				consensus.ReadySignal(NewProposal(AsyncProposal, blk.NumberU64()+1), "preCommitAndPropose rotation", "proposing new block which will wait on the full commit signatures to finish")
+			}
+		}
 	}()
 
 	return nil
 }
 
+func (consensus *Consensus) isRotation(epoch *big.Int) bool {
```
sophoah, line 622: why re-creating a new function and not use directly IsLeaderRotationInternalValidators ?

Frozen: i don't need to know which kind of rotation, i need the fact.

sophoah:
@Frozen please fix travis CI:
```
# github.com/harmony-one/harmony/consensus [github.com/harmony-one/harmony/consensus.test]
consensus/consensus_test.go:100:27: too many arguments in call to consensus.finalCommit
	have (number, number, bool)
	want (bool)
```

consensus/consensus_v2.go
sophoah: ok and why recreating the function ?

Frozen:
fixed, lost some code during rebase

consensus/consensus_v2.go
Frozen: because leader rotation does not affect blockchain anyhow. It means, that we can safely remove internal and external epochs, and left only v2. Exactly in this part of code, i don't need to know which kind of rotation we use, i need to know that we have it active.

GheisMohammadi (approved):

mur-me (approved):
Checked against localnet via make debug-ext.

Faced both pre and post leaderRotationconsensus epochs in the `!consensus.isRotation(blk.Epoch())` condition

Frozen:
This fixes certain issues, additional fixes will be in the next Prs.

PR #4817, opened 2024-12-18
Removed multiSigBitmap.

Frozen:
Removed multiSigBitmap.


sophoah:
what's the advantage of doing this ?

PR #4818, opened 2024-12-19
Add New Command to Makefile for Enhanced Localnet Debugging

GheisMohammadi:
This PR adds a new Makefile command, `debug-multi-bls-with-terminal`, to improve the debugging process for running a localnet with external validators and multiple BLS keys. Unlike the existing `debug-multi-bls` command, which runs the localnet in the background and requires manual process management to view logs or handle errors, the new command uses `screen` to run the localnet in the foreground, allowing developers to easily attach to the session (`screen -r localnet`) and view live logs. All logs are stored in `./tmp_log/localnet_terminal.log` for convenient access. Additionally, the localnet automatically shuts down when the terminal or `screen` session is closed, streamlining cleanup and making it easier to debug issues without manual intervention.


sophoah (approved):

mur-me (approved):
Nice, it should work with log aggregator as well

PR #4819, opened 2024-12-20
multi bls multi ext node

sophoah:
This PR add a new localnet config based on make debug-multi-bls but replace the s0ext/s1ext bls key folder with a split of the two validators for each shard.
- s0ext1 contains keys of validator1 https://github.com/harmony-one/harmony/blob/main/test/build-localnet-validator.sh#L33
- s0ext2 contains keys of validator2 https://github.com/harmony-one/harmony/blob/main/test/build-localnet-validator.sh#L41
- s1ext1 contains keys of validator3 https://github.com/harmony-one/harmony/blob/main/test/build-localnet-validator.sh#L49
- s1ext2 contains keys of valdidator4 https://github.com/harmony-one/harmony/blob/main/test/build-localnet-validator.sh#L57

this PR can be used to test view process


mur-me:
I was rechecking info 2 or 3 times, so this is just a set of:
shard 0: 
* 2 bls external with separate node process
* 1 bls external with separate node process
shard 1: 
* 2 bls external with separate node process
* 1 bls external with separate node process

+ now I understand that `test/configs/local-multi-bls.txt` external validators are actually `aggregate-sig=false` validators

mur-me (approved):

Frozen (approved):

PR #4821, opened 2024-12-21
Fix for view id setting

Frozen:


sophoah:
The PR looks fine.  It's important to note that this change only applies when the node is syncing.  Consensus operations should not be active during synchronization.

sophoah:
cc @GheisMohammadi thanks for review and approval

sophoah (approved):

GheisMohammadi:
`syncNotReadyChan` notifies the consensus that it is not fully synchronized, transitioning it to Syncing mode. But why do we need to increment the consensus block and viewID?

PR #4822, opened 2024-12-21
chore: suggest using errors.New instead of fmt.Errorf with no parameters

RiceChuan:
suggest using errors.New instead of fmt.Errorf with no parameters


Frozen:
pls fix errors

PR #4823, opened 2024-12-28
Fixed excess propose on epoch change.

Frozen:
Fixed excess propose on epoch change.


sophoah:
we can close this one now right @Frozen

PR #4830, opened 2025-01-14
feat: store localnet logs between runs

mur-me:
What was done:
- change rm and clean commands to remove only subfolder for tmp_log folder- justification - tmp_log folder needed for the log aggregator, prevent removal on the go when log stack is running
- fix promtail config to point it to the right place
- add named volume to the docker-compose to store Loki indexed between localnet runs
- echance debug-start-log - it checking that localnet is running, exits
  if not
- add debug-delete-volume-log to  make, it removes loki named volume,
  warns about sudo rm for the loki files

UPD by 24 Jan:
- fixed loki config path
- tweak loki limits config

## Test
Main use case:
1. You started localnet via `make debug-ext` or similiar
2. You started log stack - debug-start-log
3. Open grafana and search against this run, e.g. `{filename=~"/var/log/log-20250116-195803/.*"}`
4. You stopped localnet run
5. You started again with `make clean debug-ext`
6. Check the fresh run with e.g. `{filename=~"/var/log/log-20250116-205803/.*"}`
7. Recheck that old run is still here

Tested commands 
* make debug-start-log 
```
$ make debug-start-log 
bash ./test/logs_aggregator/start_log_aggregator.sh
working in /home/user/.gvm/pkgsets/go1.22.5/global/src/github.com/harmony-one/harmony/test/logs_aggregator
/home/user/.gvm/pkgsets/go1.22.5/global/src/github.com/harmony-one/harmony/test/logs_aggregator
starting docker compose for log aggregation
[+] Running 4/4
 ✔ Volume "logs_aggregator_loki_data"  Created                                                                                                  0.0s 
 ✔ Container loki                      Started                                                                                                  1.4s 
 ✔ Container grafana                   Started                                                                                                  1.4s 
 ✔ Container promtail                  Started                                                                                                  1.3s 
Whole list of log folders
/home/user/.gvm/pkgsets/go1.22.5/global/src/github.com/harmony-one/harmony/tmp_log/log-20250116-195803
Opening Grafana
```
* make debug-delete-volume-log
```
$ make debug-delete-volume-log 
docker volume rm logs_aggregator_loki_data
logs_aggregator_loki_data
[WARN] - it needs sudo to remove folder created with loki docker image user
sudo rm -rf test/logs_aggregator/loki
[sudo] password for user: 
``` 

### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

8. **Describe how the plan was tested.**

9. **How much minimum baking period after the last flag epoch should we allow on Pangaea before promotion onto mainnet?**

10. **What are the planned flag epoch numbers and their ETAs on Pangaea?**

11. **What are the planned flag epoch numbers and their ETAs on mainnet?**

12. **What should node operators know about this planned change?**

13. **Does this PR introduce backward-incompatible changes *NOT* related to on-disk data structure and/or over-the-wire protocol?** (If no, continue to question 11.)

    **YES|NO**

14. **Does the existing `node.sh` continue to work with this change?**

15. **What should node operators know about this change?**

16. **Does this PR introduce significant changes to the operational requirements of the node software, such as >20% increase in CPU, memory, and/or disk usage?**

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,10 +55,13 @@ help:
 	@echo "travis_rpc_checker - run the Travis RPC checker script, defaulting the test branch to 'master' unless overridden by TEST_REPO_BRANCH"
 	@echo "travis_rosetta_checker - run the Travis Rosetta checker script, defaulting the test branch to 'master' unless overridden by TEST_REPO_BRANCH"
 	@echo "debug_external - cleans up environment, rebuilds the binary, and deploys with external nodes"
+	@echo "debug-multi-bls - cleans up environment, rebuilds the binary, and deploys with external nodes in configuration 1 harmony process -> 2 validators"
 	@echo "build_localnet_validator - imports validator keys, funds validator accounts, waits for the epoch, and creates external validators on a local network"
-	@echo "debug-start-log - start a docker compose Promtail->Loki->Grafana stack against localnet logs, needs docker compose and started localnet"
+	@echo "debug-start-log - start a docker compose Promtail->Loki->Grafana stack against localnet logs, creates"\
+		"persistent volume to store parsed logs between localnet runs, needs docker compose and started localnet"
 	@echo "debug-stop-log - stops a docker compose Promtail->Loki->Grafana stack"
 	@echo "debug-restart-log - restart a docker compose Promtail->Loki->Grafana stack"
+	@echo "debug-delete-log - removes persistent volume for the Loki and host folder for it"
 
 libs:
 	make -C $(TOP)/mcl -j8
@@ -101,7 +104,7 @@ debug-multi-bls:
 	echo sleep 10s before creating the external validator
 	sleep 10
 	bash ./test/build-localnet-validator.sh
-	
+
 debug-multi-bls-with-terminal:
 	# add VERBOSE=true before bash or run `export VERBOSE=true` on the shell level for have max logging
 	# add LEGACY_SYNC=true before bash  or run `export LEGACY_SYNC=true` on the shell level to switch to the legacy sync
@@ -111,7 +114,7 @@ debug-multi-bls-with-terminal:
 	bash ./test/build-localnet-validator.sh
 	screen -r localnet
 
-debug-multi-bls-multi-ext-node: pre-external
+debug-multi-bls-multi-ext-node:
 	# add VERBOSE=true before bash or run `export VERBOSE=true` on the shell level for have max logging
 	# add LEGACY_SYNC=true before bash  or run `export LEGACY_SYNC=true` on the shell level to switch to the legacy sync
 	./test/debug.sh ./test/configs/local-multi-bls-multi-ext-node.txt &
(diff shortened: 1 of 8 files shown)


Frozen (approved):

test/debug.sh
```diff
@@ -6,7 +6,6 @@ Localnet_Blocks_Per_Epoch=$2
 Localnet_Blocks_Per_Epoch_V2=$3
 
 ./test/kill_node.sh
-rm -rf tmp_log* 2> /dev/null
```
sophoah, line 9: let's still do the rm here so we have a proper cleanup

The assumption here is that we keep the loki logs in the DB, but we don't have to keep the old txt raw log.

mur-me: hm, good point, I'll double check this.

removed this rm, because we have an explicit `make clean` command, but probably you are right, parsed logs should be already in the loki DB

Makefile
```diff
@@ -236,7 +236,7 @@ travis_rosetta_checker:
 	TEST_REPO_BRANCH='master'
 	bash ./scripts/travis_rosetta_checker.sh
 
-debug_external: clean
+debug_external:
```
sophoah, line 239: we use make debug-ext which also call the the script to build the external validator 

hence let's remove this completely and delete https://github.com/harmony-one/harmony/blob/dev/Makefile#L57

test/debug.sh
mur-me: well, if we are going this way, then it will be shift from storing logs between runs to store a Loki db.

sophoah: add in the distclean the db clean up

mur-me: yes, will do the full idea with the upcoming commit :pray:

Makefile
mur-me: done in the last commit

mur-me:
> ```
> 2025-01-22 16:10:22 level=warn ts=2025-01-22T09:10:22.112218886Z caller=client.go:419 component=client host=loki:3100 msg="error sending batch, will retry" status=429 tenant= error="server returned HTTP status 429 Too Many Requests (429): Ingestion rate limit exceeded for user fake (limit: 4194304 bytes/sec) while attempting to ingest '3375' lines totaling '1048446' bytes, reduce log volume or contact your Loki administrator to see if the limit can be increased"
> ```

Hey @sophoah, how did you face this?

sophoah:
just start the localnet, it doesn't happen all the time i tried

mur-me:
> just start the localnet, it doesn't happen all the time i tried

Fixed in https://github.com/harmony-one/harmony/pull/4830/commits/2cda0aed1d4270c19045e88cc6ca9c5d6074730d commit

Proof - 150 MB/700K lines of logs ingested in a meter of 1-2 minutes and no more 420 errors in the logs:
image

sophoah (approved):

PR #4820, opened 2024-12-20
1 second finality

Frozen:
1 second finality

1. Created config `IsOneSecondEpoch`
2. Reward reduced 2 times.

PR #4835, opened 2025-01-20
Use DNS for localnet.

Frozen:
Use DNS for localnet instead of `NewLocalSyncingPeerProvider`


Frozen:
without this PR you have to hardcode node addresses. Each command potentially have it's own set of nodes. I already mentioned that, here you can see the problem https://github.com/harmony-one/harmony/pull/4820/files#diff-65c92fad7c09270b02e7cb7b0d66f7a7a4145645cdea36a9061bb45c62e9670fR209

Frozen:
Found issue with shard peer discovering, converted to draft

sophoah:
@Frozen do we still need this PR ? if not let's close it

PR #4837, opened 2025-01-23
Improve multiple propose at epoch change

Frozen:
Fix for doulbe proposing on epoch change.

Note: 
1) tested with legacy sync only
2) better contains [this](https://github.com/harmony-one/harmony/pull/4820/files#diff-65c92fad7c09270b02e7cb7b0d66f7a7a4145645cdea36a9061bb45c62e9670fR209) change to reduce side effects


sophoah (approved):

sophoah:
it's not perfect but i am at epoch 78 and there was no consensus loss running with `make debug-multi-bls-multi-ext-node`. However we still have two async proposal as per screenshot 
image

sophoah:
@GheisMohammadi please review/approve

PR #4839, opened 2025-01-30
Stream Removal if its Sanity Check Fails

GheisMohammadi:
## Issue

This PR removes stream from peer store if its sanity check fails.

### DIFF
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -696,21 +696,22 @@ func (host *HostV2) Connected(net libp2p_network.Network, conn libp2p_network.Co
 	}
 }
 
+// TODO: this function could be used later to get peer topics and filter them based on the topics
 // getPeerTopics returns the list of topics a peer is subscribed to
-func (host *HostV2) getPeerTopics(peerID libp2p_peer.ID) map[string]bool {
-	topics := make(map[string]bool)
-	host.lock.Lock()
-	defer host.lock.Unlock()
-	for topic, t := range host.joined {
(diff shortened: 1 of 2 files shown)


p2p/host.go
```diff
@@ -693,6 +696,22 @@ func (host *HostV2) Connected(net libp2p_network.Network, conn libp2p_network.Co
 	}
 }
 
+// getPeerTopics returns the list of topics a peer is subscribed to
+func (host *HostV2) getPeerTopics(peerID libp2p_peer.ID) map[string]bool {
```
sophoah, line 700: that function is currently not used anywhere.  Is this required in this PR ? also please write the unit test for this function

GheisMohammadi: This function has been added for potential future use. If we later need to retrieve peer topics, we can use it. I included it preemptively because I believe it might be useful.

sophoah: noted please the add unit test

Frozen (approved):

p2p/host.go
sophoah: @GheisMohammadi can you add unit test for getPeerTopics or maybe create a seperate PR for that function with its test ? what do you think ?

Frozen (approved):

p2p/host.go
sophoah: @GheisMohammadi

sophoah (approved):

PR #4845, opened 2025-02-07
Prevent Explorer Nodes from Adding Future Blocks

GheisMohammadi:
This PR ensures that explorer nodes do not accept future blocks that are not the immediate next block during the COMMIT and PREPARE phase. Previously, there was no strict check on whether a received block was sequential, which could lead to inconsistencies.

### DIFF
--- a/node/harmony/node_explorer.go
+++ b/node/harmony/node_explorer.go
@@ -43,7 +43,15 @@ func (node *Node) explorerMessageHandler(ctx context.Context, msg *msg_pb.Messag
 				Msg("[Explorer] onCommitted unable to parse msg")
 			return err
 		}
-
+		// If a future block is received, it logs a warning and discards the block
+		// TODO: can be done with isRightBlockNumAndViewID(recvMsg)
+		if recvMsg.BlockNum != node.Blockchain().CurrentBlock().NumberU64()+1 {
+			utils.Logger().Warn().
+				Uint64("Received BlockNum", recvMsg.BlockNum).
+				Uint64("Consensus BlockNum", node.GetConsensusBlockNum()).
+				Msg("[Explorer] received a future block on COMMIT phase")
+			return nil
+		}
 		aggSig, mask, err := node.Consensus.ReadSignatureBitmapPayload(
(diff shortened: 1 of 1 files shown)


node/harmony/node_explorer.go
```diff
@@ -43,7 +43,15 @@ func (node *Node) explorerMessageHandler(ctx context.Context, msg *msg_pb.Messag
 				Msg("[Explorer] onCommitted unable to parse msg")
 			return err
 		}
-
+		// If a future block is received, it logs a warning and discards the block
+		// TODO: can be done with isRightBlockNumAndViewID(recvMsg)
+		if recvMsg.BlockNum != node.Blockchain().CurrentBlock().NumberU64()+1 {
+			utils.Logger().Warn().
+				Uint64("Received BlockNum", recvMsg.BlockNum).
+				Uint64("Consensus BlockNum", node.GetConsensusBlockNum()).
```
Frozen, line 51: Explorer node does not participate in consensus, so getting consensus block number doesn't make much sense for me

GheisMohammadi: Fixed

node/harmony/node_explorer.go
```diff
@@ -43,7 +43,15 @@ func (node *Node) explorerMessageHandler(ctx context.Context, msg *msg_pb.Messag
 				Msg("[Explorer] onCommitted unable to parse msg")
 			return err
 		}
-
+		// If a future block is received, it logs a warning and discards the block
+		// TODO: can be done with isRightBlockNumAndViewID(recvMsg)
+		if recvMsg.BlockNum != node.Blockchain().CurrentBlock().NumberU64()+1 {
```
Frozen, line 48: if current := node.Blockchain().CurrentBlock().NumberU64(); recvMsg.BlockNum != current +1 }
...

Frozen (approved):

node/harmony/node_explorer.go
GheisMohammadi: done

sophoah (approved):

PR #4852, opened 2025-02-17
Build Static Linux Executable from macOS

GheisMohammadi:
This PR introduces a solution for macOS users to build Linux static executables for the Harmony project. The primary goal of this change is to address the issue where macOS developers, who are unable to natively build Linux binaries on their machines, can now generate static Linux executables by leveraging Docker. 

The main addition is a new shell script that facilitates building the Linux static binary inside a Docker container. This allows developers on macOS to run the build process in a Linux environment without the need for a native Linux setup. The script automates the Docker image build, mounts the necessary directories, and executes the `make linux_static` command within the container, ensuring the resulting binaries are placed in a designated output directory. 

In addition to the shell script, a new `linux_static_from_macos` command has been added to the Makefile. This new command simplifies the process for macOS users by providing a convenient, one-line command to initiate the build process for Linux static executables. 

Please note that the first time the `linux_static_from_macos` command is run, the Docker image will need to be built, which may take some time. However, after the Docker image is successfully built and stored locally, subsequent builds will be much faster as the image will be reused without needing to rebuild it.

This PR ensures that developers can focus more on the code and less on managing cross-platform build environments.


sunwavesun (approved):

sophoah:
@Frozen

sophoah:
@Frozen

sophoah:
@Frozen

Frozen (approved):

PR #4863, opened 2025-02-27
Remove Staged DNS Sync

GheisMohammadi:
This PR removes "Staged DNS Sync". So, if node is set to use DNS sync, regardless of the StagedSync parameter, it uses DNS sync.


sophoah:
hey @GheisMohammadi this is also removing stagedsync for streamsync right ? If yes can we also update the config section ?

GheisMohammadi:
> hey @GheisMohammadi this is also removing stagedsync for streamsync right ? If yes can we also update the config section ?

No, it only deletes staged dns sync. Those config still needed for staged stream sync.

sophoah (approved):

PR #4867, opened 2025-03-10
Update to eth 1.9.25

Frozen:
Updates the Harmony codebase to be compatible with Ethereum 1.9.25, aligning core VM functionality and API signatures with the updated Ethereum client.

Updated EVM context creation from NewEVMContext to NewEVMBlockContext
Migrated from *big.Int to *uint256.Int for value handling in tracers
Updated tracer interfaces to match new Ethereum VM tracer signatures


Frozen:
rebased

PR #4871, opened 2025-03-11
fix(CI) - pin goimports and gencodec versions

mur-me:
## Issue

Both latest goimports and gencodec dependencies wants to have `golang.org/x/tools@v0.31.0` library, which supports only golang >= 1.23, but we are running 1.22.5.

Solution is consisted of 2 parts:
* pin goimports version to the 0.30.0 - previous latest and supports 1.22.5, thanks @Frozen for the fix
* start to use [gencodec@v0.1.0](https://github.com/fjl/gencodec/releases/tag/v0.1.0) :
  * no more vendors folder - previous reason to fork original repo
  * v0.1.0 release supports go 1.22.5

Additionally:
* travis ci time to time creates symlink from `$GOPATH/src/github.com/harmony-one/` to the `/home/travis/gopath/src/github.com/harmony-one/harmony` and it is crashing all our relative paths -solution is to remove if it exists
 
### Unit Test Coverage

Before:

```
```

After:

```
```

## Operational Checklist

    **YES|NO**

    **YES|NO**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,11 @@ install:
   - (cd $GOPATH/src/github.com/harmony-one/bls; make BLS_SWAP_G=1 -j4)
 #  - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.41.1
   - make go-get
-  - go install golang.org/x/tools/cmd/goimports@latest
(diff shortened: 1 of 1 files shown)


GheisMohammadi:
Great 👍

sophoah (approved):

Frozen (approved):

PR #4891, opened 2025-04-23
Optimize and Clean-Up Stages in Staged Stream Sync, Thread Safe P2P Stream

GheisMohammadi:
This PR improves the reliability and clarity of the sync process. The P2P stream is now thread-safe and resets automatically on failure, which helps prevent sync hangs. Short range sync was updated to accept partial blocks and properly break the loop when done, improving flexibility and performance during catch-up.

Several refactors and clean-ups were made to staged sync: initSync now uses internal parameters instead of external variables, and the unused LogProgress option was removed. We also fixed DB path handling for stream sync and re-added last mile blocks to ensure full chain coverage.

On the beacon side, we cleaned up redundant shard checks, fixed beacon helper initialization, and clarified internal comments. Minor improvements include fixing stream failure type handling and simplifying epoch tracking with a dedicated variable.

### DIFF
--- a/api/service/stagedstreamsync/beacon_helper.go
+++ b/api/service/stagedstreamsync/beacon_helper.go
@@ -176,10 +176,6 @@ func (bh *beaconHelper) getNextBlock(expBN uint64) *types.Block {
 		if b.NumberU64() < expBN {
 			continue
 		}
-		// if b.NumberU64() > expBN {
-		// 	bh.lastMileCache.push(b)
-		// 	return nil
-		// }
 		return b
 	}
 	return nil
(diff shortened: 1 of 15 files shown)


api/service/stagedstreamsync/beacon_helper.go
```diff
@@ -176,10 +176,10 @@ func (bh *beaconHelper) getNextBlock(expBN uint64) *types.Block {
 		if b.NumberU64() < expBN {
 			continue
 		}
-		if b.NumberU64() > expBN {
-			bh.lastMileCache.push(b)
-			return nil
-		}
+		// if b.NumberU64() > expBN {
```
mur-me, line 179: Do we need this this commented code?

Frozen (approved):

api/service/stagedstreamsync/beacon_helper.go
GheisMohammadi: removed

mur-me (approved):

PR #4892, opened 2025-04-24
Improved consensus logging

Frozen:
Feature/improved consensus logging


sunwavesun (approved):

internal/utils/timer.go
mur-me, line 1: Can our current issue with a new `dev` grow from this new timer?

PR #4901, opened 2025-05-14
Fully Disable Libp2p Resource Manager by Setting ResourceManager to nil

GheisMohammadi:
This PR ensures the libp2p resource manager is fully disabled by explicitly passing `libp2p.ResourceManager(nil)` in the host configuration. Although the existing configuration flag (`ResourceMgrEnabled = false`) was intended to disable the resource manager, libp2p enables it by default unless explicitly set to `nil`. This change prevents connection blocks caused by default resource limits, ensuring unrestricted P2P communication during development or when custom resource control is not required.

### DIFF
--- a/p2p/host.go
+++ b/p2p/host.go
@@ -266,6 +266,8 @@ func NewHost(cfg HostConfig) (Host, error) {
 			return nil, fmt.Errorf("failed to open resource manager: %w", err)
 		}
 		p2pHostConfig = append(p2pHostConfig, libp2p.ResourceManager(rmgr))
+	} else {
+		p2pHostConfig = append(p2pHostConfig, libp2p.ResourceManager(nil))
 	}
 
 	// Set host security


p2p/host.go
```diff
@@ -260,12 +260,14 @@ func NewHost(cfg HostConfig) (Host, error) {
 		libp2p.AutoNATServiceRateLimit(10, 5, time.Second*60),
 	}
 	if cfg.ResourceMgrEnabled {
-		rmgr, err := makeResourceMgr(cfg.ResourceMgrEnabled, cfg.ResourceMgrMemoryLimitBytes, cfg.ResourceMgrFileDescriptorsLimit, cfg.ConnManagerHighWatermark)
+		rmgr, err := makeResourceMgr(false, cfg.ResourceMgrMemoryLimitBytes, cfg.ResourceMgrFileDescriptorsLimit, cfg.ConnManagerHighWatermark)
```
Frozen, line 263: ```makeResourceMgr(cfg.ResourceMgrDisabled```

Frozen (approved):

p2p/host.go
GheisMohammadi: The `if cfg.ResourceMgrEnabled` condition determines whether resource management is enabled. Also, the first parameter in `makeResourceMgr` now sets infinite limits and it no longer controls enable/disable behavior.

mur-me (approved):

PR #4902, opened 2025-05-14
Refactor libp2p muxer config: use mplexC6 and yamux as default multiplexers

GheisMohammadi:
This PR updates the libp2p multiplexing configuration by removing support for the older `/mplex/0.7.0` protocol and setting `/mplex/6.7.0` (mplexC6) and `/yamux/1.0.0` (yamux) as the default muxers. The change reflects a more modern and stable setup aligned with current recommendations in the libp2p ecosystem.

The older mplex version was retained primarily for backward compatibility, but it's considered inefficient and outdated. The newer `/mplex/6.7.0` implementation offers better stream handling, improved interoperability, and is the version recommended for environments that still require TCP-based multiplexing without QUIC. Yamux is included for fallback and legacy compatibility, although it is now marked deprecated by the libp2p maintainers and may be removed in future refactors.

With this change, peer handshakes and stream management should be more reliable across nodes using newer libp2p stacks. It also simplifies the muxer configuration going forward while preserving minimal backward compatibility until QUIC becomes fully viable across the entire peer set.


mur-me:
Good point about backward compatibility :+1:

mur-me:
Quick question - do we have any statistics to understand what multiplexors we have in the wild rn?

GheisMohammadi:
> Quick question - do we have any statistics to understand what multiplexors we have in the wild rn?

Unfortunately not

Frozen (approved):

sunwavesun (approved):

mur-me (approved):
+ already installed on the devnet on all nodes

PR #4918, opened 2025-06-26
Log active muxer in p2p host

GheisMohammadi:
This PR logs which stream muxer is negotiated for each new connection.


Frozen (approved):

mur-me (approved):
Huge thanks for this one :pray:

PR #4939, opened 2025-08-13
Scope Discovery: Custom Kad-DHT + Connection Gating

joycaleb9705:
## Problems
1. Nodes discover unrelated ecosystems via broad/public discovery, leading to thousands of extra peers and off-namespace PubSub subscriptions.
2. Resource pressure (FDs/CPU/memory) correlates with dial failures like failed to negotiate security protocol: `EOF`.

## Rationale
A custom Kad-DHT protocol prefix isolates our routing table from the public DHT and keeps discovery Harmony-only.

Connection gating blocks or immediately tears down non-Harmony peers at the transport layer (pre-dial/accept and post-identify), preserving capacity for legitimate Harmony traffic.

## Changes
- Add config for a Harmony-scoped DHT prefix (default: `/harmony/kad/1.0.0`) and use only configured Harmony bootnodes.
- Implement connection gating:
  - Pre-dial / inbound-accept policy to allow only vetted peers/addresses (e.g., Harmony bootnodes/validators).
  - Post-identify filter: require advertised Harmony protocol IDs / chain marker; close non-conforming peers and purge their addrs from the peerstore.
  - Basic metrics/logging for gated/closed peers to validate effectiveness.

Notes:
No changes to PubSub topics, scoring, or connmgr here (follow-ups if needed).

## Rollout
1. Bootnodes first: enable dual-prefix (old + new) to avoid partitions.
2. Validators/regular nodes: switch to the new prefix; verify peer counts and dial success.
3. Converge (a new PR required): remove the old prefix on bootnodes after majority adoption.
4. Enable connection gating in “observe” or permissive mode (log only), then tighten to enforce close policies once metrics look healthy.

## Success criteria
- [ ] Peerstore and active connections drop to expected ranges; off-namespace subscriptions materially reduced.
- [ ] Non-Harmony peers are rejected at connect or immediately post-identify (visible in gating metrics).
- [ ] Dial success rate to Harmony peers improves; handshake EOFs materially reduced.

### DIFF
--- a/cmd/bootnode/main.go
+++ b/cmd/bootnode/main.go
@@ -176,6 +176,7 @@ func main() {
 		DialTimeout:                     time.Minute,
 		Muxer:                           *muxer,
 		NoRelay:                         *noRelay,
+		IsBootNode:                      true,
 	})
 	if err != nil {
 		utils.FatalErrMsg(err, "cannot initialize network")
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -613,6 +613,7 @@ func createGlobalConfig(hc harmonyconfig.HarmonyConfig) (*nodeconfig.ConfigType,
 		DialTimeout:                     hc.P2P.DialTimeout,
 		Muxer:                           hc.P2P.Muxer,
 		NoRelay:                         hc.P2P.NoRelay,
+		IsBootNode:                      false,
 	})
 	if err != nil {
 		return nil, errors.Wrap(err, "cannot create P2P network host")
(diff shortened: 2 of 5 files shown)


mur-me:
Hey, tests are failing with
```
p2p/host.go:394:31: cannot use disc.GetRawDiscovery() (value of type []"github.com/libp2p/go-libp2p/core/discovery".Discovery) as "github.com/libp2p/go-libp2p/core/discovery".Discovery value in argument to libp2p_pubsub.WithDiscovery: []"github.com/libp2p/go-libp2p/core/discovery".Discovery does not implement "github.com/libp2p/go-libp2p/core/discovery".Discovery (missing method Advertise)
```

PR #4940, opened 2025-08-16
Implement Progress-based Timeout System and Enhance Sync Detection Logic

GheisMohammadi:
This PR also addresses critical issues discovered during devnet testing of the Staged Stream Sync and introduces several improvements to enhance synchronization reliability and performance.

The current stream sync implementation suffers from short timeouts that causes partial data corruption and RLP parsing failures. When syncing large blocks over slow network connections, the system times out mid-message, leaving nodes with incomplete block data. This results in `rlp: too few elements for types.extblock` errors and causes nodes to stop syncing completely.

The existing timeout system uses fixed, non-adaptive timeouts that don't distinguish between "stalled connections" and "slow but active connections." When a connection is slow due to network conditions, the aggressive timeout triggers before the complete message is received, leading to partial data that fails RLP parsing.

This PR introduces a progress-based timeout system that intelligently manages stream timeouts:
Chunked Reading: Data is read in 2KB chunks with individual 30-second timeouts instead of a single aggressive timeout for the entire message. The system monitors data transfer progress and extends timeouts when data is actively flowing. Only truly stalled connections are timed out, while slow but active connections are allowed to complete

Significant improvements have been made to the staged stream sync range switching logic and error handling mechanisms. The downloader now properly handles failures by checking for minimum stream requirements and implementing more robust error recovery. We've also resolved semantic conflicts in sync status reporting and improved the overall sync state lifecycle management.

Additional enhancements include better consensus coordination through direct sync status management, and cleaner stream health monitoring with improved timeout handling.


mur-me (approved):

mur-me:
Great job on the stream/sync and crosslinks :pray:

PR #4941, opened 2025-08-25
Fix cross-link processing infinite loop by adding sync status checks and retry mechanism

GheisMohammadi:
## Problem
The node was getting stuck in an infinite loop when processing cross-links from epochs it couldn't handle due to being out of sync. This caused:
- **RPC unresponsiveness** - Node became unusable for API calls
- **Resource exhaustion** - CPU and memory usage skyrocketed
- **Sync failure** - Node couldn't complete synchronization
- **Service degradation** - Consensus and other services were affected

## Root Cause
When a node is out of sync (e.g., local epoch 5755, network epoch 5757), it receives cross-links for epochs it doesn't have shard state data for. The original code would:
1. Attempt to verify cross-links
2. Fail with "failed to read shard state from DB" 
3. Retry the same cross-links indefinitely
4. Never complete synchronization

## Solution
Implemented a comprehensive fix with two main components:

### 1. Sync Status Check (Early Filtering)
- **Epoch validation**: Check if cross-link epoch is within node's sync range
- **Tolerance window**: Allow cross-links up to 1 epoch ahead for minor sync delays
- **Immediate filtering**: Skip cross-links that can't be processed due to sync status
- **Resource savings**: Avoid unnecessary verification attempts

### 2. Retry Mechanism with Cleanup
- **Limited retries**: Maximum 3 attempts per cross-link
- **Exponential backoff**: 5-second delay between retries
- **Automatic cleanup**: Delete failed cross-links after max retries
- **Progress tracking**: Monitor retry attempts per cross-link

### DIFF
--- a/node/harmony/node_cross_link.go
+++ b/node/harmony/node_cross_link.go
@@ -1,6 +1,9 @@
 package node
 
 import (
+	"sync"
+	"time"
+
 	common2 "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/rlp"
 	ffi_bls "github.com/harmony-one/bls/ffi/go/bls"
@@ -9,7 +12,6 @@ import (
 	"github.com/harmony-one/harmony/internal/utils"
 	"github.com/harmony-one/harmony/shard"
 	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
 )
 
 var CrosslinkOutdatedErr = errors.New("crosslink signal is outdated")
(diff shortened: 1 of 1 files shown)


mur-me (approved):

node/harmony/node_cross_link.go
```diff
@@ -181,50 +257,121 @@ func (node *Node) ProcessCrossLinkMessage(msgPayload []byte) {
...
+				utils.Logger().Warn().
+					Int("processed", i).
+					Int("total", len(crosslinks)).
+					Int("limit", crossLinkBatchSize*2).
+					Msg("[ProcessingCrossLink] Batch size limit reached, stopping processing")
 				break
 			}
 
-			if _, ok := existingCLs[cl.Hash()]; ok {
-				nodeCrossLinkMessageCounterVec.With(prometheus.Labels{"type": "duplicate_crosslink"}).Inc()
```
mur-me, line 193: Question, do we need to delete this metric?

mur-me:
Thanks!

PR #4947, opened 2025-09-13
Fix staged stream sync RLP decoding to handle both extblock and BlockWithSig formats

GheisMohammadi:
This PR fixes a critical issue where Harmony nodes get stuck during staged stream sync due to RLP decoding errors. The problem occurs when different peers send blocks in different RLP formats - some send the older extblock format while others send the newer BlockWithSig format. When nodes try to sync with peers that send blocks in the BlockWithSig format, they fail with "rlp: too few elements for types.extblock" errors because the staged stream sync was only trying to decode blocks as the extblock format. This causes sync to stall and nodes to get stuck.

This fix implements a fallback RLP decoding mechanism that tries to decode blocks in the extblock format first, and if that fails, falls back to the BlockWithSig format. This approach is already successfully used in DNS sync and epoch block processing, so we're applying the same proven pattern to staged stream sync.

This resolves the sync stalling issues and eliminates the getBlockHashes errors that were preventing nodes from completing their synchronization process. The fix is backward compatible and doesn't affect existing functionality.

### DIFF
--- a/api/service/synchronize/stagedstreamsync/stage_bodies.go
+++ b/api/service/synchronize/stagedstreamsync/stage_bodies.go
@@ -469,14 +469,27 @@ func (b *StageBodies) verifyBlockAndExtractReceiptsData(batchBlockBytes [][]byte
 		if blockBytes == nil {
 			continue
 		}
-		if err := rlp.DecodeBytes(blockBytes, &block); err != nil {
-			b.configs.logger.Error().
-				Uint64("block number", i).
-				Msg("block size invalid")
-			return ErrInvalidBlockBytes
-		}
-		if sigBytes != nil {
-			block.SetCurrentCommitSig(sigBytes)
+		var err error
+		// Try to decode as *types.Block first (extblock format)
+		if err = rlp.DecodeBytes(blockBytes, &block); err != nil {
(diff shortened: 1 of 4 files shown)


api/service/synchronize/stagedstreamsync/stage_bodies.go
```diff
@@ -469,14 +469,27 @@ func (b *StageBodies) verifyBlockAndExtractReceiptsData(batchBlockBytes [][]byte
...
 		}
-		if err := rlp.DecodeBytes(blockBytes, &block); err != nil {
-			b.configs.logger.Error().
-				Uint64("block number", i).
-				Msg("block size invalid")
-			return ErrInvalidBlockBytes
-		}
-		if sigBytes != nil {
-			block.SetCurrentCommitSig(sigBytes)
+		var err error
```
Frozen, line 472: we already have function for this `RlpDecodeBlockOrBlockWithSig`

Frozen (approved):

api/service/synchronize/stagedstreamsync/stage_bodies.go
GheisMohammadi: good suggestion, have updated it

Frozen (approved):

PR #4953, opened 2025-10-06
Update 1.10: London and Berlin forks

Frozen:


consensus/view_change_test.go
mur-me, line 1: Seems this fixed the segfault on the trace RPC calls

PR #5014, opened 2026-03-05
Enable P2P peer scoring in host and add concrete sync request score updates

GheisMohammadi:
Peer scoring in the P2P stack is now fully wired into the host startup path instead of remaining effectively dormant. The change introduces configuration-driven controls for score retention and minimum accepted score, and uses those values during host construction to activate score-based connection gating when enabled. This keeps current behavior unchanged by default, while making scoring an explicit and operationally controllable feature.
On top of the host-level wiring, the sync request/response client path now updates peer scores using concrete runtime outcomes. Successful requests increase the valid-response signal, while transient failures increase an error-response signal. Responses that look malformed or indicate protocol mismatch are treated more strictly and counted as rejected payload behavior. This gives the scoring system meaningful, ongoing input derived from actual protocol interactions rather than static plumbing alone.
The implementation also includes defensive handling for test and mock environments where a full host or peerstore may not be present. In those cases, scoring hooks safely no-op instead of panicking, which preserves test stability and allows the broader suite to run normally. Overall, this PR moves peer scoring from “defined but mostly inactive” to “actively producing signal and enforceable when configured,” with backward-compatible defaults so operators can roll it out deliberately.


mur-me:
Overall looks good, but I have one improvement idea, but I've probably already guess it - we need metrics on that, something like `peer_score{instance="p2p_id/ipv4", remote_p2p_id="p2p_id", score='"}`

WDYT?

GheisMohammadi:
> peer_score{instance="p2p_id/ipv4", remote_p2p_id="p2p_id", score='"}

[PR#5017](https://github.com/harmony-one/harmony/pull/5017)

PR #5016, opened 2026-03-08
Reduced block production timeout

Frozen:
Reduced block production timeout. On my local machine produced 1639 block for 1664 seconds which is no more than 1.5% deviation for the ideal scenario

### DIFF
--- a/consensus/leader.go
+++ b/consensus/leader.go
@@ -337,6 +337,12 @@ func (consensus *Consensus) onCommit(recvMsg *FBFTMessage) {
 		if maxWaitTime > waitTime {
 			waitTime = maxWaitTime
 		}
+		if consensus.Blockchain().Config().IsOneSecond(consensus.Blockchain().CurrentBlock().Epoch()) {
+			waitTime = time.Until(consensus.NextBlockDue) - 200*time.Millisecond
+			if waitTime < 0 {
+				waitTime = 0
+			}
+		}
 		go consensus.finalCommit(waitTime, viewID, consensus.isLeader())
 
 		consensus.msgSender.StopRetry(msg_pb.MessageType_PREPARED)


mur-me:
Let's check this on the devnet as we discussed

mur-me:
Well, what we have right now after the run in the devnet:
* it was good in the logs - all 60 blocks were produced and validated, **but only in the logs**.
* but in fact, external validators, especially far away geographically from the current leader was signing only 12-15% of the blocks and were kicked off from the committee.
* only shard 0 was affected, when the shard 1 nodes were good generally.

Right now I'm suspecting that there should be more investigation or/and code change for the shard 0

mur-me:
Deployed on devnet, hardfork in 9 epochs

mur-me:
Tested, result is the same, we still have issue on the shard 0 like previous time.

Some of geographically far away nodes are still signing less blocks than 66.6% and getting kicked from the committee:
| Address | Epoch | Signed | To-sign | Percent | Status |
|---|---:|---:|---:|---:|---|
| `one1tedqsp6fm47cmp7xk5m44stvz8pxjmzursl2r5` | 50370 | 1132 | 1774 | 63.81% | active |
| `one1d95s9qrvrryfawysa9jvpxysxwazyt45zd6su3` | 50370 | 736 | 896 | 82.14% | active |
| `one13djpgmsunxm3dzdjq6855hx905r8a86fa0wzhv` | 50370 | 2101 | 2688 | 78.16% | active |
| `one1mgjpfcz8w7nkakwty7errmcf956yp239umnfv6` | 50370 | 738 | 1774 | 41.60% | active |
| `one1ymyg4d0y4w8g6ythply0lkmp5dr6dvdxj5e67w` | 50370 | 507 | 887 | 57.16% | active |

GheisMohammadi:
This idea doesn't work.

With waitTime = 0, the leader freezes the bitmap as soon as 2/3 is reached. Validators whose commit messages are still in flight are recorded as missing, even if they signed prepare and sent commit on time. So they will be excluded from the next epoch committee:

Frozen:
@copilot rebase on dev

mur-me:
@Frozen copilot have done strange rebase, it is including 2 commits already in the dev into you PR

mur-me:
Other than that, I've tested this change on the devnet, both validation + block producing are around 60 blocks per minute for 24 hours, before it was around 35-42:
image

Let's do a final decision on the tomorrow meeting

Frozen:
@mur-me rebased

mur-me:
https://github.com/harmony-one/harmony/pull/5097 - fixed hooray counting on the hmy_consensus_bingo metric

PR #5019, opened 2026-03-18
fix: add partnerV4 instance for devnet 

Frozen:
fix: add partnerV4 instance for devnet

### DIFF
--- a/internal/configs/sharding/partner.go
+++ b/internal/configs/sharding/partner.go
@@ -40,6 +40,8 @@ const (
 
 func (ps partnerSchedule) InstanceForEpoch(epoch *big.Int) Instance {
 	switch {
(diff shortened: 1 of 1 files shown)


internal/configs/sharding/partner.go
```diff
@@ -137,3 +139,12 @@ var partnerV3_1 = MustNewInstance(
 	hip30CollectionAddressTestnet, partnerReshardingEpoch,
 	PartnerSchedule.BlocksPerEpoch(),
 )
+
+var partnerV4 = MustNewInstance(
+	2, 20, 4, 0,
```
mur-me, line 144: Maybe even 0 or 1 will be okay here for the internal validators

PR #5031, opened 2026-03-31
fix legacy structLog JSON encoding

GheisMohammadi:
This PR makes error field truly optional in tracer logger. `StructLogRes.Error` was a string. Even when there was no error, JSON had "error": "". After changes, It’s a *string with `omitempty`, and we only set it when there is a real error message. Many tracers/INDEXERS/specs expect the field to be absent when no error occurs, not present with "". This fixes schema/decoder issues and false “error” detections. 
Current memory was chunked into 32‑byte slices and encoded with ```fmt.Sprintf("%x", slice)```. A partial word (e.g. 2 bytes) gave "0xaabb" instead of a 32‑byte (64‑hex‑char) word. After changes each memory chunk is:
- Cut into up to 32 bytes,
- Zero‑padded to 32 bytes,
- Hex‑encoded via `hexutil.Bytes(word).String()` → 0x + 64 hex chars.
Tools that assume 32‑byte, zero‑padded, 0x‑prefixed words (as in Geth) no longer break or misinterpret short words at the end of memory.

Also, the PR helps the storage and return data hex encoding is now consistent. In current code:
Storage keys/values were `fmt.Sprintf("%x", hash)` → no 0x prefix, inconsistent with most Ethereum JSON APIs.
`ExecutionResult.ReturnValue` was a plain hex string and manually blanked out on failure.
But the PR changes it to:
- Storage uses `hash.Hex()` for both key and value → consistent 0x‑prefixed hex.
- ReturnValue is `hexutil.Bytes`, and we zero the underlying bytes on hard failure.

So, now debuggers and explorers that expect standard 0x‑prefixed hex (like the Execution APIs / Geth) can work with your traces without custom handling.

The PR fixes **debug_traceTransaction** style output’s JSON shape and hex formats now match what existing Ethereum tooling already knows how to parse.

Note: the same changes on Ethereum can be found on [PR#34093](https://github.com/ethereum/go-ethereum/pull/34093/changes)

### DIFF
--- a/hmy/tracers/logger/logger.go
+++ b/hmy/tracers/logger/logger.go
@@ -81,6 +81,33 @@ type StructLog struct {
 	CallerAddress   common.Address `json:"callerAddress"`
 }
 
+// structLogLegacy stores a structured log emitted by the EVM while replaying a
+// transaction in debug mode.
+type structLogLegacy struct {
+	Pc            uint64             `json:"pc"`
+	Op            string             `json:"op"`
+	Gas           uint64             `json:"gas"`
+	GasCost       uint64             `json:"gasCost"`
+	Depth         int                `json:"depth"`
+	Error         string             `json:"error,omitempty"`
+	Stack         *[]string          `json:"stack,omitempty"`
+	ReturnData    string             `json:"returnData,omitempty"`
+	Memory        *[]string          `json:"memory,omitempty"`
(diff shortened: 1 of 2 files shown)


mur-me:
@mur-me - check the failed rpc tests, fix should be simple, see the golang tests from this PR

Frozen (approved):

PR #5037, opened 2026-04-21
Await until next second for block timestamp validation

Frozen:
This pull request makes several improvements to the block proposing logic in the consensus mechanism, primarily focusing on more precise block timestamp handling and improved sleep timing. The changes ensure that block timestamps are strictly increasing and reflect accurate wall-clock time, especially when using NTP, and refactor the code to pass the current time explicitly.

**Block timestamp handling and accuracy:**

* The logic for determining the block timestamp is moved from `ProposeNewBlock` to the proposer loop (`WaitForConsensusReadyV2`), ensuring that the block timestamp is always strictly greater than the parent block's timestamp. The code now explicitly sleeps until the next second if necessary.
* The `now` time value is passed explicitly to `ProposeNewBlock`, rather than being determined inside the method, improving testability and clarity. [[1]](diffhunk://#diff-2668cc7eabbb47387732df8057e06ea5364a7216a77b5b02864364d65827133fL21-R25) [[2]](diffhunk://#diff-2668cc7eabbb47387732df8057e06ea5364a7216a77b5b02864364d65827133fL35-R35) [[3]](diffhunk://#diff-bbc0fda6ab77138378c2a3470d12c91cb080269ae82b349e2cae87dd8d8858fdL79-R103)
* fix ntp servers parsing
* add random pick of ntp server to get time 
* lower future block tolerance to the 1s, we are not Ethereum with 12s block finality 
* exit if no ntp servers provided
* update ntp list to the broader set of servers, remove example server from the beevik library
* block proposal readiness is checked with consensus.registry.Now() guaranteeing that blocks will have only increasing timestamps

### DIFF
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -682,7 +682,7 @@ func setupChain(hc harmonyconfig.HarmonyConfig, nodeConfig *nodeconfig.ConfigTyp
 		chainDBFactory = &shardchain.LDBFactory{RootDir: nodeConfig.DBDir}
 	}
 
-	engine := chain.NewEngine()
+	engine := chain.NewEngineWithTime(registry.Now)
 	registry.SetEngine(engine)
 
 	chainConfig := nodeConfig.GetNetworkType().ChainConfig()
--- a/cmd/harmony/main.go
+++ b/cmd/harmony/main.go
@@ -257,6 +257,7 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 		impl, err := ntptime.TryNew(nodeConfig.NtpServer, 3)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error initializing NTP time provider: %v\n", err)
+			os.Exit(1)
 		} else {
 			go impl.Run(context.Background(), time.Minute)
 			ntpService = impl
--- a/consensus/proposer.go
+++ b/consensus/proposer.go
@@ -42,22 +42,10 @@ func (p *Proposer) WaitForConsensusReadyV2(stopChan chan struct{}, stoppedChan c
 				for retryCount := 0; retryCount < 3 && consensus.IsLeader(); retryCount++ {
 					var (
 						currentHeader = p.consensus.Blockchain().CurrentHeader()
-						nowEpoch      = currentHeader.Epoch()
 						parentTime    = currentHeader.Time().Int64()
+						now           = consensus.registry.Now()
+						timestamp     = now.Unix()
 					)
-
-					// Update worker's current header and
-					// state data in preparation to propose/process new transactions.
-					// After NTPEpoch, use the NTP-corrected clock stored in the registry
-					// so that the block timestamp reflects true wall-clock time even when
-					// the local system clock drifts.
-					var now time.Time
-					if consensus.Blockchain().Config().IsNTP(nowEpoch) {
-						now = consensus.registry.Now()
-					} else {
-						now = time.Now()
-					}
-					timestamp := now.Unix()
 					if timestamp <= parentTime {
 						// Current time is within the same second as the parent block.
 						// Sleep until the next second so the child timestamp is strictly
--- a/consensus/proposer.go
+++ b/consensus/proposer.go
@@ -40,7 +40,31 @@ func (p *Proposer) WaitForConsensusReadyV2(stopChan chan struct{}, stoppedChan c
 				return
 			case proposal := <-consensus.GetReadySignal():
 				for retryCount := 0; retryCount < 3 && consensus.IsLeader(); retryCount++ {
-					time.Sleep(SleepPeriod)
+					var (
+						currentHeader = p.consensus.Blockchain().CurrentHeader()
+						nowEpoch      = currentHeader.Epoch()
+						parentTime    = currentHeader.Time().Int64()
+					)
+
+					// Update worker's current header and
+					// state data in preparation to propose/process new transactions.
+					// After NTPEpoch, use the NTP-corrected clock stored in the registry
+					// so that the block timestamp reflects true wall-clock time even when
+					// the local system clock drifts.
+					var now time.Time
+					if consensus.Blockchain().Config().IsNTP(nowEpoch) {
+						now = consensus.registry.Now()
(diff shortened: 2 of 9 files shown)


mur-me:
:heavy_check_mark: tested on the devnet nodes, and we don't have this problem anymore, see the screenshot: 

image

mur-me:
Hey @Frozen, please fix https://github.com/harmony-one/harmony/blob/ddc96ae953a8185c908c8846450a5613c7ea9539/cmd/harmony/main.go#L250-L270 because it is interpreted as string without any slicing into separate dns names.

From the node log:
```
Error initializing NTP time provider: lookup 1.pool.ntp.org,0.beevik-ntp.pool.ntp.org: no such host
```

See the https://github.com/harmony-one/harmony/blob/cfe12540894cd5c888008cbab173b44ed66e3a85/common/ntp/ntp.go#L57-L67 as example in our codebase

cmd/harmony/main.go
```diff
@@ -257,6 +257,7 @@ func setupNodeAndRun(hc harmonyconfig.HarmonyConfig) {
 		impl, err := ntptime.TryNew(nodeConfig.NtpServer, 3)
 		if err != nil {
 			fmt.Fprintf(os.Stderr, "Error initializing NTP time provider: %v\n", err)
+			os.Exit(1)
```
mur-me, line 260: 👍

mur-me:
Hey @Frozen after the last commit I've seeing time to time the following error message and a view change after it:
```
[finalCommit] channel not received after 6s for commitSigAndBitmap
```
I believe the reason is the following: 
1. NTP provider actually works with multi-server config
2. registry.Now() uses real ClockOffset
3. same-second timestamp guard triggers more often
4. proposer delays before creating receiver
5. finalCommit sometimes times out sending commitSigAndBitmap

How is look like on the Hooray dashboard, see the 1 on the few validators:
image

mur-me:
Hey @Frozen, I've tried to debug the issue on the current leader and found the diff between linux ntp aka local vs golang ntp time aka registry and found that there is a ` "registryMinusLocal": -4.554831` difference between them. 
So I was right that this 2 time sources are different. 

```json
{
  "level": "warn",
  "shardID": 1,
  "myBlock": 40411052,
  "myViewID": 40422315,
  "phase": "Commit",
  "mode": "Normal",
  "parentTimeUnix": 1777905324,
  "registryNow": "2026-05-04T09:35:24.910835186-05:00",
  "localNow": "2026-05-04T09:35:24.915390017-05:00",
  "registryMinusLocal": -4.554831,
  "targetTime": "2026-05-04T09:35:25-05:00",
  "sleepByRegistry": 89.164814,
  "sleepByLocalUntil": 84.609983,
  "registryTimestampUnix": 1777905324,
  "retryCount": 0,
  "proposalBlockNum": 40411053,
  "isLeader": true,
  "asyncProposal": true,
  "called": "/home/um/harmony/consensus/consensus_v2.go:628",
  "caller": "/home/um/harmony/consensus/proposer.go:67",
  "time": "2026-05-04T09:35:24.915410385-05:00",
  "message": "[timestamp-guard] waiting for next second before proposing block"
}
```
Will add more after this comment

consensus/proposer.go
```diff
@@ -40,7 +40,19 @@ func (p *Proposer) WaitForConsensusReadyV2(stopChan chan struct{}, stoppedChan c
...
+						currentHeader = p.consensus.Blockchain().CurrentHeader()
+						parentTime    = currentHeader.Time().Int64()
+						now           = consensus.registry.Now()
+						timestamp     = now.Unix()
+					)
+					if timestamp <= parentTime {
+						// Current time is within the same second as the parent block.
+						// Sleep until the next second so the child timestamp is strictly
+						// greater, satisfying the engine.go validation check.
+						time.Sleep(time.Until(time.Unix(parentTime+1, 0)))
```
mur-me: Here we are mixing 2 time sources together.

And my debug got the following:
```
registryNow = 09:35:24.999764573
localNow    = 09:35:25.004318402
targetTime  = 09:35:25.000000000
```
Meaning:
```
consensus clock says: 09:35:24.999764573
local clock says:     09:35:25.004318402
target is:            09:35:25.000000000
```

So at the exact same moment:
```
registry.Now() says: not yet, wait 0.235 ms
time.Now() says:     already passed, no need to wait
```

On the last retry, this happened:
1. retryCount = 2
2. registry clock: still too early
3. local clock: already past target
4. sleep: 0ms
5. continue
5. loop ends
6. *the code never reached the proposal path*

mur-me:
So after test screw in both ways - -2 minutes and +2 minutes node is still validating and producing blocks using golang NTP time :heavy_check_mark:

mur-me:
@Frozen, please:
1. remove https://github.com/harmony-one/harmony/pull/5037/changes/c5d782aff5fb2c658d1db26ac71b7c04ca6c2437 because this is not working + overhead

2. Additionally, please add more ntp servers here: https://github.com/harmony-one/harmony/blob/dev/cmd/config/default.go#L146-L148
Why? 
* remove `0.beevik-ntp.pool.ntp.org` is just a demo for the library
* add `time.cloudflare.com`: 
> Today, **Cloudflare announces its free time service to anyone on the Internet**. We intend to solve the limitations with the existing public time services, in particular by increasing availability, robustness and security. Source: https://blog.cloudflare.com/secure-time/#use-it
* set of ntp.org servers - https://support.ntp.org/Servers/NTPPoolServers
* so final should be like: `time.cloudflare.com,0.pool.ntp.org,1.pool.ntp.org,2.pool.ntp.org,3.pool.ntp.org`

Frozen:
removed 0.beevik-ntp.pool.ntp.org

mur-me:
@Frozen, I've amend commit message and force pushed, please update your local :pray:

mur-me:
I've updated the PR description

mur-me:
All features non related for the separate NTP goroutine is done in https://github.com/harmony-one/harmony/pull/5042

Team decided to move on with the chrony daemon solution, it is the best possible way to achieve NTP sync:
- open source
- uses median time from multiple source
- exist on all Linux distros
- we do hardening for our case

PR #5043, opened 2026-05-19
core: fix cross-link signature verifier cache key to include shard ID

GheisMohammadi:
Cross-link and header signature verification uses an in-memory LRU cache to avoid repeating expensive BLS checks. The cache key previously included only the block hash, signature, and bitmap.

Signature verification also depends on the shard committee context (loaded by shard ID). This PR adds `shardID` to `verifiedSigKey` so cached results are scoped to the correct shard.


Frozen (approved):

Frozen:

PR #5046, opened 2026-05-21
core: duplicate cross-link rejection

GheisMohammadi:
Beacon blocks can carry multiple cross-links in Header.CrossLinks. Today we validate sorting and committed storage, but we do not ensure each (`shardID`, `blockNum`) pair appears only once within a single header. This change adds that constraint for blocks at or after a new chain-config epoch. A sorted cross-link list can be scanned in linear time for repeated shard block references. CrossLinks gains `HasDuplicateShardBlocks()` for that check.


Frozen (approved):

Frozen:

PR #5042, opened 2026-05-15
Remove NTP configurations and functionality

Frozen:
## Issue

This pull request removes the NTP time synchronization logic from the codebase. All code related to using NTP-corrected time for block timestamps, including the `ntptime` service, the `NTPEpoch` configuration, and registry support for NTP time, has been deleted. As a result, block timestamps now always use the local system clock.

The most important changes are:

**Removal of NTP time synchronization:**
* Deleted the entire `ntptime` service, including `api/service/ntp/service.go` and `api/service/ntp/local_time.go`, which handled NTP queries and local time fallback. [[1]](diffhunk://#diff-dcaefa4872eedbdd1952e79aa50c18e292443640c7e4c2783e360137f86c0d04L1-L108) [[2]](diffhunk://#diff-e56d6b90d1b1cadd2cfffd760b77b5090cfb25082ea66d9e4dacce0eae525ba1L1-L12)
* Removed all logic in `cmd/harmony/main.go` that initialized and used the NTP time provider, including setting it in the registry and running periodic updates. [[1]](diffhunk://#diff-864c5118b935b2cb1b63e6a371f24a9390844972ba6bf01cb739160ab1b2e018L23) [[2]](diffhunk://#diff-864c5118b935b2cb1b63e6a371f24a9390844972ba6bf01cb739160ab1b2e018L250-L264) [[3]](diffhunk://#diff-864c5118b935b2cb1b63e6a371f24a9390844972ba6bf01cb739160ab1b2e018L280)
* Changed block proposal logic in `consensus/consensus_block_proposing.go` to always use `time.Now()` for block timestamps, instead of optionally using NTP-corrected time.

**Configuration cleanup:**
[...]

PR #5048, opened 2026-05-27
consensus: fix cascading block-time skew with stall-safe timestamp caps

GheisMohammadi:
Harmony already requires each block timestamp to be after its parent’s, and not more than about 15 seconds ahead of the validator’s own clock. That catches obviously bad headers, but it does not stop chain time from jumping forward several seconds in one block when a leader’s clock is fast. The next leader must still pick a time after that parent, so block time can ratchet ahead of real wall time. Peers with slightly slow clocks may reject blocks or wait a long time before they can vote or propose, which hurts liveness and makes view changes messy. If a block with a very high timestamp is already on chain and a node’s clock is still behind that time, view-change logic used to take a non-deterministic fallback path so nodes could disagree on the next leader.

There is no shared network clock. Each node only sees its own `time.Now()` when validating or proposing. This PR keeps using wall clock (same as today’s 15-second rule) but aligns what the leader writes with what other nodes will accept, and limits how far one block can push time past its parent in the common case.

On header verification, when `TimestampValidationEpoch` is active, a block may not advance more than 12 seconds past its parent’s timestamp unless the node’s local time has already moved much further ahead (for example after a long period with no blocks). In that stall case the limit can follow local time so recovery does not take many blocks of tiny steps.

The leader uses matching rules before it proposes. If its clock is only moderately ahead of the parent, it clamps the timestamp so peers are likely to accept the block. If the gap is very large and looks like a real stall rather than mild skew (beyond the 45-second view-change slot band), it may propose at full local time so the chain can catch up in one block. If the leader’s clock is behind the parent time, it still waits until it can use a later second, but it will not sleep without limit; after 17 seconds it gives up that round instead of hanging the proposer.

View change is updated so that when timestamp validation is on, a block time in the future of local wall clock no longer triggers the old fallback for computing the next view ID. Nodes stay on the same deterministic timing path instead of diverging.

Before `TimestampValidationEpoch`, only the older leader behavior remains (sleep until parent+1, no clamp or give-up cap). Engine timestamp checks were already gated on that epoch; proposer and view-change behavior now follow the same gate.

**Notes**
- Constants (12s step, 30s stall heuristic, 17s sleep cap) are paired between engine and proposer; changing one side should be reflected on the other and in tests.
- If mainnet rollout needs a **separate fork flag** instead of piggybacking on `TimestampValidationEpoch`, that can be a follow-up; this PR implements the behavioral fix behind the existing timestamp-validation gate.

### DIFF
--- a/consensus/proposer.go
+++ b/consensus/proposer.go
@@ -7,6 +7,71 @@ import (
 	"github.com/harmony-one/harmony/node/harmony/worker"
 )
 
+const (
+	// maxProposerForwardStep mirrors internal/chain.maxBlockTimeStep (12s).
+	// When local wall is only moderately ahead of parent.Time, clamp the
+	// proposed unix time to parent+maxProposerForwardStep so the header stays
+	// inside the engine's per-block step bound relative to that parent.
+	// When wall is far ahead of parent (see stallRecoveryThreshold), do not
+	// clamp so the proposed time can follow local wall (matches engine's
+	// max(parent+step, localWall) step ceiling).
+	maxProposerForwardStep int64 = 12
+	// stallRecoveryThreshold separates the two clamp behaviors above, in
+	// seconds of (localWall - parent.Time). Large gaps often mean no block was
+	// committed for a while (parent time is stale vs this leader's clock), not
+	// only sub-second skew. The threshold is chosen above typical skew but
+	// below multi-minute scales; it is a heuristic, not consensus state.
+	stallRecoveryThreshold int64 = 30
+	// maxProposerCatchupWait caps sleeping until parent.Time+1 when the leader
+	// wall is behind the parent header time. Without a cap, a far-future
+	// parent would block this goroutine for a long time. 17s covers waiting
+	// out a parent timestamp that is still valid under allowedFutureBlockTime
+	// (15s) plus typical inter-second boundary (+1s), on the same machine.
+	maxProposerCatchupWait = 17 * time.Second
+)
+
+// proposalTiming describes the next wait/propose action from the leader's
+// local wall time and parent header unix time only (no chain or peer state).
+// Exactly one mode applies: ready (propose), sleep then retry, or giveUp.
+type proposalTiming struct {
+	// ready: when true, the leader should propose with timestamp = proposeAt.
+	ready     bool
+	proposeAt time.Time
+	// sleep: when non-zero, the leader should sleep for this duration and retry.
+	sleep time.Duration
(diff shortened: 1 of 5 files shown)


mur-me:
Checking on the devnet shard 1

mur-me:
Tests are the same as https://github.com/harmony-one/harmony/pull/5042#issuecomment-4486884815

All tests done on the 1 active current leader. 

| Leader skew | Required catch-up time | Observed / expected behavior |
|---:|---:|---|
| `-120s` | view change | view change |
| `-30s` | view change | view change |
| `-14s` | `~14s` | leader waited 14 seconds to get the time older than parent block |
| `-2s` | `~2s` | leader waited 2 seconds to get the time older than parent block   |
| `+5s` | `5s` | Skewed leader block accepted, leader waited 5 seconds to produce new block |
| `+14s` | `12s` | Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+15s` | `12s` | Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+16s` | `12s` | Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up  |
| `+30s` | `12s` | Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+35s` | `15s` | view change |
| `+42s` | `~35s` | A single +42s leader can poison the next view by creating a future parent/prepared block.  |
| `+45s` | view change | view change |
| `+120s` | view change | view change |
|All active validators stoppage|| Recovered :heavy_check_mark: |

mur-me:
With a fresh logic, I'll focus explicitly on following tests:

| Leader skew | Required catch-up time | Observed / expected behavior |
|---:|---:|---|
| `+5s` | `+5s` | Skewed leader block accepted, leader waited 5 seconds to produce new block |
| `+30s` | `12s` |Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+35s` | `12s` |Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+42s` | `12s` |Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+45s` | `12s` |Skewed leader block accepted; timestamp clamped to parent+12; leader waits for local wall time to catch up |
| `+50s` | view change | view change |
| `+60s` | view change | view change |

mur-me (approved):

PR #5050, opened 2026-05-28
core: Fix double-sign external slash denominator to use snapshot self-stake

GheisMohammadi:
Double-sign slashing already takes validator self-slash debt and total stake from the epoch snapshot, but the external delegator denominator was built by subtracting the validator’s live self-delegation from snapshot total stake. A validator could add self-stake after the snapshot and before slash settlement without increasing their own slash debt, while shrinking the external denominator and pushing disproportionate losses onto outside delegators (and inflating the slash beneficiary payout when those debts are capped to available balances).

This change aligns the denominator with the snapshot: `totalExternalStake` is computed as snapshot total stake minus snapshot self-stake, matching the snapshot amounts used in `makeSlashList` and in each external delegator’s slash share. Validator slashing still runs against current delegations so balances and `undelegations` reflect live state; only the proportional split for externals is corrected.


Frozen:

Frozen (approved):

PR #5051, opened 2026-06-03
Staking: bind validator wrapper RLP address to storing account

GheisMohammadi:
Harmony stores staking validator state as RLP-encoded `ValidatorWrapper` data in account code. The wrapper carries its own Address field, which must always match the account that holds that code. This change enforces that invariant on load and on every update, and applies the same check during delegate verification so mismatched wrappers cannot be accepted or persisted.


Frozen:

Frozen (approved):

PR #5054, opened 2026-06-09
core: Canonical slash group ordering for beacon finalization

GheisMohammadi:
Beacon-chain slash records are applied in groups during block finalization. The existing sort comparator did not define a consistent ordering across slash groups, so different nodes could apply the same slash payload in different orders and compute different state roots. This PR introduces a canonical lexicographic ordering by shard ID, height, view ID, and epoch.


Frozen:

Frozen (approved):

PR #5057, opened 2026-06-12
refactor(tests): improve transaction waiting logic and epoch checks i…

Frozen:
This pull request improves the reliability and efficiency of test scripts by replacing fixed sleep intervals with readiness and polling scripts. The changes ensure that scripts wait for services or blockchain epochs to be ready before proceeding, reducing flakiness and unnecessary waiting time.

**Test orchestration improvements:**

* Replaced fixed `sleep 10s` delays with a call to `./test/wait_rpc_ready.sh` in several `Makefile` targets (`debug-ext`, `debug-multi-bls`, `debug-multi-bls-with-terminal`, `debug-multi-bls-multi-ext-node`), ensuring that the RPC endpoint is ready before starting the validator setup. [[1]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L102-R116) [[2]](diffhunk://#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L132-R129)

* In `test/build-localnet-validator.sh`, replaced the epoch polling loop with a call to the new `./test/wait_till_n_epoch.sh` script for waiting until the blockchain reaches a specific epoch.

**Script enhancements:**

* Added a robust `test/wait_till_n_epoch.sh` script that polls the node for the current epoch, supports configurable timeouts and intervals, and fails gracefully if the target epoch is not reached in time.

* Updated `scripts/api_test.sh` to replace fixed `sleep 20s` waits with a call to `./test/wait_tx_receipt.sh`, polling for transaction receipts instead of relying on arbitrary delays. [[1]](diffhunk://#diff-6fa17b2c6718795716b568ecadb71f6e2241ecf78dfa0524093e58aab361a509L72-R72) [[2]](diffhunk://#diff-6fa17b2c6718795716b568ecadb71f6e2241ecf78dfa0524093e58aab361a509L116-R113)

### DIFF
--- a/.github/workflows/ci-pr.yaml
+++ b/.github/workflows/ci-pr.yaml
@@ -58,6 +58,21 @@ jobs:
           go-version: '1.24.2'
           cache-dependency-path: go/src/github.com/harmony-one/harmony/go.sum
 
+      - &cache-native-libs
+        name: Cache native mcl/bls libraries
+        uses: actions/cache@v4
+        with:
+          path: |
+            go/src/github.com/harmony-one/mcl/bin
+            go/src/github.com/harmony-one/mcl/lib
+            go/src/github.com/harmony-one/mcl/obj
+            go/src/github.com/harmony-one/bls/bin
+            go/src/github.com/harmony-one/bls/lib
+            go/src/github.com/harmony-one/bls/obj
+          key: ${{ runner.os }}-native-libs-${{ hashFiles('go/src/github.com/harmony-one/mcl/**', 'go/src/github.com/harmony-one/bls/**') }}
+          restore-keys: |
+            ${{ runner.os }}-native-libs-
+
       - &build-mcl
         name: Build mcl
         run: make -j4
@@ -82,6 +97,7 @@ jobs:
       - *checkout-mcl
       - *checkout-bls
       - *setup-go
+      - *cache-native-libs
       - *build-mcl
       - *build-bls
       - *build-harmony
@@ -107,6 +123,7 @@ jobs:
       - *checkout-mcl
       - *checkout-bls
       - *setup-go
+      - *cache-native-libs
       - *build-mcl
       - *build-bls
       - *build-harmony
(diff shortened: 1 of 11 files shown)


.github/workflows/ci-pr.yaml
```diff
@@ -58,6 +58,21 @@ jobs:
           go-version: '1.24.2'
           cache-dependency-path: go/src/github.com/harmony-one/harmony/go.sum
 
+      - &cache-native-libs
```
mur-me, line 61: Idea is good, but you aren't changing the way how CI works rn, because bls and mcl are still building **every time**.

Main idea of cache is to cache some actions one time and skip them on the next run if they are present.

What we are seeing on the 2 runs of the same PR is almost same timing results.

+ one thing - `make` on the harmony repo automatically build mcl and bls again.

Example diagram of the build stage:
```mermaid
flowchart TD
    A["Checkout harmony/mcl/bls"] --> B["Set up Go"]
    B --> C["actions/cache restores native mcl/bls cache"]

    C --> D{"Exact cache hit?"}

    D -->|yes| E["Cached bin/lib/obj restored"]
    D -->|no| F["No exact cached native outputs"]

    E --> G["Build mcl runs anyway"]
    F --> G

    G --> H["Build bls runs anyway"]
    H --> I["Build harmony runs"]

    I --> J["Job completes"]

    J --> K{"Cache miss and job success?"}
    K -->|yes| L["actions/cache may save native outputs"]
    K -->|no| M["No new native cache saved"]
```

.github/workflows/ci-pr.yaml
```diff
@@ -58,6 +58,21 @@ jobs:
...
+        name: Cache native mcl/bls libraries
+        uses: actions/cache@v4
```
mur-me, line 63: Lest use best practice here - last version + `sha1 # v1.1.1 comment`

.github/workflows/ci-pr.yaml
```diff
@@ -121,6 +138,10 @@ jobs:
           path: go/src/github.com/harmony-one/harmony-test
           persist-credentials: false
 
+      - &setup-docker-buildx
```
mur-me, line 141: why do we need this one? 
all github runners has docker by default

mur-me (changes requested):

PR #5066, opened 2026-06-23
Handle stale snapshots by returning ErrSnapshotStale

Frozen:
This pull request adds additional stale state checks to the diff layer's account and storage retrieval methods, improving the robustness of snapshot state management. Now, before attempting to access account or storage data, the code verifies if the diff layer is stale and returns an appropriate error if so.

Improvements to snapshot diff layer safety:

* Added a check for `Stale()` in the `AccountRLP` method of `diffLayer` to prevent access and return `ErrSnapshotStale` if the layer is stale.
* Added a similar `Stale()` check in the `Storage` method of `diffLayer` to ensure safe handling of stale state.


Frozen:
duplicate

PR #5083, opened 2026-07-14
Staking V2: batch operations, precompile validation, and safer staking failures

GheisMohammadi:
This PR introduces **Staking V2** — a staking upgrade that fixes several Staking V1 pain points, adds batch staking operations, and makes EVM precompile staking calls safer and cheaper to fail.

---

### 1. Precompile staking txs were under-validated
Native Harmony staking transactions (delegate, undelegate, collect rewards) are checked in the tx pool before inclusion. **EVM calls to the staking precompile (`0xfc`) were not.**

A transaction could pass the pool with only enough ONE to pay gas, get mined, and then **fail on-chain** — for example when redelegating undelegated funds that are not yet available, or when liquid balance is insufficient.

Users saw confusing failures: `status: 0` with **full gas burned** (e.g. 100,000 / 100,000), which looks like out-of-gas rather than a clear staking error.

### 2. Misleading redelegation balance in RPC
`hmy_getAvailableRedelegationBalance` could report tokens as available when the chain would not actually allow redelegation yet (same-epoch undelegations). Wallets and bots could build transactions doomed to fail.

### 3. Redelegation bookkeeping edge cases
Staking V1 redelegation logic could leave undelegation entries in inconsistent states during partial consumption. Staking V2 rewrites this path with explicit entry removal and clearer handling.

### 4. No batch or “undelegate all” operations
Rebalancing across many validators required **one transaction per delegate/undelegate**, increasing gas overhead, nonce management, and failure surface — especially for wallets and staking services managing large delegator portfolios.

### 5. UndelegateAll missed same-block delegations
`UndelegateAll` could miss delegations created in the same block, leaving stake behind when users expected a full exit.

---

### New native staking directives
- **BatchDelegate** — delegate to multiple validators in one transaction
- **BatchUndelegate** — undelegate from multiple validators in one transaction
- **UndelegateAll** — undelegate all active stake in one transaction (including same-block delegations)

### Precompile hardening
- **Tx pool validation** for precompile delegate / undelegate / collect-rewards calls (same checks as native staking txs)
- **Correct redelegation balance RPC** aligned with on-chain verifier rules
- **Precompile revert behavior** (after `StakingV2Epoch`): failures return a standard EVM revert with an encoded reason and **refund unused gas** instead of burning the full gas limit

---

## Benefits and cost savings

| Area | Before (V1) | After (V2) |
|------|-------------|------------|
| Failed precompile delegate | Often burns **100% of gas limit** | Reverts with reason; pays only **intrinsic + precompile cost** (~44k in typical delegate case vs 100k limit) |
| Doomed precompile txs | Can be included and fail on-chain | **Rejected in tx pool** — no wasted gas fee |
| Multi-validator rebalance | N separate transactions | **1 batch transaction** — fewer base tx costs and less nonce churn |
| Full exit | UndelegateAll could miss stake | Reliable full undelegation including same-block delegations |
| Developer / wallet UX | Opaque failures, wrong balance hints | Clear revert reasons and accurate redelegation balance |

**Example:** A mainnet precompile delegate of 68,500 ONE failed with `gasUsed = 100,000` after Bloom. Under Staking V2 pool validation, that tx would not be accepted. After activation, a similar on-chain failure would revert with a readable reason and cost roughly **half the gas** instead of the full limit.

---

## Hardfork notes

| Change | Requires `StakingV2Epoch`? |
|--------|----------------------------|
| Batch delegate / undelegate / undelegate-all | Yes |
| Precompile revert + gas refund | Yes |
| Tx pool precompile validation | No (node release) |
| RPC redelegation balance fix | No (node release) |


Frozen:
@copilot rebase dev

PR #5086, opened 2026-07-25
Replace Harmony BLS forks with official Herumi package

Frozen:
## Summary

- replace `github.com/harmony-one/bls` with the official `github.com/herumi/bls-eth-go-binary` module
- remove separate Harmony BLS/MCL checkout, build flags, shared-library packaging, and runtime library paths
- centralize Harmony consensus configuration in `crypto/bls/core`
- preserve the historical generator, serialization, group layout, hash-to-curve mode, address derivation, and public-key subtraction
- add golden wire-compatibility vectors for secret keys, public keys, message signatures, and hash signatures
- update CI, Docker, Makefile, scripts, and documentation for the single static Go dependency

The resulting Harmony binary does not require external `libbls` or `libmcl` files at runtime.

## Validation

- `go test ./crypto/bls/core ./crypto/bls ./crypto/vrf/bls ./consensus/quorum ./consensus/votepower ./staking/slash`
- compile `node/harmony` tests
- direct `go build ./cmd/harmony` without BLS/MCL CGO flags
- verified on macOS arm64 that the resulting binary has no dynamic dependency on `libbls` or `libmcl`
- full `go test ./...` completed all compile paths; remaining local failures require network or localhost bind access unavailable in the sandbox


GheisMohammadi (approved):

GheisMohammadi:
Great change! It simplifies the stack. We should test it carefully to ensure full compatibility—especially that the same BLS keys, signatures, serialization/deserialization, and all existing behavior remain identical to the previous implementation.

mur-me:
Tests done:
- [x] Pre-5086 and #5086 nodes run together and finalize blocks normally.
- [x] #5086 nodes successfully load and use unchanged pre-5086 BLS keys.
- [x] A new BLS key generated using the Herumi package is successfully added and used by a validator.
- [x] Pre-5086 nodes accept signatures produced by #5086 validators.
- [x] #5086 nodes accept signatures produced by pre-5086 validators.
- [x] Blocks signed by a mixed set of pre-5086 and #5086 validators are finalized.
- [x] A #5086 validator successfully acts as block leader.
- [x] The mixed-version devnet successfully crosses epoch boundaries.
- [x] A #5086 validator restarts with the same existing BLS key and resumes signing.
- [x] A validator is rolled back from #5086 to the pre-5086 release using the same database and BLS key.
- [x] The rolled-back node is alive and actively participating as a validator.

